PR #131, opened 2026-03-21
fix: make codex send replays idempotent

TheVortex8:
## Summary
- make `send_agent_message_request` idempotent when the same client `messageId` is replayed
- reject `messageId` reuse with different text instead of silently starting a new run
- allow Codex terminal events to end a stream even if a fresh `turn_started` event was not observed first
- add regressions for duplicate-send replay and terminal-before-start Codex behavior

## Root cause
On reconnect/replay paths, the daemon could accept the same logical user message twice because `messageId` was only used for timeline recording, not run idempotency. That could start a second Codex run for the same message.

The Codex app-server provider also buffered all pre-`turn_started` events except permissions. If the replayed/resumed run surfaced only a terminal event (`turn_completed`, `turn_failed`, or `turn_canceled`), the provider ignored it and could leave the agent stuck in `running`.

## Testing
- `npx vitest run src/server/agent/providers/codex-app-server-agent.test.ts`
- `npx vitest run src/server/daemon-e2e/send-message-id-replay.e2e.test.ts --maxWorkers=1 --minWorkers=1`

### DIFF
--- a/packages/app/src/contexts/session-context.tsx
+++ b/packages/app/src/contexts/session-context.tsx
@@ -49,6 +49,7 @@ import { derivePendingPermissionKey, normalizeAgentSnapshot } from "@/utils/agen
 import { resolveProjectPlacement } from "@/utils/project-placement";
 import { buildDraftStoreKey } from "@/stores/draft-keys";
 import type { AttachmentMetadata } from "@/attachments/types";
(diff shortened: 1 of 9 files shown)


boudra:
Closing for now: this needs before/after evidence from a real Codex reconnect/replay showing one run per message and completion when no turn-start event arrives. Please reopen with that evidence and regression-test results.

boudra:

PR #201, opened 2026-04-05
fix(server): reset session ID on query restart to prevent overwrite crash

skevetter:
When `ensureQuery()` restarts a query (triggered by interrupt, config change, or rewind failure), it tears down the old query but retains the old `claudeSessionId`. If the new query connects to a different Claude session (e.g. after a hook or reconnect), `captureSessionIdFromMessage()` detects the mismatch and throws, terminating the agent.

- Fixes a race condition where a Claude session ID overwrite during a query restart kills the agent mid-turn with a fatal `"CRITICAL: Claude session ID overwrite detected!"` error
- Resets `claudeSessionId` and `persistence` in `ensureQuery()` when performing a query restart, so the new query can establish its own session identity

Closes #200


boudra:
thanks for the explanation!

Issue #250, opened 2026-04-11
Desktop app cannot open multiple windows


hissincn:
## Summary
The Electron desktop app currently behaves as a single-window application. Attempting to open a second window just focuses the existing window instead of creating a new one.

## Current behavior
- Launching the desktop app a second time does not open another window.
- The existing window is focused/restored instead.
- There is no obvious "New Window" entry in the desktop app menu.

## Expected behavior
Users should be able to open multiple desktop windows at the same time, especially when working with different agents or projects in parallel.

## Why this matters
Paseo is built around managing multiple coding agents and workspaces. A single-window limitation makes it harder to:
- compare agents side by side
- monitor multiple sessions at once
- keep different projects or contexts open simultaneously

## Suspected cause
The desktop app uses Electron's single-instance lock and currently routes second launches back to the first window instead of creating a new one. The app menu also appears to be missing a dedicated "New Window" action.

## Reproduction
1. Open the Paseo desktop app.
2. Try to launch the app again from the dock, app launcher, or command line.
3. Observe that no new window is created.
4. The existing window is focused instead.

## Acceptance criteria
- Users can open more than one desktop window.
- Launching the app again creates a new window when appropriate.
- The app exposes a clear "New Window" action in the menu.
- Existing open-project routing continues to work without breaking current behavior.
- Multi-window behavior works on macOS, Windows, and Linux, or platform-specific limitations are documented.

arieel-ost:
Opened #1355 which fixes this: File → New Window (⌘N), relaunching the app now opens a new window instead of focusing the existing one, plus an "Open in new window" action on sidebar projects. Works across macOS/Windows/Linux via the existing single-instance + second-instance paths. cc @boudra

PR #254, opened 2026-04-11
Fix desktop workspace multi-window handling

hissincn:
## Summary

This PR fixes the desktop multi-window workflow so workspaces can be moved into their own windows without being duplicated across windows.

It also fixes a dev-only desktop issue where the desktop-managed daemon could stay alive with stale CORS origins after the Metro port changed, causing the renderer to time out while connecting.

## What changed

- added main-process workspace ownership tracking per window
- made all desktop window creation paths use tracked window creation
- added renderer-side desktop window state syncing
- filtered sidebar workspaces by window ownership
- guarded host-root auto-redirects so a window only opens workspaces it owns
- added workspace route guards to prevent a moved workspace from staying open in the old window
- added drag-outside handling on web to move a workspace into a new window
- added "Move to new window" menu actions as a fallback
- claimed newly opened projects and newly created worktrees in the current window
- restarted the desktop-managed dev daemon automatically when the Metro origin changes so dev renderer connections do not fail with stale CORS config

## Behavior

- a brand-new window opens to the startup/open-project flow
- moved workspaces only appear in their owning window
- unowned workspaces stay visible in the primary window only
- closing a secondary window releases its workspace ownership back to the primary window
- desktop dev no longer gets stuck with `Rejected connection from origin` after Metro restarts on a new port

## Validation

- `npm run build:main --workspace=@getpaseo/desktop`
- manually verified desktop dev startup with isolated `PASEO_HOME`
- manually verified second-instance creates a new window
- manually verified the desktop-managed daemon reconnect path after Metro origin changes
- manually tested the multi-window workflow end-to-end


hissincn:
[#250](https://github.com/getpaseo/paseo/issues/250)

hissincn:
Emm...maybe,it's a feature

hissincn:
The conflict has been resolved.

arieel-ost:
Opened #1355 as a smaller alternative approach to #250. Rather than per-window workspace ownership (the filtering / route guards / hydration this PR carries), it's a "land-on" hybrid: ⌘N / relaunch / an "Open in new window" action open new windows, but every window keeps the full sidebar — no ownership map, no filtering. Roughly 2 commits / ~560 lines.

Sharing for comparison, not to step on this work — whichever direction the maintainers prefer. cc @boudra

boudra:
superseded by https://github.com/getpaseo/paseo/pull/1355

PR #256, opened 2026-04-11
Fix agent module loading issue in CI

volfiros:
This change resolves a crash occurring when tests load the agent module. The issue was caused by the missing `@sinclair/typebox` dependency in the dependency graph.

**Changes:**
* Added `@sinclair/typebox` to `packages/server/package.json`.
* Verified that the `pi-direct-agent.ts` code remains unchanged.
* Confirmed that all packages compile correctly and the CI test runner crash is resolved.


volfiros:
@boudra Please Review

Issue #264, opened 2026-04-12
Enter key during IME composition triggers message send


nexmoe:
## Description

When typing in the input field using an IME (Input Method Editor) for Chinese/Japanese/Korean, pressing Enter to dismiss or confirm IME candidate words also triggers the send action, causing the message to be sent prematurely.

## Steps to Reproduce

1. Focus on the input field
2. Type using a CJK input method (e.g. Chinese Pinyin)
3. IME shows candidate word list
4. Press Enter to confirm/dismiss the candidates
5. The input content is sent immediately instead of just confirming the IME selection

## Expected Behavior

When the IME is in composing state, the Enter key should only interact with the IME (confirm candidate selection) and should NOT trigger form submission / message send. The send action should only be triggered when the IME composition is finished.

## Suggested Fix

Check the `isComposing` state before handling Enter key submission:

\`\`\`js
let isComposing = false

inputEl.addEventListener('compositionstart', () => { isComposing = true })
inputEl.addEventListener('compositionend', () => { isComposing = false })
inputEl.addEventListener('keydown', (e) => {
  if (e.key === 'Enter' && isComposing) {
    e.preventDefault()
    return
  }
  // normal submit logic
})
\`\`\`

Alternatively, check \`e.isComposing\` directly in the keydown handler.

nexmoe:
Duplicates found in this repo: #312, #309, #314, #241 — all report the same IME composition Enter key issue.

PR #242, opened 2026-04-11
fix(desktop): ignore Enter key during IME composition in message inputs

hissincn:
[Fixes #241](https://github.com/getpaseo/paseo/issues/241)

### Summary
This is a small, focused bug fix for an IME (Input Method Editor) UX issue. 

When typing with an IME (e.g., Chinese Pinyin, Japanese), users press `Enter` to finalize their composition. Currently, pressing `Enter` accidentally submits the message instead of just confirming the text into the input field. 

This PR simply checks `nativeEvent.isComposing` and bails out early from the `onKeyPress` handler if an IME session is active in `message-input.tsx` and `agent-input-area.tsx`.

### PR Checklist
- [x] Focused change (one bug only)
- [x] Solves a clear regression/UX bug (IME composition)
- [x] Typecheck passes
- [x] Tested on Web / Desktop

PR #270, opened 2026-04-12
fix: skip Enter key handling during IME composition

### LINKED ISSUE #241: Bug: Pressing Enter during IME composition sends the message prematurely
**Describe the bug**
When typing in the chat input using an Input Method Editor (IME) for languages like Chinese, Japanese, or Korean, pressing the `Enter` key to confirm the composition or input English characters mistakenly triggers the message submission. 

**To Reproduce**
1. Focus on the message input area.
2. Switch your keyboard to an IME (e.g., Chinese Pinyin).
3. Type some characters (an IME composition window appears).
4. Press `Enter` to confirm the phonetic spelling or English text.
5. The message is sent immediately instead of just confirming the text into the input field.

**Expected behavior**
The `Enter` key should only finalize the IME composition. It should not trigger the "send message" or "queue message" action if the user is in the middle of composing text with an IME.

**Proposed Solution**
We can fix this by checking the `nativeEvent.isComposing` property in the `onKeyPress` handlers (specifically in `message-input.tsx` and `agent-input-area.tsx`) and bailing out early if an IME session is active.

cjh-store:
## Summary

- Skip Enter key handling when `isComposing` is true during IME input (Chinese, Japanese, Korean, etc.)
- Propagate `isComposing` through the key event chain: `message-input` → `agent-input-area` → `use-autocomplete`
- Follow the same pattern already used in `terminal-emulator-runtime.ts`

## Changes

| File | Change |
|------|--------|
| `message-input.tsx` | Add `isComposing` to event type + guard before Enter handling |
| `agent-input-area.tsx` | Accept `isComposing` in key event type |
| `use-autocomplete.ts` | Skip Tab/Enter handling during composition |

## Test plan

- [ ] Type Chinese characters using Pinyin IME, press Enter to confirm character selection — should NOT send message
- [ ] Press Enter without IME active — should send message as before
- [ ] Shift+Enter — should still insert newline
- [ ] Cmd/Ctrl+Enter — should still trigger alternate action
- [ ] Autocomplete selection via Enter/Tab during composition — should not trigger

Fixes #241, #264


cjh-store:
### Comparison with #242

I noticed #242 addresses the same IME issue. Here's how this PR differs:

| | #242 | This PR |
|---|---|---|
| **`keyCode === 229` fallback** | ❌ | ✅ Matches react-native-web's own `isEventComposing` logic |
| **Type safety** | Uses `as any` casts | Properly typed with `isComposing?: boolean` and `keyCode?: number` |
| **Autocomplete coverage** | Not covered | ✅ `use-autocomplete.ts` + `use-agent-autocomplete.ts` also guarded |
| **Check scope** | Blocks all key events during composition (including Escape, arrows) | Only blocks Enter — IME users can still use other keys normally |
| **isComposing propagation** | Not passed to parent handlers | ✅ Propagated through the full event chain |
| **Interface alignment** | — | ✅ All 4 related interfaces updated consistently |

JingxuanKang:
I really need this feature. I was already enjoying using this, but without this functionality, it is impossible to work with.

I hope the author can merge this PR😊

hissincn:
> I really need this feature. I was already enjoying using this, but without this functionality, it is impossible to work with.
> 
> I hope the author can merge this PR😊

i need toooooooo!

Issue #276, opened 2026-04-12
Support defaultModel override in provider runtime settings for AWS Bedrock


skevetter:
## Problem

When using Claude Code with AWS Bedrock (`CLAUDE_CODE_USE_BEDROCK=1`), the Paseo app (in daemon mode) always sends `claude-opus-4-6` (the hardcoded `isDefault: true` model from `claude-models.ts`) to the Claude SDK. Bedrock rejects this with:

```
API Error (claude-opus-4-6): 400 The provided model identifier is invalid…
Try --model to switch to us.anthropic.claude-opus-4-1-20250805-v1:0.
```

Bedrock requires fully-versioned model IDs like `anthropic.claude-sonnet-4-5-20250929-v1:0` or cross-region inference profile IDs like `us.anthropic.claude-opus-4-1-20250805-v1:0`.

Setting `ANTHROPIC_MODEL` in the provider env config does not help because the Paseo app explicitly passes the selected model ID to `createAgent`, which overrides the env var.

## Root cause

In `claude-agent.ts`, `base.model = this.config.model` is set unconditionally when a model is present. Since the app always sends a model (the default), `ANTHROPIC_MODEL` is never consulted.

## Proposed fix

Add a `defaultModel` field to `ProviderRuntimeSettings` (alongside `command` and `env`). When set, it overrides `this.config.model` before passing to the SDK. This lets Bedrock users configure the correct model ID in `paseo.config.json` without needing code changes or a wrapper script.

**`provider-launch-config.ts`:**
```ts
export const ProviderRuntimeSettingsSchema = z
  .object({
    command: ProviderCommandSchema.optional(),
    env: z.record(z.string()).optional(),
    defaultModel: z.string().optional(),  // add this
  })
  .strict();
```

**`claude-agent.ts`** (in `buildQueryOptions`):
```ts
if (this.config.model) {
  base.model = this.runtimeSettings?.defaultModel ?? this.config.model;
}
```

**Config usage:**
```json
{
  "agents": {
    "providers": {
      "claude": {
        "env": {
          "CLAUDE_CODE_USE_BEDROCK": "1",
          "AWS_REGION": "us-west-2",
          "AWS_BEARER_TOKEN_BEDROCK": "..."
        },
        "defaultModel": "anthropic.claude-sonnet-4-5-20250929-v1:0"
      }
    }
  }
}
```

Happy to submit a PR with this change if the approach looks good.

dgallitelli:
[...]

PR #277, opened 2026-04-12
feat(server): support defaultModel override in provider runtime settings

skevetter:
## Summary

- Add support for `defaultModel` override in provider runtime settings for Bedrock providers
- Allows users to specify a default model when launching Claude agents through Bedrock

## Changes

- Modified provider launch config to include `defaultModel` in runtime settings
- Updated Claude agent provider to use `defaultModel` from runtime settings when available
- Added schema definition for `defaultModel` in paseo.config.v1.json

### DIFF
--- a/packages/server/src/server/agent/provider-launch-config.ts
+++ b/packages/server/src/server/agent/provider-launch-config.ts
@@ -34,6 +34,7 @@ export const ProviderRuntimeSettingsSchema = z
   .object({
     command: ProviderCommandSchema.optional(),
     env: z.record(z.string()).optional(),
+    defaultModel: z.string().optional(),
   })
   .strict();
 
--- a/packages/server/src/server/agent/providers/claude-agent.ts
+++ b/packages/server/src/server/agent/providers/claude-agent.ts
@@ -2141,7 +2141,7 @@ class ClaudeAgentSession implements AgentSession {
     }
 
     if (this.config.model) {
-      base.model = this.config.model;
+      base.model = this.runtimeSettings?.defaultModel ?? this.config.model;
     }
     this.lastOptionsModel = base.model ?? null;
     if (this.claudeSessionId) {
--- a/packages/website/public/schemas/paseo.config.v1.json
+++ b/packages/website/public/schemas/paseo.config.v1.json
@@ -166,6 +166,9 @@
                     "additionalProperties": {
                       "type": "string"
                     }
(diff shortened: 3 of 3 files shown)


dgallitelli:
This would unblock Bedrock users — I'm hitting the exact problem described in #276. Small change, big impact. Would love to see this merged.

dgallitelli:
## Workaround: AWS Bedrock works in the UI today (without this PR)

While this PR is pending, I found that **provider profiles** (from #290) already work for Bedrock in the Paseo web UI — the model selection issue only affects the CLI fallback path.

### Setup

Add a custom provider to `~/.paseo/config.json`:

```json
{
  "agents": {
    "providers": {
      "claude-bedrock": {
        "extends": "claude",
        "label": "Claude (Bedrock)",
        "description": "Claude Code via AWS Bedrock",
        "env": {
          "CLAUDE_CODE_USE_BEDROCK": "1",
          "AWS_REGION": "us-east-1"
        },
        "models": [
          {
            "id": "us.anthropic.claude-opus-4-6-v1[1m]",
            "label": "Opus 4.6 (1M)",
            "isDefault": false
          },
          {
            "id": "us.anthropic.claude-sonnet-4-6[1m]",
            "label": "Sonnet 4.6 (1M)",
            "isDefault": true
          },
          {
            "id": "us.anthropic.claude-haiku-4-5-20251001-v1:0",
            "label": "Haiku 4.5",
            "isDefault": false
          }
        ]
      }
    }
  }
}
```

Restart the daemon (`paseo daemon restart`), then select **"Claude (Bedrock)"** as the provider in the UI. The Bedrock model IDs from the profile flow through correctly because the UI always explicitly passes the selected model.

**Why this PR is still needed:** The CLI path (`paseo run --provider claude-bedrock "..."`) without an explicit `--model` flag falls back to hardcoded short model IDs (e.g., `claude-opus-4-6`) which Bedrock rejects. This PR's `defaultModel` override would fix that fallback.

If using AWS credentials via instance profile or `~/.aws/credentials`, you can omit `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` from the `env` block.

boudra:
thanks for the pull request.

paseo's custom providers config supports per-provider model lists with an isDefault flag, which covers this use case. see https://paseo.sh/docs/custom-providers

PR #283, opened 2026-04-12
feat: add 'default' model option to respect provider config files

zccrs:
Add a special 'default' model entry for each provider (claude, codex, opencode) that defers to the provider's own config file for model selection:

- Claude: ~/.claude/settings.json
- Codex: ~/.codex/config.toml
- OpenCode: ~/.opencode/opencode.json

This allows users to explicitly select their configured default rather than having to choose from Paseo's hardcoded model list.

Changes:
- claude-models.ts: Add 'default' as first option in model list
- claude-agent.ts: Treat 'default' as null (don't override SDK)
- agent-manager.ts: Normalize 'default' to undefined
- codex-app-server-agent.ts: Add 'default' entry, skip override when default
- opencode-agent.ts: Add 'default' entry to model list

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ packages/server/src/server/fixtures/dictation/dictation-debug-largest.transcript
 /artifacts
 packages/desktop/.cache/
 packages/desktop/src-tauri/resources/managed-runtime/
+.omx/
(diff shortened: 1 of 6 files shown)


boudra:
thanks for the pull request.

#979 just landed and discovers claude models from settings.json without a 'default' pseudo-model. closing this one as superseded.

PR #289, opened 2026-04-12
feat(provider): add Hermes ACP provider support

### LINKED ISSUE #278: Support Hermes Agent, please

polkn:
## Summary

Adds Hermes as a first-class Paseo provider via ACP.

Closes #278.

This is intentionally a small provider PR:
- one new provider: `hermes`
- one integration path: `hermes acp`
- no new settings UI
- no new protocol surface
- no unrelated runtime or bootstrap changes

## Why this fits Paseo

Hermes already exposes an ACP server over stdio, and Paseo already has an ACP provider base class.

That makes Hermes a good fit for the existing provider architecture without introducing a new transport or bespoke integration layer. This PR follows the same pattern Paseo already uses for ACP-backed providers instead of adding a one-off implementation.

## What this adds

Server:
- `HermesACPAgentClient` implemented as a thin `ACPAgentClient` subclass
- `hermes` added to the provider manifest
- `hermes` added to the provider registry
- Hermes added to the shared real-provider E2E config list

Tests:
- provider-specific unit tests covering:
  - default model discovery from Hermes `config.yaml`
  - empty-model fallback
  - runtime `HERMES_HOME` override behavior
  - no static modes

Docs:
- README now lists Hermes as a supported agent CLI
- architecture doc now includes Hermes in the provider table
- provider authoring doc now reflects current ACP providers and shows how Hermes profiles can be targeted through runtime env overrides

## Implementation details

The provider launches Hermes with:

```bash
hermes acp
```

Hermes currently does not expose rich enough model metadata through ACP for Paseo to populate the model picker cleanly from the protocol alone.

To keep the provider usable without adding Hermes-specific protocol handling, the provider synthesizes a default model entry from `<HERMES_HOME>/config.yaml` by reading:
- `model.default`
- `model.provider`

That gives Paseo a stable default model entry while keeping the integration small and local to the provider.

## Profile support

This PR supports non-default Hermes profiles through existing Paseo provider runtime settings rather than through new UI.

Example:

```json
{
  "agents": {
    "providers": {
      "hermes": {
        "env": {
          "HERMES_HOME": "/Users/you/.hermes/profiles/work"
        }
      }
    }
  }
}
```

Why this approach:
- it supports multiple Hermes installs/profiles immediately
- it uses an existing Paseo configuration mechanism
- it avoids expanding the product surface in this PR

## Deliberate non-goals

Not included here:
- profile-selection UI
- custom Hermes icon assets
- Hermes-side ACP changes
- unrelated cleanup in other providers or bootstrap/runtime code

Those may be reasonable follow-ups, but they would broaden a PR that can stay focused as “add Hermes as an ACP provider.”

## Verification

Hermes-specific verification completed:

- `cd packages/server && npx vitest run src/server/agent/providers/hermes-acp-agent.test.ts`
  - passes (`4 tests`)
- live smoke through Paseo’s daemon test harness using the real local Hermes binary
  - agent creation succeeded with provider `hermes`
  - a real round-trip completed successfully
  - final assistant text was `HERMES_SMOKE_OK`
  - persistence returned with provider `hermes`

## Notes on repo-wide checks

I also ran broader checks, but this checkout currently has unrelated failures outside the Hermes change:
- `npm run typecheck` fails in unrelated server/website code
- broad server test entrypoints currently hit unrelated existing failures before they reach the Hermes-specific path

I did not bundle fixes for those unrelated failures into this PR, to keep the scope narrow and reviewable.

## UI impact

This changes provider availability and provider documentation, but does not add new UI components or layouts. Hermes currently uses the existing generic provider icon fallback in the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
   </a>
 </p>
 
-<p align="center">One interface for all your Claude Code, Codex and OpenCode agents.</p>
+<p align="center">One interface for all your Claude Code, Codex, OpenCode, and Hermes agents.</p>
 
 <p align="center">
   <img src="https://paseo.sh/hero-mockup.png" alt="Paseo app screenshot" width="100%">
@@ -34,7 +34,7 @@
 Run agents in parallel on your own machines. Ship from your phone or your desk.
 
 - **Self-hosted:** Agents run on your machine with your full dev environment. Use your tools, your configs, and your skills.
-- **Multi-provider:** Claude Code, Codex, and OpenCode through the same interface. Pick the right model for each job.
+- **Multi-provider:** Claude Code, Codex, OpenCode, and Hermes through the same interface. Pick the right model for each job.
 - **Voice control:** Dictate tasks or talk through problems in voice mode. Hands-free when you need it.
 - **Cross-device:** iOS, Android, desktop, web, and CLI. Start work at your desk, check in from your phone, script it from the terminal.
 - **Privacy-first:** Paseo doesn't have any telemetry, tracking, or forced log-ins.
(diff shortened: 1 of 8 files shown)


polkn:
Closing this in favor of #290. The maintainer-owned provider profiles work appears to be the right upstream direction, and it overlaps heavily with the same provider/registry/config surfaces. Thanks for the review path here — I'll test #290 directly instead of asking you to reconcile both.

PR #303, opened 2026-04-12
[codex] Fix first-run dev setup

Varuas37:
## Summary

Fresh checkouts were not able to get from `npm install` to `npm run dev` successfully. This PR makes the default dev path self-bootstrapping so contributors do not need undocumented global setup or manual workspace builds.

## Root Cause

There were three separate first-run assumptions in the current setup:

1. `scripts/dev.sh` invoked `portless`, but `portless` was not installed by the repository.
2. The script expected the shared `portless` proxy to already be running before calling `portless get ...` / `portless run ...`.
3. The daemon imports workspace packages through published `dist/*` entries, but fresh checkouts do not have `packages/highlight/dist/*` or `packages/relay/dist/*` yet.

That meant a new contributor would hit at least one of these failures immediately:
- `portless: command not found`
- proxy-not-running failures from `portless`
- `ERR_MODULE_NOT_FOUND` for `@getpaseo/highlight/dist/index.js`

## What Changed

- added `portless` as a root dev dependency
- updated `scripts/dev.sh` to ensure the `portless` proxy is running before resolving service URLs
- updated `scripts/dev.sh` to build `@getpaseo/highlight` and `@getpaseo/relay` on first run when their `dist` outputs are missing

## Why This Shape

This keeps the fix local to the existing dev entrypoint instead of requiring separate contributor setup steps or broader package-script changes. It also preserves the current dev flow and only adds the missing bootstrap work when needed.

## Validation

- `npm install`
- `npm run dev`
- `npm run typecheck`
- `npm run format`

During validation, `npm run dev` successfully reached a usable state with:
- daemon listening and bootstrapped
- Expo/Metro running
- only non-blocking local speech model warnings remaining on this machine

## Notes

A separate `portless` stale-route ownership issue can still happen if `app.localhost` / `daemon.localhost` are left behind by a previous session. That behavior is outside the scope of this PR. This PR fixes the original clean-checkout boot failure, not route cleanup inside `portless`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "get-port-cli": "^3.0.0",
         "knip": "^5.82.1",
         "patch-package": "^8.0.1",
+        "portless": "^0.10.1",
         "react": "19.1.4",
         "react-dom": "19.1.4",
         "typescript": "^5.9.3"
@@ -17466,7 +17467,6 @@
       "dev": true,
       "license": "BSD-3-Clause",
       "optional": true,
-      "peer": true,
       "engines": {
         "node": ">=0.10.0"
       }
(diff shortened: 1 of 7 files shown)


boudra:
thanks for the pull request.

Issue #312, opened 2026-04-13
IME composition: Enter key triggers message send instead of confirming candidate


muzhi1991:
## Description

When using a CJK (Chinese/Japanese/Korean) input method on macOS, pressing Enter to confirm an IME candidate in the chat input field triggers message send instead of just committing the text.

For example, typing English with a Chinese IME active → pressing Enter to commit the raw text → message gets sent immediately instead of just inserting the text into the input field.

## Root Cause

In `packages/app/src/components/message-input.tsx`, the `handleDesktopKeyPress` function (line ~885) does not check `event.nativeEvent.isComposing` before handling Enter:

```typescript
function handleDesktopKeyPress(event: WebTextInputKeyPressEvent) {
  // ...
  if (event.nativeEvent.key !== "Enter") return;
  if (shiftKey) return;
  // Directly triggers send — no isComposing check
  event.preventDefault();
  handleDefaultSendAction();
}
```

Interestingly, the terminal emulator component (`packages/app/src/terminal/runtime/terminal-emulator-runtime.ts`, line ~284) already handles this correctly:

```typescript
if (event.type !== "keydown" || event.isComposing) {
  return true;
}
```

## Suggested Fix

Add an `isComposing` guard before processing Enter:

```typescript
if (event.nativeEvent.key !== "Enter") return;
if (event.nativeEvent.isComposing) return; // ← add this line
```

## Environment

- macOS with CJK input method (e.g. Pinyin, Sogou, etc.)
- Paseo desktop app (Electron)

nexmoe:
Related: #264, #309, #314, #241 — all describe the same root cause: the Enter keydown handler does not check `event.isComposing`, so confirming an IME candidate also fires the send action.

muzhi1991:
重复了，sorry

Issue #309, opened 2026-04-13
Enter key submits input when using CJK IME


peidayu:
**Description:**
When using a CJK input method (Chinese/Japanese/Korean), pressing Enter to confirm composition also triggers submission.

**Expected:**
Enter should only confirm IME input.

**Actual:**
Enter confirms input and submits at the same time.

**Suggestion:**
Ignore Enter key events while IME is composing (e.g., `event.isComposing`).

nexmoe:
Duplicate of #264. Also related: #312, #314, #241 — same IME composition Enter key issue.

PR #340, opened 2026-04-13
fix: prevent IME composition Enter from triggering send on macOS

### LINKED ISSUE #314: macos下回车按键的适配问题
在macos系统下面，使用中文输入法输入英文内容默认回车确定英文输入的时候会直接换成发送功能。希望可以提供中文输入法适配，或者自定义发送按键为cmd+enter

feiandxs:
## Summary

On macOS, pressing Enter to commit CJK (Chinese/Japanese/Korean) IME composition also triggers the message send action.

The codebase already has an `isComposing` guard in `handleDesktopKeyPress`, but it does not work on macOS due to a platform-specific event ordering difference:

- **Windows / Linux**: `compositionend` fires *after* the Enter `keydown`, so `isComposing` is still `true` when the key handler runs — the guard works correctly.
- **macOS**: `compositionend` fires *before* the Enter `keydown` that commits the composition — so by the time the handler runs, `isComposing` is already `false` and the guard is bypassed.

Fix: attach `compositionstart` / `compositionend` listeners directly to the underlying DOM element and maintain `isComposingRef` independently of `event.nativeEvent.isComposing`. The `compositionend` handler defers the flag clear with `setTimeout(0)` so the keydown handler still sees the composition as active on that tick.

## Changes

- `packages/app/src/components/message-input.tsx`
  - Add `isComposingRef` to track IME composition state independently
  - Add `useEffect` (web-only) to attach native `compositionstart` / `compositionend` listeners via `textInputRef` / `getNativeRef()`
  - Update the IME guard in `handleDesktopKeyPress` to also check `isComposingRef.current`

## Test Plan

Requires manual testing with a real CJK input method. Automated tests cannot simulate IME composition event sequences reliably.

Steps to reproduce the bug (before fix) and verify the fix:

1. Open the web or Electron desktop app
2. Switch system input method to Chinese (e.g. macOS Simplified Chinese Pinyin)
3. Click the message input field
4. Type some English letters — they appear in the IME candidate buffer
5. Press Enter to commit the text to the input
6. **Before fix**: text is committed *and* the message is sent
7. **After fix**: text is committed, message is NOT sent

Typecheck and unit tests pass:
```
npm run typecheck --workspace=@getpaseo/app  ✓
npm run test --workspace=@getpaseo/app       ✓  127 test files, 767 tests
```

Fixes: #314
Related: #241, #264, #309, #312

### DIFF
--- a/packages/app/src/components/message-input.tsx
+++ b/packages/app/src/components/message-input.tsx
@@ -306,6 +306,9 @@ export const MessageInput = forwardRef<MessageInputRef, MessageInputProps>(funct
   const overlayTransition = useSharedValue(0);
   const sendAfterTranscriptRef = useRef(false);
   const valueRef = useRef(value);
+  // Tracks IME composition state independently of event.isComposing, which is
+  // unreliable on macOS (compositionend fires before keydown there).
+  const isComposingRef = useRef(false);
   const serverInfo = useSessionStore(
     useCallback(
       (state) => {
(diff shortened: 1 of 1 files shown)


boudra:
thanks for the pull request.

PR #402, opened 2026-04-14
fix(server): load Codex repo-local skills from the agent workspace

### LINKED ISSUE #401: Codex repo-local skills load from daemon cwd instead of agent workspace
## Summary

Paseo's Codex provider does not reliably load repo-local Codex skills from the agent workspace.

## Reproduction

1. Create a repo with a repo-local skill at `repo/.codex/skills/repo-skill/SKILL.md`.
2. Start Paseo and create a Codex agent whose `cwd` points at that repo.
3. Open the slash-command list or call `listCommands()`.

## Expected

The repo-local skill appears for that agent.

## Actual

The skill list reflects the daemon process working directory instead of the agent workspace, so repo-local Codex skills are missing or come from the wrong repo.

## Root Cause

Paseo launches `codex app-server` without setting its process `cwd` to the agent session workspace. In practice, Codex discovers repo-local `.codex/skills` based on the app-server process cwd. The current `skills/list` request also passes `cwd` in a shape that does not affect Codex app-server discovery.

## Proposed Fix

- Launch `codex app-server` with the agent session `cwd`.
- Pass `skills/list` the session cwd in the expected string shape.
- Add a regression test that proves a repo-local `.codex/skills/.../SKILL.md` is listed for the matching workspace.

lldxflwb:
## Summary

Fix Codex skill discovery so repo-local `.codex/skills` are loaded from the agent workspace instead of the daemon process working directory.

Closes #401.

## Root cause

Paseo launched `codex app-server` without setting its process cwd to the Codex session workspace. In practice, Codex app-server discovers repo-local skills from its own process cwd, so skills from the active repo were missing or replaced by skills from the daemon checkout.

The existing `skills/list` request also passed `cwd` as an array shape that did not affect Codex app-server discovery.

## Changes

- launch `codex app-server` with the agent session cwd
- pass `skills/list` the session cwd as a string
- add a regression unit test for the `skills/list` request shape
- add a real e2e regression test proving repo-local `.codex/skills/.../SKILL.md` appears for the matching workspace

## Validation

- `npx vitest run src/server/agent/providers/codex-app-server-agent.test.ts src/server/agent/providers/codex-app-server-agent.features.test.ts`
- `npx vitest run src/server/agent/providers/codex-app-server-agent.e2e.test.ts -t "lists repo-local Codex skills from the session workspace"`
- `npm run typecheck`
- `npm run format`

## Scope

This change is Codex-only. It does not change Claude, OpenCode, or client schemas.

### DIFF
--- a/packages/server/src/server/agent/providers/codex-app-server-agent.e2e.test.ts
+++ b/packages/server/src/server/agent/providers/codex-app-server-agent.e2e.test.ts
@@ -3,7 +3,7 @@ import { execFileSync } from "node:child_process";
 import os from "node:os";
 import path from "node:path";
 import { createServer } from "node:http";
-import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { once } from "node:events";
 
 import { CodexAppServerAgentClient } from "./codex-app-server-agent.js";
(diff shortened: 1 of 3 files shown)


lldxflwb:
Confirmed locally in a direct-connection Paseo setup.

I tested this against a patched daemon by creating two separate workspaces, each with a different repo-local Codex skill under `.codex/skills/.../SKILL.md`, then connecting directly to that daemon and creating one Codex agent per workspace.

Observed behavior after this fix:
- each Codex agent listed the repo-local skill from its own workspace
- skills from the daemon checkout were no longer leaking into other workspaces
- the targeted Codex regression tests added in this PR also pass locally

This was tested in direct mode against the local daemon, not only through the isolated e2e harness.

chinameok:
One detail to re-check before merging: the current Codex app-server schema I generated locally exposes `SkillsListParams` as `cwds?: string[]`, not `cwd?: string`.

I tested this directly against `codex app-server` with a temp repo containing `.codex/skills/repo-skill/SKILL.md`:
- `skills/list` with `cwds: [repoRoot]` returns `repo-skill` with `scope: "repo"`
- `skills/list` with `cwd: [repoRoot]` is ignored and falls back to the app-server process cwd

So the process `cwd` change in this PR is still useful and addresses the daemon-checkout leakage, but the request-shape change here may still be version-sensitive/outdated. I think this line should likely use:

```ts
this.config.cwd ? { cwds: [this.config.cwd] } : {}
```

rather than `cwd: this.config.cwd`, unless Paseo intentionally targets an older Codex app-server protocol that accepted singular `cwd`.

lldxflwb:
Updated the PR to use the current Codex app-server request shape for skills discovery:

- changed `skills/list` params from `cwd` to `cwds`
- kept the app-server process cwd fix in place
- re-ran the targeted unit tests, the repo-local Codex skills e2e regression, and full workspace typecheck locally

So the PR now fixes both pieces:
1. the daemon-checkout cwd leakage
2. the request-shape mismatch with current Codex app-server

boudra:
thanks for the pull request.

PR #412, opened 2026-04-15
fix: update lockfile signatures and Nix hash

eljojo:
hi! somehow this commit fixing the hash is missing from main, it just caught me during an upgrade.
it's weird because I didn't generate the commit, I updated my fork, and I saw it was there. Looks like a GitHub issue?

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -42,7 +42,7 @@ buildNpmPackage rec {
 
   # To update: run `nix build` with lib.fakeHash, copy the `got:` hash.
   # CI auto-updates this when package-lock.json changes (see .github/workflows/).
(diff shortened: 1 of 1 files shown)


boudra:
@eljojo hmm right, looks like the fix hash action ran as I probably pushed to main so it failed to push: https://github.com/getpaseo/paseo/actions/runs/24404092556/job/71282420048

not sure whats the ebst way to make this more resilient, maybe rebase when it happens and retry? or maybe GH has built in retry options

eljojo:
@boudra thanks for the context, that's good to know! 

if the hash is broken, other PRs should fix it automatically so it's not that big of a deal, I discovered this as I was debugging what was going on so I decided to submit the PR anyways.

PR #456, opened 2026-04-16
Fix provider snapshot cache collisions across workspaces

aysuio:
## Summary
- key provider snapshot cache entries by both `serverId` and `cwd`
- request provider snapshots with the current workspace cwd when the agent form or status bar needs workspace-specific provider state
- ignore `providers_snapshot_update` events that belong to a different cwd
- add a focused test covering Windows path normalization and per-cwd cache isolation

## Root Cause
The daemon serves provider snapshots per workspace `cwd`, but the app cached them only by `serverId`.
That let one workspace overwrite another workspace's provider state, which could make `codex` appear available in one folder and missing in another.

## Verification
- reproduced against a live daemon where both `E:\evolution` and `E:\lzzy` supported `codex` server-side
- added a focused unit test for the new cache key behavior

## Notes
I could not complete a full local app typecheck on this Windows machine because a fresh workspace install is currently blocked by an npm workspace symlink failure:
`EISDIR: illegal operation on a directory, symlink '...packages\\website' -> '...node_modules\\@getpaseo\\website'`

### DIFF
--- a/packages/app/src/components/agent-status-bar.tsx
+++ b/packages/app/src/components/agent-status-bar.tsx
@@ -879,7 +879,7 @@ export const AgentStatusBar = memo(function AgentStatusBar({
     isLoading: snapshotIsLoading,
     isFetching: snapshotIsFetching,
     invalidate: invalidateSnapshot,
-  } = useProvidersSnapshot(serverId);
+  } = useProvidersSnapshot(serverId, agent?.cwd);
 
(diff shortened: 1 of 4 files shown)


boudra:
Implemented in [#4401](https://github.com/getpaseo/paseo/pull/4401), which is merged. Closing this PR.

PR #459, opened 2026-04-16
fix(cli): include custom providers in provider ls

chenmijiang:
## Summary

Before this change, the CLI connected to the daemon without an `appVersion`. The server then treated it as a legacy client and filtered provider snapshot entries down to legacy built-in providers only, so custom providers could appear in the app settings but not in `paseo provider ls`.

This change:

- sends the CLI `appVersion` during daemon connection
- restores custom provider visibility in `paseo provider ls`
- adds a regression test for daemon-backed custom providers
- extracts CLI version resolution into a shared helper reused by both the CLI entrypoint and daemon client

## Test plan

- [x] `npm run build --workspace=@getpaseo/cli`
- [x] `npx tsx packages/cli/tests/15-provider.test.ts`
- [x] `npx tsx packages/cli/tests/33-cli-version.test.ts`


boudra:
thanks for the patch. landed the same fix independently in 9e3d8f0, shipped in v0.1.61. closing as duplicate.

PR #470, opened 2026-04-17
Pass MCP servers to structured helper agents

randyjtw:
## Summary
- structured helper agents used for commit message and PR text generation did not inherit `mcpServers` from the live agent session
- this meant helper runs could lose access to MCP-backed tools even when the primary agent in the same workspace had them available

## Reproduction
1. Start an agent in a workspace with MCP servers configured.
2. Trigger a flow that uses a structured helper, such as commit message generation or PR title/body generation.
3. Observe that the helper session is created without inherited `mcpServers`, so MCP-backed tools are unavailable inside the helper.

## Fix
- resolve the best matching live agent for the same `cwd`
- copy that agent's `mcpServers` into the structured helper `agentConfigOverrides`
- apply the same inheritance for both commit message generation and PR text generation

### DIFF
--- a/packages/server/src/server/session.structured-helper-mcp.test.ts
+++ b/packages/server/src/server/session.structured-helper-mcp.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, test } from "vitest";
+
+import { resolveStructuredHelperMcpServers } from "./session.js";
+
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

PR #503, opened 2026-04-19
Support GitHub SSH host aliases for PR actions

Valian:
## Summary
- support GitHub SSH host aliases by resolving SSH config hostnames locally via `ssh -G`
- centralize GitHub remote parsing/resolution in a new server utility and reuse it across PR status + PR creation paths
- add focused tests for direct remotes, SSH aliases, and workspace GitHub detection

## Problem
Paseo currently only treats remotes that literally look like `github.com` as GitHub remotes.

That breaks setups like:

```bash
git@github-work:owner/repo.git
```

where `github-work` is an SSH alias that resolves to `github.com` via `~/.ssh/config`.

In that setup, Git and `gh` work correctly, but Paseo reports GitHub as disconnected and disables `Create PR`. Even if the UI were bypassed, PR creation would still fail because the server could not derive the GitHub repo from the aliased remote.

## Solution
This change introduces a centralized GitHub remote resolver that:

1. parses standard GitHub remotes directly
2. for SSH remotes with non-literal hosts, resolves the effective SSH hostname with `ssh -G`
3. treats aliases that resolve to `github.com` or `ssh.github.com` as GitHub remotes

That resolver is now used by:
- workspace GitHub feature detection / PR status loading
- GitHub repo resolution for PR creation
- workspace Git metadata's direct GitHub parsing path

This keeps the fast path local for standard remotes and adds local-only SSH alias resolution for alias-based setups, without introducing a `gh repo view` network/API dependency.

## Testing
- `npm exec --workspace=@getpaseo/server -- vitest run src/utils/github-remote.test.ts src/server/workspace-git-service.test.ts src/server/workspace-git-metadata.test.ts --bail=1`

## Notes
I also manually verified the real-world alias case with:

```bash
ssh -G mindnexus.github.com | rg '^hostname '
```

which resolves to `github.com`, and the new resolver correctly maps:

```bash
git@mindnexus.github.com:JakubMindNexus/postline.git
```

to `JakubMindNexus/postline`.

### DIFF
--- a/packages/server/src/server/session.ts
+++ b/packages/server/src/server/session.ts
@@ -5190,7 +5190,6 @@ export class Session {
           base: msg.baseRef,
         },
         this.github,
-        this.workspaceGitService,
       );
       await this.notifyGitMutation(cwd, "create-pr", { invalidateGithub: true });
 
--- a/packages/server/src/server/workspace-git-metadata.ts
+++ b/packages/server/src/server/workspace-git-metadata.ts
@@ -1,4 +1,5 @@
 import { basename } from "path";
+import { parseGitHubRemoteUrl } from "../utils/github-remote.js";
 import { slugify } from "../utils/worktree.js";
 
 export interface WorkspaceGitMetadata {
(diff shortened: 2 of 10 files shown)


Valian:
@boudra would be great if you could look into this, I know you're a victim of a successful project :D but hopefully this change is not controversial. 

It's as easy as defining a custom host in ~/.ssh/config and using it as git remote. Example entry in config:

```
Host example.github.com
    IdentityFile ~/.ssh/id_rsa
    HostName github.com
```

boudra:
thanks @Valian looking into this

boudra:
thanks for this, clean refactor and the tests made it easy to extend. pushed a small follow-up that validates the host before passing it to ssh -G, otherwise an ssh alias starting with `-` would get parsed as an option flag.

Valian:
@boudra thank you !! Keep up great work with Paseo :)

PR #613, opened 2026-04-29
fix(server): preserve fatal log on crash and guard socket.send against close races

### LINKED ISSUE #612: Daemon crashes silently — fatal log lost due to async pino + immediate process.exit in fatal handlers
## Symptom

Daemon process dies abruptly. Clients (mobile app + Electron) see "connection timeout" because port 6767 is no longer listening. `~/.paseo/daemon.log` ends mid-stream with **no** `FATAL` / `uncaughtException` / `unhandledRejection` / shutdown markers, even though those handlers are installed.

Observed last entries from one such crash:
- `1777428402611 warn` — `ProcessTransport is not ready for writing` (Claude SDK, *caught and logged*, not fatal)
- `1777428407969 info` — `relay_data_connected`
- `1777428410118 warn` — `relay_data_disconnected` (code 1001)
- `1777428411576 info` — `ws_runtime_metrics` (1 running agent, 2 inflight requests)
- *log ends here mid-stream*

System was up, no reboot, no jetsam/OOM events, no crash dump in `~/Library/Logs/DiagnosticReports/`.

## Root cause #1 — fatal log lost on async pino

`packages/server/src/server/index.ts:208-216`:

\`\`\`ts
process.on(\"uncaughtException\", (err) => {
  logger.fatal({ err }, \"Uncaught exception — daemon crashing\");
  process.exit(1);                       // ← logs lost: pino is async
});

process.on(\"unhandledRejection\", (reason) => {
  logger.fatal({ err: reason }, \"Unhandled promise rejection — daemon crashing\");
  process.exit(1);                       // ← same problem
});
\`\`\`

Both pino streams are async (`packages/server/src/server/logger.ts:277,281-285`):
- console: `pino.destination({ dest: 1, sync: false })`
- file:    `createRotatingFileStream(...)`

The team already knows this pattern needs special handling — `packages/server/src/server/index.ts:219-225` does it correctly for `main().catch`:

\`\`\`ts
main().catch((err) => {
  process.stderr.write(...);
  // Give pino streams a moment to flush the fatal log entry to daemon.log
  // before the process exits. Without this, async file streams may lose the
  // last few entries that explain why the daemon crashed.
  setTimeout(() => process.exit(1), 200);
});
\`\`\`

…but the `uncaughtException` / `unhandledRejection` handlers don't apply the same delay. So whenever the daemon crashes through those paths, the fatal log entry is queued but the process exits before it's flushed — leaving exactly the silent-crash signature observed.

## Root cause #2 — likely crash trigger: unprotected `socket.send()`

Two locations call `socket.send(...)` without try/catch. `ws` throws synchronously on send-after-close, and the `readyState === 1` check in `sendToClient` has a race window (state can transition between the check and the call).

**`packages/server/src/server/relay-transport.ts:421-441`:**

\`\`\`ts
function createRelayTransportAdapter(socket: WebSocket): RelayTransport {
  const relayTransport: RelayTransport = {
    send: (data) => socket.send(data),     // ← line 423, unprotected
    close: (code?, reason?) => socket.close(code, reason),
    onmessage: null, onclose: null, onerror: null,
  };
  ...
}
\`\`\`

**`packages/server/src/server/websocket-server.ts:740-754`:**

\`\`\`ts
private sendToClient(ws: WebSocketLike, message: WSOutboundMessage): void {
  if (ws.readyState === 1) {
    ws.send(JSON.stringify(message));      // ← line 743, race + unprotected
    this.recordOutboundMessage(message, ws);
  }
}

private sendBinaryToClient(ws: WebSocketLike, frame: Uint8Array): void {
  if (ws.readyState !== 1) return;
  ws.send(frame);                           // ← line 752, race + unprotected
  this.recordOutboundBinaryFrame(ws);
}
\`\`\`

In the observed crash there was an active foreground turn streaming `agent_stream` outputs at the time of termination — exactly the high-traffic state where a peer-side disconnect race is most likely.

## Why this matters

The two issues compound: cause #2 produces an uncaught throw, cause #1 swallows the resulting fatal log. Without #1's fix, every future crash through the process-level error handlers will be invisible.

## Proposed fix

Three small, targeted changes:

1. **`index.ts`** — wrap fatal handlers in a `setTimeout(() => process.exit(1), 200)` (mirroring the existing `main().catch` pattern) so the fatal entry has time to flush.
2. **`relay-transport.ts:423`** — try/catch around `socket.send(data)` in the relay transport adapter; warn-log on failure and let `onclose`/`onerror` handle cleanup.
3. **`websocket-server.ts:740-754`** — try/catch around the two `ws.send(...)` calls; the `readyState === 1` check stays as a fast path.

Happy to send a PR.

## Environment

- Daemon: installed via `@getpaseo/cli` (npm)
- Node: v25.8.2 (Homebrew)
- macOS: Darwin 25.3.0 (Apple Silicon)

yuruiz:
## Summary

- `index.ts` — pull `setTimeout(() => process.exit(1), 200)` into a shared `exitAfterPinoFlush()` and use it from `uncaughtException` / `unhandledRejection` so fatal log entries actually reach `daemon.log` before the process dies (pino is async).
- `relay-transport.ts` — wrap `socket.send(...)` in the relay adapter with try/catch and warn-log; lets `onclose`/`onerror` drive cleanup instead of an uncaughtException.
- `websocket-server.ts` — same guard on `sendToClient` / `sendBinaryToClient`. Keep the `readyState === 1` fast path; it's still useful, but it's a fast path, not a guarantee.

Closes #612.

## Why

Observed a daemon that vanished mid-stream with no fatal entry in the log. Investigation traced it to two compounding bugs:

1. The fatal handlers in `index.ts:208-216` exit immediately after `logger.fatal(...)`, but pino is configured async (`logger.ts:277,281-285` — `sync: false` console + rotating-file-stream). The same code base already worked around this for `main().catch` at line 224 — the process-level handlers just hadn't been updated to match.
2. `socket.send()` and `ws.send()` were unprotected at three call sites. `ws` throws synchronously on send-after-close, and the `readyState === 1` check has a race window. During an active foreground turn streaming `agent_stream` output, a peer-side disconnect could escalate into an uncaughtException — which #1 then ate.

Together: the bug had no diagnostic trail. Fix #1 makes the next crash recoverable; fix #2 prevents the most likely trigger.

## Test plan

- [x] `npm run typecheck` (clean — pre-commit hook ran)
- [x] `npm run lint -- <changed files>` (0 warnings)
- [x] `npm run format:check:files -- <changed files>` (clean)
- [ ] Smoke test on local daemon: start `npm run dev`, force-quit the Electron app while a turn is streaming, confirm daemon stays up and `ws_send_failed` warns appear in `daemon.log`
- [ ] Optionally: deliberately throw from a setImmediate to verify the fatal entry now lands in `daemon.log` before exit

## Not in scope

- Server-side WebSocket ping/pong heartbeats — separate concern.
- The `ProcessTransport is not ready for writing` warning from the Claude SDK observed near the crash — already caught and logged; not the trigger.
- The `read_project_config_request` validation noise from forward-compat clients — harmless, separate issue.

### DIFF
--- a/packages/server/src/server/index.ts
+++ b/packages/server/src/server/index.ts
@@ -207,19 +207,23 @@ async function main() {
 
   process.on("uncaughtException", (err) => {
     logger.fatal({ err }, "Uncaught exception — daemon crashing");
-    process.exit(1);
+    exitAfterPinoFlush();
   });
 
   process.on("unhandledRejection", (reason) => {
     logger.fatal({ err: reason }, "Unhandled promise rejection — daemon crashing");
-    process.exit(1);
+    exitAfterPinoFlush();
   });
+}
+
+// Give pino async streams a moment to flush the fatal log entry to daemon.log
+// before the process exits. Without this, the last few entries that explain
+// why the daemon crashed can be lost.
+function exitAfterPinoFlush(): void {
+  setTimeout(() => process.exit(1), 200);
 }
 
 main().catch((err) => {
   process.stderr.write(`${err instanceof Error ? (err.stack ?? err.message) : String(err)}\n`);
-  // Give pino streams a moment to flush the fatal log entry to daemon.log
-  // before the process exits. Without this, async file streams may lose the
-  // last few entries that explain why the daemon crashed.
-  setTimeout(() => process.exit(1), 200);
+  exitAfterPinoFlush();
 });
--- a/packages/server/src/server/relay-transport.ts
+++ b/packages/server/src/server/relay-transport.ts
@@ -396,7 +396,7 @@ async function attachEncryptedSocket(
   metadata?: ExternalSocketMetadata,
 ): Promise<void> {
   try {
-    const relayTransport = createRelayTransportAdapter(socket);
+    const relayTransport = createRelayTransportAdapter(socket, logger);
     const emitter = new EventEmitter();
     const channel = await createDaemonChannel(relayTransport, daemonKeyPair, {
       onmessage: (data) => emitter.emit("message", data),
@@ -418,9 +418,18 @@ async function attachEncryptedSocket(
   }
 }
 
-function createRelayTransportAdapter(socket: WebSocket): RelayTransport {
+function createRelayTransportAdapter(socket: WebSocket, logger: pino.Logger): RelayTransport {
   const relayTransport: RelayTransport = {
-    send: (data) => socket.send(data),
+    send: (data) => {
+      try {
+        socket.send(data);
+      } catch (err) {
+        // Socket likely transitioned to closed between checks; let onclose/onerror
+        // drive cleanup. Without this guard the synchronous throw would propagate
+        // up as an uncaughtException and take down the daemon.
+        logger.warn({ err }, "relay_socket_send_failed");
+      }
+    },
     close: (code?: number, reason?: string) => socket.close(code, reason),
     onmessage: null,
     onclose: null,
--- a/packages/server/src/server/websocket-server.ts
+++ b/packages/server/src/server/websocket-server.ts
@@ -738,10 +738,17 @@ export class VoiceAssistantWebSocketServer {
   }
 
   private sendToClient(ws: WebSocketLike, message: WSOutboundMessage): void {
-    // WebSocket.OPEN = 1
-    if (ws.readyState === 1) {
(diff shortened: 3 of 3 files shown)


boudra:
good catch

PR #738, opened 2026-05-05
fix(app): add aria-checked to Switch so E2E toBeChecked() works

boudra:
## Summary

- React Native Web doesn't map `accessibilityState.checked` to `aria-checked` for `role="switch"`, so Playwright's `toBeChecked()` found the switch element but always saw it as unchecked (missing attribute).
- `expectDaemonManagementEnabled` in the cancel-path tests always timed out (10 s) because the attribute was never present.
- `expectDaemonManagementDisabled` (`not.toBeChecked()`) always passed for the same reason — making the confirm-path test a false positive.
- Fix: add `aria-checked={value}` directly to the `Pressable` in `switch.tsx`. RN Web passes explicit `aria-*` props through to the DOM.
- Update `switch.test.tsx` mock to accept and forward the explicit `aria-checked` prop, keeping it faithful to the fixed component.

Fixes tests at lines 58, 76, and 126 in `desktop-updates.spec.ts`.

## Test plan

- [ ] `npx vitest run packages/app/src/components/ui/switch.test.tsx` green
- [ ] typecheck + lint + format green (pre-commit hook passed)
- [ ] CI green on this PR

### DIFF
--- a/packages/app/src/components/ui/switch.test.tsx
+++ b/packages/app/src/components/ui/switch.test.tsx
@@ -32,6 +32,7 @@ vi.mock("react-native-reanimated", () => ({
 
 vi.mock("react-native", () => ({
   Pressable: ({
+    "aria-checked": ariaChecked,
     accessibilityLabel,
     accessibilityRole,
     accessibilityState,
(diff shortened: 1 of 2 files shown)


boudra:
Switch aria-checked fix is included in #739 (cherry-picked) and merged. Closing as superseded.

PR #745, opened 2026-05-05
fix: prevent macOS desktop unlock freeze after display sleep

### LINKED ISSUE #741: Desktop UI freeze on macOS unlock after extended display sleep (visibility resume cascade)
## Summary

The Paseo desktop app (Electron, macOS) reliably enters a state where the UI is unresponsive to **both mouse and keyboard input** after the user unlocks the Mac from an extended display-sleep window during which the daemon and Claude Code agents kept producing output in the background.

The Electron processes are healthy throughout the freeze — the bug is a renderer-side resume cascade that saturates the JS main thread.

## Reproduction

1. Have the desktop app open with several active agents (observed with 10 idle agents and ~2000 timeline items at freeze time, but smaller setups likely also reproduce).
2. Lock the Mac / let the display sleep for ≥ 1 minute (the longer, the more reliably it reproduces).
3. While locked, agents continue to run / produce output (or new agents are created).
4. Unlock the Mac.

Result: the Paseo UI does not respond to clicks or keypresses. `Cmd+,`, `Cmd+Q`, etc. all dead. Only Force Quit recovers.

## Diagnostic evidence (collected from a real freeze)

System-wide spindump (`sudo spindump <browserPID> 5`) and per-process `sample` runs showed:

- **Electron browser main thread**: 99.7 % idle in `__CFRunLoopServiceMachPort → mach_msg2_trap` — not blocked.
- **Renderer V8 main thread**: 99.9 % idle in `mach_msg`, but two snapshots 10 minutes apart had different non-idle leaves and the renderer was still sending `client_heartbeat` / `ping` / `schedule/list` to the daemon every 30 s. **Event loop alive.**
- **Daemon**: healthy, all 10 agents in \`idle\` state, \`bufferedAmount.max=0\` for the entire frozen window.
- **\`pmset -g log\`**: only \`Display is turned off/on\` events (no system sleep / Wake from Sleep) — daemon and renderer were never OS-suspended, only window-occluded.
- **\`lsappinfo Paseo\`**: reports \`Foreground\`, no \"not responding\" flag.
- **Other GUI processes** (Spotlight, Ghostty, loginwindow) were also showing momentarily-blocked event threads, but only Paseo stayed wedged after unlock — the system itself recovered.

So: the freeze is **not** an OS event-routing failure, **not** a daemon backlog, **not** an Electron main-process hang. It is the **renderer's JS main thread being saturated by a self-induced cascade triggered on resume**.

## Root cause

\`packages/app/src/contexts/session-context.tsx\` — \`handleAppResumed\` (~line 748):

\`\`\`ts
const handleAppResumed = useCallback(
  (awayMs: number) => {
    scheduleAuthoritativeRevalidation();
    // ...
    if (awayMs < HISTORY_STALE_AFTER_MS) {
      return;
    }
    bumpHistorySyncGeneration(serverId);
  },
  ...
);
\`\`\`

Triggered by \`useClientActivity\` (\`packages/app/src/hooks/use-client-activity.ts:91-117\`) on \`AppState change(\"active\")\` and web \`visibilitychange\`.

When \`awayMs >= HISTORY_STALE_AFTER_MS\` (60 s), \`bumpHistorySyncGeneration(serverId)\` flips a global generation counter consumed by **every mounted \`<AgentPanel>\`** (\`packages/app/src/panels/agent-panel.tsx:751-756\`). Every panel's \`needsAuthoritativeSync\` becomes true on the same tick, so:

1. Each mounted panel calls \`ensureAgentIsInitialized(agentId)\` → fires a \`fetchAgentTimeline\` request.
2. Each response runs \`processTimelineResponse\` → big snapshot commit into \`useSessionStore\`.
3. Each commit re-evaluates 8+ \`useSessionStore\` selectors per panel.
4. Content-size changes drive recursive double-rAF chains in \`packages/app/src/components/use-bottom-anchor-controller.ts:113-138\` and \`packages/app/src/utils/web-focus.ts:38-48\`. While the renderer was occluded, Chromium throttled rAF (~1 Hz) — on resume, deferred rAF closures fire back-to-back and stack on top of the new ones.

Net effect: the JS main thread cannot drain user input events until the cascade settles, so the UI looks frozen.

\`packages/app/src/runtime/host-runtime.ts:1517-1522\` adds a second \`bumpHistorySyncGeneration\` site on every WS reconnect, which can compound the storm if reconnect happens to coincide with unlock.

## Why the pattern matches \"only on long unlocks\"

Switching tabs / brief app blur → \`awayMs < 60 s\` → the early return runs → no \`bumpHistorySyncGeneration\` → no cascade. After ≥ 1 min locked, every panel is forced to re-bootstrap simultaneously — which scales linearly with the number of mounted panels and each agent's timeline size.

## Suggested fix direction

- Make \`bumpHistorySyncGeneration\` lazy: mark history as stale, but defer \`ensureAgentIsInitialized\` until each panel actually re-enters the user's view.
- Coalesce \`handleAppResumed\` so multiple resume signals (e.g. \`visibilitychange\` + \`AppState change\`) within a short window produce a single revalidation.
- Cancel/reset rAF chains in \`use-bottom-anchor-controller\` when the document becomes hidden so they don't accumulate while occluded.

I will be opening a PR implementing the lazy-revalidation approach. Filing this issue first for visibility and to gather feedback on the approach before merging.

## Environment

- Paseo desktop 0.1.67
- macOS 26.3.1 (build 25D2128), Mac mini M1 (8 cores, 16 GB)
- Daemon embedded in desktop bundle

## Artifacts available on request

- 14 MB system-wide spindump captured during freeze
- Per-process \`sample\` outputs (browser / renderer / GPU / daemon)
- Daemon log around the freeze window

yuruiz:
## Summary

Fixes #741.

On macOS, locking the screen until the display sleeps can leave the Paseo
desktop window frozen after unlock — clicks and keypresses do not land —
even though the renderer and every process stay alive. It self-recovers
after a few minutes. Strongest repro: the Mac mini / external-display wake path.

**Root cause:** macOS display sleep can leave Chromium's GPU-process display
link — the vsync source that drives frame production — stuck on a stale
display. The compositor then stops producing frames, so the window looks
frozen while actually being alive. Restarting the GPU process makes Chromium
rebuild the display link, and frames resume.

## Changes

`setupDarwinCompositorWatchdog` (`packages/desktop/src/window/window-manager.ts`)
replaces the previous `setupDarwinPaintRefresh`:

- Polls the renderer for frame production every 2 s (`requestAnimationFrame`,
  bounded by a `setTimeout` so the probe resolves even when frame production
  has stopped).
- After 3 consecutive stalled probes (~6 s) — only while the window is visible,
  not minimized, and the screen is unlocked — restarts the GPU process with
  `SIGKILL`; Chromium relaunches it automatically.
- Bounded by a 60 s cooldown and a 3-attempt cap.

`main.ts` wires it in; `window-manager.test.ts` covers the
`shouldRecoverFromFrameStall` decision logic; `docs/development.md` documents
the gotcha. The previous `setupDarwinPaintRefresh` only nudged the window size /
invalidated the surface, which does not address the dead display link.

## Note on the earlier revision

An earlier version of this PR attributed the freeze to a renderer-side
history-sync cascade and changed five app-layer files. Diagnostic builds showed
the freeze is frame-production death in the GPU process, not renderer
main-thread starvation — so the app-layer changes were dropped and this PR now
carries only the GPU-watchdog fix.

## Verification

- Built the desktop app from this commit and ran it until a real freeze
  occurred. The watchdog detected the frame-production stall, restarted the GPU
  process once (no second attempt needed), and the window recovered — versus
  the ~3-minute unmitigated freeze.
- `npx vitest run packages/desktop/src/window/window-manager.test.ts` — 18 tests pass.
- `npm run typecheck` (all workspaces), `npm run lint`, `npm run format:check` — pass.

## Risk / compatibility

- Desktop-only, gated to macOS (`process.platform === "darwin"`). No WebSocket
  or schema change.
- The watchdog acts only on a sustained stall while the window is visible and
  unlocked; cooldown + attempt cap bound it.
- The probe timer and all `powerMonitor` listeners are removed when the window closes.

## Test plan

- [ ] Mac mini + external display: lock until the display sleeps, unlock,
      interact — the window recovers within seconds instead of staying frozen.
- [ ] MacBook Pro: normal lock/unlock — no regression.
- [ ] Normal use across display-sleep cycles — confirm no spurious GPU restarts
      when there is no real freeze.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -43,6 +43,21 @@ In any worktree-style or portless setup, never assume default ports.
 `http://127.0.0.1:9223` so renderer CPU profiles can be captured through CDP.
 Override the port with `PASEO_ELECTRON_REMOTE_DEBUGGING_PORT` when `9223` is busy.
 
+### Desktop macOS compositor watchdog
+
+macOS display sleep can leave Chromium's GPU-process display link — the vsync
+source that drives frame production — stuck on a stale display. The compositor
+then stops producing frames and the window looks frozen: unresponsive to clicks
+and keys even though the renderer and every process stay alive. It self-recovers
+after a few minutes, which is too long for a foreground app.
+
+`setupDarwinCompositorWatchdog` (`packages/desktop/src/window/window-manager.ts`)
+guards against this. It polls the renderer for frame production every couple of
+seconds and, after a sustained stall while the window is visible and unlocked,
+restarts the GPU process so Chromium rebuilds the display link. The probe is
+skipped while the screen is locked or the window is hidden or minimized, since a
+window legitimately stops producing frames then.
+
 ### Daemon logs
 
 Check `$PASEO_HOME/daemon.log` for daemon logs. The default level is `info`; set
--- a/docs/development.md
+++ b/docs/development.md
@@ -51,12 +51,13 @@ then stops producing frames and the window looks frozen: unresponsive to clicks
 and keys even though the renderer and every process stay alive. It self-recovers
 after a few minutes, which is too long for a foreground app.
 
-`setupDarwinCompositorWatchdog` (`packages/desktop/src/window/window-manager.ts`)
-guards against this. It polls the renderer for frame production every couple of
-seconds and, after a sustained stall while the window is visible and unlocked,
-restarts the GPU process so Chromium rebuilds the display link. The probe is
-skipped while the screen is locked or the window is hidden or minimized, since a
-window legitimately stops producing frames then.
+`setupDarwinCompositorWatchdog`
+(`packages/desktop/src/window/compositor-watchdog/index.ts`) guards against
+this. It polls the renderer for frame production every couple of seconds and,
+after a sustained stall while the window is visible and unlocked, restarts the
+GPU process so Chromium rebuilds the display link. The probe is skipped while
+the screen is locked or the window is hidden or minimized, since a window
+legitimately stops producing frames then.
 
 ### Daemon logs
 
--- a/packages/desktop/src/main.ts
+++ b/packages/desktop/src/main.ts
@@ -19,7 +19,7 @@ import {
   getMainWindowChromeOptions,
   getWindowBackgroundColor,
   resolveSystemWindowTheme,
-  setupDarwinPaintRefresh,
+  setupDarwinCompositorWatchdog,
   setupWindowResizeEvents,
   setupDefaultContextMenu,
   setupDragDropPrevention,
@@ -398,7 +398,7 @@ async function createMainWindow(): Promise<void> {
     app.dock?.setBadge(devWorktreeName);
   }
 
-  setupDarwinPaintRefresh(mainWindow);
+  setupDarwinCompositorWatchdog(mainWindow);
   setupWindowResizeEvents(mainWindow);
   setupDefaultContextMenu(mainWindow);
   setupDragDropPrevention(mainWindow);
--- a/packages/desktop/src/main.ts
+++ b/packages/desktop/src/main.ts
@@ -19,12 +19,12 @@ import {
   getMainWindowChromeOptions,
   getWindowBackgroundColor,
   resolveSystemWindowTheme,
-  setupDarwinCompositorWatchdog,
   setupWindowResizeEvents,
   setupDefaultContextMenu,
   setupDragDropPrevention,
   buildStandardContextMenuItems,
 } from "./window/window-manager.js";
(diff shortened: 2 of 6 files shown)


boudra:
hey @yuruiz apologies for the force push on main, i was going to rebase this but i see that you already merged. will look into your PRs in a moment.

PR #781, opened 2026-05-06
Steer active agent turns from follow-up messages

boudra:
### Linked issue

None

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Follow-up messages sent while an agent is already running now steer the active turn when the provider supports it, instead of always replacing the run. Claude follow-ups are pushed into the existing input stream with next priority, Codex app-server follow-ups call the turn steering endpoint, and unsupported providers keep the existing replace behavior.

The shared send path now centralizes out-of-band command handling, unarchive/load behavior, user-message recording, and run startup so MCP, CLI, and websocket sends use the same policy. The app also marks local steering messages with a small “Steering conversation” hint while the daemon catches up.

### How did you verify it

- `npm install`
- `npm run build:daemon`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

CI is being watched by the existing babysit schedule for PR #781.

Extra review surface: this touches active-run concurrency and the client/server stream shape. The new message field is optional for old clients, but it is worth smoke-checking one active Claude or Codex follow-up from the app before merge.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/agent-stream/view.tsx
+++ b/packages/app/src/agent-stream/view.tsx
@@ -423,6 +423,7 @@ const AgentStreamViewComponent = forwardRef<AgentStreamViewHandle, AgentStreamVi
             client={client}
             isFirstInGroup={layoutItem.isFirstInUserGroup}
             isLastInGroup={layoutItem.isLastInUserGroup}
+            deliveryHint={item.deliveryHint}
           />
         );
       },
--- a/packages/app/src/components/message.tsx
+++ b/packages/app/src/components/message.tsx
@@ -63,7 +63,11 @@ import Animated, {
 import Svg, { Defs, LinearGradient as SvgLinearGradient, Rect, Stop } from "react-native-svg";
 import { createMarkdownStyles } from "@/styles/markdown-styles";
 import { Fonts } from "@/constants/theme";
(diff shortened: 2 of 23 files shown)


alpc91:
Would love to see this merged soon if the remaining blockers are manageable. Real-time steering is a pretty important workflow for Paseo: when an agent is already running, being able to send a follow-up that steers the active turn instead of replacing/restarting it makes long-running remote/mobile sessions much more usable.

This is especially useful for correcting direction from a phone, adding constraints after seeing partial output, or stopping an agent from going too far down the wrong path without losing the current turn. Happy to test this on the latest desktop/mobile builds once there is a build available.

Sumkor:
This steering capability is extremely important.

I recently ran into a case where an agent was working on a long-running task, and I realized midway through that I needed to correct its direction. Since steering was unavailable, I used the interrupt feature instead. Context compaction then occurred, and after compaction, Paseo continued from the interruption message while completely losing sight of the original long-running task.

True mid-turn steering would have allowed me to correct the agent without terminating the active turn or risking the loss of the original task context. I would really appreciate seeing this PR move forward.

PR #805, opened 2026-05-07
Fix child workspace resolution for created agents

CitrusIce:
## Summary
- stop `create_agent_request` from reusing a parent directory workspace when the requested cwd is itself a distinct workspace root
- add a regression test covering a nested git project under an active parent directory workspace
- preserve the existing parent workspace while creating the child workspace on demand

## Testing
- `npm run build:daemon`
- `npx vitest run packages/server/src/server/session.workspaces.test.ts --bail=1`
- `npm run typecheck:daemon`
- `npm run lint -- packages/server/src/server/session.ts packages/server/src/server/session.workspaces.test.ts`

### DIFF
--- a/packages/server/src/server/session.ts
+++ b/packages/server/src/server/session.ts
@@ -3000,10 +3000,18 @@ export class Session {
         worktreeName,
         firstAgentContext,
       );
-      let resolvedWorkspace = msg.workspaceId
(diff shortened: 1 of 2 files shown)


boudra:
removed the behaviour entirely https://github.com/getpaseo/paseo/pull/808

PR #827, opened 2026-05-09
feat(codex): archive native Codex thread when archiving agent

### LINKED ISSUE #822: Feat: Archive native Codex thread when archiving a Codex agent from Paseo
## Problem

When I close/archive a Codex-backed agent from Paseo, Paseo correctly removes it from the active agent list and marks the Paseo agent record as archived.

However, the underlying Codex thread does not appear to be archived in Codex itself. When I later open the Codex app, I still see many sessions that I had already closed from Paseo. After using Paseo heavily, this makes the Codex app session list noisy and hard to use.

## Expected behavior

Archiving or closing a Codex agent in Paseo should also archive the native Codex thread, so the same session is treated as archived by the Codex app.

## Actual behavior

Paseo archives the local Paseo agent record, but the corresponding Codex thread still appears as an active/unarchived session in the Codex app.

## Reproduction

1. Start a Codex agent from Paseo.
2. Send at least one message so a Codex thread is created.
3. Close/archive the agent from Paseo.
4. Open the Codex app.
5. Observe that the corresponding session still appears in the normal session list instead of being archived.

## Suggested direction

Codex app-server appears to expose a native archive API:

- `thread/archive`
- params: `{ threadId }`

Paseo already stores the Codex thread id in the agent persistence handle (`persistence.sessionId` / `nativeHandle`), so a narrow fix might be:

- Add a provider-level archive hook, e.g. `archivePersistedSession(handle)`.
- Implement it for the Codex provider by calling `thread/archive`.
- Call that hook when Paseo archives a live agent and when it archives a stored Codex agent.
- Treat native archive failures as best-effort: log a warning but still archive the Paseo agent record, so the local close/archive flow does not get blocked.

## Scope

This issue is only about keeping Paseo's archive action in sync with Codex's native archive state. It is not proposing hard-deleting Codex sessions or changing the general archive UI.

If this direction looks acceptable, I can prepare a small PR that only adds Codex native thread archiving during Paseo agent archive.

32r4:
## Summary
- Archive the native Codex thread when a Codex agent is archived from Paseo
- Calls `thread/archive` JSON-RPC method (best-effort, failures don't block)
- Covers both live agent archive and stored agent archive paths

Closes #822

## Test plan
- [x] Archive a Codex agent from Paseo, verify thread disappears from Codex app session list
- [ ] Verify archive still works when Codex app-server is unavailable (best-effort)
- [ ] CI passes typecheck + lint

### DIFF
--- a/packages/server/src/server/agent/agent-manager.ts
+++ b/packages/server/src/server/agent/agent-manager.ts
@@ -975,6 +975,9 @@ export class AgentManager {
       attentionReason: null,
       attentionTimestamp: null,
     });
+
+    await this.archiveNativeSessionBestEffort(agent.provider, stored.persistence);
+
     this.notifyAgentState(agentId);
     await this.closeAgent(agentId);
 
@@ -1118,6 +1121,9 @@ export class AgentManager {
       attentionTimestamp: null,
     };
     await registry.upsert(nextRecord);
+
+    await this.archiveNativeSessionBestEffort(record.provider, record.persistence);
+
     return nextRecord;
   }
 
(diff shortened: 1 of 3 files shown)


boudra:
thanks

PR #888, opened 2026-05-10
Extract permission response command

boudra:
## Summary
- extract shared permission-response orchestration into an agent-domain command
- route both WebSocket Session and MCP respond_to_permission through the same command
- preserve follow-up prompt handling so MCP now starts the same follow-up run as Session

## Verification
- npx vitest run packages/server/src/server/agent/permission-response.test.ts --bail=1
- npx vitest run packages/server/src/server/agent/mcp-server.test.ts --bail=1
- npm run format
- npm run typecheck
- npm run lint


boudra:
Closing in favor of the single integration PR #893, which contains the full Session/MCP command-boundary stack and is being held until after the stable release.

PR #895, opened 2026-05-10
fix(server): recover completed opencode turns after SSE EOF

### LINKED ISSUE #861: [System Error] OpenCode event stream ended before the turn reached a terminal state
Image

I updated to the latest Paseo version v0.1.71 and title is the only message I can see in chat. OpenCode works just fine. Other agents with Paseo work just fine.

> OpenCode
>   Binary: C:\Users\%USERNAME%\.bun\bin\opencode.EXE
>   Version: 1.14.45

atomlink-ye:
## Summary
- recover OpenCode turns from persisted completion markers when the SSE stream closes before Paseo observes a terminal idle event
- scope recovery to the current foreground turn so older completed assistant messages cannot be reused by mistake
- preserve recovered usage on the terminal `turn_completed` event and cover the regression with focused unit tests

## Verification
- `npx vitest run packages/server/src/server/agent/providers/opencode-agent.test.ts --bail=1 -t "OpenCode adapter startTurn error handling"`
- `npm run format`
- `npm run build:daemon`
- `npm run typecheck`
- `npm run lint`

Closes #861

### DIFF
--- a/packages/server/src/server/agent/providers/opencode-agent.test.ts
+++ b/packages/server/src/server/agent/providers/opencode-agent.test.ts
@@ -566,6 +566,321 @@ describe("OpenCode adapter context-window normalization", () => {
 });
 
 describe("OpenCode adapter startTurn error handling", () => {
+  test("recovers SSE EOF into turn_completed when persisted assistant completion exists", async () => {
+    const storageRoot = mkdtempSync(path.join(os.tmpdir(), "opencode-storage-"));
+    const cwd = "/tmp/test";
+    const dateNowSpy = vi.spyOn(Date, "now").mockReturnValue(2000);
+
+    writeOpenCodeJson(storageRoot, "session/project-1/ses_unit_test.json", {
+      id: "ses_unit_test",
+      directory: cwd,
+      time: { created: 1000, updated: 3000 },
+    });
+    let releaseStream!: () => void;
+    const streamMayEnd = new Promise<void>((resolve) => {
+      releaseStream = resolve;
+    });
+
+    const fakeClient = {
+      event: {
+        subscribe: vi.fn().mockResolvedValue({
+          stream: {
+            [Symbol.asyncIterator]: () => {
+              let index = 0;
+              const events: OpenCodeEvent[] = [
+                {
+                  type: "message.updated",
+                  properties: {
+                    info: {
+                      id: "msg_assistant",
+                      sessionID: "ses_unit_test",
+                      role: "assistant",
+                    },
+                  },
+                } as OpenCodeEvent,
+                {
+                  type: "message.part.delta",
+                  properties: {
+                    sessionID: "ses_unit_test",
+                    messageID: "msg_assistant",
+                    partID: "prt_text",
+                    field: "text",
+                    delta: "Recovered ",
+                  },
+                } as OpenCodeEvent,
+              ];
+
+              return {
+                next: async () => {
+                  if (index < events.length) {
+                    return { done: false, value: events[index++] };
+                  }
+                  await streamMayEnd;
+                  return { done: true, value: undefined };
+                },
+              };
+            },
+          },
+        }),
+      },
+      provider: {
+        list: vi.fn().mockResolvedValue({ data: { connected: [], all: [] }, error: undefined }),
+      },
+      session: {
+        create: vi.fn().mockResolvedValue({ data: { id: "ses_unit_test" }, error: undefined }),
+        promptAsync: vi.fn().mockImplementation(async () => {
+          writeOpenCodeJson(storageRoot, "message/ses_unit_test/msg_assistant.json", {
+            id: "msg_assistant",
+            sessionID: "ses_unit_test",
+            role: "assistant",
+            finish: "stop",
+            time: { created: 2000, completed: 2500 },
+          });
+          writeOpenCodeJson(storageRoot, "part/msg_assistant/prt_text.json", {
+            id: "prt_text",
+            sessionID: "ses_unit_test",
+            messageID: "msg_assistant",
+            type: "text",
+            text: "Recovered assistant reply",
+            time: { start: 2100, end: 2400 },
+          });
+          writeOpenCodeJson(storageRoot, "part/msg_assistant/prt_finish.json", {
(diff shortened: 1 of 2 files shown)


atomlink-ye:
A bit more detail on **why this started happening after newer OpenCode**, and why the fix is written this way.

## What changed on the OpenCode side

I could not point this to a specific upstream source commit because the distributed OpenCode package is mostly a binary wrapper, but the **observable contract change** is clear from repros:

- with OpenCode **1.14.39**, the same `paseo run --provider opencode/openai/gpt-5.4 ...` repro completed normally
- with OpenCode **1.14.46**, the underlying OpenCode session still completed successfully, but Paseo sometimes saw the SSE stream end **before** it observed the session-level terminal signal it was waiting for

In other words, from Paseo's point of view, newer OpenCode can now do this ordering:

1. persist the final assistant reply
2. persist completion markers such as `finish: "stop"`, `step-finish`, and completed assistant message state
3. end the SSE stream
4. only later expose `session.idle` / idle status, or expose it outside the timing assumptions the adapter previously depended on

That timing difference is enough to break the older adapter logic.

## Why the old adapter failed

The previous implementation treated **session-level idle** as the authoritative terminal signal.

So the old logic was effectively:

- stream events until we see `turn_completed`
- map `session.idle` / `status.type === "idle"` to `turn_completed`
- if the stream ends first, fail the turn with
  `OpenCode event stream ended before the turn reached a terminal state`

That was safe as long as OpenCode always delivered terminal idle before stream EOF.
It is **not** safe once final reply persistence can happen before idle becomes visible to Paseo.

## What this patch changes

This patch keeps the old fast path intact:

- if Paseo sees the normal terminal idle event, it still completes immediately the old way

It only adds an **EOF recovery path**:

- if SSE ends first
- read persisted OpenCode state for the **current foreground turn only**
- if that current turn already has strong completion evidence
  - completed assistant text
  - `finish: "stop"`
  - `step-finish`
  - recovered usage
- then finish as `turn_completed` instead of hard-failing

It also explicitly avoids a false positive where an older completed assistant message from a previous turn could be reused for the current turn.

## Sequence diagram

```mermaid
sequenceDiagram
    participant P as Paseo adapter
    participant S as OpenCode SSE
    participant D as OpenCode persisted state

    Note over P,S: Old flow
    P->>S: subscribe
    S-->>P: assistant deltas and tool events
    S-->>P: session.idle
    P-->>P: complete turn

    Note over P,D: Newer OpenCode flow
    P->>S: subscribe
    S-->>P: partial assistant output
    S->>D: persist final assistant text
    S->>D: persist finish stop and step finish
    S-->>P: stream EOF
    alt old adapter
        P-->>P: fail turn
    else patched adapter
        P->>D: read current turn state
        D-->>P: completed reply markers
        P-->>P: complete turn
    end
```

## Why this is compatible with older OpenCode

This patch does **not** replace the original `session.idle` completion path.
It only adds a fallback when EOF happens first.

So on older OpenCode versions that still deliver idle in the old order, behavior stays the same.

atomlink-ye:
Follow-up with a more concrete upstream mapping.

I dug further into the OpenCode side and the smallest upstream regression window is **not** `1.14.39 -> 1.14.46` in general — it is **`1.14.39 -> 1.14.42`**.

## Upstream mapping

- npm package: `opencode-ai`
- upstream repo: `anomalyco/opencode`
- `opencode-ai@1.14.39` -> tag `v1.14.39` -> commit `d35cdffd2e6a3c3d7214f25e349ab399541d1de5`
- `opencode-ai@1.14.42` -> tag `v1.14.42` -> commit `23c6b69abf35314781b9a55a6530527094075505`
- `opencode-ai@1.14.46` -> tag `v1.14.46` -> commit `d802b0a277f4e7f113b5efd8d5446fc1db22f4a4`

## Most likely upstream change that exposed this

The strongest candidate is the serve/backend refactor in the `v1.14.39..v1.14.42` window, especially:

- commit `28b03595b` — `research: delete Hono backend (do not merge) (#25667)`
- follow-up cleanup commit `b24a4e897` — `chore(server): clean up post-Hono-deletion scar tissue (#26542)`

The important part is that this changed the **default server + SSE implementation**:

### Before (`v1.14.39`)
- `packages/opencode/src/server/backend.ts`
  - default backend selection returned **Hono** unless experimental HTTP API was explicitly enabled
- `packages/opencode/src/server/server.ts`
  - `Server.Default()` could still route to the legacy Hono implementation
- `packages/opencode/src/server/routes/instance/event.ts`
  - legacy Hono `/event` SSE route

### After (`v1.14.42`)
- `packages/opencode/src/server/backend.ts`
  - removed
- `packages/opencode/src/server/server.ts`
  - `Server.Default()` always uses the Effect HTTP API path
- `packages/opencode/src/server/routes/instance/event.ts`
  - deleted
- `packages/opencode/src/server/routes/instance/httpapi/event.ts`
  - became the only `/event` SSE implementation
- `packages/opencode/src/server/routes/instance/httpapi/server.ts`
  - new HTTP API stack/layers around the handler
- `packages/opencode/src/server/routes/instance/httpapi/lifecycle.ts`
  - response-disposal path in the new stack

## Why that matters

I did **not** find evidence that OpenCode intentionally changed the meaning of terminal session state:

- `packages/opencode/src/session/run-state.ts` still transitions the session to idle
- `packages/opencode/src/session/status.ts` still publishes both `session.status` and deprecated `session.idle`

So this does **not** look like an intentional contract change such as “idle is no longer emitted”.

What it *does* look like is a **tail-delivery / EOF timing regression** caused by the server/SSE implementation swap:

- final assistant reply is persisted
- completion markers such as `finish: "stop"` / `step-finish` are persisted
- SSE stream can terminate
- external client may fail to observe the idle event it was previously relying on

That lines up with the live repro here:

- `1.14.39` works
- `1.14.42+` fails in the server-backed integration path
- direct `opencode run ...` still works

So my current best read is:

> this is most likely an upstream **regression introduced by the default Hono -> Effect HTTP API backend/SSE migration**, rather than a deliberate OpenCode terminal-state contract change.

That is also why this Paseo patch keeps the original idle fast-path and only adds an EOF recovery fallback.

PR #905, opened 2026-05-10
fix (app): compact composer web interactions

nikuscs:
## Summary & Features

-  **Mobile Web View**:  icons were overflowing, now you get a little 3 dots icon at the right that opens a preferences pane, making more suitable for smaller screens without sacrificing any option.
- **Mobile Web View**: Model + Reasoning labels are hidden in mobile to save space.
- **Accessibility/UX**: Lock compact web viewport zoom ( Note that this is a bit anti-accessibility, take it with a grain of salt , but when we were type screen on iphone was zooming in due to font size ) 
-  **Accessibility/UX**: Sidebar open gestures now is triggered on the edges only, this was breaking the text selection on mobile ( on iphone you couldnt select text as it would trigger sidebars ), now should trigger near the edges only
- **Accessibility/UX**: DND workspaces/worktrees are now required to be pressed for a few seconds and then they can be moved/adjusted, on mobile a simple slide from left to right was triggering DND ( or even selecting ). Ideially we should have a little option at the top to toggle ordering, but this should be ok for now.
- **Accessibility/UX Mobile**: Return ( or Enter ) in mobile should not trigger anymore sending prompt, this is made so in mobile ( like Claude and other apps ) its required to press the send button so we can give paragraphs with enter instead ( expected UX ) 

## Screenshots

CleanShot 2026-05-10 at 17 32 18@2x
CleanShot 2026-05-10 at 17 33 00@2x

## Verification
- `npm run lint`
- `npm run typecheck`
- `npx vitest run src/components/draggable-list.web.test.tsx --bail=1`
- pre-commit hook: format check, lint, typecheck
- Manual compact web preview on `http://localhost:8001`


boudra:
merged partially in https://github.com/getpaseo/paseo/pull/1048

PR #919, opened 2026-05-11
Fix iPad hardware Enter submit in composer

### LINKED ISSUE #810: iPad hardware keyboard: Enter should send, while soft keyboard Return inserts newline
## Problem

On iPadOS, the native app currently treats the Return/Enter key in the composer as a newline. That is right for the on-screen keyboard, but feels wrong with a Magic Keyboard or other external keyboard: the iPad behaves like a small laptop, while sending a message still requires tapping the send button.

Desktop/web already supports the expected keyboard flow: Enter sends, Shift+Enter inserts a newline.

## Desired behavior

When using an external keyboard on iPadOS:

- Enter sends the message.
- Shift+Enter inserts a newline.
- Existing queue/send semantics continue to apply if an agent is running.

When using the iPad on-screen keyboard:

- Return continues to insert a newline.

IME behavior should remain safe:

- Pressing Enter to confirm/dismiss CJK/IME composition must not send prematurely.
- This is related to, but distinct from, #264 and #314, which track Enter sending too eagerly during IME composition.

## Implementation notes / constraints

The current app gates desktop-style composer Enter handling to web/Electron only.

From a quick code read:

- `packages/app/src/constants/platform.ts` defines `isWeb` as `Platform.OS === "web"`, so Electron desktop follows the web path.
- `packages/app/src/components/message-input.tsx` currently sets `shouldHandleDesktopSubmit = isWeb`.
- React Native's JS `TextInput.onKeyPress` event appears to expose only `nativeEvent.key`, which may not reliably distinguish hardware-keyboard Return from soft-keyboard Return.

Because of that, the robust fix may need a small iOS-native hook or hardware-keyboard-specific path rather than simply enabling the existing web key handler for all native iOS text inputs. A JS-only setting such as "Enter sends on iPad" could be simpler, but it risks turning the on-screen keyboard Return into send unless hardware vs software input can be detected reliably.

## Acceptance criteria

- On iPadOS with an external keyboard, Enter sends the composer message.
- On iPadOS with an external keyboard, Shift+Enter inserts a newline.
- On iPadOS with the on-screen keyboard, Return still inserts a newline.
- CJK/IME composition confirmation does not send prematurely.
- Existing desktop/web Enter behavior is unchanged.
- The change is covered by targeted tests where practical, plus iPad simulator/manual verification notes.

kongjiadongyuan:
## Summary

Fixes #810.

This adds a narrow iOS-only Expo module for iPad hardware keyboard submit behavior in the composer:

- bare hardware Return emits a submit event while the composer is focused and the default send action is enabled
- Shift+Return is not registered natively, so it continues through the normal multiline TextInput path as a newline
- the on-screen keyboard Return remains untouched, because the implementation uses `UIKeyCommand` rather than JS `TextInput.onKeyPress`
- IME composition is guarded by checking the current `UITextInput.markedTextRange` before emitting submit
- iPhone is left unchanged; the key command is only registered when `UIDevice.current.userInterfaceIdiom == .pad`

The JS side keeps the existing composer send/queue semantics by calling the same default send action used by the send button.

## Testing

- `npx vitest run packages/app/src/hooks/use-ios-hardware-keyboard-submit.test.tsx --bail=1`
- `npm run lint -- packages/app/src/components/message-input.tsx packages/app/src/hooks/use-ios-hardware-keyboard-submit.ts packages/app/src/hooks/use-ios-hardware-keyboard-submit.test.tsx packages/app/src/native/ios-hardware-keyboard-submit.ts packages/app/src/native/ios-hardware-keyboard-submit.ios.ts`
- pre-commit hook: `npm run lint`, `npm run format:check:files`, and full workspace `npm run typecheck --workspaces --if-present`
- `APP_VARIANT=development npx expo prebuild --platform ios --non-interactive --no-install`
- `pod install` from the generated `packages/app/ios` directory; confirmed `PaseoHardwareKeyboard (0.1.0)` is installed and listed in ExpoModulesProvider
- `xcodebuild -workspace packages/app/ios/PaseoDebug.xcworkspace -scheme PaseoDebug -configuration Debug -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPad Pro 11-inch (M5),OS=latest' -derivedDataPath /tmp/paseo-ios-hardware-keyboard-derived build` -> `BUILD SUCCEEDED`

Manual note: I could validate the native module is autolinked and the iPad simulator build succeeds, but I did not have a physical iPad + external keyboard available to perform a real hardware keypress check.

### DIFF
--- a/packages/app/modules/paseo-hardware-keyboard/expo-module.config.json
+++ b/packages/app/modules/paseo-hardware-keyboard/expo-module.config.json
@@ -1,0 +1,7 @@
+{
+  "platforms": ["ios"],
+  "ios": {
+    "modules": ["PaseoHardwareKeyboardModule"],
+    "reactDelegateHandlers": ["PaseoHardwareKeyboardReactDelegateHandler"]
+  }
+}
--- a/packages/app/modules/paseo-hardware-keyboard/ios/PaseoHardwareKeyboard.podspec
+++ b/packages/app/modules/paseo-hardware-keyboard/ios/PaseoHardwareKeyboard.podspec
@@ -1,0 +1,24 @@
+require 'json'
+
+Pod::Spec.new do |s|
+  s.name           = 'PaseoHardwareKeyboard'
+  s.version        = '0.1.0'
+  s.summary        = 'Hardware keyboard shortcuts for Paseo'
+  s.description    = 'Hardware keyboard shortcuts for Paseo'
+  s.license        = 'AGPL-3.0-or-later'
+  s.author         = 'Paseo'
+  s.homepage       = 'https://paseo.sh'
+  s.platforms      = { :ios => '13.4' }
+  s.swift_version  = '5.4'
+  s.source         = { :path => '.' }
+  s.static_framework = true
+
+  s.dependency 'ExpoModulesCore'
+
+  s.pod_target_xcconfig = {
+    'DEFINES_MODULE' => 'YES',
+    'SWIFT_COMPILATION_MODE' => 'wholemodule'
+  }
+
+  s.source_files = "**/*.{h,m,swift}"
+end
--- a/packages/app/modules/paseo-hardware-keyboard/ios/PaseoHardwareKeyboardModule.swift
+++ b/packages/app/modules/paseo-hardware-keyboard/ios/PaseoHardwareKeyboardModule.swift
@@ -1,0 +1,99 @@
+import ExpoModulesCore
+import UIKit
+
+private let hardwareSubmitEventName = "onHardwareKeyboardSubmit"
+
+private weak var activeModule: PaseoHardwareKeyboardModule?
+private var isHardwareSubmitEnabled = false
+
+@objc
+public class PaseoHardwareKeyboardReactDelegateHandler: ExpoReactDelegateHandler {
+  public override func createRootViewController() -> UIViewController? {
+    return PaseoHardwareKeyboardRootViewController()
+  }
+}
+
+public class PaseoHardwareKeyboardModule: Module {
+  public func definition() -> ModuleDefinition {
+    Name("PaseoHardwareKeyboard")
+
+    Events(hardwareSubmitEventName)
+
+    OnCreate {
+      activeModule = self
+    }
+
+    Function("setHardwareKeyboardSubmitEnabled") { (enabled: Bool) in
+      DispatchQueue.main.async {
+        isHardwareSubmitEnabled = enabled
+      }
+    }
+
+    OnDestroy {
+      if activeModule === self {
+        activeModule = nil
+      }
+      isHardwareSubmitEnabled = false
+    }
+  }
+
+  fileprivate func emitHardwareKeyboardSubmit() {
+    sendEvent(hardwareSubmitEventName, [:])
+  }
+}
+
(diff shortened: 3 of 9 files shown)


nikuscs:
Should be solved here: https://github.com/getpaseo/paseo/pull/905 if this lands merged, along side with other UI / UX polishes

Edit: partially, maybe this can be merged , i think i just did for the web bits.

kongjiadongyuan:
Thanks! I think #905 and this PR cover adjacent but different surfaces.

#905 seems to address compact/mobile web behavior. This PR is specifically for native iPadOS with an external hardware keyboard, where bare Enter should submit while Shift+Enter and the soft keyboard Return should still insert a newline. That split needs the iOS key-command path because RN TextInput events do not reliably distinguish hardware Return from the software keyboard Return on iPadOS.

Keeping this as draft until I finish testing on a real iPad + external keyboard.

kongjiadongyuan:
Manual real-device verification completed on a physical iPad with an external hardware keyboard. The composer behavior felt correct in use, including the core Enter submit / Shift+Enter newline path, and no issues were found during the check. Marking this ready for maintainer review.

boudra:
wow seems crazy that you'd need to reach into native code to do this, did you exhaust all other possibilities? not against having native modules (we already have some), just trying to keep them at a minimum

Plus if you did a RN-level solution this would work with Android too

kongjiadongyuan:
Thanks, this concern is fair.

I also wanted this to be RN-level if possible. I checked the native `TextInput` paths, but I don’t think RN exposes enough information to implement this exact behavior cleanly.

The closest RN-only option is `submitBehavior="submit"` / `onSubmitEditing`. That can make Return submit, but for a multiline native input it treats the Return key as one submit path. It does not distinguish hardware Return from the software keyboard Return, so it would also make the iPad software keyboard Return submit instead of inserting a newline.

The other option is `onKeyPress`, but native RN only gives the key text. It does not expose modifier/source information like “this came from a hardware keyboard” or Shift state in the way the web `KeyboardEvent` path does. On Android, RN’s own TextInput docs/source also note that `onKeyPress` handles soft keyboard input, not hardware keyboard input.

So RN-only seems to support a coarser behavior, but not the behavior from #810:
- hardware Enter submits
- Shift+Enter inserts newline
- software keyboard Return inserts newline
- IME composition should not submit

I intentionally opened this PR narrowly against #810 because I’m new to the project and didn’t want to start with a broad native shortcut layer PR without maintainer feedback. I also opened #920 for the broader direction: iPad hardware keyboard global shortcuts. My hope was to fix this concrete composer issue first, then discuss whether the broader hardware-key bridge is a direction you’d want.

That said, I agree the current native module is probably too submit-specific. I’m happy to reshape it into the smallest version of that bridge:
- native only captures/emits hardware key commands
- RN/JS keeps the key mapping and action routing
- this PR only registers the iPad hardware Enter command for #810
- broader iPad hardware shortcuts and Android support stay out of scope here and can continue under #920

Would that shape address your concern better?

boudra:
makes sense, RN can't disambiguate hardware vs soft return on ios.

i'd rather keep this narrow than reshape into a generic bridge, that's more abstraction for one use case. if #920 lands later, extract then.

thanks, going to merge.

PR #922, opened 2026-05-11
Fix iPad Settings sidebar safe area

### LINKED ISSUE #921: bug: Settings sidebar Back header overlaps the iPadOS status bar
## Problem

On a physical iPad in the wider/tablet layout, the Settings sidebar `Back` header renders underneath the iPadOS status bar instead of below the top safe area.

The result is that the iPadOS status text overlaps the `Back` icon/label, and the control becomes hard or impossible to tap reliably.

## Environment

Observed on a physical iPad running the native iPadOS app:

- iPad Pro 13-inch
- iPadOS 26.4.2
- Landscape/wide tablet layout
- External keyboard was attached during testing, but this appears to be a layout/safe-area issue, not a keyboard issue

## Repro

1. Open Paseo on iPadOS in the wider/tablet layout.
2. Navigate into Settings.
3. Look at the top-left Settings sidebar `Back` row.

## Actual behavior

The `Back` row starts at the very top of the screen and overlaps the iPadOS status bar area.

Reporter screenshot observation:

- the status bar text (`8:56`, date) is drawn over the same vertical area as the `Back` icon/label
- the `Back` affordance appears partially under the status bar rather than below it
- the `General` row starts below a separator, so the issue is specifically the top sidebar header/safe-area placement

## Expected behavior

The Settings sidebar header should reserve the native top safe area on iPadOS, so the `Back` row sits below the iPadOS status bar and remains fully tappable.

## Initial code-level suspicion

This looks specific to the desktop-style Settings sidebar path rather than the shared screen header:

- `ScreenHeader` / `BackHeader` already use `useSafeAreaInsets().top` and appear designed to reserve native top safe area.
- `LeftSidebar` also appears to account for native top insets in its desktop sidebar style.
- `SettingsSidebar` in desktop layout appears to use `useWindowControlsPadding("sidebar")`, which is intended for Electron window controls. On native iPadOS that top padding is zero.

So the likely fix is to make the desktop-style Settings sidebar include native `useSafeAreaInsets().top` on iPadOS, matching the treatment used by other native sidebar/header shells, while preserving Electron window-control padding behavior.

## Related context

This was noticed during real-device testing for iPad hardware keyboard work, but it appears to be an independent native iPad layout bug rather than a regression from the keyboard change.

kongjiadongyuan:
## Summary

Fixes #921.

This keeps the Settings sidebar `Back` header below the native iPadOS top safe area when the app is using the wider/tablet sidebar layout.

The change mirrors the existing native sidebar treatment used elsewhere: native safe-area padding belongs on the sidebar container, while the existing Electron window-control spacer remains separate.

## Testing

- `npm run format:files -- packages/app/src/screens/settings-screen.tsx`
- `npm run lint -- packages/app/src/screens/settings-screen.tsx`
- `npm run build:workspace-deps --workspace=@getpaseo/app`
- pre-commit hook: targeted lint, targeted format check, full workspace `npm run typecheck --workspaces --if-present`
- loaded the branch through a development client on a physical iPad and confirmed the Settings sidebar `Back` row no longer overlaps the iPadOS status bar

### DIFF
--- a/packages/app/src/screens/settings-screen.tsx
+++ b/packages/app/src/screens/settings-screen.tsx
@@ -720,9 +720,16 @@ function SettingsSidebar({
   }, [hosts, localServerId]);
   const isDesktopApp = isElectronRuntime();
   const items = SIDEBAR_SECTION_ITEMS.filter((item) => !item.desktopOnly || isDesktopApp);
+  const insets = useSafeAreaInsets();
   const padding = useWindowControlsPadding("sidebar");
   const isDesktop = layout === "desktop";
-  const containerStyle = isDesktop ? sidebarStyles.desktopContainer : sidebarStyles.mobileContainer;
+  const containerStyle = useMemo(
+    () => [
+      isDesktop ? sidebarStyles.desktopContainer : sidebarStyles.mobileContainer,
+      isDesktop ? { paddingTop: insets.top } : null,
+    ],
+    [insets.top, isDesktop],
+  );
   const selectedSectionId = view.kind === "section" ? view.section : null;
   const selectedServerId = view.kind === "host" ? view.serverId : null;
   const isProjectsSelected = view.kind === "projects" || view.kind === "project";


kongjiadongyuan:
Manual verification screenshot from the physical iPad after this fix:

iPad Settings sidebar safe-area fixed

The Settings sidebar `Back` row now sits below the iPadOS status bar, while the detail header remains aligned on the right.

boudra (approved):

PR #927, opened 2026-05-11
Bridge Pi extension dialogs into Paseo question permissions

thatdaveguy1:
## Summary
- bridge Pi extension `ui.select()`, `ui.input()`, and `ui.confirm()` calls into Paseo's existing permission/question request flow
- make Pi sessions bind extensions with a lightweight `uiContext` so dialog-based extensions can surface questions to Paseo users
- add a unit test covering the new question-permission bridge behavior

## Why
Paseo already has a solid question/permission UI for ACP providers, but the Pi provider was binding extensions without any `uiContext`, so Pi extensions that need user input could not surface questions to the user.

This adds a minimal Pi-side bridge that:
- emits `permission_requested` / `permission_resolved` events from dialog-style extension UI calls
- reuses Paseo's existing `kind: "question"` permission card format
- returns `undefined` from `ui.custom()` so extensions like `pi-ask-user` can fall back to their simpler dialog protocol (`select` / `input`)

## Notes
- **Actual Pi + Paseo integration is untested.**
- I did run the repo checks locally via the pre-commit hook, and they passed:
  - format
  - lint
  - typecheck
- I also ran the targeted provider test file locally:
  - `npx vitest run packages/server/src/server/agent/providers/pi-direct-agent.test.ts --bail=1`

## Scope / limitations
- This is intentionally a minimal bridge for dialog-style extension UI.
- It does **not** implement full Pi TUI rendering inside Paseo.
- `ui.custom()` intentionally resolves to `undefined` so extensions that support RPC/headless fallbacks can degrade into `select()` / `input()` instead of trying to render Pi's TUI directly.

### DIFF
--- a/packages/server/src/server/agent/providers/pi-direct-agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi-direct-agent.test.ts
@@ -34,6 +34,7 @@ function createPiSession(prompt: () => Promise<void>): PiDirectSessionAdapter {
       getSessionFile: () => "/tmp/pi-session.json",
       getCwd: () => "/tmp/paseo-pi-test",
     },
+    bindExtensions: vi.fn(async () => undefined),
     subscribe: vi.fn(),
     prompt,
     abort: vi.fn(),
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

this is untested in real pi + paseo. i don't merge changes without testing evidence. i'm closing the pull request.

opened #980 to follow up on this myself. if you send a PR with testing evidence against a real pi extension, i would accept it.

PR #940, opened 2026-05-12
Fix native diff row expansion

bolasblack:
## Summary

- Fix structured diff parsing for git headers without `a/` / `b/` prefixes, e.g. `diff --git path path`.
- Keep support for the normal prefixed form, e.g. `diff --git a/path b/path`.
- Apply the same parser fix to both the server and app parser copies.
- Add regression tests for no-prefix diff headers.

## Root cause

This was not a row press/tap handling issue.

Git patch output usually includes source/destination prefixes:

```diff
diff --git a/packages/app/src/git/diff-pane.tsx b/packages/app/src/git/diff-pane.tsx
```

However, git can emit no-prefix paths when patch output is configured to omit source/destination prefixes. In the affected checkout, `git diff` emitted headers like:

```diff
diff --git packages/app/src/git/diff-pane.tsx packages/app/src/git/diff-pane.tsx
```

The server-side structured diff parser only understood the prefixed `a/... b/...` form. When it saw the no-prefix form, it failed to extract the file path, so checkout diff assembly could not match the parsed hunks back to the file entry.

The row click worked, but the expanded body had no hunks to render, making it look like the row only expanded by about 1px.

The app parser had the same header assumption, so it was updated too to keep parser behavior consistent and avoid the same bug if that parser path is used later.

## How to reproduce

1. In a checkout, configure git diff output to omit source/destination prefixes:

   ```sh
   git config diff.noprefix true
   ```

2. Make any tracked file change.

3. Confirm git now emits no-prefix diff headers:

   ```sh
   git diff --patch --raw
   ```

   The patch header will look like:

   ```diff
   diff --git path/to/file.ts path/to/file.ts
   ```

   instead of:

   ```diff
   diff --git a/path/to/file.ts b/path/to/file.ts
   ```

4. Open the checkout in Paseo and go to the Changes diff panel.

5. Click the changed file row to expand it.

Before this fix, the click was handled but the structured diff had no hunks for that file, so the row appeared to expand only slightly / render an empty body.

After this fix, the same row expands with the expected hunk content.

## Test plan

- [x] `agent-browser` verification with isolated daemon: collapsing removes `diff-file-0-body`; expanding restores a populated body with hunk text.
- [x] Direct `getCheckoutDiff(..., { includeStructured: true })` check: affected file returns populated hunks instead of `hunks: []`.
- [x] `npx vitest run packages/server/src/server/utils/diff-highlighter.test.ts packages/app/src/utils/diff-highlighter.test.ts --bail=1`
- [x] `npm run lint`
- [x] `npm run typecheck`

### DIFF
--- a/packages/app/src/utils/diff-highlighter.test.ts
+++ b/packages/app/src/utils/diff-highlighter.test.ts
@@ -185,6 +185,18 @@ describe("parseDiff", () => {
     expect(hunk.lines[3].content).toBe("const bar = 3;");
   });
 
+  it("parses no-prefix git diff headers", () => {
+    const files = parseDiff(
+      SIMPLE_DIFF.replace("a/example.ts b/example.ts", "example.ts example.ts"),
+    );
+
+    expect(files).toHaveLength(1);
+    expect(files[0].path).toBe("example.ts");
+    expect(files[0].additions).toBe(1);
+    expect(files[0].deletions).toBe(1);
+    expect(files[0].hunks).toHaveLength(1);
+  });
+
   it("parses a diff with multiple hunks", () => {
     const files = parseDiff(MULTI_HUNK_DIFF);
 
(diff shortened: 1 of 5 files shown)


bolasblack:
@boudra, thanks. Really appreciated the help improving and merging this PR.

PR #952, opened 2026-05-12
Fix iPad host switcher popover width

### LINKED ISSUE #946: bug: iPad host switcher popover can render too narrow
### What's broken

The host switcher popover in the iPad app can become too narrow to read in the non-compact sidebar layout.

Expected: the popover should have a readable width, similar to the desktop host switcher.

### Steps to reproduce

1. Open the iPad app in a non-compact layout.
2. Use the sidebar footer host switcher.
3. Open the host switcher popover.

### Where did this happen

iOS app

### Paseo version

0.1.74

### OS version

iPadOS

### Agent provider

Not relevant

### Provider configuration

N/A

### Logs

N/A

### Screenshots or video

Image

### Proposed direction

This looks like a `Combobox` desktop/native popover sizing issue rather than a host-switcher-only issue.

Current `Combobox` desktop sizing is anchored around `desktopMinWidth ?? referenceWidth ?? 200`, with a max width, but no explicit content-based width strategy. On React Native Web/Electron, CSS layout can still make the popover wide enough for its content. On native iPad, the popover can settle close to the measured trigger width instead.

A clean fix may be to add an opt-in `Combobox` width strategy for non-compact popovers, e.g. content/preferred width with a readable minimum and max-width clamp, then use it for the host switcher. Other menu-like components already use measured content size when positioning their popovers, so this would align `Combobox` with an existing pattern without changing every combobox by default.

kongjiadongyuan:
## Summary

Fixes #946.

This adds an opt-in content-width strategy for desktop/non-compact `Combobox` popovers and uses it for the sidebar host switcher. Instead of estimating a width from character counts or hard-coding a wider host picker, the combobox renders an offscreen measurement copy of the menu rows and uses the measured native layout width, clamped by the existing available-width/max-width logic.

For the host picker, the measurement set includes every host label paired with every possible connection status label, so status changes do not make the popover width jump or truncate unexpectedly.

Before
iPad 问题
After
截屏 2026-05-12 下午7 42 59

## Why this shape

- Keeps the default combobox behavior unchanged (`desktopWidthStrategy="anchor"`).
- Uses actual React Native layout measurement on iPad instead of web shrink-to-fit assumptions.
- Hides the popover until content width is measured, avoiding a one-frame narrow render.
- Matches the native combobox row margins in the hidden measurement copy so measured width tracks real row geometry.

## Validation

- `npm run format:files -- packages/app/src/components/ui/combobox.tsx packages/app/src/components/left-sidebar.tsx`
- `npm run lint -- packages/app/src/components/ui/combobox.tsx packages/app/src/components/left-sidebar.tsx`
- `npm run build:daemon`
- Pre-commit hook passed: targeted lint, format check, and `npm run typecheck --workspaces --if-present`
- Built a Release iOS app and installed/launched it on a real iPad Pro 13-inch (M5), bundle id `sh.paseo.zbl`

### DIFF
--- a/packages/app/src/components/left-sidebar.tsx
+++ b/packages/app/src/components/left-sidebar.tsx
@@ -67,6 +67,28 @@ import { SidebarCalloutSlot } from "./sidebar-callout-slot";
 import { SidebarWorkspaceList } from "./sidebar-workspace-list";
 
 const MIN_CHAT_WIDTH = 400;
+const HOST_PICKER_STATUS_MEASURE_LABELS = [
+  formatConnectionStatus("idle"),
+  formatConnectionStatus("connecting"),
+  formatConnectionStatus("online"),
+  formatConnectionStatus("offline"),
+  formatConnectionStatus("error"),
+];
+
+function buildHostContentMeasureOptions(hostOptions: ComboboxOption[]): ComboboxOption[] {
+  const measureOptions: ComboboxOption[] = [];
+  for (const option of hostOptions) {
+    for (const description of HOST_PICKER_STATUS_MEASURE_LABELS) {
+      measureOptions.push({
+        id: `${option.id}:${description}`,
+        label: option.label,
+        description,
+        kind: option.kind,
+      });
+    }
+  }
+  return measureOptions;
+}
 
 type SidebarShortcutModel = ReturnType<typeof useSidebarShortcutModel>;
 type SidebarTheme = ReturnType<typeof useUnistyles>["theme"];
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

the bug is real, but for this one i'd rather hardcode a `desktopMinWidth` (around 280) on the host-switcher combobox usage and keep the combobox primitive untouched. smaller surface to maintain.

closing this one. if you want to send the smaller fix yourself, go for it. otherwise tell me and i'll do it.

PR #981, opened 2026-05-13
fix(app): widen host switcher popover

### LINKED ISSUE #946: bug: iPad host switcher popover can render too narrow

kongjiadongyuan:
## Summary

- Set a host-switcher-only desktop popover minimum width of 280px
- Keep the shared `Combobox` primitive untouched, following the maintainer feedback on #952

Fixes #946.

## Verification

- `npm run format:check:files -- packages/app/src/components/left-sidebar.tsx`
- `npm run lint -- packages/app/src/components/left-sidebar.tsx`
- `npm run build:workspace-deps --workspace=@getpaseo/app`
- pre-commit hook: `lint`, `format`, `typecheck`

### DIFF
--- a/packages/app/src/components/left-sidebar.tsx
+++ b/packages/app/src/components/left-sidebar.tsx
@@ -472,6 +472,7 @@ function SidebarFooter({
         searchable={false}
         title="Switch host"
         searchPlaceholder="Search hosts..."
+        desktopMinWidth={280}
         open={isHostPickerOpen}
         onOpenChange={setIsHostPickerOpen}
         anchorRef={hostTriggerRef}


boudra:
thanks!

PR #1014, opened 2026-05-14
fix(desktop): bound external CLI output capture

kongjiadongyuan:
## Summary

- Bound stdout/stderr captured from desktop external CLI child processes to the last 64 KiB.
- Preserve the tail of oversized output and include a truncation note in failure messages.
- Add regression coverage for large stdout and stderr output.

## Why

Desktop daemon management shells out to the bundled CLI for commands like `daemon status`, `daemon status --json`, and `daemon pair --json`. That helper previously appended child stdout/stderr into unbounded strings, so pathological CLI output could grow the Electron main-process heap until V8 throws `RangeError: Invalid string length`.

This mirrors the existing desktop daemon startup-output guard, using the same 64 KiB tail-capture style.

## Platform impact

This affects the Electron desktop helper, so it applies to desktop builds on macOS, Windows, and Linux. It does not affect iOS/iPadOS/Android clients, which do not run this desktop CLI child-process path.

## Test plan

- `NODE_PATH=/Users/zbl/Desktop/paseo/node_modules /Users/zbl/Desktop/paseo/node_modules/.bin/vitest run packages/desktop/src/daemon/cli/external.test.ts --bail=1`
- `npm run format:check:files -- packages/desktop/src/daemon/cli/external.ts packages/desktop/src/daemon/cli/external.test.ts`
- `npm run lint -- packages/desktop/src/daemon/cli/external.ts packages/desktop/src/daemon/cli/external.test.ts`
- `npm run typecheck --workspace=@getpaseo/desktop`

Note: local pre-commit lint/format passed after wiring this isolated worktree to the main checkout's `node_modules`. The hook's full workspace typecheck failed in this isolated worktree due unrelated workspace dependency resolution/type errors; the targeted desktop typecheck above passed.

### DIFF
--- a/packages/desktop/src/daemon/cli/external.test.ts
+++ b/packages/desktop/src/daemon/cli/external.test.ts
@@ -88,4 +88,36 @@ describe("external CLI", () => {
       localDaemon: "running",
     });
   });
+
+  it("bounds captured stdout from text commands", async () => {
+    const suffix = "daemon still running\n";
+    const largeOutput = `${"x".repeat(512 * 1024)}${suffix}`;
+    mockExternalCliOutput({ stdout: largeOutput });
+
+    const output = await runExternalCliTextCommand(["daemon", "status"]);
+
+    expect(output.length).toBeLessThanOrEqual(64 * 1024);
+    expect(output.endsWith(suffix.trimEnd())).toBe(true);
+  });
+
+  it("bounds captured stderr in failure messages", async () => {
+    const suffix = "fatal: daemon exploded\n";
+    mockExternalCliOutput({
+      stdout: "",
+      stderr: `${"x".repeat(512 * 1024)}${suffix}`,
+      exitCode: 1,
+    });
+
+    await runExternalCliTextCommand(["daemon", "status"]).then(
+      () => {
+        throw new Error("Expected external CLI command to fail");
+      },
+      (error: unknown) => {
+        const message = error instanceof Error ? error.message : String(error);
+        expect(message).toContain(`output truncated to the last ${64 * 1024} chars`);
+        expect(message).toContain(suffix.trimEnd());
+        expect(message.length).toBeLessThan(70 * 1024);
+      },
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

before i take this, can you say more about what you hit? i looked at the callers, daemon stop / status / pair are all bounded one-shot commands and i couldn't find a path that produces anywhere near 64 KiB of output, so i want to make sure i'm not waving off a real problem.

if you saw `RangeError: Invalid string length` in practice, what was the command and roughly how much output? if it's defensive against something that could happen later, that's worth knowing too, the fix might belong at the producing CLI command rather than the desktop helper.

kongjiadongyuan:
Thanks, that’s a fair question. I rechecked the original incident and I think my PR description was too broad.

What I hit in practice was this: after using Paseo from mobile/iPad for a long-running session, I later opened the macOS desktop app and immediately got this uncaught exception. I don’t know yet whether the “long mobile session” is causally related, but that was the real-world sequence.

The exact error was:

```txt
Uncaught Exception:
RangeError: Invalid string length
at Socket.<anonymous> (/Applications/Paseo.app/Contents/Resources/app.asar/dist/daemon/daemon-manager.js:277:24)
at Socket.emit (node:events:508:28)
at addChunk (node:internal/streams/readable:563:12)
at readableAddChunkPushByteMode (node:internal/streams/readable:514:3)
at Readable.push (node:internal/streams/readable:394:5)
at Pipe.onStreamRead (node:internal/stream_base_commons:189:23)
```

After looking at the stack again, I agree this does not clearly prove that `daemon status`, `daemon stop`, or `daemon pair` produced huge output. The stack points at a child-process pipe `data` handler in `daemon-manager.js`, so the direct crash was more likely in the desktop daemon startup/management output capture path, not necessarily the external CLI helper changed in this PR.

So I think there are two separate things here:

1. The actual crash I saw was a real desktop main-process crash caused by unbounded string accumulation from a child-process pipe.
2. This PR applies the same bounded-output pattern to the external CLI helper, but I can’t honestly say it was the exact path that caused my original crash.

Given that, I’m happy to reframe this PR as defensive hardening rather than “the” fix for the crash. If you think the right direction is to keep output bounding only at the producer/caller that can actually emit large output, I can close or reshape this PR. My current read is that any desktop helper which captures child stdout/stderr into memory should have a small tail buffer instead of unbounded string concatenation, but I agree the original report should be tied to `daemon-manager.js`, not over-attributed to `status`/`stop`/`pair`.

boudra:
Closing this defensive-hardening PR, as the discussion did not establish a failing external CLI command that needs this change.

PR #1017, opened 2026-05-14
Refine user message bubble contrast

### LINKED ISSUE #986: feat: distinguish user message bubbles from agent responses
### Prior search

### What's the problem

In long conversations my own messages and the agent's responses share the same `surface2` bubble background, so it's hard to scan back through a thread and pick out where my prompts are. Right-alignment + the corner radius help, but on light themes the contrast against the page is already low and the eye doesn't catch the boundary at a glance.

### What would solve it

A way to make the user message bubble visually distinct from the agent's. No strong opinion on the shape — could be:

- a setting under General with a small curated palette (default / blue / green / purple), or
- just a different default for the user bubble across all themes (e.g. a subtle accent or border), with no setting at all.

Happy to defer to whichever fits the product direction better.

Relying on alignment alone — works for the last few messages, falls apart scrolling through a long thread, especially on the lighter themes.

32r4:
### Linked issue

Closes #986

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

This PR improves the visual distinction between user messages and assistant responses in the conversation view.

Changes included:
- adds semantic theme tokens for user message bubbles instead of tying the styling directly to shared surface tokens
- keeps the light theme on a border-only treatment so it stays calm while still being easier to scan
- keeps the default dark theme neutral with clearer contrast
- keeps lighter, restrained tinting for the specialty dark themes (`zinc`, `midnight`, `claude`, and `ghostty`)
- keeps the implementation scoped to theme values so each theme can be tuned independently later

### How did you verify it

I verified this change by:
- checking the final diff for 'packages/app/src/styles/theme.ts'
issue-986-before-vs-latest-part-1
issue-986-before-vs-latest-part-2

Known verification limits:
- `npm run typecheck` still fails on unrelated pre-existing repository issues in `packages/server`, `packages/app`, and `packages/cli`
- `npm run lint` / `npm run format` are blocked in this local Windows environment because the `oxlint` / `oxfmt` native bindings are unavailable here

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/components/message.tsx
+++ b/packages/app/src/components/message.tsx
@@ -337,9 +337,9 @@ const userMessageStylesheet = StyleSheet.create((theme) => ({
     marginBottom: theme.spacing[4],
   },
   bubble: {
-    backgroundColor: theme.colors.surface3,
+    backgroundColor: theme.colors.userMessageBubble,
     borderWidth: theme.borderWidth[1],
-    borderColor: theme.colors.border,
+    borderColor: theme.colors.userMessageBubbleBorder,
     borderRadius: theme.borderRadius["2xl"],
     borderTopRightRadius: theme.borderRadius.sm,
     paddingHorizontal: theme.spacing[4],
--- a/packages/app/src/components/message.tsx
+++ b/packages/app/src/components/message.tsx
@@ -337,7 +337,9 @@ const userMessageStylesheet = StyleSheet.create((theme) => ({
     marginBottom: theme.spacing[4],
   },
   bubble: {
-    backgroundColor: theme.colors.surface2,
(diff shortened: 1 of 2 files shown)


JasonLandbridge:
Very nice, I was hoping someone would fix this!

boudra:
Thanks for taking a swing at this @32r4! I went a different direction in ed7944cf9 — instead of bumping the bubble to a heavier surface, the user-message bubble now uses a new `accentSubtle` theme token (accent darkened + slightly desaturated + low alpha), so it picks up each theme's accent without clashing.

Closing this in favor of that commit. Appreciate the contribution.

PR #1039, opened 2026-05-15
Harden default host allowlist

joaosa:
## Summary
- keep default Host header allowlist to localhost, .localhost, and loopback IP literals
- require explicit hostnames opt-in for LAN or public IP literals
- reuse a shared server loopback-host helper for Host validation and workspace service binding
- cover loopback, rejected non-loopback, explicit IP allowlist, and service binding cases

## Tests
- npm run test:unit --workspace=@getpaseo/server -- src/server/hostnames.test.ts src/server/workspace-service-env.test.ts --bail=1
- npm run format:check:files -- packages/server/src/server/hostnames.ts packages/server/src/server/hostnames.test.ts packages/server/src/server/loopback-host.ts packages/server/src/server/workspace-service-env.ts packages/server/src/server/workspace-service-env.test.ts
- npm run lint -- packages/server/src/server/hostnames.ts packages/server/src/server/hostnames.test.ts packages/server/src/server/loopback-host.ts packages/server/src/server/workspace-service-env.ts packages/server/src/server/workspace-service-env.test.ts
- npm run typecheck --workspace=@getpaseo/server

### DIFF
--- a/packages/server/src/server/hostnames.test.ts
+++ b/packages/server/src/server/hostnames.test.ts
@@ -11,9 +11,16 @@ describe("hostnames (vite-style)", () => {
     expect(isHostnameAllowed("foo.localhost:6767", undefined)).toBe(true);
   });
 
-  it("allows IP addresses by default", () => {
+  it("allows loopback IP addresses by default", () => {
     expect(isHostnameAllowed("127.0.0.1:6767", undefined)).toBe(true);
+    expect(isHostnameAllowed("127.42.0.1:6767", undefined)).toBe(true);
     expect(isHostnameAllowed("[::1]:6767", undefined)).toBe(true);
+  });
+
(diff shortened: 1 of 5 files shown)


boudra:
thanks for the pull request.

before i merge this i want to understand the exact attack this closes. hostnames is the DNS rebinding control, but rebinding uses an attacker-controlled domain in the Host header, not an IP literal. a direct request from attacker JS to http://192.168.1.5:6767 is cross-origin and is handled by CORS.

the cost of this change is real: anyone running paseo on the LAN today has to add their IP to hostnames in addition to CORS to keep things working. before i ship that, i want to see the concrete attack scenario this prevents that isn't already covered.

joaosa:
This was more of a defense in-depth approach. I agree the cost is real and this is likely not truly needed.

PR #1075, opened 2026-05-17
feat(push): add ServerChan notifications

xiaobaifly7:
## Summary

Adds optional ServerChan (Server酱) push notifications for agent completion events.

When `PASEO_SERVERCHAN_SENDKEY` is set, the server sends a ServerChan notification in addition to the existing Expo push flow. This gives users in environments where Expo/FCM push delivery is unreliable a server-side notification fallback without changing mobile client behavior.

## Behavior

- Uses `PASEO_SERVERCHAN_SENDKEY` as the opt-in configuration switch.
- Sends ServerChan notifications even when no Expo push tokens are registered.
- Keeps the existing Expo push token flow intact.
- Logs and swallows ServerChan failures so notification delivery cannot interrupt agent lifecycle handling.

## Validation

- `rtk npx --yes oxfmt@0.46.0 --check packages/server/src/server/push/notifications.ts packages/server/src/server/push/notifications.test.ts packages/server/src/server/push/serverchan-service.ts packages/server/src/server/push/serverchan-service.test.ts`
- `rtk npx --yes oxlint@1.61.0 packages/server/src/server/push/notifications.ts packages/server/src/server/push/notifications.test.ts packages/server/src/server/push/serverchan-service.ts packages/server/src/server/push/serverchan-service.test.ts`
- `rtk git diff --check`

Full Vitest/typecheck were not run in this checkout because the workspace install currently fails on local dependency resolution issues unrelated to this change.

### DIFF
--- a/packages/server/src/server/push/notifications.test.ts
+++ b/packages/server/src/server/push/notifications.test.ts
@@ -1,0 +1,51 @@
+import type pino from "pino";
+import { afterEach, describe, expect, test, vi } from "vitest";
+
+import { createPushNotificationSender } from "./notifications.js";
+import type { PushTokenStore } from "./token-store.js";
+
+function createLogger(): pino.Logger {
+  const logger = {
+    child: () => logger,
+    info: vi.fn(),
+    warn: vi.fn(),
+  };
+  return logger as unknown as pino.Logger;
+}
+
+function createTokenStore(tokens: string[]): PushTokenStore {
+  return {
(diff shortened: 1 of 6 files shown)


boudra:
this doen't belong in Paseo

the best way to get this done is to allow setting "agent completion" hooks which can run a command, you'd put your specific ServerChan notification there instead

please open an issue for agent completion hooks

PR #1078, opened 2026-05-17
feat(server): skip model/mode fetching during provider refresh

dougmorato:
### Linked issue

N/A — small, focused performance improvement surfaced from latency profiling.

### Type of change

(Technically a small feature addition to the snapshot manager API, but the user-visible effect is purely a latency reduction.)

### What does this PR do

Adds a `skipModelsAndModes` option to `ProviderSnapshotManager` that allows refreshing a provider's status (enabled/disabled, metadata, heartbeat) **without** re-fetching the full models and modes lists from the provider API.

**The problem:** `refreshProvider` is called on several hot paths — provider settings updates, periodic health checks, daemon startup provider validation. On every call, it was unconditionally invoking `fetchModels()` and `fetchModes()` against the provider. For remote providers (OpenAI, Anthropic, Google, etc.) this means a blocking HTTPS roundtrip. For local providers hosting many models, the local enumeration is similarly wasteful. No caller of `refreshProvider` actually needs the models/modes lists — they only need to know the provider is alive and what its current configuration looks like.

**The fix:** Introduce a `skipModelsAndModes` boolean on the internal options object. When `true`, the snapshot entry is written with `status: "ready"` and `enabled: true` immediately after the provider definition is resolved, skipping the model+mode fetch entirely. The flag defaults to `false` so **all existing callers of `loadProviderSnapshot` are unaffected**. Only `refreshProvider` sets it to `true`.

### Design decisions

| Decision | Rationale |
|---|---|
| Flag is opt-**out** (defaults `false`) | Preserves full backwards compatibility. `loadProviderSnapshot` continues to fetch models/modes as before. |
| Only `refreshProvider` opts in | `refreshProvider` is the only call site where models/modes are provably not consumed downstream. |
| Sets `models: undefined, modes: undefined` rather than preserving stale values | The snapshot entry already has stale data from a prior full load. Setting them to `undefined` explicitly signals to consumers "this entry was fast-loaded" and prevents silent drift. |
| Still resolves the provider definition and checks connectivity | The `resolve()` call still happens, so provider auth errors or unreachable endpoints are still caught. Only the model enumeration is skipped. |
| No new public API surface | The option is added only to internal interfaces (`ProviderLoadOptions`, the private `loadEntry` signature). It is not exposed on the public `ProviderSnapshotManager.loadProviderSnapshot()` method. |

### How did you verify it

1. **Typecheck** — `npm run typecheck` passes across all workspace packages (server, cli, app, desktop, website). Zero errors.
2. **Lint** — `npm run lint` (oxlint) passes with 0 warnings and 0 errors across 1755 files.
3. **Format** — `npm run format:check` (oxfmt) passes on all 1897 files.
4. **Manual code audit of all call sites:**
   - `loadProviderSnapshot` (public) — does **not** set `skipModelsAndModes`, so models/modes are still fetched on initial load. ✅
   - `refreshProvider` — only consumer of provider status, never reads models/modes from the snapshot it writes. ✅
   - `loadEntry` — new early-return path is guarded by the flag and only executes when explicitly opted in. ✅
5. **Behavior preservation** — the early-return path writes the same `ProviderEntry` shape that the full path would write (minus models/modes), so any code reading `entry.status` or `entry.enabled` works identically.
6. **Edge cases inspected:**
   - What if a consumer later calls `loadProviderSnapshot` after a fast refresh? → `loadProviderSnapshot` does not set the flag, so it will do a full fetch and overwrite the fast-loaded entry. No stale data leak.
   - What if `resolve()` throws (bad auth, unreachable)? → The early-return is placed **after** the resolve guard, so errors still propagate. Provider health check semantics are preserved.
   - What if `force: true` is combined with `skipModelsAndModes`? → This combination is not used anywhere; `refreshProvider` always passes `force: false`.

### Checklist

- [x] `npm run lint` passes
- [x] Not a UI change — no screenshots/video needed.

### DIFF
--- a/packages/server/src/server/agent/provider-snapshot-manager.ts
+++ b/packages/server/src/server/agent/provider-snapshot-manager.ts
@@ -22,6 +22,7 @@ interface ProviderLoadOptions {
   cwd: string;
   providers: AgentProvider[];
   force: boolean;
+  skipModelsAndModes?: boolean;
 }
 interface ProviderLoad {
   promise: Promise<void>;
@@ -207,6 +208,7 @@ export class ProviderSnapshotManager {
       cwd,
       providers: providersToRefresh,
       force: false,
+      skipModelsAndModes: true,
     });
   }
 
@@ -243,6 +245,7 @@ export class ProviderSnapshotManager {
           definition,
           load,
           force: options.force,
+          skipModelsAndModes: options.skipModelsAndModes,
         }),
       )
       .finally(() => {
@@ -263,8 +266,9 @@ export class ProviderSnapshotManager {
     definition: ProviderDefinition;
     load: ProviderLoad;
     force: boolean;
+    skipModelsAndModes?: boolean;
   }): Promise<void> {
-    const { cwd, provider, definition, load, force } = options;
+    const { cwd, provider, definition, load, force, skipModelsAndModes } = options;
     const snapshot = this.getOrCreateSnapshot(options.cwd);
     const base = {
       provider,
@@ -295,6 +299,18 @@ export class ProviderSnapshotManager {
       );
       if (!available) {
         setEntry({ ...base, status: "unavailable", enabled: true });
+        return;
+      }
+
+      if (skipModelsAndModes) {
+        setEntry({
+          ...base,
+          status: "ready",
+          enabled: true,
+          models: undefined,
+          modes: undefined,
+          fetchedAt: new Date().toISOString(),
+        });
         return;
       }
 


dougmorato:
Hey @boudra,

Thanks for building Paseo — really enjoying the product.

**Why this PR exists:** I noticed the CLI would hang with a spinning indicator whenever it tried to look up agents through a provider that had even one model disabled. Under the hood, `refreshProvider` was unconditionally fetching the full models and modes list on every health check and settings update, even though none of those code paths actually consume model/mode data. For remote providers that is an unnecessary network roundtrip; for local providers with many models it is wasteful enumeration.

This patch adds a small internal `skipModelsAndModes` flag — it defaults to `false` so `loadProviderSnapshot` (the initial-load path) is completely untouched. Only `refreshProvider` opts into the fast path, writing a ready entry immediately after resolving the provider definition, without touching the model API.

**The result:** provider status checks become instant. No more spinner when browsing agents via the CLI.

**Verification:** typecheck, lint, and format all pass with zero errors. Existing tests continue to validate the full snapshot path. Happy to add a dedicated unit test for the fast path if you like the direction.

Let me know if there is anything you would like adjusted. 🍻

boudra:
thanks for the pull request.

i'm having trouble lining this up with what i'm seeing in the code, and i could use your help. you mentioned:

> the CLI would hang with a spinning indicator whenever it tried to look up agents through a provider that had even one model disabled.

i couldn't find a CLI command that fits this. `paseo agent ls`, `agent run`, `agent inspect`, and `agent mode --list` don't read the provider snapshot. the only command that does is `paseo provider models <name>`, and it only blocks on first warmup after a daemon restart; afterwards it returns from cache instantly, including cached errors.

i'm also not seeing a per-model disabled concept in paseo. providers have an enabled flag, individual models don't.

on the refresh side, the only path that actually re-fetches models/modes is the explicit settings refresh button, which is the user-facing escape hatch for fresh data. a flag that skips the refresh on that path would weaken what users expect when they click refresh.

could you share repro steps? the exact command or app action, what you saw, and what you expected. i'd rather understand the user problem before deciding on the fix shape, because i can't reproduce it against the current code.

PR #1082, opened 2026-05-18
Fix CLI/daemon environment inconsistency in `daemon status` and `daemon pair` commands

### LINKED ISSUE #1081: bug: CLI daemon subcommands use CLI process environment instead of daemon's environment
### What's broken

Two paseo daemon subcommands execute environment-dependent checks using the CLI process's process.env, which may differ from the running daemon's environment. This produces misleading output or broken functionality when the CLI and daemon run in different environments (e.g. daemon started via systemd/launchd/desktop app, CLI run from an interactive shell).

1. `paseo daemon status` — Provider detection uses CLI's PATH

`status` reports provider availability by calling `findExecutable()` (which uses which against the CLI's `PATH`) at [`packages/cli/src/commands/daemon/status.ts:170-195`](https://github.com/getpaseo/paseo/blob/77b92b58f7d51a93b2a6dc034904622e77ba568e/packages/cli/src/commands/daemon/status.ts#L170-L195). This reports what the CLI can find, not what the daemon can find. When the daemon is started by a process with a minimal PATH (`systemd`, `launchd`, etc.), the CLI may show all providers as available while the daemon can't resolve any of them.

2. `paseo daemon pair` — Relay config may come from CLI's env vars

`pair` reads relay configuration (`PASEO_RELAY_ENDPOINT`, `PASEO_RELAY_USE_TLS`, etc.) at [`packages/server/src/server/pairing-offer.ts`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/server/pairing-offer.ts). If these are set only as environment variables and not written to config.json, the CLI reads them from the current shell while the daemon reads from its own startup environment. The pairing URL can point to a different relay than the daemon is actually connected to, causing pairing to fail silently.

### Steps to reproduce

Reproducing the status issue:

1. Start daemon from a shell where providers (claude, codex, opencode) are on PATH
2. Verify providers show as available: `paseo daemon status`
3. In a separate shell with a minimal PATH (e.g. `env -i PATH=/usr/bin:/bin paseo daemon status`)
4. Observe providers showing as "not found" while the daemon is still running and could itself have different PATH resolution

Reproducing the pair issue:

1. Set `PASEO_RELAY_ENDPOINT=custom.relay.example.com:443` in your shell
2. Run paseo daemon pair — the QR code URL will contain the custom relay
3. If the daemon was started from a different environment without this env var, it connects to the default `relay.paseo.sh:443` instead
4. Scanning the QR code fails because the phone and daemon are on different relays

### Where did this happen

Daemon

### Paseo version

0.1.74

### OS version

macOS Darwin 25.4.0 (also affects Linux with systemd)

### Agent provider

Not relevant

### Provider configuration

N/A — issue is in CLI environment resolution, not provider behavior

### Logs

```text
N/A — no crash, the commands succeed but produce incorrect results. Example status output showing the inconsistency:

  Local Daemon      stopped
  Connected Daemon  reachable
  ...
  Note              Connected daemon is reachable at 127.0.0.1:6767 but no local daemon PID file was found

Providers
  Claude          /etc/profiles/per-user/yzx9/bin/claude (2.1.137 (Claude Code))
  Codex           /etc/profiles/per-user/yzx9/bin/codex (codex-cli 0.130.0)
  OpenCode        /etc/profiles/per-user/yzx9/bin/opencode (1.14.35)

The Providers section reflects the CLI's PATH (/etc/profiles/per-user/yzx9/bin/), not the daemon's. The daemon has no PID file (started from a different environment), yet it's reachable — meaning the daemon's environment is unknown to the CLI.
```

### Screenshots or video

_No response_

yzx9:
### Linked issue

Closes #1081

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

When the daemon runs in a different environment than the CLI (e.g. systemd), `status` reported providers from the CLI's `PATH` and `pair` generated pairing URLs using CLI's env vars — both potentially wrong from the daemon's perspective.

Add two new daemon RPCs (`daemon.status`, `daemon.pairing_offer`) that let the CLI query the daemon's own provider availability and relay config. Both commands fall back to local checks when the daemon is unreachable.

Also strip `PASEO_RELAY_*` env vars in `resolveLocalDaemonState()` so the CLI reads relay config from config.json rather than the current shell env.

### How did you verify it

I have add and run all the tests, and run the checklist

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/cli/src/commands/daemon/pair.ts
+++ b/packages/cli/src/commands/daemon/pair.ts
@@ -1,6 +1,8 @@
 import { Command } from "commander";
 import chalk from "chalk";
 import { generateLocalPairingOffer, loadConfig, resolvePaseoHome } from "@getpaseo/server";
+import { tryConnectToDaemon } from "../../utils/client.js";
+import { resolveLocalDaemonState, resolveTcpHostFromListen } from "./local-daemon.js";
 import { addJsonOption } from "../../utils/command-options.js";
 
 interface PairOptions {
@@ -22,6 +24,33 @@ export async function runPairCommand(options: PairOptions): Promise<void> {
   }
 
   const paseoHome = resolvePaseoHome();
+  const state = resolveLocalDaemonState({ home: paseoHome });
+  const host = resolveTcpHostFromListen(state.listen);
+
+  // Try to get the pairing offer from the running daemon first.
+  if (host) {
+    const client = await tryConnectToDaemon({ host, timeout: 1500 });
+    if (client) {
+      const supportsDaemonStatusRpc =
+        client.getLastServerInfoMessage()?.features?.daemonStatusRpc === true;
+      if (supportsDaemonStatusRpc) {
+        try {
+          const offer = await client.getDaemonPairingOffer();
+          await client.close().catch(() => {});
+          outputPairingResult(
+            { relayEnabled: offer.relayEnabled, url: offer.url, qr: offer.qr ?? null },
+            options,
+          );
+          return;
+        } catch {
+          // Fall through to local generation
+        }
+      }
+      await client.close().catch(() => {});
+    }
+  }
+
+  // Fall back to local pairing offer generation.
   const config = loadConfig(paseoHome);
   const pairing = await generateLocalPairingOffer({
     paseoHome,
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/daemon/pair.ts
```diff
@@ -22,6 +24,33 @@ export async function runPairCommand(options: PairOptions): Promise<void> {
...
+          return;
+        } catch {
+          // Fall through to local generation
+        }
+      }
+      await client.close().catch(() => {});
+    }
+  }
+
+  // Fall back to local pairing offer generation.
```
yzx9, line 55: I’m not sure we should keep this fallback, as it could potentially be problematic

PR #1103, opened 2026-05-19
fix: stop leaking CSS rules from dynamic UI styles (#1084)

### LINKED ISSUE #654: High RAM Usage Issue
This app is awesome basically, with the capability to agentic dev on iOS and MacOS is just really good.

But I think we have memory usage issue here.

Image

<br/>
with just 2 project and 4 workspace opened, and remote Paseo server (OpenCode provider).
<br/><br/>

Image

<br/>
I open more project and workspace with OpenCode desktop (remote OpenCode server) but never hit more than 2gigs of RAM. and yes, I know that OC uses Tauri as their GUI. idk whats Paseo uses for GUI.
<br/><br/>
is this normal guys?

### LINKED ISSUE #1084: Observed sustained memory growth in app.paseo.sh during extended use
## Environment

- app.paseo.sh web app
- Firefox 150.0.3 on Linux
- Normal interactive use for about 84 minutes
- Process/page metrics were sampled locally. Raw logs are not attached here because they may include local workspace paths or page state.

## What I observed

- Browser memory grew from about 3.4 GB RSS / 1.5 GB PSS to about 26.1 GB RSS / 22.1 GB PSS.
- The largest growth was in the Paseo content process and the Firefox parent process.
- Paseo content process near the end: about 12.5 GB RSS / 12.3 GB PSS.
- Firefox parent process near the end: about 11.0 GB RSS / 8.9 GB PSS.
- Swap stayed at 0 MB during the capture.

## Page-level metrics during the same capture

- DOM peaked around 19.7k elements / 23.9k nodes; near the end it was around 10.3k elements / 12.0k nodes.
- Storage estimate stayed small, roughly 56-73 KB. localStorage stayed around 13-15 KB.
- The recorder did not observe fetch, WebSocket, or EventSource message accumulation.
- The extension process stayed small compared with total browser memory, around 189 MB PSS near the end.

## Memory report snapshot

- The page process for paseo.sh reported about 13.5 GB resident / 13.3 GB resident-unique.
- The largest reported bucket in that process was `explicit/layout/style-sheet-cache/document-shared`, about 11.5 GB.
- JS strings/objects and DOM orphan nodes were much smaller by comparison; DOM orphan nodes were around 59 MB.

## Additional note

I have also experienced the app becoming slower / using more memory during longer sessions in desktop and mobile clients, so I am not sure whether this is one issue across clients or separate symptoms. I can provide anonymized sampling summaries or a memory report if useful.

boudra:
Closes #1084.
Closes #654.

This PR stops stylesheet growth from high-churn UI styles during normal web/Electron interaction.

What changed:
- Adds `inlineUnistylesStyle` as the focused escape hatch for dynamic style values that must stay on normal React Native components.
- Centralizes the floating UI seam: `FloatingSurface.frameStyle` and `FloatingScrollView.style` now own the escape hatch, so hover cards, menus, tooltips, and comboboxes do not need web/plain variants or caller-level Unistyles details.
- Removes the separate `floating.web.tsx` raw DOM wrapper implementation.
- Applies the same escape hatch to the web desktop scrollbar, diff viewer, Changes diff layout controls, inline review geometry, file preview gutters, tool-call detail max heights, whitespace/wrap/expand controls, and git action split-button controls.
- Documents the rule in `docs/unistyles.md`.

Rule going forward:
- Static theme/layout styles can use Unistyles normally.
- Runtime-changing values such as measured size, transform, hover/open/pressed colors, opacity, transition, and cursor styles must use `inlineUnistylesStyle(...)` or a component-owned prop that applies it at the seam.
- Raw DOM wrappers are only for real DOM infrastructure, such as terminal hosts, virtualized web rows, or third-party drag wrappers.

Verification:
- `npm run format`
- `npm run lint`
- `npm run typecheck`
- Targeted Playwright diagnostic for stylesheet growth passed before removing the temporary spec: repeated control cycles settled at `delta: 0`, scrollbar dynamic rules were `[]`, and inline marker warnings were `[]`.

Risk:
- Main risk is visual polish around the changed web/Electron controls; static styles remain on the existing components, while only high-churn values move through the inline escape hatch.


faithleysath:
Thanks a lot for digging into this and putting together the fix.

This looks like it also covers the high-memory reports in #654. That issue has had similar RAM growth observations for a while, including desktop/web usage reports, so it may be worth closing or linking it when this lands.

Really appreciate the quick turnaround here.

PR #1115, opened 2026-05-20
fix(cli): avoid heavy fetches in status and worktree ls

### LINKED ISSUE #1114: CLI status/worktree ls time out because fetchAgents refreshes git metadata for non-active agents
## Summary

`paseo daemon status` can report the local daemon as unstable/unresponsive, and `paseo worktree ls` can time out, even though the daemon WebSocket is reachable and normal ping/heartbeat traffic is healthy.

The failure appears to come from both CLI commands calling the heavy default `fetch_agents_request` path, which refreshes git checkout/project placement metadata for many persisted agent cwd values before returning. On a host with many historical/non-active agents in Paseo-managed worktrees, this can exceed the CLI's 10s request timeout.

## Environment

- Desktop daemon: `0.1.76`
- CLI: `0.1.75`
- macOS desktop app daemon on `127.0.0.1:6767`
- Provider mix includes Codex app-server agents

## Symptoms

`paseo daemon status` shows WebSocket reachable but daemon API unavailable:

```text
Local Daemon      unresponsive
Connected Daemon  reachable
Agents            Unavailable (daemon API not reachable)
Note              Local daemon PID is running but API requests to 127.0.0.1:6767 failed
```

`paseo worktree ls` fails before listing worktrees:

```text
Error: Failed to list worktrees: Timeout waiting for message (10000ms)
```

Daemon logs show slow `fetch_agents_request` calls, while ping/client heartbeat remain fast:

```text
requestType="fetch_agents_request" durationMs=48347
requestType="fetch_agents_request" durationMs=51022
requestType="fetch_agents_request" durationMs=190859
requestType="fetch_agents_request" durationMs=207793
```

Runtime metrics in the same period showed the daemon was still processing pings/heartbeats normally and had only about 15 live agents, so this was not a full daemon hang.

## Investigation Notes

There were initially many archived/stale persisted agent records. After deleting 319 archived records and 113 closed records whose cwd no longer existed, the problem improved but did not disappear.

After cleanup, persisted agent state still had roughly:

```text
108 total agent JSON records
72 idle
34 closed
2 running
0 archived
```

Many of these remaining non-archived records pointed at Paseo-managed worktrees under `~/.paseo/worktrees/.../feat-issue-*`.

Process inspection while a `fetch_agents_request` was in flight showed the daemon continuously spawning git commands such as:

```text
git rev-parse --show-toplevel
git rev-parse --abbrev-ref HEAD
git rev-parse --git-common-dir
git show-ref --verify --quiet refs/remotes/origin/main
git rev-list --left-right --count ...
git ls-files --others --exclude-standard
git diff --shortstat ...
```

Direct `git worktree list --porcelain` was fast, so `worktree ls` was not blocked by raw worktree listing. It was blocked by the earlier agent fetch.

## Root Cause Hypothesis

Both affected CLI commands call the heavy default agent list path:

- `packages/cli/src/commands/daemon/status.ts`
  - calls `client.fetchAgents({ filter: { includeArchived: true } })`
- `packages/cli/src/commands/worktree/ls.ts`
  - calls `client.fetchAgents({ filter: { includeArchived: true } })` before `getPaseoWorktreeList({})`

On the server side:

- `packages/server/src/server/session.ts`
  - `handleFetchAgents`
  - `listFetchAgentsEntries`
  - `collectFetchAgentsEntries`
  - `buildProjectPlacementForCwd`
  - `resolveWorkspaceDirectory`
- `resolveWorkspaceDirectory()` calls `workspaceGitService.getCheckout()` unless `refreshGit:false` is used.
- `WorkspaceGitServiceImpl.getCheckout()` calls `getCheckoutStatus()`, which runs multiple git commands for each cwd.

Even `fetchAgents({ page: { limit: 1 } })` can be slow because `collectFetchAgentsEntries()` resolves placement for batches of up to 25 candidates before it can determine matches.

A direct WebSocket comparison showed the difference clearly:

```text
fetch_agents_request limit:1, default path       ~33s
fetch_agents_request limit:1, scope:"active"    ~0.4s
```

So the daemon is responsive, but the default agent listing path is too expensive for lightweight CLI health/status/worktree operations.

## Suggested Fixes

1. Make `paseo daemon status` use a lightweight health check.
   - Prefer ping/server_info for daemon health.
   - If agent counts are needed, use `fetchAgents({ scope: "active" })` or a dedicated lightweight count API.
   - Do not mark the daemon unresponsive just because the full historical agent directory cannot be hydrated within 10s.

2. Make `paseo worktree ls` avoid full historical `fetchAgents({ includeArchived: true })`.
   - At minimum use `scope: "active"` for mapping active agents to worktrees.
   - Better: have the worktree-list RPC return associated active agent ids, or add a lightweight cwd-to-agent mapping API that does not refresh git metadata for every historical agent cwd.

3. Consider making the default fetch-agents path less eager about git refresh.
   - Use cached/peeked workspace snapshots for list placement when possible.
   - Add an explicit `refreshGit` option for clients that truly need fresh checkout metadata.
   - Apply structural filters that do not need project placement before resolving placement.

4. Consider storage/cache hygiene for externally deleted agent JSONs.
   - `AgentStorage` loads records into memory; deleting JSON files from disk alone does not remove them from a running daemon's in-memory cache. That is expected, but it made manual cleanup less intuitive during diagnosis.

## Why This Matters

This makes the daemon look unstable when it is actually reachable and processing realtime traffic. It also breaks `paseo worktree ls` for users who accumulate many historical agents/worktrees, because the command times out before it reaches the actual worktree-list RPC.

dafang:
## Summary

- make `paseo daemon status` fetch only active agents instead of archived agents
- avoid marking the daemon unresponsive when the websocket is reachable but agent counts fail
- make `paseo worktree ls` avoid the daemon worktree RPC and list managed worktrees locally with `git worktree list --porcelain`
- make `paseo worktree archive` resolve targets locally first, while still delegating the actual archive/delete operation to the daemon
- keep a daemon-list fallback for archive target resolution when local resolution misses or resolves a path the daemon rejects
- add coverage for parsing git worktree porcelain output and resolving worktrees by directory or branch name

## Why

`daemon status` and `worktree ls` previously fetched archived agents, which can force expensive project placement and git checkout refresh work when a daemon has many persisted agents. `worktree ls` also called the daemon worktree-list RPC, which computes checkout status/shortstat data and can become slow enough to time out.

`worktree archive` used the same daemon worktree-list RPC just to resolve `<name>` to a path before calling the archive RPC. This keeps the archive operation daemon-owned, but avoids the heavy list path in the common local case.

## Verification

- `npm run format:files -- packages/cli/src/commands/daemon/status.ts packages/cli/src/commands/worktree/ls.ts packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run format:files -- packages/cli/src/commands/worktree/archive.ts packages/cli/src/commands/worktree/ls.ts packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run format:check:files -- packages/cli/src/commands/daemon/status.ts packages/cli/src/commands/worktree/ls.ts packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run lint -- packages/cli/src/commands/daemon/status.ts packages/cli/src/commands/worktree/ls.ts packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run lint -- packages/cli/src/commands/worktree/archive.ts packages/cli/src/commands/worktree/ls.ts packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run typecheck --workspace=@getpaseo/cli`
- `node --import tsx packages/cli/tests/20-worktree-ls-paths.test.ts`
- `npm run build:daemon`
- pre-commit hooks: format, lint, full workspace typecheck

Fixes #1114

### DIFF
--- a/packages/cli/src/commands/daemon/status.ts
+++ b/packages/cli/src/commands/daemon/status.ts
@@ -242,7 +242,7 @@ async function probeDaemonOverWebsocket(args: {
   const supportsDaemonStatusRpc =
     client.getLastServerInfoMessage()?.features?.daemonStatusRpc === true;
   try {
-    const agentsPayload = await client.fetchAgents({ filter: { includeArchived: true } });
+    const agentsPayload = await client.fetchAgents({ scope: "active" });
     const agents = agentsPayload.entries.map((entry) => entry.agent);
     const runningAgents = agents.filter((a) => a.status === "running").length;
     const idleAgents = agents.filter((a) => a.status === "idle").length;
@@ -290,10 +290,9 @@ async function probeDaemonOverWebsocket(args: {
     return {
       connectedDaemon: "reachable",
       daemonVersion,
-      localDaemonOverride: state.running ? "unresponsive" : undefined,
       note: state.running
-        ? `Local daemon PID is running but API requests to ${host} failed`
-        : `Connected daemon websocket is reachable at ${host} but fetch_agents failed`,
+        ? `Local daemon PID is running and websocket is reachable at ${host}, but active agent counts are unavailable`
+        : `Connected daemon websocket is reachable at ${host} but active agent counts are unavailable`,
     };
   } finally {
     await client.close().catch(() => {});
--- a/packages/cli/src/commands/worktree/archive.ts
+++ b/packages/cli/src/commands/worktree/archive.ts
@@ -8,6 +8,7 @@ import type {
   OutputSchema,
   CommandError,
 } from "../../output/index.js";
+import { findManagedWorktreeByNameOrBranch, listManagedWorktrees } from "./ls.js";
 
 /** Result type for worktree archive command */
 export interface WorktreeArchiveResult {
@@ -34,6 +35,33 @@ export interface WorktreeArchiveOptions extends CommandOptions {
 }
 
 export type WorktreeArchiveCommandResult = SingleResult<WorktreeArchiveResult>;
+
+async function resolveWorktreeFromDaemon(client: DaemonClient, nameArg: string) {
+  const listResponse = await client.getPaseoWorktreeList({});
+
+  if (listResponse.error) {
+    const error: CommandError = {
+      code: "WORKTREE_LIST_FAILED",
+      message: `Failed to list worktrees: ${listResponse.error.message}`,
+    };
+    throw error;
+  }
+
+  return findManagedWorktreeByNameOrBranch(
+    listResponse.worktrees.map((worktree) => ({
+      worktreePath: worktree.worktreePath,
+      createdAt: worktree.createdAt,
+      branchName: worktree.branchName ?? null,
+      head: worktree.head ?? null,
+    })),
+    nameArg,
+  );
+}
+
+async function resolveWorktreeForArchive(client: DaemonClient, nameArg: string) {
+  const localMatch = findManagedWorktreeByNameOrBranch(await listManagedWorktrees(), nameArg);
+  return localMatch ?? (await resolveWorktreeFromDaemon(client, nameArg));
+}
 
 export async function runArchiveCommand(
   nameArg: string,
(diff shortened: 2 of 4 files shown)


boudra:
thanks for the pull request.

the diagnosis is right, but the worktree `ls` / `archive` part needs a daemon-owned rpc shape rather than local cli filesystem scanning. i do not want to ask you to design that boundary.

i would accept a smaller pr that only fixes `paseo daemon status`: use the active-agent fetch path, and do not mark the daemon unresponsive when the websocket is reachable but agent counts fail. i will handle the worktree rpc shape separately.

PR #1142, opened 2026-05-22
Guard assistant file link parsing against malformed URI sequences

tariqismail:
This fixes a crash where assistant-rendered file links containing malformed percent-encoding could throw URIError from decodeURIComponent during file-link classification. The parser now safely rejects malformed URI paths instead of crashing the workspace tab.

## Changes

- Added `safeDecodeURIComponent` wrapper that returns `null` instead of throwing on invalid percent-encoded sequences
- Replaced bare `decodeURIComponent` call in `normalizeFileUrlPath` with the safe wrapper
- Extracted absolute-href parsing in `parseAssistantFileLink` into `parseAbsoluteHrefAsFileLink` helper (keeps complexity within lint limit) and guarded the `decodeURIComponent` call there
- Added two regression tests: one for `parseFileProtocolUrl` and one for `parseAssistantFileLink` with malformed `%file` sequences

## Test

```
cd packages/app && npx vitest run src/assistant-file-links/parse.test.ts
# 36 tests pass
```


boudra:
thanks for the pull request.

PR #1146, opened 2026-05-23
fix(app): prune stale file-explorer paths during hydration

### LINKED ISSUE #1145: File explorer ENOENT-fails on stale expanded paths after external deletion (v0.1.80)
## Summary

The file explorer panel persists the list of expanded subdirectory nodes across sessions, but does not gracefully handle the case where one of those subdirectories has been deleted outside of Paseo. On reopening the panel for the workspace, the daemon attempts to `stat` each previously-expanded path and the panel fails to render with an `ENOENT` error, requiring a manual **Retry** click before the parent listing recovers.

## Environment

- App version: **0.1.80** (also reproduced on 0.1.78)
- OS: macOS (Darwin 25.5.0)
- Transport: `relay` (web UI at `https://app.paseo.sh` connecting to local daemon)

## Steps to reproduce

1. Open a workspace in Paseo (e.g. `/Users/me/project/claude-workspace`).
2. In the file explorer panel, expand some nested subdirectories, e.g. `some_dir/` → `sub_a/` → `sub_b/`.
3. Outside Paseo, delete one of the expanded directories: `rm -rf some_dir`.
4. Close the file explorer panel (or reload the app) and reopen it.

## Expected

Missing nodes are silently pruned from the persisted expanded-paths state. The panel renders the workspace root and any still-existing expanded subdirectories normally.

## Actual

The panel shows an error banner:

```
ENOENT: no such file or directory, stat '/Users/me/project/claude-workspace/some_dir/sub_a/sub_b'
```

with a **Retry** button. Clicking Retry falls back to listing only the workspace root, so contents become visible again — but the same error returns on every subsequent open of the panel until the stale paths are manually cleared from the front-end state.

## Daemon log excerpt

From `~/.paseo/daemon.log` — note that several stale paths are stat'd on each open, not just one:

```
msg: "Failed to fulfill file explorer request for workspace /Users/.../claude-workspace"
path: "some_dir"
path: "some_dir/sub_a"
path: "some_dir/sub_a/sub_b"
path: "some_dir/sub_c"
path: "some_dir/sub_d"
err.code: ENOENT
err.syscall: stat
```

## Workaround

Open devtools on the `app.paseo.sh` page and clear the relevant key(s) from `localStorage` (e.g. anything containing the stale path), then reload. UI preferences are lost but project/session data in `~/.paseo` is unaffected.

## Suggested fix

When hydrating the file-tree expanded state on panel open, filter out paths whose `stat` fails with `ENOENT` (instead of letting one ENOENT fail the whole panel render), and persist the pruned list back so subsequent opens are clean.

muzhi1991:
Fixes #1145.

## Problem

Opening the file-explorer panel for a workspace fires a directory listing for every entry in `expandedPathsByWorkspace`. When one of those subdirectories has been removed outside the app, the listing returns an ENOENT, which writes to the shared `lastError` and renders the whole panel as an error banner with a Retry button. Retry recovers the current session, but the failed path stays in `expandedPathsByWorkspace`, so the same error returns on the next open.

I hit this on macOS with 0.1.80 after deleting a directory I'd previously expanded. Once I confirmed the stale paths were sitting in `panel-state` localStorage and that pruning them by hand made the banner go away, the cause was clear.

## Fix

In the hydration `useEffect` of `FileExplorerPane`:

1. `await` the persisted listings instead of fire-and-forget.
2. Pass a new `silent` flag to `requestDirectoryListing` so a failed listing does not overwrite the panel's `lastError`.
3. Remove any failed paths from `expandedPathsByWorkspace` after hydration so the persisted state self-heals.

The interactive paths (toggle, Retry, ancestor expand, navigation) don't pass `silent`, so their error-reporting behaviour is unchanged.

A small pure helper `dropFailedExpandedPaths` is exported and unit-tested so the prune logic is covered without rendering the React tree.

## Verification

Only tested on **macOS desktop (relay transport)** — other platforms not exercised.

- `npm run typecheck --workspace=@getpaseo/app` ✓
- `npm run format:check` ✓
- `npm run lint --workspace=@getpaseo/app -- src/hooks/use-file-explorer-actions.ts src/hooks/use-file-explorer-actions.test.ts src/components/file-explorer-pane.tsx` — no new findings (8 pre-existing warnings in `file-explorer-pane.tsx` on `main` are unrelated)
- `npx vitest run src/hooks/use-file-explorer-actions.test.ts` ✓ 4/4

Bug repro & root-cause confirmation:

1. Inspected `panel-state` in localStorage; found the stale paths under the workspace key.
2. Cross-checked `~/.paseo/daemon.log`: every open of the panel emits `Failed to fulfill file explorer request for workspace … path: <stale-path>` for every stale entry.
3. Manually pruning the stale entries from `expandedPathsByWorkspace` made the banner stop appearing — confirming that the persisted state is the trigger, which is exactly what this patch removes automatically on hydration.

I didn't rebuild the Electron desktop app to run the patched code end-to-end, but the patch only changes the front-end hydration flow and is covered by the unit test plus the manual root-cause evidence above.

### DIFF
--- a/packages/app/src/components/file-explorer-pane.tsx
+++ b/packages/app/src/components/file-explorer-pane.tsx
@@ -40,8 +40,11 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
-import { useFileExplorerActions } from "@/hooks/use-file-explorer-actions";
-import { buildWorkspaceExplorerStateKey } from "@/hooks/use-file-explorer-actions";
+import {
+  buildWorkspaceExplorerStateKey,
+  dropFailedExpandedPaths,
+  useFileExplorerActions,
+} from "@/hooks/use-file-explorer-actions";
 import { usePanelStore, type SortOption } from "@/stores/panel-store";
 import { formatTimeAgo } from "@/utils/time";
 import { buildAbsoluteExplorerPath } from "@/utils/explorer-paths";
@@ -929,7 +932,7 @@ async function initializeExplorer({
   workspaceStateKey: string | null;
   requestDirectoryListing: (
     path: string,
-    opts?: { recordHistory?: boolean; setCurrentPath?: boolean },
+    opts?: { recordHistory?: boolean; setCurrentPath?: boolean; silent?: boolean },
   ) => Promise<boolean>;
 }): Promise<void> {
   if (!hasWorkspaceScope || hasInitializedRef.current) {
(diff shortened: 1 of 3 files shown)


muzhi1991:
@boudra rebased onto the latest `main` — resolved the conflict from the i18n migration and kept the `silent` guard consistent across the new `t(...)` error strings.

Still reproduces on the **0.1.96** release: expand a subdirectory in the file explorer, delete that directory outside the app, then reopen the panel. The whole tree fails with `ENOENT … stat '<deleted path>'` and only **Retry** clears it, because the stale path is replayed from the persisted `expandedPathsByWorkspace` on every open.

This PR makes the hydration listings `silent` and prunes the dead paths from the persisted set, so the panel renders normally and self-heals on the next open. Interactive paths (toggle, Retry, ancestor expand, navigation) keep their existing error reporting. Verified locally with the new unit test plus `typecheck` / `lint` / `app-tests`; CI was green on the pre-rebase commit and is re-running now.

Issue #648, opened 2026-05-01
cannot select and copy words in phone app


KitschYoung:
cannot select and copy words in phone app

kongjiadongyuan:
@boudra I’m also hitting this in real iOS mobile use on both iPhone and iPad, and I wanted to clarify the exact workflow I’m looking for.

The ideal long-term solution is probably what #21 describes: truly selectable native Markdown. But that feels like a larger renderer project.

For this issue, the smaller thing I need as a mobile user is:

- I’m reading an assistant response on iPhone or iPad
- I want to copy one command, path, error snippet, or a few words
- copying the entire message is often too coarse
- I don’t necessarily need rich Markdown selection in-place as the first step

I looked at how https://github.com/slopus/happy handles this. Their workaround is pragmatic:

- long-press a Markdown message on mobile
- open a dedicated text-selection screen/sheet
- show the raw message text in a read-only native TextInput
- let iOS/Android handle normal text selection/copy there
- keep copy-all as a shortcut

For Paseo specifically, another discoverable entry point could be a small “select/expand text” action next to the existing copy icon on assistant messages. That would make the feature visible without relying only on long-press.

Would you be open to a focused interim PR that adds a mobile-only “Select text” / “Expand text” action for assistant messages using this pattern?

It would not replace #21, and I’d keep the scope narrow: no Markdown renderer rewrite, no WebView/native rich text work, just a reliable escape hatch for selecting/copying partial assistant text on mobile.

Issue #238, opened 2026-04-10
can't select text and copy at ios


atian25:
long press is not responsible

Issue #21, opened 2026-02-06
Investigate native selectable Markdown w/ tables via Compose/KMP + RN wrapper


boudra:
Context
- We shipped assistant message rendering via Expo DOM/WebView (PR #17) to get fully-selectable Markdown (tables/lists/code) on mobile.
- Longer-term, we want the same UX as Claude/ChatGPT: snappy + native selection + tables, ideally without a WebView per message.

Findings
1) Claude Android app appears to use native rich text for chat + WebView for special surfaces
- APK inspected: `com.anthropic.claude` versionName `1.260202.27` (file: `/tmp/com.anthropic.claude-1.260202.27-APK4Fun.com.apk`).
- Strong signals of Jetpack Compose + native selectable text: Compose `AnnotatedString` + `SelectableText` nodes, markdown parsing pipeline (`MarkdownRichText`, `getMarkdownAst`, `MarkdownParseOptions`).
- Syntax highlighting appears to be tokenization via shipped JS (`assets/highlight.min.js`, `assets/token-highlight.js`) + applying token ranges natively (also includes `androidx.javascriptengine.*`).
- WebView usage exists but seems oriented around artifact/sandbox surfaces (`SandboxWebView`, artifact sheet, embedded HTML templates), not necessarily per chat bubble.

2) React Native has a known selection limitation
- Most RN Markdown/HTML renderers produce many `<Text>` nodes; text selection cannot span across multiple text surfaces, making “select entire message with tables/lists/code” hard without a single native surface or WebView.

3) Kotlin/Compose solution exists: `mikepenz/multiplatform-markdown-renderer`
- Compose Multiplatform Markdown renderer with GFM support incl tables (>=0.30.0).
- Tables are implemented as layout (rows/cells) and horizontal-scroll when wider than available width.

4) Potential wrapping approaches (RN <- Compose)
- `voize-gmbh/reakt-native-toolkit`: exposes Compose Multiplatform composables to React Native via `@ReactNativeViewManager` (experimental).
- `mgcrea/react-native-jetpack-compose`: Android-only Compose components exposed to RN via Fabric (New Architecture).

Suggested next step (spike)
Goal: prototype a “NativeMarkdownView” that renders assistant transcripts natively with selection + tables, starting Android-first.

- Android:
  - Implement a small native module/view that hosts a Compose view (or Compose Multiplatform), rendering markdown using `multiplatform-markdown-renderer`.
[...]

PR #1153, opened 2026-05-24
fix(app): native iOS text selection in assistant messages via UITextView

muzhi1991:
## Why

Long-press text selection on iOS assistant messages doesn't work — users can only "copy the whole message" via the button, not pick out a command, path, error snippet, or a phrase. Reported in #238 and #648. This is more painful than it sounds when you're vibe-coding from mobile and want to grab one line out of a long agent response.

## Root cause

RN's `<Text>` renders as `UILabel` on iOS, which doesn't support word selection / drag handles — only "copy the entire block". macOS and web work because CSS `user-select: text` is the default.

## What this PR does

Swap the inline `<Text>` used by the markdown renderer for [`react-native-uitextview`](https://github.com/bluesky-social/react-native-uitextview)'s `<UITextView uiTextView selectable>`, which renders as a real `UITextView` on iOS (native word selection + drag handles + system menu) and falls back to base `<Text>` on Android/web.

`react-native-uitextview` is a tiny MIT library (zero runtime deps, ~400★) maintained by Bluesky and used in production by their app for the exact same problem. Smallest patch I could find that gives genuine UITextView behavior without a renderer rewrite.

## Why not other approaches

- **PR #17 / Expo DOM / WebView** — closed by maintainers as "park[ed]… in favor of native"; perf/integration cost too high.
- **#21 / Compose Multiplatform + KMP** — large architectural change, iOS path via \`reakt-native-toolkit\` is experimental, no PRs in 3+ months. This PR could take some pressure off that long-term effort.
- **#648's "selection sheet"** — separate screen to view raw text is a UX regression vs native selection.
- **\`TextInput multiline editable={false}\`** — gives native selection but loses all markdown formatting (TextInput can't render nested styled Text children).

## Files changed

- \`packages/app/package.json\` — add \`react-native-uitextview@^2.2.0\`
- \`packages/app/src/components/message.tsx\`
  - \`MarkdownInheritedText\` (text / textgroup / code_inline rules): always \`UITextView\` (auto-falls back to Text on non-iOS).
  - \`MarkdownParagraphView\`: iOS uses \`UITextView\`; other platforms keep \`View\` (the library's non-iOS fallback to Text would break Android, since Text can't contain View children).
- \`packages/app/src/components/highlighted-code-block.tsx\`
  - \`<Text>\` wrapping code/tokens → \`<UITextView uiTextView selectable>\` so users can select substrings inside a code block.

Diff is small (+37 / -4 across 4 files including lockfile).

## Known limitation

Selection cannot cross paragraph or code-block boundaries — each block is an independent \`UITextView\` and iOS native selection can't span sibling native views. Selection within a single paragraph (including across inline \`**bold**\` / \`*italic*\` / \`\` \`code\` \`\` / \`[link](url)\`) or within a single code block works.

Supporting cross-block selection would require rendering an entire message as a single \`UITextView\`, losing block-level layout (list indent, code-block padding, etc.) — not worth the trade-off in my opinion.

## Cross-platform safety

\`react-native-uitextview\` source confirms a hard \`Platform.OS !== 'ios'\` early-return that yields base \`RNText\` — the native components are never instantiated off-iOS:

\`\`\`tsx
export function UITextView(props) {
  if (Platform.OS !== 'ios') {
    return <RNText {...props} />        // Android / web / desktop
  }
  return <UITextViewInner {...props} />  // iOS only
}
\`\`\`

Additionally \`MarkdownParagraphView\` has its own \`Platform.OS === 'ios'\` guard, so on Android/web/desktop the original \`<View>\` path is preserved (important because paragraph children can include image \`<View>\` nodes that wouldn't be valid Text children).

Per-platform behavior summary:

| Platform | Behavior after this PR | Mechanism |
|---|---|---|
| iOS | ✅ Native UITextView word selection | UITextViewInner |
| Android | Unchanged — base RN Text (same as before) | library Platform guard |
| Web (react-native-web) | Unchanged — base RN Text rendered with default CSS \`user-select: text\` (same as before) | library Platform guard |
| Desktop (Electron) | Unchanged — runs web bundle, same as Web | same as Web |

Bluesky's social-app uses this library across iOS / Android / RN-web in production, so the cross-platform path is battle-tested upstream.

## Architecture notes

- \`react-native-uitextview@2.x\` requires RN New Architecture. paseo already has \`newArchEnabled: true\` ✓
- No \`<View>\` children flow into \`<UITextView>\` — verified that \`AssistantMarkdownLink\` / \`AssistantMarkdownCodeLink\` / \`AssistantInlineCodePathLink\` all render as \`<Text>\` on native.

## Disclosure

I'm not an iOS engineer — this fix came out of digging through the existing issues / your \`message.tsx\` rules with the help of an LLM to identify the right library, then verifying behavior in the simulator myself. Diff is small and contained, but I'd appreciate a sanity check from someone with deeper RN/iOS context, especially on:

1. Whether \`UITextView\` interacts cleanly with existing handlers (\`onPress\` on \`AssistantMarkdownLink\`, \`CopyButton\` overlay in \`HighlightedCodeBlock\`).
2. Anything in your style pipeline that might assume \`<Text>\` vs \`UITextView\`.

## Test plan

Verified on **iPhone 17 Pro Simulator (iOS 26.5)** with a built dev client:

- [x] Long-press on assistant paragraph → drag handles, can select arbitrary word/range within paragraph
- [x] Selection spans inline \`**bold**\` / \`*italic*\` / \`\` \`code\` \`\` / \`[link](url)\` within the same paragraph
- [x] Long-press on syntax-highlighted code block → can select substrings, copy works
- [x] System menu (Copy / Look Up / Translate / Share) appears
- [x] User messages unchanged (already used \`<Text selectable>\`)
- [x] TypeScript typecheck passes; pre-commit lint/format pass
- [x] **Web bundle verified**: \`expo export --platform web\` succeeds (9.76 MB index.js, no \`codegenNativeComponent\` errors); library's Platform guard prevents any non-iOS native call.
- [ ] Not exercised on Android or in a running web/desktop session — but no behavior change is expected (see Cross-platform safety above), and the upstream library is in production at Bluesky on all those platforms.
- [ ] Not verified on physical iOS device (I don't have a paid Apple Developer account).

## Screenshot

Simulator Screenshot - iPhone 17 Pro - 2026-05-24 at 10 25 35

Refs #21, #238, #648


muzhi1991:
Updated to fix the playwright CI failure.

The web **dev** bundle was failing because importing `react-native-uitextview` transitively pulls in `react-native/Libraries/Utilities/codegenNativeComponent` → `setUpReactDevTools` → `ReactDevToolsSettingsManager`, which doesn't resolve under the web target in dev mode. (Production `expo export --platform web` tree-shakes dev tools, which is why my initial local web-bundle check passed but playwright — running the web dev bundle — failed.)

Wrapped the import behind a platform-split shim following the existing `composer-height-mirror.{native,web,d}.ts` convention in this repo:

- `packages/app/src/components/selectable-text.native.tsx` — re-exports `react-native-uitextview`
- `packages/app/src/components/selectable-text.web.tsx` — falls back to base RN `Text` (which already renders with default `user-select: text` on web)
- `packages/app/src/components/selectable-text.d.ts` — TS module shim

`message.tsx` and `highlighted-code-block.tsx` now import from `@/components/selectable-text` instead of `react-native-uitextview` directly, so the web bundle never touches the upstream library.

Re-verified locally:

- TypeScript typecheck ✓
- `expo export --platform web` ✓
- `expo export --platform ios` ✓ (13.8 MB iOS bundle)
- pre-commit hook (lint + format + typecheck) ✓
- iPhone 17 Pro Simulator behavior unchanged

muzhi1991:
Hi @boudra — sorry to ping, just wanted to surface this when you have a moment.

Quick recap: small PR (+62/−4) addressing #238 / #648 using the same approach Bluesky ships in production (`react-native-uitextview`). All CI is green; behavior verified on iPhone 17 Pro Simulator (screenshot in the description). The web and Android paths are gated so off-iOS behavior is unchanged — Android falls back to base RN Text via the library's own `Platform.OS` guard, web uses a `.web.ts` shim that never imports the upstream package.

Mostly looking for a maintainer take on whether this is a direction you'd accept in principle before going deeper on iteration. Happy to adjust naming, file layout, scope, etc. — and totally fine to wait if there are deeper architectural concerns I should know about (e.g. if you'd rather hold off until #21 lands and avoid an interim selection-text component).

No rush, just flagging so it doesn't get lost in the queue. Thanks!

PR #1184, opened 2026-05-26
fix(app): archive agent before closing tab to prevent race condition

### LINKED ISSUE #1182: bug: Closing last empty agent tab creates multiple "New Agent" tabs
## Bug

When closing the last empty agent tab in a workspace, multiple "New Agent" tabs are automatically created instead of just one (or none, if the user intentionally closed all tabs).

## Expected behavior

Closing the last tab should either:
- Leave the workspace in a clean empty state, or
- Create at most **one** "New Agent" draft tab

## Steps to reproduce

1. Open a workspace with a single empty agent tab
2. Close that tab
3. Observe multiple "New Agent" tabs appearing

## Root cause analysis

Race condition between synchronous tab close and async agent archive in `handleCloseAgentTab` (`workspace-screen.tsx`):

1. `closeWorkspaceTabWithCleanup()` removes the tab **synchronously** from the layout store
2. `archiveAgent()` fires **asynchronously** — the agent remains in the session store
3. React re-render triggers `reconcileWorkspaceTabs`, which calls `addMissingEntityTabs`
4. `addMissingEntityTabs` sees the agent still in `autoOpenAgentIds` but with no tab → creates a new tab
5. When archive completes, `collapseStaleEntityTabs` removes the auto-created tab
6. Now all counts are 0, so `shouldSeedEmptyWorkspaceDraft` fires → creates yet another "New Agent" tab
7. Net result: multiple unwanted tabs appear

### Key files
- `packages/app/src/screens/workspace/workspace-screen.tsx` — `handleCloseAgentTab` (L2321), reconciliation effect (L1849), empty seed effect (L1984)
- `packages/app/src/stores/workspace-layout-actions.ts` — `addMissingEntityTabs` (L1681), `reconcileWorkspaceTabs` (L1732)
- `packages/app/src/workspace-tabs/agent-visibility.ts` — `deriveWorkspaceAgentVisibility`

## Related

- Issue #94: "Multiple drafts created when opening a new workspace with no agents" — may share the same underlying cause

vlln:
## Summary

Fixes #1182 — when closing the last empty agent tab, multiple "New Agent" tabs were automatically created instead of just one.

## Root cause

Race condition in `handleCloseAgentTab`: `closeWorkspaceTabWithCleanup` removed the tab **synchronously** while `archiveAgent` ran **asynchronously**. The reconciliation effect fired between the two operations, saw the agent still in `autoOpenAgentIds` but missing from layout, and re-created the tab. After the archive completed, the auto-created tab was collapsed, triggering the empty workspace seed to create yet another "New Agent" draft.

## Fix

Reverse the order: `await archiveAgent()` first, then `closeWorkspaceTabWithCleanup()`. This ensures the agent is already archived when the layout updates, so reconciliation won't re-create the tab.

If archive fails, the tab is not closed — better UX than the previous behavior of closing the tab and silently failing the archive.

## Test plan

- [x] Typecheck (app package — pre-existing errors in CLI/terminal are unrelated)
- [x] Lint: 0 warnings, 0 errors
- [x] Format: passed
- [x] Unit tests: 7/7 passing (close-tab-policy, workspace-empty-draft-seed, workspace-subagents-integration)
- [x] Manual test: rebuilt desktop web bundle, verified closing last empty agent tab creates only one "New Agent" draft on desktop (Linux)

### DIFF
--- a/packages/app/src/screens/workspace/workspace-screen.tsx
+++ b/packages/app/src/screens/workspace/workspace-screen.tsx
@@ -2351,19 +2351,22 @@ function WorkspaceScreenContent({
 
         setHoveredTabKey((current) => (current === tabId ? null : current));
         setHoveredCloseTabKey((current) => (current === tabId ? null : current));
+
+        if (closePolicy.kind === "archive-on-close") {
+          try {
+            await archiveAgent({ serverId: normalizedServerId, agentId });
+          } catch {
+            // If archive fails, don't close the tab.
+            return;
+          }
+        }
+
         if (persistenceKey) {
           closeWorkspaceTabWithCleanup({
             tabId,
             target: { kind: "agent", agentId },
           });
         }
-
-        if (closePolicy.kind === "layout-only") {
-          return;
-        }
-
-        // Errors (e.g. timeout) are handled by the mutation's onSettled callback
-        void archiveAgent({ serverId: normalizedServerId, agentId }).catch(() => {});
       });
     },
     [archiveAgent, closeTab, closeWorkspaceTabWithCleanup, normalizedServerId, persistenceKey],


boudra:
thanks for the pull request.

PR #1204, opened 2026-05-28
feat(app): add Archived Sessions entry for workspace recovery

zhaoshe:
## Summary

Closing a root-agent tab archives the agent — that's the documented behavior ([docs/agent-lifecycle.md](https://github.com/getpaseo/paseo/blob/main/docs/agent-lifecycle.md#tabs-vs-archive) line 51). But Import session hides every sessionId already known to paseo (via `collectImportedProviderSessionHandles` in `packages/server/src/server/agent/import-sessions.ts`), so a freshly archived session has no obvious way back inside the workspace where it lived.

The existing recovery path is: left sidebar → Sessions → find the row (Archived badge) → tap in → pull to the bottom → Unarchive. Most users don't discover this.

## Change

Adds an **"Archived Sessions(N)"** pill next to the existing **Import session** pill on the New agent draft tab. N counts archived agents whose cwd matches the workspace, sourced from `fetch_agent_history` data **already in the React Query cache** (populated by the archived-sessions sheet or the sidebar Sessions screen). Tapping a row calls the existing `refresh_agent_request` RPC to unarchive and immediately opens the agent as a tab via the same handler used by Import session.

The pill stays hidden when N === 0 so it doesn't add noise.

### Why cache-only and not an eager fetch

Earlier rounds tried two endpoints:

- **Eager fetch from workspace-screen** (flagged P2): "Every focused workspace screen issues a background history fetch even for users who never interact with archived sessions."
- **Always-visible pill without count** (flagged P1): The pill rendered in every workspace, including ones with no archived sessions — the exact noise the spec aimed to avoid.

This iteration reads via `useAgentHistory({ enabled: false })`, which subscribes to the React Query cache for `agentHistoryQueryKey(serverId)` but never triggers a fetch. The cache is filled the first time the user opens the archived-sessions sheet or visits the sidebar Sessions screen. Until then the count is 0 and the pill stays hidden. After that, every workspace activation reads from the same cache without extra network cost.

The discoverability trade-off is intentional: the sidebar Sessions screen is the always-available entry point that bootstraps the cache; the draft-tab pill is a fast-path reminder that only surfaces once the user has stepped into the feature.

## Files

- `packages/app/src/components/archived-sessions-sheet.tsx` (new) — sheet that lists archived agents for current cwd and dispatches `refreshAgent` on row tap; invalidates `agentHistoryQueryKey` so the count refreshes immediately. Separates the restore and tab-open phases so a tab-open failure doesn't misreport as a restore failure. Resets transient row state when the sheet hides.
- `packages/app/src/composer/draft/archived-pill.tsx` (new) — pill button mirroring `ComposerImportPill` style, renders `Archived Sessions(N)`.
- `packages/app/src/composer/draft/workspace-tab.tsx` — renders both pills side-by-side via the `DraftComposerPillRow` sub-component (keeps `WorkspaceDraftAgentTab` under the 20-complexity cap). `resolveArchivedPillPress` gates on `count > 0`.
- `packages/app/src/panels/pane-context.tsx`, `screens/workspace/workspace-pane-content.tsx`, `panels/agent-panel.tsx` — thread `openArchivedSheet` + `archivedSessionCount` alongside the existing `openImportSheet`.
- `packages/app/src/screens/workspace/workspace-screen.tsx` — sheet visibility state; uses `useAgentHistory({ enabled: false })` to read the cache without triggering a fetch; reuses `handleImportedAgent` so the post-restore flow (open tab, focus it) matches the import flow exactly.

## Test plan

- [x] Open a workspace with at least one archived agent for the cwd, then visit the sidebar Sessions screen (or open the archived sheet via any other entry) so the cache populates → "Archived Sessions(N)" pill shows next to "Import session" on the New agent draft tab
- [x] Tap the pill → sheet lists the archived agents, sorted by lastActivityAt desc
- [x] Tap a row → sheet closes, agent restores, becomes the focused tab; subsequent restore-success → count decrements
- [x] Workspace with no archived sessions (cache populated, but matching cwd has 0 archived) → pill hidden
- [x] Workspace before any cache load → pill hidden; opening the sidebar Sessions screen or the sheet itself populates the cache and the pill reappears
- [x] Submitting the draft composer (`isSubmitting=true`) → pill suppressed like Import session
- [x] Reopening the sheet after a restore error → stale "Could not restore" banner cleared


zhaoshe:
Pushed 9e851c71 to address the new P1.

The previous round dropped the count display to neutralise the older P2 about the eager fetch — that left the pill visible in every workspace, which Greptile correctly flagged as defeating the original "hide when N === 0" intent.

This round wires the count back in but sources it from the React Query cache only (`useAgentHistory({ enabled: false })`). No fetch is triggered from workspace-screen; the cache is filled by the archived-sessions sheet itself when opened, and by the sidebar Sessions screen. Until then `count === 0` and the pill stays hidden — which is exactly what the spec wanted. The PR description has been updated to spell out the discoverability trade-off.

The other Greptile threads (stale error state, misleading error attribution, redundant `new Date()` wrap) were addressed in the previous push (`d81a2434`).

zhaoshe:
Pushed 5387d62a for the cold-start P1.

You're right — the cache-only read was circular: empty cache on a fresh session → count stays 0 → pill never renders → sheet (the only cache filler reachable from the draft tab) can't be opened. Dead loop for exactly the users who need recovery.

Took your suggested fix: the count query now runs with `enabled: isRouteFocused`, so the first focus of a workspace issues one fetch and populates the cache. To keep the earlier P2 (per-activation / 30s revalidation cost) from coming back, `useAgentHistory` now accepts an optional `staleTimeMs` and workspace-screen passes 5 minutes — so repeat activations read the cache instead of refetching. The sheet keeps its own 30s-staleTime observer, so opening it still pulls a fresh list.

zhaoshe:
@boudra ready for review when you have a moment. All Greptile feedback is addressed (last P1 on the cold-start/zero-archive pill logic resolved in 5387d62a), and the branch is merged up to date with main. No outstanding threads.

zhaoshe:
@boudra When you have a moment, could you take a look at this one? It's mergeable with no conflicts, and the Greptile review feedback has been addressed. Thanks!

boudra:
Closing in favor of the archived-session recovery implementation in [#4033](https://github.com/getpaseo/paseo/pull/4033).

PR #1243, opened 2026-05-31
perf(app): idle the working indicator when no agent is running

fireblue:
### Linked issue

_No filed issue — found while dogfooding the desktop app on a high-refresh (180Hz) display._

### Type of change

- [x] Bug fix (performance / battery)

### What does this PR do

`SyncedLoader` — the dot-snake "working" indicator shown in the turn footer, sidebar workspace list, and tabs — starts a module-level `withRepeat(..., -1)` Reanimated loop via a `sharedLoopStarted` flag and **never tears it down**. Once the first working indicator appears, the loop runs for the app's whole life, driving `requestAnimationFrame` at the display's refresh rate (120/180Hz) continuously — even after every loader unmounts and no agent is working.

Two changes, both in `synced-loader.tsx`:

- **Stop when idle:** reference-count mounted loaders and cancel the driver when the last one unmounts, so the renderer idles when no working indicator is on screen.
- **Don't over-sample:** drive the snake's head with a low-rate, clock-derived stepper (~`DOT_COUNT`/sec) instead of a per-frame `withTiming`. The dot opacities are already a step function of `Math.floor(progress)`, so the motion is **visually identical** while producing ~6 updates/sec instead of one per display frame. The fixed epoch keeps all loaders in phase across mounts/clients.

### How did you verify it

- **Root cause:** `sharedStepProgress` / `sharedLoopStarted` are only ever started, never cancelled — the mount effect had no cleanup. Confirmed `SpinnerDot` opacity is a pure step function of `Math.floor(progress.value)`, so stepping the shared value discretely (rather than per-frame) is visually lossless.
- App `typecheck` + `lint` clean; `format` ran.
- Built the desktop app with the change and installed it for on-device dogfooding on the affected display; the spinner motion looks unchanged. (Power/refresh confirmation is via dogfooding — no synthetic fps claim here.)

### Checklist

- [x] One focused change (synced-loader.tsx only).
- [x] `typecheck` passes (app workspace)
- [x] `lint` passes
- [x] `format` ran (Biome)
- [ ] UI change: motion is visually identical (the underlying opacity step function is unchanged) — dogfooding the installed build rather than attaching a video
- [ ] Tests: no dedicated test (timer/animation-driven); the change is mount/unmount ref-counting + discrete step derivation

### DIFF
--- a/packages/app/src/components/synced-loader.tsx
+++ b/packages/app/src/components/synced-loader.tsx
@@ -1,12 +1,5 @@
 import { View } from "react-native";
-import Animated, {
-  Easing,
-  makeMutable,
-  type SharedValue,
-  useAnimatedStyle,
-  withRepeat,
-  withTiming,
-} from "react-native-reanimated";
+import Animated, { makeMutable, type SharedValue, useAnimatedStyle } from "react-native-reanimated";
 import { useEffect, useMemo } from "react";
 
 const SYNCED_LOADER_DURATION_MS = 950;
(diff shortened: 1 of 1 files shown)


boudra:
thanks for the pull request.

PR #1258, opened 2026-05-31
fix(server): stop the daemon supervisor from hot-looping on permanent crashes

fireblue:
## Problem

A desktop/CLI-managed Paseo daemon could peg a CPU core indefinitely. Observed in the wild: load average **31**, a `Paseo Daemon` worker respawning ~2×/second for **9+ hours**.

Root cause: the supervisor restarts its worker on **every** non-zero exit with **no backoff and no limit**. When the worker can never start — e.g. it hits `EADDRINUSE` because an orphaned daemon (a worker that outlived its supervisor and reparented to launchd) still owns `127.0.0.1:6767` — that unrecoverable failure becomes an infinite hot respawn loop, each cycle a full daemon bootstrap thrown away in milliseconds.

## Fix

A crash-loop circuit breaker, expressed as pure, unit-tested helpers so the decision logic is testable without forking:

- `decideRestart(crashTimestamps, now, policy)` — windowed crash count → `{ shouldRestart, delayMs }` with exponential backoff.
- `planWorkerExit(input)` — folds the existing shutdown / intentional-restart / crash / clean-exit branches together with the breaker into one `ExitPlan`.
- `supervisor.ts` close handler executes the plan: backoff via `setTimeout`, give up (clean non-zero exit) after `maxRestarts` crashes within the window, and cancel a pending restart on shutdown.

Default policy: **5 restarts within 10s**, exponential backoff (200ms → 5s cap). An unrecoverable failure now retries ~5× over ~6s, then the supervisor exits with a clear `Worker crash-looped …` log instead of looping forever. Intentional restarts and clean/SIGTERM exits are unaffected.

## Tests

- `restart-policy.test.ts` — unit tests for `decideRestart` / `isCrash` / `planWorkerExit` (backoff, give-up, window expiry, intentional-restart bypass).
- `supervisor.logging.test.ts` — subprocess integration test: an always-crashing worker backs off twice, then the supervisor gives up with exit 1 and logs `Worker crash-looped (3 crashes within 5000ms)`.

## Scope / follow-up

This is the defense-in-depth half. A follow-up will add **cooperative handoff** — the worker detects `EADDRINUSE` from a healthy Paseo daemon and yields; the desktop attaches to the existing daemon — so a collision resolves on the first attempt instead of after the backoff budget.

Known limitation: intentional restarts (`paseo:restart` IPC) bypass the breaker by design; they're externally triggered so they don't self-loop, but bounding their frequency could be a later hardening step.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -140,6 +140,12 @@ The supervisor rotates `daemon.log`. Persisted `log.file.rotate` settings in
 `PASEO_LOG_ROTATE_SIZE` and `PASEO_LOG_ROTATE_COUNT` env vars override the
 defaults. The default rotation is `10m` x `3` files everywhere.
 
+The supervisor also bounds crash restarts. When the daemon worker exits
+non-zero repeatedly, the supervisor applies exponential backoff and then exits
+instead of respawning forever. Look for `Worker crash-looped` in
+`daemon.log` when a desktop or CLI-managed daemon stops after repeated startup
+failures such as a stale port owner.
+
 ## paseo.json service scripts
 
 `worktree.setup` and `worktree.teardown` accept either a multiline shell script or an array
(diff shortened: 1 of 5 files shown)


boudra:
thanks for the pull request.

PR #1259, opened 2026-05-31
fix(server): yield to an existing healthy daemon on EADDRINUSE

fireblue:
## Problem

When a Paseo daemon worker can't bind its listen port because **a healthy daemon already owns it** — most commonly an orphaned daemon still serving `127.0.0.1:6767` — the worker throws on `listen`, the supervisor treats it as a crash, and (without #1258) respawns forever.

## Fix

On `EADDRINUSE`, the worker probes the occupant's unauthenticated `/api/health` endpoint. If a live Paseo daemon answers, the worker logs and **exits cleanly (code 0)** so the supervisor stops without restarting — the existing daemon keeps serving its agents. Any other listen failure (foreign process, unix socket) falls through to the original throw.

- `daemon-port-conflict.ts` — `isAddressInUseError` + `isHealthyPaseoDaemonAt` (probe `/api/health`, expects 200 + `{status:"ok"}`), unit-tested against real ephemeral HTTP servers.
- `daemon-worker.ts` — thin wiring in the start-`catch`; the yield runs before the SIGTERM handlers register, and the code-0 exit means the supervisor won't restart.

## Relationship to #1258

Complementary and independent (different files; branches off `main`):
- **#1258** (circuit breaker) bounds *any* crash loop — the safety net.
- **This** resolves the common port collision on the **first** attempt with zero wasted restarts — the cooperative path.

Together: a collision yields immediately; anything unexpected is still bounded.

Known limits (non-blocking): unix-socket listen targets aren't probed; an IPv6-literal listen host isn't bracketed in the probe URL (default is IPv4 `127.0.0.1`). Both fall through safely.

### DIFF
--- a/packages/server/src/server/bootstrap.ts
+++ b/packages/server/src/server/bootstrap.ts
@@ -9,6 +9,7 @@ import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/
 import { isInitializeRequest } from "@modelcontextprotocol/sdk/types.js";
 import type { Logger } from "pino";
 import { createBranchChangeRouteHandler } from "./script-route-branch-handler.js";
+import { PASEO_HEALTH_HEADER, PASEO_HEALTH_HEADER_VALUE } from "./daemon-port-conflict.js";
 
 export type ListenTarget =
   | { type: "tcp"; host: string; port: number }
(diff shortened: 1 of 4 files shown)


boudra:
thanks for the pull request.

Issue #1264, opened 2026-05-31
feat: show Markdown source from preview tabs


damselem:
## Prior search

- [x] I searched existing issues and discussions, and this isn't already proposed.

## Problem

Paseo renders Markdown files directly in file tabs. That is useful for reading generated reports, READMEs, and agent artifacts, but it leaves no in-tab way to inspect or copy the raw Markdown source.

Users currently have to open the file through another editor/context or rely on line-targeted links to get the highlighted source view. That makes it harder to review generated Markdown, debug formatting, or copy exact source text while staying in the workspace tab.

## Example

An agent writes `client_report.md`. Paseo opens it as a rendered report preview. The user wants to check the raw Markdown around a blockquote, list, or table without leaving the tab.

Today the tab menu only has generic tab actions, so the rendered preview cannot be switched to source from the file tab itself.

## Expected behavior

Markdown file tabs should offer a tab-menu action to switch the same tab between rendered preview and source view.

In preview mode, the menu should show `Show source`.

In source mode, the menu should show `Show preview`.

The mode should belong to the file tab state, not create a second tab for the same file.

## Actual behavior

Rendered Markdown file tabs only show the preview. There is no menu action to inspect the raw Markdown source from that tab.

## Proposed direction

Add an explicit file tab render mode, such as `preview` or `source`, defaulting to `preview` for existing file open paths and persisted tabs. Show the toggle only for rendered Markdown tabs that are not line-targeted, because line-targeted Markdown links already need the highlighted source view.

## Related PR

Draft PR: https://github.com/getpaseo/paseo/pull/1263

PR #1267, opened 2026-06-01
fix(app): selectable text in file preview (code + markdown), with large-file OOM guard

muzhi1991:
## Why

iOS file preview had no usable text selection: code files only did whole-line copy, and rendered Markdown wasn't selectable at all. #447 (merged) added line-level `selectable` to code files and explicitly left Markdown as a follow-up — this is that follow-up, plus word-level selection for code, reusing the UITextView-backed spans from #1153.

## What this PR does

**Code files**: render the whole file as ONE UITextView (lines joined by `\n`) + a separate line-number gutter column + an absolute line-highlight overlay, so iOS selection can span multiple lines (the old per-line `<Text>` couldn't select across rows). Mobile and desktop both horizontal-scroll (no wrap) so the gutter and overlay stay row-aligned — matching the existing `DiffViewer`.

**Markdown files**: custom `rules` route `text` / `paragraph` / `code_inline` / `fence` through the selectable spans (`MarkdownTextSpan` / `MarkdownParagraphView`); headings, lists and quotes inherit selectability via `textgroup`.

## Large-file handling (OOM guard) — please read

The rendered-Markdown path emits **one UITextView per block** (paragraph / heading / list item). A large document spawns hundreds of them, and on iOS that **OOM-kills the app** — confirmed during testing: opening a large `.md` terminated the app to the home screen with **no crash report**, the signature of a jetsam OOM kill (`pendingTermination` -> home-screen in the runningboard logs).

The guard is size- and platform-aware:

| Condition | Behavior |
|---|---|
| Markdown <= ~15KB | Full Markdown render (selectable), all platforms |
| **Markdown > ~15KB, native (iOS/Android)** | Falls back to a **source view: one UITextView holding the whole file as a single attributed string** — still word-selectable (TextKit-paged), but a single native view instead of hundreds, so no OOM. Also surfaces the raw Markdown source (cf #1264). |
| Markdown > ~15KB, web/desktop | Full Markdown render — the DOM/Chromium handles large docs fine, no native-view explosion, so no downgrade. |
| Rendered Markdown (native) | `maxTopLevelChildren` caps blocks as a second safety net |
| Plain-text fallback | Capped at 200KB with a truncation note so a pathological multi-MB file can't stall text layout |

Why the source-view fallback is safe where the rendered path isn't: the OOM comes from the **number of native views** (one UITextView per block, or one child view per syntax token in `CodeBody`). A single UITextView holding one big attributed string is one native view + one text run — it scales with content size, not block count — so it stays selectable without the per-node explosion.

`LARGE_MARKDOWN_CHARS` (15KB) is deliberately conservative and tunable; it's a character-count proxy for block count, set well before the observed OOM cliff.

## Cross-platform behavior

Per the existing `markdown-text.{ios,android,web}` platform split:
- **iOS**: UITextView (word selection)
- **Android**: `<Text selectable>` (word selection; ReactTextViews are lighter than UITextView but still real native views, so the >15KB native fallback applies here too — conservative since I couldn't test Android)
- **web/desktop (Electron)**: base `<Text>` with CSS `user-select`; full rendering at any size

Code body is forced no-wrap + web `user-select` (matches `DiffViewer`); gutter uses `userSelect:none` so line numbers never end up in copied text.

## Test plan

Verified on **iPhone 17 Pro Simulator (iOS 26.5)** with a dev client:
- [x] Code file: long-press selects across multiple lines + drag handles; line numbers excluded from copy; gutter aligned; syntax highlight intact
- [x] Small/medium `.md`: renders + word-selectable
- [x] **Large `.md` (>15KB): no longer OOM-crashes; shows selectable plain-text source** (this was the crash that's now fixed)
- [x] TypeScript typecheck, lint, format all pass
- [x] `expo export` for web + ios both bundle clean
- [ ] **Not tested on Android or a running web/desktop session** — logic is platform-gated and follows the existing markdown-text split, but please sanity-check on Android (the >15KB fallback and the nested horizontal scroll especially)
- [ ] Not tested on a physical iOS device (no paid Apple Developer account)

## Known limitations
- Selection can't span paragraph/code-block boundaries (each block is its own native text view; iOS can't select across sibling views) — same trade-off as #1153.
- Large native Markdown shows source instead of rendered output (the OOM trade-off; doubles as the #1264 "view source" ask).

Refs #447, #21, #1264

### DIFF
--- a/packages/app/src/components/file-pane.tsx
+++ b/packages/app/src/components/file-pane.tsx
@@ -1,14 +1,21 @@
-import React, { useEffect, useMemo, useRef } from "react";
+import React, { useEffect, useMemo, useRef, type ReactNode } from "react";
 import { useQuery } from "@tanstack/react-query";
 import type { FileReadResult } from "@getpaseo/client/internal/daemon-client";
-import Markdown, { MarkdownIt } from "react-native-markdown-display";
+import Markdown, {
+  MarkdownIt,
+  type ASTNode,
+  type RenderRules,
+} from "react-native-markdown-display";
 import {
   ActivityIndicator,
   Image as RNImage,
   ScrollView as RNScrollView,
   Text,
   View,
+  type TextStyle,
+  type ViewStyle,
 } from "react-native";
+import { ScrollView as GHScrollView } from "react-native-gesture-handler";
 import { StyleSheet, useUnistyles } from "react-native-unistyles";
 import { useIsCompactFormFactor } from "@/constants/layout";
 import { useSessionStore, type ExplorerFile } from "@/stores/session-store";
@@ -20,8 +27,10 @@ import { inlineUnistylesStyle } from "@/styles/unistyles-inline-style";
 import { lineNumberGutterWidth } from "@/components/code-insets";
 import { CODE_SURFACE_DATASET } from "@/styles/code-surface";
 import { isRenderedMarkdownFile } from "@/components/file-pane-render-mode";
-import { isWeb } from "@/constants/platform";
+import { isNative, isWeb } from "@/constants/platform";
 import { createMarkdownStyles } from "@/styles/markdown-styles";
+import { MarkdownParagraphView, MarkdownTextSpan } from "@/components/markdown-text";
+import { HighlightedCodeBlock } from "@/components/highlighted-code-block";
 import type { AttachmentMetadata } from "@/attachments/types";
 import { useAttachmentPreviewUrl } from "@/attachments/use-attachment-preview-url";
 import { persistAttachmentFromBytes } from "@/attachments/service";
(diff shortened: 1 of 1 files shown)


muzhi1991:
Thanks for the review — both points were valid and are fixed in fa4d85bc:

**1. Hardcoded Chinese string** — the truncation note is now English (`… file too large — showing the first N KB of source`). My oversight; the rest of the UI is English.

**2. Lost jump-to-line on large native Markdown** — fixed. `LargeSourcePreview` has no `highlightedLines`/`lineSelection`, so the scroll effect now resolves its target from `location.lineStart` directly (via a small `resolveScrollTargetLine` helper) and scrolls there too.

One honest caveat on #2: it's **best-effort**. The plain-text source body wraps, so `(line - 1) * lineHeight` is an approximate offset rather than an exact row — a long-wrapping doc can land a few rows off. The reason it can't just reuse the precise code-view path: that path is exactly what OOMs on large files (one highlighted token view per token), which is what this PR exists to avoid. So the trade-off is "scrolls to roughly the right place, never crashes" vs. "scrolls exactly, but crashes on large files." If you'd prefer exact scrolling, the alternative is making the source body non-wrapping (horizontal-scroll, like the code view) so row math is exact — happy to do that if you want it.

Still pending: I haven't been able to test on Android or a physical iOS device (no paid Apple Developer account), so the >15KB native fallback and the nested horizontal scroll would benefit from a sanity check there.

Issue #1253, opened 2026-05-31
bug: iOS silently drops bold/italic/strikethrough and line breaks in assistant messages (regression in v0.1.84)


outofrange-consulting:
### Summary

In the **iOS app**, assistant/agent messages **silently drop inline content** — the characters are *missing*, not merely unstyled:

- `**bold**`, `*italic*`, and `~~strikethrough~~` runs, and
- **hard and soft line breaks** (a hard break vanishes; a softbreak between two lines joins the words, `one\ntwo` → `onetwo`).

Plain text and inline `code` render fine, so the loss is invisible — no indication anything was omitted. Dropping an emphasized word can **invert meaning** (observed: *"…does **not** compose…"* rendered as *"…does compose…"*); dropping a softbreak produces the "adjacent words jammed together" symptom.

> Inline **links / bare URLs** are dropped by the same root cause but are tracked separately in #1256 (the fix there changes a shared component's API and has an open tap-handling dimension, #21).

### Environment

- Client: **Paseo iOS app**, v0.1.84+ (reproduced 2026-05-31)
- Daemon: `@getpaseo/cli` 0.1.85
- Provider: Claude (provider-independent)
- Platforms: **iOS only.** Web and Android render correctly.

### Steps to reproduce

1. Have an agent emit a paragraph with `**bold**`, `*italic*`, `~~strikethrough~~`, a hard line break, and a soft-wrapped multi-line paragraph.
2. View the message in the iOS app.

### Expected

All of it renders (matching web/desktop/Android).

### Actual

- `**bold**` / `*italic*` / `~~strikethrough~~` text is **missing** (`- **Web**: …` renders as `- : …`).
- Hard breaks vanish; soft-wrapped lines lose the break and jam words together.
- Plain text and inline `code` render fine.

### Root cause

Introduced in **v0.1.84** by #1153 (native iOS text selection via `UITextView`). On iOS, `MarkdownParagraphView`/`MarkdownTextSpan` are native `react-native-uitextview` `UITextView`s (`packages/app/src/components/markdown-text.ios.tsx`); nested `UITextView`s hoist into `UITextViewChild` fragments so one native text view composes the paragraph. A plain RN `<Text>` nested in that `UITextView` is **not** hoisted → its content is dropped.

[...]

PR #1289, opened 2026-06-02
fix(app): keep iOS inline markdown emphasis & links inside UITextView

### LINKED ISSUE #1287: Inline code spans (single backticks) are stripped from rendered chat output
## What I'm seeing

In a Claude Code session viewed through Paseo, any assistant text wrapped in single backticks (Markdown inline code) renders as **completely empty** — the text inside the backticks vanishes, leaving an empty space where the inline code should be.

The effect is consistent and reproducible across many messages in a single session.

## Example

The assistant emitted (Markdown):

```
The only output-altering hook script is `credential-guard-message.sh`, but it's not wired into `settings.json` and its own header says "unregistered".
```

What appeared in the Paseo UI:

```
The only output-altering hook script is , but it's not wired into  and its own header says "unregistered".
```

Both backticked tokens were dropped.

A second example: a Markdown table where the first column contained backtick-wrapped filenames (one per row) rendered with that entire first column empty — every row's leading code span gone.

## Expected

Inline code spans should render as monospaced inline text (their content preserved), the standard Markdown / CommonMark behaviour.

## Environment

- Paseo session connected to a Claude Code session running on macOS (Mac mini, "Minecraft-Server.local").
- Reported by the assistant in-session; the user confirmed visually.
- Triple-backtick fenced code blocks appear to render correctly — only single-backtick inline spans are affected. (Worth double-checking, but that's the impression from session output.)

## Workaround

Avoid inline code formatting for short identifiers/filenames; use plain text or block code instead.

## Suspected cause

A Markdown renderer or sanitiser stripping `<code>` tags (or matching `\`…\`` spans) outright rather than passing the inner text through. Could also be a CSS rule with `display: none` on inline-code elements.

muzhi1991:
## Problem

On iOS, assistant chat messages drop inline Markdown:

- **Inline code that resolves to a file/URL link** (e.g. `settings.json`, `*.sh`, or any autolink) renders **completely empty** — the text inside the backticks vanishes (issue #1287).
- **Bold / italic / strikethrough** don't render.

Plain text and fenced code blocks are fine. **Web and Android are unaffected.**

## Root cause

The assistant paragraph is a native `UITextView` (`react-native-uitextview`, from the iOS native text-selection work). That view only draws text from its `UITextViewChild` nodes, which are produced exclusively by the library's `UITextView` / `MarkdownTextSpan`. Two inline render paths still emitted a plain RN `<Text>`, whose text is dropped inside the paragraph:

1. `strong` / `em` / `s` used `react-native-markdown-display`'s **default** rules (there were no overrides) → plain `<Text>` wrapper.
2. `AssistantMarkdownLink`'s native branch rendered a plain `<Text>`; for inline-code file links / autolinks (whose child is a raw string) that produced **no** `UITextViewChild`, so the inline code rendered empty.

## Fix

- Add custom `strong` / `em` / `s` rules that render through `MarkdownTextSpan`.
- Route `AssistantMarkdownLink`'s native text through `MarkdownTextSpan` as well (extended to forward `onPress` / `accessibilityRole`).

`MarkdownTextSpan` is a `UITextView` on iOS and a plain selectable `<Text>` on web/Android, so **web and Android behavior is unchanged**. Emphasis styling still reaches the leaf text via `inheritedStyles` (the AstRenderer derives it from the parent emphasis node).

## Verification

- `format` ✅ · `lint` ✅ · `typecheck` (all workspaces) ✅ · existing tests ✅ (58)
- **iOS Simulator (iPhone 17, Xcode 26.5):** a throwaway preview route placed a plain RN `<Text>` between `MarkdownTextSpan` spans inside the paragraph — its text vanished (confirming the mechanism), while the fixed `AssistantMarkdownLink` plus **bold / italic / strike** all rendered. The preview route is **not** part of this PR.

Fixes #1287

### DIFF
--- a/packages/app/src/assistant-file-links/link.tsx
+++ b/packages/app/src/assistant-file-links/link.tsx
@@ -9,6 +9,7 @@ import {
 } from "react-native";
 import { StyleSheet } from "react-native-unistyles";
 import { isNative, isWeb } from "@/constants/platform";
+import { MarkdownTextSpan } from "@/components/markdown-text";
 import { Shortcut } from "@/components/ui/shortcut";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
 import { useStableEvent } from "@/hooks/use-stable-event";
@@ -57,16 +58,22 @@ export function AssistantMarkdownLink({
   );
 
   if (isNative) {
+    // Route the text through MarkdownTextSpan, not a plain RN <Text>. On iOS the
+    // assistant paragraph is a native UITextView that only renders UITextViewChild
+    // nodes; a bare <Text> here (e.g. inline-code file links / autolinks, whose
+    // children are a raw string) produces no UITextViewChild and renders empty.
+    // MarkdownTextSpan is a UITextView on iOS and a selectable <Text> on Android.
+    // See issue #1287. (data-pmono is a web-only font marker — not needed here.)
     return (
       <FileLinkHoverTooltip filePath={tooltipPath}>
-        <Text
+        <MarkdownTextSpan
           accessibilityRole="link"
-          dataSet={monoSurface ? CODE_SURFACE_DATASET : undefined}
+          monoSurface={monoSurface}
           onPress={onPress}
           style={style}
         >
           {children}
-        </Text>
+        </MarkdownTextSpan>
       </FileLinkHoverTooltip>
     );
   }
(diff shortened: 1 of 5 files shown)


muzhi1991:
Closing as superseded — this duplicates fixes already merged into `main`:

- #1254 — bold/italic/strikethrough (and hard/soft line breaks) via `MarkdownTextSpan` on iOS
- #1257 — inline links (markdown links, bare URLs, inline-code path links) via `MarkdownTextSpan` on iOS

Same root cause (a plain RN `<Text>` inside the iOS paragraph `UITextView` isn't hoisted into a `UITextViewChild`, so its text is silently dropped) and the same `MarkdownTextSpan` approach; #1254 additionally covers line breaks. I branched off a stale local `main` and missed that these had already landed. The behavior reported in #1287 is fixed by #1253/#1254/#1257 on current `main`.

PR #1294, opened 2026-06-02
fix(app): load older history when chat is shorter than viewport

kongjiadongyuan:
## Summary

- load older agent history after viewport/content measurement, not only after user scroll
- share the history-start scroll math across forward web and inverted native stream strategies
- cover short-content cases with pure strategy tests and a web regression test that triggers without a scroll event

## Why

When the initial chat content is shorter than the viewport, there is no scrollable top edge. The existing lazy history loader only ran from scroll handlers, so users could get stuck with a short partial timeline even though `hasOlderHistory` was true.

## Tests

- `npm run test --workspace=@getpaseo/app -- src/agent-stream/render-strategy.test.ts src/agent-stream/strategy-web.test.tsx --bail=1`
- `npm run lint -- packages/app/src/agent-stream/strategy.ts packages/app/src/agent-stream/strategy-native.tsx packages/app/src/agent-stream/strategy-web.tsx packages/app/src/agent-stream/render-strategy.test.ts packages/app/src/agent-stream/strategy-web.test.tsx`
- pre-commit: lint, format check, full workspace typecheck

### DIFF
--- a/packages/app/src/agent-stream/render-strategy.test.ts
+++ b/packages/app/src/agent-stream/render-strategy.test.ts
@@ -10,6 +10,7 @@ import {
   getStreamNeighborIndex,
   getStreamNeighborItem,
   isNearBottomForStreamRenderStrategy,
+  isNearHistoryStartForStreamRenderStrategy,
   orderHeadForStreamRenderStrategy,
   orderTailForStreamRenderStrategy,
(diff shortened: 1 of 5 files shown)


boudra:
thanks for the pull request.

PR #1305, opened 2026-06-03
fix(server): don't treat a recycled PID as a running daemon

### LINKED ISSUE #1304: Daemon fails to start after unclean shutdown: stale paseo.pid + PID reuse leaves clients stuck on "connecting"
## Summary

After the daemon exits uncleanly (kill / sleep / crash), the stale `~/.paseo/paseo.pid` is left behind. The PID lock's liveness check only verifies that *some* process with that PID exists — not that it's actually the daemon. Once the OS recycles that PID to an unrelated process, the check passes, so a new daemon refuses to start with `Another Paseo daemon is already running (PID …)`, exits 1, and never binds `127.0.0.1:6767`. Every client (desktop **and** CLI) then hangs on **"connecting"** indefinitely.

## Steps to reproduce

1. Start the daemon, then kill it uncleanly (e.g. `kill -9`, or let the machine sleep) so `releasePidLock` never runs — `paseo.pid` stays behind.
2. Wait until the OS reuses the recorded PID for some other process (`ps -p <pid>` shows e.g. `printtool`).
3. Launch the desktop app, or run `paseo daemon status`.

Result: daemon won't start; clients stuck on "connecting".

## Evidence (from my machine)

`~/Library/Logs/paseo/main.log`:

```
[desktop daemon] detached startup grace period completed {
  exitedEarly: true,
  stderr: 'Another Paseo daemon is already running (PID 2742, started 2026-06-02T14:11:02.499Z)',
  exitCode: 1,
}
```

But PID 2742 was no longer the daemon — the OS had recycled it:

```
$ ps -p 2742 -o pid,lstart,command
 2742  Tue Jun  2 22:27:30 2026  .../PrintCore.framework/.../printtool
```

(The daemon's last `daemon.log` line was `2026-06-02 22:21`; `printtool` grabbed PID 2742 at 22:27.) Nothing was listening on 6767.

## Root cause

`packages/server/src/server/pid-lock.ts` — `isPidRunning()` uses only `process.kill(pid, 0)`, which is identity-blind. `acquirePidLock()` and `isLocked()` both rely on it, so a recycled PID is treated as a live daemon.

## Workaround

Delete `~/.paseo/paseo.pid` and relaunch.

## Possible fix

Verify process identity before trusting the lock — e.g. compare the live process's start time (`ps -o lstart`) against the lock's `startedAt`; if they diverge beyond a tolerance, the PID was reused and the lock is stale. I have a small patch + tests for this and am happy to open a PR if this direction looks right.

chosen1hyj:
Closes #1304.

## Problem

The PID lock's liveness check is identity-blind: `isPidRunning()` (used by both `acquirePidLock()` and `isLocked()` in `packages/server/src/server/pid-lock.ts`) only does `process.kill(pid, 0)`, which tells you *a* process with that PID exists — not that it's still the daemon. After an unclean shutdown the stale `~/.paseo/paseo.pid` survives; once the OS recycles that PID to an unrelated process, the daemon refuses to start (`Another Paseo daemon is already running`), exits 1, never binds 6767, and every client hangs on "connecting". Full evidence in #1304.

## Fix

Verify process identity before trusting the lock: compare the live process's start time (`ps -o lstart`, the portable keyword on both macOS and Linux; `LC_ALL=C` for a parseable timestamp) against the lock's `startedAt`. If they diverge beyond a 60s tolerance, the PID was reused → clear the stale lock and start.

`lstart` is second-granularity and the lock is written a beat after the daemon starts, so a genuine daemon's two timestamps sit within a few seconds, while a recycled PID differs by the daemon's whole lifetime. When the start time can't be determined (e.g. `ps` unavailable on Windows), it falls back to the previous conservative behaviour, so there's no regression there.

## Testing

- Added two cases in `pid-lock.test.ts`: one reproduces the recycled-PID bug (fails on `main`, passes with the fix), one confirms a genuinely-live daemon is still rejected (guards against over-correcting).
- `oxfmt`, `oxlint`, and `tsgo` typecheck across the whole workspace all pass (via the repo's lefthook pre-commit hook).

No protocol/schema changes.

I actually hit this myself — both the desktop app and the CLI were stuck on "connecting" indefinitely. Digging into the logs, the daemon had exited uncleanly the night before and left ~/.paseo/paseo.pid behind, and the OS had since recycled that PID to an unrelated printtool process, so the lock check thought a daemon was still running. Deleting ~/.paseo/paseo.pid and relaunching fixed it immediately, which confirmed the root cause. I tested this change locally: the new test reproduces the recycled-PID case (fails on main, passes with the fix), the guard test confirms a genuinely-live daemon is still rejected, and typecheck/lint/format all pass via the pre-commit hook. Happy to adjust the approach (tolerance value, or verifying identity a different way) if you'd prefer.

### DIFF
--- a/packages/server/src/server/pid-lock.test.ts
+++ b/packages/server/src/server/pid-lock.test.ts
@@ -1,9 +1,20 @@
-import { mkdtemp, rm } from "node:fs/promises";
-import { tmpdir } from "node:os";
+import { execFileSync } from "node:child_process";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { hostname, tmpdir } from "node:os";
 import { join } from "node:path";
 import { describe, expect, test } from "vitest";
 
 import { acquirePidLock, getPidLockInfo, releasePidLock, updatePidLock } from "./pid-lock.js";
+
+// Independently derive the real OS start time of a live PID, so the staleness
+// tests don't depend on the implementation's own start-time helper.
+function realProcessStartIso(pid: number): string {
+  const out = execFileSync("ps", ["-o", "lstart=", "-p", String(pid)], {
+    env: { ...process.env, LC_ALL: "C" },
+    encoding: "utf8",
+  }).trim();
+  return new Date(out).toISOString();
+}
 
 describe("pid-lock ownership", () => {
   test("writes and releases lock for explicit owner pid", async () => {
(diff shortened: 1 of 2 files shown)


boudra:
should be fixed in 0.1.91

PR #1318, opened 2026-06-03
Add repository-based thread organization

ebg1223:
## What changes

This adds a workspace organization preference for how the app presents active work.

The existing **Workspaces** organization remains the default. It keeps the current sidebar behavior: projects expand into workspaces/branches, and workspace tabs continue to represent the active work for that workspace.

The new **Threads** organization gives users a repo-centered way to work when several unrelated agent tasks are happening in the same checkout. In this mode:

- The sidebar shows repositories/projects with active threads underneath.
- Thread rows show their branch/workspace context and status.
- Opening a thread creates or focuses a view tab for that thread.
- Closing a tab only hides that view; it does not archive the thread.
- Closing a thread from the sidebar archives it and removes it from active lists.
- Repository rows can start a new thread in an existing workspace/branch, including when that branch does not currently have any active thread.

Workspace-specific organization still exists where it is useful. Status grouping and numbered workspace shortcuts remain tied to the default Workspaces organization, where those controls still map directly to visible workspace rows.

## Why

Worktrees are not always the right organizing boundary. Users may run multiple unrelated tasks from the same checkout and resolve conflicts later, which makes branch/workspace tabs a weak proxy for the work they are actually switching between.

This keeps the current workflow intact while adding an alternate organization that treats threads as the primary units of work.


boudra:
thanks for the pull request.

PR #1333, opened 2026-06-04
fix(claude): preserve alwaysLoad on MCP server configs

### LINKED ISSUE #1332: bug: Claude provider drops alwaysLoad on MCP server config, tools never surface
### What's broken

The Claude provider configures MCP servers via `mcpServers` in the agent provider config (and via `extra.claude.mcpServers`), but tools from those servers never appear in the agent's tool list. The agent reasons over a "deferred tools list" and concludes the tools are not available — even when the MCP server is running, connected, and discoverable on disk.

Root cause: `toClaudeSdkMcpConfig` in `packages/server/src/server/agent/providers/claude/agent.ts` (lines 805-828) drops the `alwaysLoad` field when normalizing config for `@anthropic-ai/claude-agent-sdk`. The SDK's `McpStdioServerConfig` / `McpHttpServerConfig` / `McpSseServerConfig` types all expose `alwaysLoad?: boolean`, and per the SDK docs, without it `defer_loading` defaults to true and tools are hidden behind tool search. Tool search in practice doesn't surface them (we have not been able to reproduce a case where it does), so the agent treats the server as unavailable.

This affects every MCP server configured through paseo on the Claude provider — chrome-devtools, markitdown, aws-docs, and several others in our setup.

The other providers (Codex, Kiro via ACP, Copilot via ACP) are unaffected because they shell out to their own CLIs which read their own MCP config files directly; paseo's normalization isn't on the hot path for them.

### Steps to reproduce

1. On a host with `paseo` running, configure any MCP server in `~/.paseo/config.json` under `agents.providers.claude.mcpServers`. Example:

   ```json
   {
     "agents": {
       "providers": {
         "claude": {
           "mcpServers": {
             "chrome-devtools": {
               "type": "stdio",
               "command": "npx",
               "args": ["-y", "chrome-devtools-mcp@latest", "--browserUrl", "http://127.0.0.1:9222"],
               "alwaysLoad": true
             }
           }
         }
       }
     }
   }
   ```

2. `systemctl restart paseo-daemon` (or restart the daemon however you run it).

3. Run a prompt that requires the tool:

   ```
   paseo run --provider claude \
     "Use chrome-devtools to navigate to https://example.com and return only the page title. \
      If chrome-devtools tools are not available, reply exactly 'chrome-devtools UNAVAILABLE' and nothing else."
   ```

4. Inspect the agent's reasoning with `paseo logs <agent-id>`. The agent reports it sees only `mcp__paseo__*` tools and concludes chrome-devtools is unavailable, even though `ps aux` confirms `chrome-devtools-mcp` is running and connected.

5. The same prompt against `paseo run --provider codex` (or `--provider kiro`, `--provider copilot`) succeeds and returns "Example Domain", because those providers don't go through `toClaudeSdkMcpConfig`.

### Where did this happen

Daemon

### Anything else

PR with the fix and unit tests is ready: nodomain/paseo#fix/claude-mcp-always-load — opening it as soon as this issue exists so the PR can link back.

The fix:

- Adds `alwaysLoad?: boolean` to `McpStdioServerConfig`, `McpHttpServerConfig`, and `McpSseServerConfig` in both `packages/protocol/src/agent-types.ts` and `packages/server/src/server/agent/agent-sdk-types.ts`, and to the matching zod schemas in `packages/protocol/src/messages.ts`.
- Updates `toClaudeSdkMcpConfig` to copy `alwaysLoad` through for all three transports.
- Adds four unit tests covering stdio/http/sse pass-through and the default-undefined case.

Backward compat: the field is optional everywhere. Old clients sending without `alwaysLoad` keep the existing behavior (tools deferred). New clients setting `alwaysLoad: true` get the tools loaded immediately, which is the SDK's documented contract.

Verified on a live devbox (Ubuntu 24.04, paseo 0.1.90) with chrome-devtools-mcp: before the fix, Claude can't see the tools; after the fix, Claude calls `chrome-devtools.navigate_page` and `chrome-devtools.evaluate_script` and returns the page title.

nodomain:
### Linked issue

Closes #1332

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

`toClaudeSdkMcpConfig` (`packages/server/src/server/agent/providers/claude/agent.ts:805-828`) was dropping the `alwaysLoad` field on its way to `@anthropic-ai/claude-agent-sdk`. Without it, every MCP server's tools default to `defer_loading: true` — they're hidden behind tool search, which in practice never surfaces them. The Claude agent ends up reasoning over a "deferred tools list" that contains only `mcp__paseo__*` and concludes user-configured servers are unavailable.

Fix:

- `packages/protocol/src/agent-types.ts` and `packages/server/src/server/agent/agent-sdk-types.ts` — add `alwaysLoad?: boolean` to `McpStdioServerConfig`, `McpHttpServerConfig`, `McpSseServerConfig`.
- `packages/protocol/src/messages.ts` — add the field to the three zod schemas so the protocol round-trips it.
- `packages/server/src/server/agent/providers/claude/agent.ts` — export `toClaudeSdkMcpConfig` and copy `alwaysLoad` through for all three transports.
- `packages/server/src/server/agent/providers/claude/agent.test.ts` — four unit tests covering stdio/http/sse pass-through and the default-undefined case.

Other providers ignore the field, consistent with the existing pattern where each provider normalizes only what its CLI/SDK supports. The field is optional everywhere, so old clients are unaffected.

### How did you verify it

What I actually ran:

- `npx vitest run src/server/agent/providers/claude/agent.test.ts -t toClaudeSdkMcpConfig` — the four new unit tests pass. They assert the `alwaysLoad` value round-trips for each of stdio/http/sse, and stays `undefined` when not provided.
- `npm run typecheck` and `npm run lint` — both clean.
- `npm run format` — clean against the touched files.

What I observed for the bug itself (the "before" state):

- On a live devbox running paseo 0.1.90, with `chrome-devtools-mcp` configured under `agents.providers.claude.mcpServers` and the server confirmed running on `127.0.0.1:9222`, Claude consistently responds:
  ```
  [Thought] Looking at the deferred tools list, I see tools prefixed with `mcp__paseo__`
            but none prefixed with `mcp__chrome` or `mcp__devtools`...
  chrome-devtools UNAVAILABLE
  ```
- The same prompt against `paseo run --provider codex|kiro|copilot` returns the page title correctly. Those providers don't go through `toClaudeSdkMcpConfig`.

What I did **not** verify:

- I did not run an end-to-end test of the fixed paseo on the live devbox. The devbox uses the npm-published paseo, and rebuilding/installing the patched build would be its own change. The fix's correctness rests on the unit tests proving `alwaysLoad` survives normalization, plus the SDK's own documented contract for what `alwaysLoad` does.

If you'd like an e2e run before merging, I can install a local build on the devbox and post the result.

### Checklist

- [x] `npm run lint` passes
- [x] No UI changes


nodomain:
Addressed greptile inline feedback in d8695746: replaced the `"alwaysLoad" in result ? ... : "missing"` ternary with a direct `expect(result.alwaysLoad).toBeUndefined()`. Since `alwaysLoad` is now declared on every branch of the discriminated union returned by `toClaudeSdkMcpConfig`, the in-operator guard is unnecessary.

PR #1341, opened 2026-06-05
fix(app): return to correct host when closing settings

### LINKED ISSUE #1320: bug: Closing settings does not return to the previously selected host
## What

After opening settings and navigating to a different host's settings, closing the settings panel does not restore the host selection that was active before entering settings.

## Steps to reproduce

1. Connect to multiple hosts
2. Select a non-default host in the main app view
3. Open settings (Cmd+, or tap settings)
4. Navigate within settings (the host picker in settings may default to a different host or change during navigation)
5. Close settings

## Expected behavior

Closing settings should return to the host that was selected before opening settings.

## Actual behavior

The host selection is lost or reset after closing settings — the user is taken to a different host than the one they were viewing before entering settings.

thedavidweng:
## What

When navigating between hosts in settings and then closing, the fallback in  used  which could point to a different host than the one the user was originally viewing.

## Changes

- Prefer  over  as the fallback host when the last workspace selection is unavailable
- Use  for host detail views so pressing back from host settings exits to the workspace instead of navigating within the settings stack

Fixes #1320

### DIFF
--- a/packages/app/src/screens/settings-screen.tsx
+++ b/packages/app/src/screens/settings-screen.tsx
@@ -648,6 +648,22 @@ function useAnyOnlineHostServerId(serverIds: string[]): string | null {
   );
 }
 
+function useLocalDaemonIfOnline(localServerId: string | null): string | null {
+  const runtime = getHostRuntimeStore();
+  return useSyncExternalStore(
+    (onStoreChange) => {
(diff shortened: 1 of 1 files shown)


boudra:
thanks for the pull request.

PR #1342, opened 2026-06-05
fix(app): disable native back-swipe on host screens

### LINKED ISSUE #1321: bug: Right swipe gesture on iOS incorrectly triggers back navigation instead of opening sidebar
## What

On iOS (native app), swiping from left to right is interpreted as a "go back" gesture, navigating to the previous chat session instead of opening the sidebar. Since the previous session may be on a different host, this creates a confusing experience where the user suddenly finds themselves in an unrelated conversation.

## Steps to reproduce

1. Open a chat session on host A
2. Open another chat session (possibly on host B)
3. Swipe from left to right on the screen

## Expected behavior

The swipe should open the sidebar / drawer for navigation.

## Actual behavior

The swipe triggers a back navigation, jumping to the previously opened chat session — which may be on a completely different host. This is disorienting and breaks the user's mental model of where they are in the app.

## Impact

- Users accidentally navigate away from their current conversation
- Cross-host navigation via swipe-back is unexpected and confusing
- No clear way to return to the original session without manually navigating back

thedavidweng:
## What

On native iOS, the default React Navigation back-swipe gesture was enabled on workspace and other host screens. Swiping right would navigate back (often to a different host screen) instead of opening the sidebar.

## Changes

Disable the native gesture on all screens inside the MobileGestureWrapper (workspace, agent, host index, sessions, open-project) so that right-swipes are handled by the wrapper Pan gesture, which opens the sidebar as expected.

Fixes #1321

### DIFF
--- a/packages/app/src/app/_layout.tsx
+++ b/packages/app/src/app/_layout.tsx
@@ -861,7 +861,12 @@ function FaviconStatusSync() {
   return null;
 }
 
-const AGENT_SCREEN_OPTIONS = { gestureEnabled: false };
+// Disable the native back-swipe gesture on screens that live inside the
+// MobileGestureWrapper.  On those screens a right-swipe should open the
+// sidebar, not navigate back.  The wrapper's own Pan gesture handles the
+// sidebar interaction; the native gesture would fight it and cause
+// unexpected back-navigation (often to a different host).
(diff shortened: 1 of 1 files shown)


boudra:
thanks for the pull request.

PR #1346, opened 2026-06-05
feat(server): make provider refresh timeout configurable via PASEO_PROVIDER_REFRESH_TIMEOUT_MS

### LINKED ISSUE #1345: Copilot provider snapshot times out at 30s on cold start with many providers/MCP servers
## Symptom

The Copilot provider snapshot fails on daemon startup with:

```
Timed out refreshing Copilot after 30000ms
```

A manual retry from the UI always succeeds — warm caches make the second probe much faster.

error

## Reproduction

Looks like a cold-start edge case that gets worse with scale:
- many providers configured (Claude, Codex, Copilot, OpenCode, Pi, plus custom providers)
- many MCP servers attached
- first `copilot --acp` invocation since boot (or after a long idle)

The 30s ceiling is hardcoded in `provider-snapshot-manager.ts`:

```ts
const DEFAULT_REFRESH_TIMEOUT_MS = 30_000;
```

`ProviderSnapshotManager` accepts a `refreshTimeoutMs` constructor option, but `bootstrap.ts` doesn't pass it, so there is no way for an operator to bump it short of rebuilding.

The same timeout was previously hit by OpenCode under similar conditions — see `docs/diagnostics/opencode-provider-snapshot-timeout-2026-05-27.md`.

## Why retry works

The first probe spawns `copilot --acp` cold (binary cache, MCP server handshakes, ACP capability negotiation). Subsequent probes hit warm filesystem and process caches, so they comfortably finish under the 30s budget.

## Proposed fix

Make the timeout configurable via env var so users hitting this can bump it without rebuilding the daemon. PR: nodomain/paseo#feat/configurable-provider-refresh-timeout

```bash
PASEO_PROVIDER_REFRESH_TIMEOUT_MS=60000 paseo daemon start
```

Defaults stay at 30s for everyone else.

## Probably edge case

This likely only hits users with many providers + many MCP servers stacked together. Still worth fixing because the workaround (retry) requires user interaction every daemon restart, and there's no obvious knob today.

nodomain:
Closes #1345

## Problem

Provider snapshot refresh has a hardcoded 30s timeout for the `isAvailable()` and `listModels`/`listModes` probes:

```
Timed out refreshing Copilot after 30000ms
```

With many providers and MCP servers configured, cold starts (first `copilot --acp` invocation since boot, etc.) can exceed this. A manual retry always succeeds because the second probe hits warm caches.

`ProviderSnapshotManager` already accepts a `refreshTimeoutMs` option, but `bootstrap.ts` never passes it — so there's no way to bump it short of rebuilding the daemon.

## Change

Add `PASEO_PROVIDER_REFRESH_TIMEOUT_MS` env var support inside `ProviderSnapshotManager` itself. Precedence:

1. explicit constructor `refreshTimeoutMs` option (used in tests)
2. `PASEO_PROVIDER_REFRESH_TIMEOUT_MS` env var (new)
3. `DEFAULT_REFRESH_TIMEOUT_MS` = 30_000 (unchanged default)

Invalid values (non-numeric, ≤ 0, NaN) silently fall back to the default.

```bash
PASEO_PROVIDER_REFRESH_TIMEOUT_MS=60000 paseo daemon start
```

## Tests

Three new tests in `provider-snapshot-manager.test.ts` exercise the timeout path with a never-resolving `isAvailable` mock and a tiny budget, asserting the resulting error status carries the expected `after Nms` message:

- explicit option wins
- env var honored when no option given
- explicit option still wins over env var

## Notes

- No protocol changes, no new RPC, no config schema changes
- Default behavior unchanged for everyone not setting the env var
- Local `npm run lint` / `npm run format:check` couldn't run (broken oxfmt/oxlint native bindings on my machine), but CI will catch any formatting issues. Server `typecheck` passed clean.

### DIFF
--- a/packages/server/src/server/agent/provider-snapshot-manager.test.ts
+++ b/packages/server/src/server/agent/provider-snapshot-manager.test.ts
@@ -156,6 +156,100 @@ describe("ProviderSnapshotManager public surface", () => {
     }
   });
 
+  test("refreshTimeoutMs option overrides the default and yields a timeout error", async () => {
+    // never-resolving isAvailable forces the timeout path
+    const isAvailable = vi.fn(() => new Promise<boolean>(() => {}));
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      refreshTimeoutMs: 1,
+      providerOverrides: {
+        claude: { enabled: false },
+        copilot: { enabled: false },
+        opencode: { enabled: false },
+        pi: { enabled: false },
+      },
+      extraClients: { codex: createExtraClient("codex", { isAvailable }) },
+    });
+    try {
+      const entry = await manager.getProvider({
+        cwd: "/tmp/project",
+        provider: "codex",
+        wait: true,
+      });
+      expect(entry.provider).toBe("codex");
+      expect(entry.status).toBe("error");
+      expect(entry.error).toMatch(/after 1ms/);
+    } finally {
+      manager.destroy();
+    }
+  });
+
+  test("PASEO_PROVIDER_REFRESH_TIMEOUT_MS env var is honored when no option is given", async () => {
+    const previous = process.env.PASEO_PROVIDER_REFRESH_TIMEOUT_MS;
+    process.env.PASEO_PROVIDER_REFRESH_TIMEOUT_MS = "1";
+    const isAvailable = vi.fn(() => new Promise<boolean>(() => {}));
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      providerOverrides: {
+        claude: { enabled: false },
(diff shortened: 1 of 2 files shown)


nodomain:
Addressed greptile feedback in 7bf50512:

- env var parsing now uses `Number()` instead of `Number.parseInt()` so scientific notation like `"6e4"` parses as 60000 instead of being silently truncated to 6
- new tests use `vi.stubEnv()` / `vi.unstubAllEnvs()` instead of manual `process.env` save/restore, matching the rest of the suite (e.g. `claude/models.test.ts`, `run-git-command.test.ts`)

`npm run typecheck:server` clean. Lint/format/tests will run in CI.

PR #1355, opened 2026-06-05
feat(desktop): support multiple windows

### LINKED ISSUE #250: Desktop app cannot open multiple windows

arieel-ost:
## Summary

Adds multi-window support to the desktop app. Closes #250.

Before this, the app was effectively single-window: relaunching just focused the existing window, and there was no New Window action.

This is a deliberately lean **"land-on" hybrid**: new windows can open focused on a specific project, but every window keeps the full sidebar. There is no per-window project ownership, no sidebar filtering, and no route guards. That keeps the change small and reviewable while still delivering the "one window per project" gesture. (It's a smaller alternative to the ownership-model approach in #254, which carried filtering/guards/hydration complexity that isn't needed to satisfy #250.)

## What's in it

**Window lifecycle (`packages/desktop/src/main.ts`)**

- `createMainWindow()` is generalized into a reusable `createWindow({ pendingOpenProjectPath?, restoreWindowState? })` returning the `BrowserWindow`.
- **File → New Window (`CmdOrCtrl+N`)** added to the app menu.
- `second-instance` (CLI `paseo [path]`, relaunch, Finder double-click) now opens a **new** window instead of focusing window 0. It waits on a `bootstrapComplete` promise rather than `app.whenReady()`, so the custom protocol + IPC handlers exist before the window loads `paseo://app/` (otherwise a second instance arriving mid-cold-start could load before the protocol handler registered).
- New `paseo:window:openNew` IPC + `window.openNew` preload bridge, wired to an **"Open in new window"** item in the sidebar project menu. It lands on the project via the same open-project flow as a CLI launch — no move, no ownership.

**Per-window deep-linking (`pending-open-project-store.ts`)**

- A per-`webContents` map of pending project paths. Each window pulls its own path on mount via `paseo:get-pending-open-project` (keyed by `event.sender.id`), replacing the previous single global + push event, so concurrent windows can't race over one pending value.

**Window state**

- Only the first window of a session restores and persists geometry; windows from ⌘N / second-instance / "Open in new window" open default-sized and OS-cascaded. This avoids every window stacking on the same restored bounds and fighting over the single window-state store.
- `setupWindowResizeEvents` now guards against sending to a destroyed `webContents` during teardown (more likely with N windows).

## Known limitations (documented in `docs/architecture.md`)

- New windows don't persist their own geometry yet (needs per-window state keys).
- In-app browser panes are not yet per-window: the active-browser id and the webview registration queue are still process-global, so with browser panes open in two windows a menu Reload can target the other window's webview. Making that subsystem window-scoped is a follow-up.

## Second commit: dev-environment isolation

`chore(desktop): isolate dev environment from a production install` fixes a separate dev-ergonomics problem this work surfaced: `npm run dev` fails while the installed app is open, because the dev build shares the default userData (loses the single-instance lock and quits) and points at the production daemon, whose CORS allowlist rejects the Metro origin. `dev.sh`/`dev.ps1` now default to a gitignored `packages/desktop/.dev/` home with its own daemon, lock, and port (6788), and seed that managed home's `config.json` (a user-supplied `PASEO_HOME` is never rewritten). Happy to split this into its own PR if preferred.

## Protocol compatibility

No WebSocket protocol/schema changes. This is desktop main↔renderer IPC that ships together.

## Testing

- `npm run build:main --workspace=@getpaseo/desktop`, `npm run typecheck` (full workspace), `npm run lint`, `npm run format` — all clean.
- Unit tests: `pending-open-project-store.test.ts`, `menu.test.ts`, `window-manager.test.ts` (20 tests).
- Manual: ⌘N, relaunch-creates-new-window, and "Open in new window" against a live daemon.
- Reviewed with both Codex and an independent adversarial pass; findings (a dev-config clobber footgun, the cold-start protocol race, a resize-after-destroy) fixed before this PR.


arieel-ost:
@boudra tagging you as maintainer for review — this is opened from a fork, so GitHub wouldn't let me add you as a formal reviewer (reviewer requests need write access on the base repo). Could you assign yourself or whoever owns the desktop package?

Closes #250. It's a smaller, "land-on" alternative to #254 (new windows + full sidebar in each, no per-window ownership/filtering). Details and known limitations are in the PR description.

arieel-ost:
Thanks @greptile-apps — both addressed in the latest push:

- **Swallowed IPC error (`sidebar-workspace-list.tsx`):** `openNew` failures now log and surface a toast ("Couldn't open a new window") instead of `catch(() => undefined)`.
- **`vi.mock` in `menu.test.ts`:** removed the test rather than add to the existing `vi.mock` debt, per `docs/testing.md`'s no-`vi.mock` rule (ports-and-adapters or real e2e only). The menu's View/Window items have electron-coupled click handlers, so a clean pure-function extraction isn't worth it here. The File → New Window wiring is covered by typecheck + manual testing, and the genuinely pure logic (pending-open-project store, window helpers) has real no-mock tests.

PR #1363, opened 2026-06-05
fix(opencode): pass --pure to spawned opencode serve to avoid silent exit on 1.16.x

### LINKED ISSUE #1362: OpenCode provider snapshot fails: "OpenCode server exited with code 0" when opencode 1.16.x is spawned as detached subprocess
### What's broken

On the first provider refresh after every daemon start, the OpenCode built-in provider snapshot fails with:

```
Error: OpenCode server exited with code 0
    at packages/server/src/server/agent/providers/opencode/server-manager.ts:startServer
```

In a typical Paseo 0.1.90 session this fires ~17 times in an 8-minute window immediately after daemon start, then stops once the relay disconnects. Manual `paseo run --provider opencode` from the same daemon also fails.

The opencode subprocess never writes "listening on" to stdout, never writes to stderr, and never logs any plugin/boot message — it just exits with code 0 within ~2s. The same `opencode serve --port N` invocation works perfectly when run manually in a TTY.

### Environment

* Paseo: 0.1.90 (npm-global, `@getpaseo/cli` from `@getpaseo/server`)
* opencode: 1.16.2 (npm-global, `opencode-ai`)
* OS: Linux aarch64
* daemon env (from /proc/<pid>/environ): HOME, PATH including `~/.npm-global/bin` first, ANTHROPIC_BASE_URL=http://localhost:8312

### Steps to reproduce

1. Install Paseo and opencode globally
2. Start the Paseo daemon: `paseo start`
3. Connect any client (CLI, app, or relay)
4. Watch `~/.paseo/daemon.log` — the OpenCode snapshot acquisition fails with exit code 0
5. From a separate terminal, run `opencode serve --port 4096` manually — works fine

### Root cause

opencode 1.16.x silently exits with code 0 when its plugin/MCP loader runs inside a child process spawned with:

```ts
spawnProcess(cmd, [...args, "serve", "--port", N], {
  detached: process.platform !== "win32",
  stdio: ["ignore", "pipe", "pipe"],
  // stdio[0] = "ignore" + detached + no controlling tty
})
```

This affects any host that spawns opencode serve non-interactively. Filed upstream against anomalyco/opencode.

### Workaround

Add `--pure` to the spawn args. opencode documents it as "run without external plugins" and it is sufficient here: Paseo only needs the opencode HTTP API; it injects its own MCP via the Agent MCP server and does not depend on plugin-loaded agents, MCPs, or skills inside the opencode subprocess.

PR: https://github.com/getpaseo/paseo/pull/<coming-soon> — adds `--pure` to the spawn args in `OpenCodeServerManager.startServer`. Verified locally by patching the compiled output and observing zero new exit-0 events over a 30s watch with the patched code loaded.

shrijit-ctrl:
## Summary

`OpenCodeServerManager.startServer` passes `--pure` to the spawned `opencode serve` so that the opencode 1.16.x subprocess boots successfully when launched detached with `stdio: ["ignore", "pipe", "pipe"]`. This avoids the silent `OpenCode server exited with code 0` rejection that fires on the first provider refresh after every daemon start.

## Symptom

In a typical Paseo 0.1.90 session with opencode 1.16.2 installed, the daemon log shows ~17 occurrences of:

```
Error: OpenCode server exited with code 0
    at packages/server/src/server/agent/providers/opencode/server-manager.ts:startServer
```

in an 8-minute window immediately after daemon start, then zero while the daemon is idle. The error fires for every `refresh_provider_snapshot` and `list_persisted_agents` request that touches the opencode provider — i.e. every time a client connects.

## Why `--pure`

opencode documents `--pure` as "run without external plugins" and ships it in 1.16.x. It is sufficient here: Paseo only needs the opencode HTTP API. The opencode subprocess does not need any plugin-loaded agents, MCPs, or skills because:

- Paseo injects its own MCP server into the opencode subprocess via the Agent MCP server (see `packages/server/src/server/agent/providers/opencode/runtime.ts` and the bootstrap log line `Agent MCP server mounted on main app` at `/mcp/agents`).
- Paseo uses the `ANTHROPIC_BASE_URL` rewrite (and the standard provider list) to drive model selection; it does not consume opencode plugin outputs.

## Root cause

This is a regression in opencode 1.16.x (released 2026-06-05, 1.16.0/1/2 all affected). The plugin/MCP loader aborts cleanly (exit code 0, no stderr) when running inside a child process spawned with `detached: true, stdio: ["ignore", "pipe", "pipe"]` and no controlling tty. The same `opencode serve` invocation runs perfectly in a TTY or under `tmux`/`script`.

Filed upstream as a regression in `anomalyco/opencode` (link in the linked issue).

## Verification

I verified the fix locally by patching the compiled output at
`~/.npm-global/lib/node_modules/@getpaseo/cli/node_modules/@getpaseo/server/dist/server/server/agent/providers/opencode/server-manager.js`
and restarting the daemon. The new worker (pid 3298829) booted in 185ms and a 30s tail of `~/.paseo/daemon.log` showed zero new `OpenCode server exited` events while the daemon was idle. The patched compiled file and the source change in this PR are identical (one arg added to the spawn args array).

## Out of scope

- No `runtimeSettings.extraArgs` config support. Keeping the fix minimal per Option A. A follow-up could add config-driven extra args for users who want to override the default.
- No tests added. Paseo AGENTS.md warns that provider tests run against a real opencode binary; CI for this change is best done on a host with opencode 1.16.x installed.

Closes #1362.

### DIFF
--- a/packages/server/src/server/agent/providers/opencode/server-manager.ts
+++ b/packages/server/src/server/agent/providers/opencode/server-manager.ts
@@ -199,7 +199,7 @@ export class OpenCodeServerManager implements OpenCodeServerManagerLike {
     return new Promise((resolve, reject) => {
       const serverProcess = spawnProcess(
         launchPrefix.command,
-        [...launchPrefix.args, "serve", "--port", String(port)],
+        [...launchPrefix.args, "serve", "--port", String(port), "--pure"],
         {
           detached: process.platform !== "win32",
           stdio: ["ignore", "pipe", "pipe"],


shrijit-ctrl:
Upstream regression filed as anomalyco/opencode#31006.

boudra:
Closing for now: this needs a regression test for detached startup and a real OpenCode before/after showing provider refresh and an agent run succeeding with the change. Please reopen with that evidence; an idle log tail does not verify the failing path.

boudra:

PR #1383, opened 2026-06-06
fix(server): skip live git in fetch_agent_history placement

### LINKED ISSUE #1382: Sessions list hangs and shows empty when sessions span slow/network filesystems
### What happens

Opening the **Sessions** list (top-left in the desktop app) sometimes hangs for ~10s and then shows an empty list — no history appears.

### Why

The daemon handles `fetch_agent_history_request` by resolving each agent's project placement through `getCheckoutStatus(cwd)`, which spawns a sequence of git subprocesses (`git rev-parse`, `git status`, …) per distinct cwd.

When agents live on slow filesystems (e.g. network mounts), a single `git status` can take seconds. Summed across all distinct cwds this exceeds the client's 10s RPC timeout (`fetch_agent_history` uses `timeout: 10000`), so the request times out and the list renders empty.

The history list doesn't even use this git data: the client requests it with `showCheckoutInfo={false}` and never reads the `project`/checkout payload, and no project filter is applied on the history path. So the per-cwd git work is pure overhead on this path.

### Repro

- Have agents whose cwds are on a slow/network filesystem, or a larger number of sessions across many repos.
- Open the Sessions list.
- Observe a long spinner followed by an empty list.

### Environment

- Real machine: ~180 sessions across 15 distinct cwds, several on network mounts.
- Measured `fetch_agent_history_request` latency: ~10.7-11.9s (over the 10s timeout).

yinaoxiong:
### Linked issue

Closes #1382

### Type of change

- [x] Bug fix

### What does this PR do

The Sessions (history) list resolves every agent's placement through `getCheckoutStatus(cwd)`, which spawns git subprocesses per cwd. On slow / network filesystems each `git status` takes seconds, and the sum across all distinct cwds blows past the client's 10s `fetch_agent_history` RPC timeout — so the list hangs and renders empty.

The history path doesn't use that git data at all: the client requests it with `showCheckoutInfo={false}`, never reads the `project`/checkout payload, and applies no project filter here. So the git work is pure overhead.

This PR makes `fetch_agent_history` resolve placement with `refreshGit: false` (cached snapshot + workspace registry lookup) instead of live git. The `fetch_agents` (active list) path is unchanged, the wire protocol is unchanged, and no client/CLI changes are needed.

### How did you verify it

Tested on a real machine with ~180 sessions across 15 distinct cwds, several on network mounts.

**Before (current code), from daemon.log:**
- `fetch_agent_history_request` latency: 10.7s – 11.9s — over the 10s client timeout, list came back empty.

**Micro-benchmark of the placement step** (real `getCheckoutStatus` against the real cwds, vs the registry lookup the fix uses):
- Before (per-cwd git): **~21,000 ms** total for 15 cwds (cold cache — the worst case the slow-FS user hits). Top offenders: one repo 4.6s, four network-mount repos ~3s each.
- After (registry lookup): **~0.03 ms**.

**End-to-end after the fix:** built the server, swapped it into my daemon, restarted, and opened the Sessions list from the app — it now loads sub-second instead of timing out.

**Automated test:** added a test that mocks `getCheckoutStatus` to throw and asserts the history path still returns entries without ever calling it (proves no git is spawned). Full `session.workspaces.test.ts` passes (53 passed, 4 pre-existing skips).

### Checklist

- [x] `npm run typecheck` passes (0 errors, all workspaces)
- [x] `npm run lint` passes (oxlint: 0 warnings, 0 errors)
- [x] `npm run format` ran
- [ ] UI changes include screenshots — N/A, server-only change

### DIFF
--- a/packages/server/src/server/session.ts
+++ b/packages/server/src/server/session.ts
@@ -6128,6 +6128,14 @@ export class Session {
         ? { ...request.filter, includeArchived: true }
         : request.filter;
     const scope = request.type === "fetch_agents_request" ? request.scope : undefined;
+    // The history list never renders checkout info (the client requests it with
+    // showCheckoutInfo=false and does not read the `project` payload), so resolving
+    // each agent's placement via live git is pure overhead. On slow filesystems
+    // (network mounts) a per-cwd `git status` can take seconds, and summed across
+    // many sessions it blows past the client's 10s RPC timeout, leaving the list
+    // blank. Resolve placement from cached snapshots + the workspace registry
+    // (refreshGit: false) instead — no git spawns on the history path.
+    const skipGitPlacement = request.type === "fetch_agent_history_request";
     const sort = this.agentsPager.normalizeSort(request.sort);
 
     let agents = await this.listAgentPayloads({
@@ -6154,7 +6162,10 @@ export class Session {
       if (existing) {
         return existing;
       }
-      const placementPromise = this.buildProjectPlacementForCwd(cwd);
+      const placementPromise = this.buildProjectPlacementForCwd(
+        cwd,
+        skipGitPlacement ? { refreshGit: false } : undefined,
+      );
       placementByCwd.set(cwd, placementPromise);
       return placementPromise;
     };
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

PR #1452, opened 2026-06-10
Authenticate injected agent MCP when daemon password is set

### LINKED ISSUE #690: Direct connection + password auth breaks Agent MCP / Inject Paseo tools with 401 on /mcp/agents
## Summary

When using a direct Tailscale connection with daemon password auth enabled, the main direct session works, but Agent MCP initialization fails when **Inject Paseo tools** is enabled.

The daemon logs show an unauthenticated `POST /mcp/agents` request (`hasToken:false`), followed by `401 Unauthorized` and `Failed to initialize Agent MCP`.

## Environment

- Paseo daemon: `0.1.66`
- Web/mobile client via `app.paseo.sh`
- VPS / headless daemon install via `@getpaseo/cli`
- Direct connection over Tailscale
- Relay disabled after verification

## Config shape

```json
{
  "version": 1,
  "daemon": {
    "listen": "100.x.y.z:6767",
    "hostnames": [
      "100.x.y.z",
      "host.tailnet.ts.net"
    ],
    "mcp": {
      "injectIntoAgents": true
    },
    "cors": {
      "allowedOrigins": [
        "https://app.paseo.sh",
        "http://100.x.y.z:6767",
        "http://host.tailnet.ts.net:6767"
      ]
    },
    "relay": {
      "enabled": false
    },
    "auth": {
      "password": "$2b$12$..."
    }
  }
}
```

## Repro steps

1. Run the daemon on a VPS or server with a direct Tailscale listener, e.g. `100.x.y.z:6767`.
2. Add the Tailscale IP / hostname to `daemon.hostnames` and `cors.allowedOrigins`.
3. Enable daemon password auth.
4. In the app, add the daemon as a **direct TCP connection** and enter the password.
5. Ensure **Inject Paseo tools** is enabled.
6. Connect successfully over the direct host.

## Actual behavior

- The direct session itself connects and works.
- The daemon logs:

```text
Rejected HTTP request with invalid daemon password
path: /mcp/agents
method: POST
hasToken: false
```

and then:

```text
Failed to initialize Agent MCP
Streamable HTTP error: Error POSTing to endpoint: {"error":"Unauthorized"}
```

- The workspace remains usable over the direct connection, but Agent MCP / injected Paseo tools do not initialize correctly.

## Expected behavior

If the app already knows the daemon password for the direct host, Agent MCP initialization should authenticate correctly as well.

In other words, when `injectIntoAgents` is enabled under direct + password-auth mode, requests to `/mcp/agents` should include the required bearer auth and should not fail with `401 Unauthorized`.

## Workaround

Set:

```json
"daemon": {
  "mcp": {
    "injectIntoAgents": false
  }
}
```

After disabling **Inject Paseo tools**, the direct connection remains healthy and I no longer see the repeated Agent MCP initialization failure.

## Why this looks like a bug / gap

The docs say that when password auth is enabled, **all HTTP requests** must include `Authorization: Bearer <password>` and unauthenticated requests receive `401 Unauthorized` (except `/api/health`). That matches the daemon behavior.

The part that seems broken is that the client appears to authenticate the direct WebSocket session, but does not carry the same auth through to the Agent MCP HTTP initialization path.

### LINKED ISSUE #914: Agent MCP injection does not forward daemon bearer auth, causing 401 Unauthorized when daemon password is enabled
Title: Agent MCP injection does not forward daemon bearer auth, causing 401 Unauthorized when daemon password is enabled

## Summary

When `daemon.mcp.injectIntoAgents` is enabled and the daemon also has `daemon.auth.password` configured, agents receive an injected MCP server URL for `/mcp/agents` but do not authenticate to it.

The injected agent-side MCP client then fails with `401 Unauthorized`, which breaks Agent MCP initialization.

## Environment

- Paseo Desktop `0.1.73`
- macOS
- Direct connection mode
- Daemon listening on `0.0.0.0:16767`
- `daemon.auth.password` configured
- `daemon.mcp.injectIntoAgents: true`

## Expected behavior

If daemon password auth is enabled, injected Agent MCP connections should still work.

The injected MCP client should forward the daemon password as:

`Authorization: Bearer <password>`

or the product should explicitly disable MCP injection when daemon auth is enabled and auth cannot be propagated.

## Actual behavior

Agents are injected with a bare MCP URL like:

```json
{
  "mcpServers": {
    "paseo": {
      "type": "http",
      "url": "http://0.0.0.0:16767/mcp/agents?callerAgentId=<agent-id>"
    }
  }
}
```

No auth metadata is included, and the daemon rejects the request with `hasToken:false`.

The agent then fails to initialize Agent MCP with a `401 Unauthorized`.

## Reproduction

1. Configure Paseo with password auth:

```json
{
  "version": 1,
  "daemon": {
    "listen": "0.0.0.0:16767",
    "mcp": {
      "injectIntoAgents": true
    },
    "auth": {
      "password": "$2b$12$..."
    }
  }
}
```

2. Restart the daemon.
3. Create an agent from the desktop app.
4. Let the agent initialize its injected MCP connection.
5. Observe `401 Unauthorized` on `/mcp/agents`.

## Relevant log output

```text
Server listening on http://0.0.0.0:16767
Daemon password authentication enabled
Rejected HTTP request with invalid daemon password
path="/mcp/agents" method="POST" hasToken=false
Failed to initialize Agent MCP
Streamable HTTP error: Error POSTing to endpoint: {"error":"Unauthorized"}
```

## Notes

- This looks like an auth propagation bug in the injected Agent MCP client path.
- The docs state that when daemon password auth is enabled, all HTTP clients must send `Authorization: Bearer <password>`.
- I also noticed the injected URL uses `0.0.0.0`, which is a listen address rather than an ideal client connect target. That is probably not the main cause of this 401, but it may still be worth revisiting.

## Additional findings

- `paseo daemon status` can report a false negative in this setup. In my environment, the daemon was actively serving desktop requests and `lsof` showed the port was listening, but the CLI still reported:

```text
Local daemon      unresponsive
Connected Daemon  unreachable
Note              Local daemon PID is running but websocket at 0.0.0.0:16767 is not reachable
```

- My CLI version was `0.1.72` while the desktop app/daemon was `0.1.73`, so this may be a version-interaction issue or a CLI health-check issue specific to `0.0.0.0`.
- I also tried changing the daemon listen address to `127.0.0.1:16767` to avoid the `0.0.0.0` client-target problem, but on this machine the daemon then failed to start with:

```text
listen EPERM: operation not permitted 127.0.0.1:16767
```

- Restoring `0.0.0.0:16767` allowed the daemon to listen again.

## Workaround

Set:

```json
"daemon": {
  "mcp": {
    "injectIntoAgents": false
  }
}
```

and restart the daemon.

## Suggested fix

When Paseo creates the injected Agent MCP client, it should pass daemon auth through to `StreamableHTTPClientTransport`, so the HTTP requests include the bearer token required by the daemon.

Separately, it may be worth hardening the CLI daemon health check so it does not attempt to validate reachability by connecting to `0.0.0.0` as if it were a client destination.

dwyanewang:
### Linked issue

Closes #914
Also fixes #690

### Type of change

- [x] Bug fix

### What does this PR do

With both `daemon.auth.password` and `daemon.mcp.injectIntoAgents` set, the MCP connection injected into agents (and the daemon's own per-session MCP client) sent no `Authorization` header, so every `/mcp/agents` request got 401 and agents lost all `mcp__paseo__*` tools, silently falling back to the CLI.

Fix: the daemon mints a per-run capability token (random per start), injects it as a bearer only into local agent configs and its own MCP client (never sent to remote/relay clients), and exempts `/mcp/agents` from the global password middleware so the route self-authorizes — accepting that token or a valid daemon-password bearer. Works for env-set and app-set (hash-only) passwords; unauthenticated `/mcp/agents` requests are still rejected. Mirrors the existing `/api/files/download` capability-token pattern.

### How did you verify it

Repro + fix against a real password-protected daemon on a spare port + separate `PASEO_HOME` (not my main one):

- Before: a new agent's `/mcp/agents` calls logged `hasToken:false → 401` and `Failed to initialize Agent MCP (401)`; the agent had no paseo tools.
- After (this branch): started the daemon with a password, created a real Claude agent — it called `mcp__paseo__list_agents` and got a result; the daemon's own MCP client initialized with 33 tools; the daemon log had zero 401 / invalid-password lines.
- Negative check: `curl -X POST /mcp/agents` with no token and with a wrong token → both `401` (remote auth not weakened).
- `npm run typecheck`, lint, and the new/updated unit + e2e tests pass.

### Checklist

- [x] `npm run lint` passes
- [ ] UI changes include screenshots or video for every affected platform (N/A — no UI)

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -992,6 +992,47 @@ test("createAgent injects paseo MCP server only into provider launch config", as
   });
 });
 
+test("createAgent injects the MCP auth token as a bearer header into the launch config", async () => {
+  const workdir = mkdtempSync(join(tmpdir(), "agent-manager-test-"));
+  const storagePath = join(workdir, "agents");
+  const storage = new AgentStorage(storagePath, logger);
+
+  class CaptureClient extends TestAgentClient {
+    lastConfig: AgentSessionConfig | null = null;
+
+    override async createSession(config: AgentSessionConfig): Promise<AgentSession> {
+      this.lastConfig = config;
+      return new TestAgentSession(config);
+    }
+  }
+
+  const client = new CaptureClient();
+  const manager = new AgentManager({
+    clients: {
+      codex: client,
+    },
+    registry: storage,
+    logger,
+    mcpBaseUrl: "http://127.0.0.1:6767/mcp/agents",
+    mcpAuthToken: "cap-token",
+    idFactory: () => "00000000-0000-4000-8000-000000000104",
+  });
+
+  const snapshot = await manager.createAgent({
+    provider: "codex",
+    cwd: workdir,
+  });
+
+  expect(manager.getMcpAuthToken()).toBe("cap-token");
+  expect(client.lastConfig?.mcpServers?.paseo).toEqual({
+    type: "http",
+    url: `http://127.0.0.1:6767/mcp/agents?callerAgentId=${snapshot.id}`,
+    headers: { Authorization: "Bearer cap-token" },
+  });
+
+  rmSync(workdir, { recursive: true, force: true });
+});
+
 test("resumeAgentFromPersistence replaces stored internal paseo MCP with current runtime URL", async () => {
   const workdir = mkdtempSync(join(tmpdir(), "agent-manager-test-"));
   const storagePath = join(workdir, "agents");
--- a/packages/server/src/server/agent/agent-manager.ts
+++ b/packages/server/src/server/agent/agent-manager.ts
@@ -196,6 +196,7 @@ export interface AgentManagerOptions {
   durableTimelineStore?: AgentTimelineStore;
   terminalManager?: TerminalManager | null;
   mcpBaseUrl?: string;
+  mcpAuthToken?: string;
   appendSystemPrompt?: string;
   agentStreamCoalesceWindowMs?: number;
   rescueTimeouts?: AgentManagerRescueTimeouts;
@@ -482,6 +483,7 @@ export class AgentManager {
   private readonly backgroundTasks = new Set<Promise<void>>();
   private readonly agentStreamCoalescer: AgentStreamCoalescer;
   private mcpBaseUrl: string | null;
+  private readonly mcpAuthToken: string | null;
   private appendSystemPrompt: string;
   private onAgentAttention?: AgentAttentionCallback;
   private onAgentArchived?: AgentArchivedCallback;
@@ -496,6 +498,7 @@ export class AgentManager {
     this.onAgentAttention = options?.onAgentAttention;
     this.onWorkspaceStateMayHaveChanged = options?.onWorkspaceStateMayHaveChanged;
     this.mcpBaseUrl = options?.mcpBaseUrl ?? null;
+    this.mcpAuthToken = options?.mcpAuthToken ?? null;
     this.appendSystemPrompt = options.appendSystemPrompt ?? "";
     this.logger = options.logger.child({ module: "agent", component: "agent-manager" });
     this.rescueTimeouts = {
@@ -552,6 +555,16 @@ export class AgentManager {
 
   setMcpBaseUrl(url: string | null): void {
     this.mcpBaseUrl = url;
+  }
+
+  /**
+   * Capability token the daemon's own MCP clients must present to the Agent MCP
+   * endpoint when a daemon password is configured. Read by the per-client
+   * session to authenticate its own MCP connection. Stays in the daemon — never
+   * sent to remote clients.
+   */
+  getMcpAuthToken(): string | null {
+    return this.mcpAuthToken;
   }
 
   setAppendSystemPrompt(prompt: string | null | undefined): void {
@@ -3567,6 +3580,7 @@ export class AgentManager {
         config: storedConfig,
         agentId,
         mcpBaseUrl: this.mcpBaseUrl,
+        mcpAuthToken: this.mcpAuthToken,
       }),
     );
     return { storedConfig, launchConfig };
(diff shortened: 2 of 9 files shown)


dwyanewang:
@greptile review

Addressed both comments in 496e86c6:

- **auth.ts (P2, constant-time compare):** Switched the capability-token comparison to `crypto.timingSafeEqual`, with a length guard and the `node:crypto` import. Heads-up that the suggested snippet as-is would break: `timingSafeEqual` throws a `RangeError` when the two buffers differ in length, so `timingSafeEqual(Buffer.from(token), Buffer.from(capabilityToken))` would throw on any wrong-length token — and since this check runs inside a fire-and-forget route handler, the rejected promise would leave the request hanging. The length-guarded form (`provided.length === expected.length && timingSafeEqual(...)`) falls through to the password check instead.
- **agent-mcp.e2e.test.ts (P2, weak assertion):** Replaced `toBeTruthy()` with `expect(typeof capabilityToken).toBe("string")`.

dwyanewang:
Hi @boudra — small nudge on this when you get a sec. It fixes the daemon-password + Agent MCP 401 bug (closes #914 & #690, both open ~5 weeks). Rebased onto latest main so it's mergeable, follows the PR template, and tests pass. The only thing holding it back is the first-time-contributor workflow approval so CI can run. Happy to adjust anything if you'd rather it look different. Thanks!

dwyanewang:
Thanks for running it, @boudra! The only red check is `playwright`, and it's a pre-existing flake on `main`, not something this PR introduces — this is a server-only change (no `packages/app` code touched). The failing case is the `welcome-screen` startup E2E (`startup-dsl.ts:143`, 15s visibility timeout). It flakes on `main` too, including on `c0614c32` — the exact commit this PR is rebased onto (actions/runs/27336375195). A re-run of the playwright job should clear it; everything else (server-tests, typecheck, lint, build, …) is green.

dwyanewang:
@boudra — quick clarification so this doesn't look like a stuck PR: the only earlier red check (`playwright`) was the app-side `welcome-screen` startup/reconnect E2E, which was red on main itself around my old base and went green after the #1472 reconnect-expectation update. This PR is server-only — zero `packages/app` changes — so it can't affect that test; the rebase was just to move past that window, not a code change. It's now on current main, mergeable, with the other 14 checks green. Ready whenever you can approve the workflow / take a look — thanks!

PR #1460, opened 2026-06-10
Use terminateWithTreeKill for ACP agent child process cleanup

### LINKED ISSUE #1459: bug: ACP agent child processes leak after session close
### What's broken

When a scheduled agent uses an ACP provider (e.g. qoder), the child process tree (npx → sh -c → node qodercli) is not fully cleaned up after the agent session ends. ACPAgentSession.close() calls child.kill("SIGTERM") which only signals the direct child (npx), leaving grandchild processes (node qodercli) alive as orphans.

Expected: all processes in the tree are killed when the agent session closes.
Actual: grandchild processes survive and accumulate. Over 3 days running 3 scheduled agents per day, 24 orphan qodercli processes accumulated consuming ~8GB RAM (~330MB each).

This is the ACP-provider counterpart of #1396 (OpenCode on Windows). The root cause is different: #1396 is about tree-kill not working through Windows .CMD wrappers; this issue is about close() never calling tree-kill at all — it uses plain child.kill("SIGTERM") while every other provider (OpenCode, Pi, Codex) uses terminateWithTreeKill().

### Steps to reproduce

1. Configure a qoder ACP provider (or any custom ACP provider that spawns via npx)
2. Create a scheduled agent: paseo schedule create "hello" --provider qoder/q35model_preview --cron "*/5 * * * *"
3. Let it run a few times
4. After agents finish: paseo agent ls shows empty
5. Check for leaked processes: ps aux | grep qodercli | grep -v grep
6. Orphan node qodercli --acp processes are still running

### Where did this happen

Daemon

### Paseo version

Paseo version: 0.1.80 (also reproduced on 0.1.90)

### OS version

Linux 5.10.112 (Alibaba Cloud, x86_64)

### Agent provider

Custom provider

### Provider configuration

Provider configuration:
"qoder": {
  "extends": "acp",
  "label": "Qoder CLI",
  "command": [
    "npx", "-y", "@qoder-ai/qodercli@1.0.4",
    "--acp", "--dangerously-skip-permissions"
  ]
}

### Logs

```text

```

### Screenshots or video

_No response_

yeshan333:
### Linked issue

Closes #1459

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

ACPAgentSession.close() used child.kill("SIGTERM"), which only signals the direct child process (for example, npx). When an ACP provider like qoder spawns a process tree (npx -> sh -c -> node qodercli), grandchild processes can survive as orphans. Over 3 days with scheduled agents, 24 orphan qodercli processes accumulated consuming ~8GB RAM.

The fix replaces direct child.kill("SIGTERM") cleanup in the ACP agent lifecycle with terminateWithTreeKill(), which recursively walks /proc on Linux and uses taskkill /T on Windows to terminate the whole process tree. This aligns ACP with OpenCode, Pi, and Codex providers that already use terminateWithTreeKill().

Affected code paths: close(), killTerminal(), releaseTerminal(), terminateChildProcess(). The PR also removes the now-unused waitForChildExit(), terminates multiple terminal children in parallel during close(), and ensures probe cleanup signals the process tree before destroying stdio so descendants are not missed if the direct child exits on stdin EOF.

### How did you verify it

1. Observed the bug in production: ran 3 scheduled qoder agents per day for 3 days. ps aux | grep qodercli showed 24 orphan processes after paseo agent ls returned empty.
2. Integration tests with real process trees (Linux, non-Windows):
  - Spawned a parent -> grandchild tree where grandchild traps SIGTERM
  - Verified child.kill("SIGTERM") kills only the parent, grandchild stays alive (bug reproduced)
  - Verified terminateWithTreeKill() kills both parent and grandchild (fix confirmed)
3. Unit tests now cover close(), killTerminal(), terminal close parallelism, and probe cleanup ordering around stdio destruction.
4. Local checks run after the review fixes:
  - npm run format
  - npx vitest run packages/server/src/server/agent/providers/acp-agent.test.ts --bail=1 — 50/50 passed
  - npm run lint
  - npm run typecheck

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -28,6 +28,7 @@ import {
   resolveACPModelSelection,
   summarizeACPRequestError,
 } from "./acp-agent.js";
+import * as treeKillModule from "../../../utils/tree-kill.js";
 import {
   COPILOT_ALLOW_ALL_MODE_ID,
   COPILOT_MODES,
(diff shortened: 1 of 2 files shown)


TommyLike:
This issue also affects the **Claude Code provider** (`packages/server/src/server/agent/providers/claude/agent.ts`). 

The `close()` method (line ~1610) only calls `query?.interrupt?.()` / `query?.return?.()` to gracefully end the Agent SDK session, but **never sends any OS-level signal to kill the underlying process tree**. There is no `process.kill()`, no `terminateWithTreeKill()`, no cleanup of spawned subprocesses.

## Production Evidence

Running paseo 0.1.93 on Ubuntu 24.04 (15GB RAM) with multiple Claude Code workspaces:

- **After 16 hours of normal use**: accumulated **19 orphaned `claude` processes** (280-780MB each) and **29 MCP child processes** (gmail-mcp via uv, playwright-mcp, etc.)
- **Total waste**: ~14GB out of 15GB consumed
- **System impact**: OOM killer triggered, killing active Claude Code sessions mid-task
- **Root cause**: When paseo closes a Claude Code session, the SDK query is interrupted but the OS-level `claude` process (and all its MCP children) keep running

## Request

The same `terminateWithTreeKill()` approach used here for ACP should be applied to the Claude Code provider. Every other provider (OpenCode, Pi, Codex) already uses tree-kill — Claude Code is the last one missing it.

This is the same class of bug as #1459, just in a different provider file that was overlooked.

yeshan333:
@boudra Hi, could you take a look when you have a moment? Thanks!

PR #1485, opened 2026-06-12
fix(desktop): stop AppImage updates hanging on quit and deleting the app

xpufx:
AppImage auto-updates were unreliable on Linux: the update hung the app on quit and deleted the running binary. This PR fixes both — the hang, and the orphaned/renamed file that broke desktop shortcuts and the `paseo` CLI symlink.

### Linked issue

Closes # n/a

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

**Disable auto-install-on-quit for AppImage (original change).** electron-updater downloads a new AppImage in the background, then on quit tries to "install" it via a blocking `execFileSync(newAppImage, { APPIMAGE_EXIT_AFTER_INSTALL })`. That env var is only honored by AppImageLauncher, so without it the new process never exits and the quit hangs — with the old binary already unlinked. Scoped to AppImage only (`process.platform === 'linux' && process.env.APPIMAGE`); `.deb`/`.rpm`/Windows/macOS keep auto-install-on-quit.

**Keep the AppImage filename stable (maintainer follow-up).** With a versioned `artifactName`, electron-updater renames the AppImage to a new path on every update and unlinks the old one — orphaning the previous binary, breaking desktop shortcuts, and dangling the `~/.local/bin/paseo` CLI symlink (which points at `$APPIMAGE`). Dropping `${version}` from the AppImage `artifactName` makes the updater overwrite the file in place. `.deb`/`.rpm` keep their versioned names. **Trade-off:** the AppImage release asset is now `Paseo-${arch}.AppImage` (no version in the filename), and existing versioned installs do one final rename on their next update, then stay stable.

**Docs + test.** The auto-install-on-quit gate is now a documented, unit-tested `shouldAutoInstallOnQuit()` helper that explains the hang in-code instead of only in the commit message.

### How did you verify it

- *(original)* Built a patch against the previous release to ensure the update fires, then applied it to latest head.
- *(maintainer)* `npm run typecheck`, `npm run lint`, and the desktop `auto-updater` test file pass locally; added a unit test covering the gate across linux-AppImage / linux-deb·rpm / macOS / Windows.

### Beyond CI

CI does not exercise the Linux desktop auto-updater end-to-end. Before merge, worth one real AppImage update check: (1) an update applied via "Update now" relaunches without hanging, and (2) the post-update file keeps the same path so a pinned shortcut and the `paseo` symlink still resolve. `.deb`/`.rpm` install-on-quit is intentionally unchanged.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/desktop/electron-builder.yml
+++ b/packages/desktop/electron-builder.yml
@@ -52,6 +52,13 @@ linux:
     - deb
     - rpm
     - tar.gz
+appImage:
+  # Keep the AppImage filename stable across versions (no ${version}) so
+  # electron-updater overwrites the file in place instead of renaming it to a new
+  # versioned path. A rename orphans the old binary, breaks desktop shortcuts, and
+  # dangles the ~/.local/bin/paseo CLI symlink, which points at $APPIMAGE. The
+  # version stays discoverable via the release tag and the in-app version.
+  artifactName: "Paseo-${arch}.${ext}"
 win:
   artifactName: "Paseo-Setup-${version}-${arch}.${ext}"
   icon: assets/icon.ico
--- a/packages/desktop/src/features/auto-updater.test.ts
+++ b/packages/desktop/src/features/auto-updater.test.ts
@@ -19,7 +19,17 @@ import {
   resolveStagingUserId,
   rolloutManifestSchema,
   shouldAdmitToRollout,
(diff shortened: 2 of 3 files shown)


xpufx:
Update should probably not be downloaded either but that's different issue.

PR #1511, opened 2026-06-13
Forward a remote dev-server port to client localhost (TCP tunnel)

marvinvr:
## What & why

Adds an in-product `ssh -L`: a service running on a **remote** daemon host becomes reachable at `http://localhost:<port>` on the client, riding the **existing** daemon WebSocket — no second connection, no DNS/reverse-proxy, no `0.0.0.0` bind, no open firewall port.

The motivation is **localhost fidelity**. Reaching a remote dev server at `http://<host>:<port>` breaks anything that assumes localhost — most importantly OAuth: Google/Microsoft only allow `http://` redirect URIs for `localhost`/`127.0.0.1`. Forwarding the same port to the client's own loopback makes the browser's origin genuinely `localhost:<port>`, so registered callbacks line up with zero app or provider-dashboard changes.

## How it works for the user

Open a running `paseo.json` `"type": "service"` script's link in the workspace scripts menu. When the connection is remote and the daemon supports it, Paseo lazily binds the same port locally, tunnels it, and opens `http://localhost:<port>` in the in-app browser instead of the remote/proxy URL. Old daemons fall back to the existing direct/proxy URL (capability-gated, no degraded path).

Desktop (Electron) only for now — it needs a main-process loopback listener.

## Design

```
browser ⇄ Electron main loopback listener ⇄ renderer controller
        ⇄ daemon client tunnel frames (existing WS) ⇄ daemon TunnelForwarder ⇄ 127.0.0.1:<port>
```

- **protocol**: new `0x20`-range binary frame family (`Open`/`Data`/`Close`) + `server_info.features.tcpTunnel` flag (`COMPAT`-gated).
- **server**: per-session `TunnelForwarder` — **loopback-only by construction** (the client supplies a port, never a host, so it can't become an open proxy), backpressure via source-socket pause/resume, graceful teardown.
- **client**: tunnel frame send/receive on `DaemonClient`.
- **desktop**: main-process `net` listener bridged to the renderer over IPC.
- **app**: tunnel controller + tunnel-aware service-link open.

## Testing

`typecheck` / `lint` / `format` all green. New coverage (34 tests): frame codec + demux round-trips, the daemon forwarder (loopback echo, connect-failure, graceful close), link resolution, and an **end-to-end test** that drives the real client over a real daemon WebSocket and forwards a 200 KB HTTP response through the tunnel intact. The Electron glue + actual OAuth flow is a two-machine manual check (see `docs/port-forwarding.md`).

## Known limitations (v1)

Documented in `docs/port-forwarding.md`: forwards reclaimed on window close (not per browser tab); two live windows can't bind the same local port (second falls back to direct); no upload-direction backpressure watermark; a >90s daemon outage orphans in-flight streams.


BetterAndBetterII:
we really need this

iamriajul:
@marvinvr any specific reason for closing this?

KennanChan:
Why was this PR closed? Port forwarding support would make Paseo perfect for me.

marvinvr:
I didn't hear a response and I'm not using paseo anymore for my workflow. Feel free to take my code and use it in a new PR as you wish.

PR #1535, opened 2026-06-14
fix: prevent IME Enter key from sending message on macOS

### LINKED ISSUE #309: Enter key submits input when using CJK IME

jkcomment:
## Problem

On macOS (Chromium/Electron), the `compositionend` event fires **before** the `keydown` event for the Enter key. This means `event.nativeEvent.isComposing` is already `false` by the time `handleDesktopKeyPressImpl` runs, so the existing IME guard doesn't fire and the message is sent when the user intended only to confirm the composition.

This affects **Korean, Japanese, and Chinese** (CJK) input on macOS. The behavior is intermittent because it depends on exact event ordering, which varies by input method and OS version.

Closes #309, related to #264, #241.

## Fix

Add a `useNativeImeComposingRef` hook in `keyboard-ime.ts` that:
- Attaches native DOM `compositionstart` / `compositionend` listeners directly to the textarea element
- Defers the flag clear in `compositionend` by one tick (`setTimeout(0)`) so the Enter `keydown` handler still sees that composition just ended

In `input.tsx`, call this hook and check `isNativeComposingRef.current` as an early return in `handleDesktopKeyPress` before delegating to `handleDesktopKeyPressImpl`.

## Testing

1. Open Paseo on macOS
2. Switch input method to Korean (or Japanese/Chinese)
3. Type a word using IME (e.g. `안녕`)
4. Press Enter to confirm the composition
5. **Before fix**: message sends immediately
6. **After fix**: composition is confirmed, Enter sends only on a second press

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/app/src/composer/input/input.tsx
+++ b/packages/app/src/composer/input/input.tsx
@@ -1657,8 +1657,10 @@ export const MessageInput = forwardRef<MessageInputRef, MessageInputProps>(
     function handleDesktopKeyPress(event: WebTextInputKeyPressEvent) {
       if (!shouldHandleWebKeyPress) return;
       // On macOS, compositionend fires before keydown so isComposing is already
-      // false when Enter is pressed. Check the native ref as a fallback.
-      if (isNativeComposingRef.current) return;
+      // false when Enter is pressed. Check the native ref as a fallback, but
+      // only for Enter — the deferred window is one tick so non-Enter keys
+      // should not be affected.
+      if (event.nativeEvent.key === "Enter" && isNativeComposingRef.current) return;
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

PR #1548, opened 2026-06-15
fix(app): keep desktop file attachment extensions dotted

### LINKED ISSUE #1547: bug: Desktop file uploads fail for files with extensions
### What's broken

On Desktop (Electron), generic file uploads failed for files with extensions. A `.zip` upload was rejected with:

`Invalid attachment extension: zip`

Expected behavior: selecting and uploading a file like `something.zip` should succeed.

This was caused by the app stripping the leading dot from the file extension before sending it through desktop IPC validation, so the desktop-side validation rejected extensions like `zip` instead of accepting `.zip`. The same root cause affects any desktop generic upload with an extension, not just `.zip`.

### Steps to reproduce

1. Open the Desktop (Electron) app.
2. Trigger a generic file upload flow.
3. Select a file with an extension, for example `test.zip`.
4. Confirm the upload.

Observed before the fix: the upload fails with `Invalid attachment extension: zip`.

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.96

### OS version

macOS 26.5.1

### Agent provider

None

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

Image

Ovler-Young:
### Linked issue

Closes #1547

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Desktop generic file uploads were stripping the leading dot from file extensions before copying files into Electron-managed attachment storage.

The desktop IPC validation expects dotted extensions such as `.zip`, `.pdf`, or `.csv`, but the app was passing values such as `zip`, `pdf`, or `csv`. That caused uploads for files with extensions to fail with errors like:

`Invalid attachment extension: zip`

This change preserves the dotted extension returned by `getFileExtension()` before calling `copyDesktopAttachmentFile()`. It applies to both direct desktop file reads and the desktop file picker path, including drag/drop flows that reuse `readDesktopFileBytes()`.

`getFileExtension()` already returns the final filename suffix with the leading dot, for example `.zip` or `.gz`. The desktop managed-storage IPC expects that same dotted form, so this change passes the value through directly instead of stripping the dot with `.slice(1)`.

### How did you verify it

- [Built an app package from this patched branch.](https://github.com/ovlerfork/paseo/actions/runs/27524869831)
- Installed that build locally.
- Reproduced the affected desktop file-upload flow with a `.zip` file.
- Confirmed the upload now succeeds; the previous `Invalid attachment extension: zip` failure no longer occurs.
- Added automated coverage for preserving dotted extensions across `.zip`, `.pdf`, and `.csv`.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/hooks/use-file-picker.test.ts
+++ b/packages/app/src/hooks/use-file-picker.test.ts
@@ -1,0 +1,116 @@
+import { describe, expect, it } from "vitest";
+import {
+  pickFilesWithDesktopDialog,
+  readDesktopFileBytes,
+  type DesktopFilePickerDependencies,
+  type DesktopFileReaderDependencies,
+} from "./use-file-picker";
+import type { DesktopDialogOpenOptions } from "@/desktop/host";
+
+interface CopiedAttachmentFile {
+  attachmentId: string;
+  sourcePath: string;
+  extension?: string | null;
+}
+
+class FakeDesktopFileReaderDependencies implements DesktopFileReaderDependencies {
+  readonly copiedFiles: CopiedAttachmentFile[] = [];
+  readonly reads: string[] = [];
+  private nextId = 1;
+  private readonly files = new Map<string, string>();
+
+  async copyAttachmentFile(input: CopiedAttachmentFile): Promise<{ path: string; byteSize: number }> {
+    this.copiedFiles.push(input);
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

PR #1554, opened 2026-06-16
Preserve subdirectory cwd when creating worktrees

nwjq:
## What changes

When an agent workspace is opened from a subdirectory of the repo root (e.g. `~/repo/packages/app`), creating a worktree from that workspace should place the new workspace at the corresponding subdirectory inside the worktree (e.g. `~/.paseo/worktrees/slug/packages/app`). Currently, `upsertWorkspaceForWorktree` hard-codes `resolve(options.worktree.worktreePath)` as the new workspace's `cwd`, losing the subdirectory offset.

This happens because the app's various worktree-creation entry points (sidebar, keyboard shortcut, new-workspace screen) send the **project root** as the `cwd` in the `create_paseo_worktree_request` RPC, not the active workspace's subdirectory. Rather than patching every UI entry point, this fix resolves the subdirectory on the server side by looking up the source workspace from the registry.

Three functions added to `paseo-worktree-service.ts`:

- **`findSourceWorkspaceForWorktree`** — Looks up the source workspace by `projectId` (when provided) or by `cwd`/`repoRoot` match. Only considers non-archived, non-worktree workspaces. This is how the server recovers the actual subdirectory even when the request sends the project root.

- **`resolveSourceRepoRoot`** — Resolves the git root of the effective input cwd via `workspaceGitService.resolveRepoRoot`. Needed to compute the relative offset without depending on `sourceRepoRoot` from `createWorktreeCore`.

- **`resolveWorktreeWorkspaceDirectory`** — Computes `path.relative(sourceRepoRoot, inputCwd)` and maps it into the new worktree path. Uses `realpathSync.native` for symlink-safe comparison (macOS `/tmp` → `/private/tmp`). Falls back to the worktree root when the relative path escapes the root or is empty.

No protocol changes, no app-side changes, no changes to `worktree-core.ts`.

## How to reproduce

1. Open a project from a subdirectory (e.g. `~/repo/packages/app`)
2. Create a worktree from that workspace via the sidebar "+" button
3. Open a terminal in the new worktree workspace
4. **Before fix:** terminal cwd is the worktree root (`~/.paseo/worktrees/slug/`)
5. **After fix:** terminal cwd is `~/.paseo/worktrees/slug/packages/app`

## How to verify beyond CI

```bash
npx vitest run packages/server/src/server/paseo-worktree-service.test.ts --bail=1
```

CI covers typecheck, lint, and the test suite. Manual verification: create a worktree from a subdirectory workspace and confirm the new workspace's "Copy path" shows the subdirectory, and the terminal opens in it.

## Risk surface

- **Workspace registry read**: `findSourceWorkspaceForWorktree` calls `workspaceRegistry.list()` once per worktree creation. This is the same pattern already used by `findWorkspaceByDirectory` and `findWorkspaceForSource` in the same function. No performance concern — worktree creation is infrequent and the list is small.
- **Fallback behavior**: When no source workspace is found (e.g. CLI usage without `projectId`), `effectiveInputCwd` falls back to the original `inputCwd`, preserving existing behavior exactly.
- **Existing tests**: All 11 existing tests pass unchanged. The `getSnapshot` spy assertion stays as `not.toHaveBeenCalled()` — the new code path uses `resolveRepoRoot`, not `getSnapshot`.

### DIFF
--- a/packages/server/src/server/paseo-worktree-service.test.ts
+++ b/packages/server/src/server/paseo-worktree-service.test.ts
@@ -1,5 +1,5 @@
 import { execFileSync } from "node:child_process";
-import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { afterEach, expect, test, vi } from "vitest";
@@ -30,7 +30,7 @@ afterEach(() => {
   }
 });
 
-test("creates a worktree and registers it in the source workspace project without git snapshot lookup", async () => {
+test("creates a worktree and registers it in the source workspace project without extra git snapshot lookup", async () => {
   const { repoDir, tempDir } = createGitRepo();
   cleanupPaths.push(tempDir);
   const events: string[] = [];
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the implementation merged in [#2098](https://github.com/getpaseo/paseo/pull/2098).

PR #1581, opened 2026-06-18
Keep connections alive through brief daemon slowdowns

boudra:
### Linked issue

No specific issue — surfaced while dogfooding the relay: the app showed a red `Liveness check timed out (5000ms)` and dropped a healthy connection while the daemon was briefly busy (an Android APK build was saturating the machine, stalling the daemon's event loop for 2–5.5s).

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Stops the client from tearing down a live connection when the daemon is alive but briefly slow.

The active connection's liveness check had no independent existence — it was a side effect of the candidate-ranking probe loop. A single 5s *latency measurement* doubled as the *death* timeout, so a pong that came back late from a stalled-but-alive daemon was counted identically to no pong. Two such cycles dropped a perfectly good connection.

This PR:
- Gives `DaemonClient` its own self-scheduling heartbeat that solely owns liveness — relaxed to a 15s timeout and 2 consecutive misses (~50s of continuous silence before teardown; ~2.7× the observed p99 stall of 5.5s).
- Makes the probe **read-only** on the active connection: it reads the heartbeat's last RTT for ranking instead of pinging the live client. Non-active candidates are still actively measured as before.
- Splits the shared ping primitive into one pure core (`sendPingAwaitRtt`) with two thin wrappers — public `measureLatency` (no side effect, candidates only) and private `livenessPing` (drives the teardown counter, heartbeat only). Nothing outside the client can drive teardown anymore.

A secondary defect — the probe switching away from a live connection on a single failed probe — dissolves structurally as a result (the active connection can no longer be marked unavailable while online).

### How did you verify it

Verified via a new deterministic test suite, **not** a live relay reproduction — I did not stand up a real stalled daemon end-to-end.

- New unit tests at the `DaemonClient` and `HostRuntimeController` boundaries, time driven by fake timers (RTT/timeouts deterministic). The headline regression test reproduces the incident: pongs arriving 5.5s late for 10 simulated minutes → connection stays online, zero teardowns, and it asserts the heartbeat actually ran (`lastLivenessRttMs === 5500`). It fails against the old 5s-timeout behavior.
- Coverage includes: silent-wedge teardown at ~50s with the `(15000ms)` reason, recovery after one missed pong, idle-quiet keep-alive, heartbeat start/stop lifecycle, the probe issuing no ping to the live client, and that candidate measurement never touches the liveness counter (two candidate timeouts → no teardown).
- `npx vitest run packages/client/src/daemon-client.test.ts` → 75 passed.
- `npm run test --workspace=@getpaseo/app -- src/runtime/host-runtime.test.ts` → 36 passed.

Worth a maintainer dogfood before relying on it: rebuild, kick off something that saturates the machine (the original trigger was an APK build), and confirm the connection now holds instead of flashing the red `Liveness check timed out` and dropping.

### Risk surface

- **Core connectivity, all clients.** The logic lives in `DaemonClient` / `HostRuntimeController`, so it applies identically to mobile, web, and desktop. CI's unit suite covers the logic; the live relay path under real load is what a manual dogfood would add.
- **Slower silent-death detection.** A genuinely dead-but-silent half-open connection now takes ~50s to detect (was ~10–20s). Hard transport closes (socket reset, daemon exit) are still instant — only the silent-wedge case waits.
- **Relay `1011 \"Control unresponsive\"` is not addressed here.** A long enough daemon stall can still trip the relay's own server-side watchdog and surface as a disconnect; that's a separate lever, intentionally out of scope.
- **No protocol/schema/wire change** — reuses the existing `ping`/`pong` frames, so old/new client↔daemon compatibility is unaffected.

### Checklist

- [x] `npm run lint` passes
- [ ] UI changes include screenshots or video for every affected platform — n/a, no UI changes

### DIFF
--- a/packages/app/src/runtime/host-runtime.test.ts
+++ b/packages/app/src/runtime/host-runtime.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import type {
   DaemonClient,
   ConnectionState,
@@ -12,16 +12,16 @@ import {
   HostRuntimeController,
   HostRuntimeStore,
   type HostRuntimeControllerDeps,
-  type HostRuntimeSnapshot,
 } from "./host-runtime";
 
 class FakeDaemonClient {
   private state: ConnectionState = { status: "idle" };
   private listeners = new Set<(status: ConnectionState) => void>();
   private error: string | null = null;
+  private heartbeatRttMs: number | null = null;
+  private latencyMeasurementFailure: Error | null = null;
+  private latencyMeasurementsRequested: Array<{ timeoutMs?: number }> = [];
   public connectCalls = 0;
-  public closeCalls = 0;
-  public ensureConnectedCalls = 0;
   public fetchAgentsCalls: FetchAgentsOptions[] = [];
   public fetchAgentsResponses: Awaited<ReturnType<DaemonClient["fetchAgents"]>>[] = [];
 
@@ -31,12 +31,10 @@ class FakeDaemonClient {
   }
 
   async close(): Promise<void> {
-    this.closeCalls += 1;
     this.setConnectionState({ status: "disconnected", reason: "client_closed" });
   }
 
   ensureConnected(): void {
-    this.ensureConnectedCalls += 1;
     if (this.state.status !== "connected") {
       this.setConnectionState({ status: "connected" });
     }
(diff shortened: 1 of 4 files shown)


boudra:
Addressed the shared-`pingProbe` edge case in 671645111.

Each ping now carries a `drivesLivenessFailure` flag — only the heartbeat sets it. `recordLivenessFailure` is gated on that flag, so a `measureLatency` timeout can never be recorded as a liveness failure, even when a heartbeat tick dedupes onto an in-flight measurement ping. Added a regression test that forces exactly that interleaving (candidate measurement still in flight when the +10s tick lands) and asserts a single genuine miss afterward does not tear the connection down — it was red before the fix.

PR #1588, opened 2026-06-18
feat(app): browse host folders in project picker

### LINKED ISSUE #1508: feat: browse host folders when opening projects from mobile
## Prior search

Related but not the same: #414 covers Windows mobile directory suggestions being unable to access paths outside HOME. This request is broader: mobile users should be able to browse/select host folders without already knowing the absolute path.

## What's the problem

When controlling a host from the phone, opening a new project currently depends on typing or searching for a directory path on the host machine. That works only if the user already knows the exact absolute path, for example `/Users/calm/Downloads/Archives/project-name`.

On desktop, the native folder picker makes this easy. On mobile, there is no equivalent host-side browser, so a user who is away from the computer cannot comfortably import an existing project folder even though the paired phone is already authorized to control that daemon.

This is especially painful for the core cross-device workflow: "ship from your phone" should include selecting a host project, not just working with projects that were already imported from the desktop.

## What would solve it

Add a host-backed directory browser to the mobile/remote Open project flow:

- Start from useful host roots such as HOME and previously/recommended project directories.
- Let the user tap folders to navigate deeper.
- Provide a clear way to select the current folder as the project, not only descend into children.
- Preserve the existing search/type-a-path behavior for users who know the path.
- Keep access bounded by the daemon's existing filesystem permissions and macOS/OS privacy rules.

The existing `directory_suggestions_request` shape already has most of the needed pieces: with `cwd` set and an empty query, it can list child directories for a host directory. The missing piece is mainly the mobile/remote picker UX around browsing and selecting folders.

## Alternatives you considered

- Typing absolute paths manually on the phone. This is error-prone and unrealistic for many nested projects.
- Importing every project from the desktop first. That defeats the remote/mobile setup flow.
- Relying only on fuzzy search. Search helps when the user remembers a folder name, but it does not replace browsing when they only know the approximate location.

xw949625012:
## Summary
- Add a host-folder browsing entry to the project picker so mobile/remote users can navigate from the host home directory.
- Reuse the existing directory suggestions RPC with cwd + empty query for child folders; no protocol or daemon changes.
- Add browse option helpers for current folder, parent folder, child folder navigation, and path dedupe across relative/absolute suggestion forms.
- Localize the new project picker labels.

Closes #1508

## Test plan
- npm run test --workspace=@getpaseo/app -- src/components/project-picker-options.test.ts src/i18n/resources.test.ts --bail=1
- npm run lint
- npm run typecheck
- npm run format

### DIFF
--- a/packages/app/src/components/project-picker-modal.tsx
+++ b/packages/app/src/components/project-picker-modal.tsx
@@ -8,7 +8,7 @@ import {
   View,
   type PressableStateCallbackType,
 } from "react-native";
-import { Folder, FolderPlus } from "lucide-react-native";
+import { Check, Folder, FolderOpen, FolderPlus, Home, Undo2 } from "lucide-react-native";
 import { StyleSheet, useUnistyles } from "react-native-unistyles";
 import { useMutation, useQuery } from "@tanstack/react-query";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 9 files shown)


boudra:
Closing in favor of the project picker implementation in [#1968](https://github.com/getpaseo/paseo/pull/1968).

PR #1605, opened 2026-06-19
fix(app): enable Save when pasting into the append-system-prompt field (#1602)

### LINKED ISSUE #1602: bug: Save button stays disabled when pasting into the append-system-prompt field
## Bug

In the daemon settings UI, the **append system prompt** field has a **Save** button. When you **paste** text into the field, the Save button stays **disabled** — it should become enabled because the field changed. Typing works fine; only paste fails to enable Save.

## Root cause (confirmed by code inspection)

The dirty/enable logic itself is correct:

`packages/app/src/screens/settings/host-page.tsx`
```ts
const persistedPrompt = config?.appendSystemPrompt ?? "";   // ~L814
const [draft, setDraft] = useState(persistedPrompt);        // ~L815
const hasChanges = draft !== persistedPrompt;               // ~L827
// ...
disabled={!hasChanges || isSaving}                          // ~L911
value={draft}
onChangeText={setDraft}                                     // ~L893-894
```

The defect is at the input layer. `SettingsTextAreaCard` (`packages/app/src/components/settings-textarea.tsx:28-37`) is a plain React Native `multiline` `TextInput` wired **only** to `onChangeText` — no `onChange`, no DOM `onPaste`, and no `.web.tsx` variant:

```tsx
<TextInput
  multiline
  value={value}
  onChangeText={onChangeText}
  ...
/>
```

On web, `multiline` renders a `<textarea>`. When a paste does not flow through RN-Web's `onChangeText` synthetic-event path, `draft` is never updated, so `hasChanges` stays `false` and Save stays disabled. There is no paste-safe fallback path, which structurally explains the symptom. There is no trimming/comparison issue — the dirty computation is sound.

## Suggested fix

Fix in the shared component `settings-textarea.tsx` (not `host-page.tsx`), e.g. add a web `onChange`/`onPaste` handler that pushes the textarea's value through `onChangeText`. The same component is reused in `project-settings-screen.tsx`, so a fix there covers both call sites.

## Notes

Confirmed via code inspection; not yet reproduced at runtime to pin whether the dropped event is RN-Web `onChangeText` behavior vs. the `AdaptiveModalSheet` portal intercepting the synthetic event. Either way the component has no paste-safe path.

nljms:
Fixes #1602

## Problem

In daemon settings, pasting into the **append system prompt** field left the **Save** button disabled. Typing worked; only paste failed.

## Root cause

The issue's original analysis (RN-Web `onChangeText` dropping paste) does not hold — I verified in a real browser that paste flows through `onChangeText` fine on plain Chromium. The actual failure is the **portal + Electron paste** combination:

- This field renders inside `AdaptiveModalSheet`, which `createPortal`s its content into `overlay-root` — a DOM node **outside** the React root container where RN-Web's delegated `onChange` listener lives.
- Trusted keystrokes still reach React through its portal handling (so typing updates the draft), but Electron's menu-driven `Cmd+V` paste produces a value change React's delegation misses. The DOM textarea value updates (the pasted text appears) yet `onChangeText` never fires, so `draft` stays equal to the persisted value, `hasChanges` stays `false`, and Save stays disabled.
- The message composer doesn't hit this because it lives inside the root container and also attaches its own direct DOM listener.

Confirmed against the reporter's environment: on the Mac desktop app, pasting shows the text but Save stays disabled; paste works in other fields (e.g. the composer).

## Fix

`SettingsTextArea` now attaches a **web-only direct DOM `input` listener** on the underlying textarea that pushes `node.value` through `onChangeText`, so the controlled value tracks the field regardless of React's delegation. It's guarded against IME composition so it doesn't disrupt CJK input. The change is gated to web (`isWeb`); native is untouched (paste fires `onChangeText` normally there). It's in the shared component, so both call sites (host settings + project settings) are covered.

## Testing

- `settings-textarea.browser.test.tsx` (real Chromium, Vitest browser mode): 4 tests — single-line paste, multi-line paste, the portal/Electron-class regression (**RED before the fix, GREEN after**), and IME-composition guard.
- `e2e/append-system-prompt-paste.spec.ts`: full real-app Playwright flow — open the sheet, real clipboard paste, assert Save enables.
- `npm run typecheck`, `npm run lint`, `npm run format` — all clean.

Supporting test-infra changes: `vitest.config.ts` gains `esbuild.jsx: "automatic"` so component browser tests can render, and `.gitignore` ignores Vitest's `.vitest-attachments/` artifacts.

## QA / platform evidence

- **Web (Chromium):** verified green via the browser tests and the Playwright E2E above.
- **Desktop (Electron, macOS):** the bug environment — desktop screen recording of the fix to be attached before marking ready.
- **Native iOS/Android:** not separately exercised; the change is web-gated and cannot affect native paste behavior.

Opening as a draft pending the desktop recording.


nljms:
closing this since its already fixed after the updates

PR #1637, opened 2026-06-20
Fix Paseo repo worktrees without global cross-env

### LINKED ISSUE #1577: Worktree setup fails with "cross-env: command not found" (exit 127) on packaged daemons
## Problem

Creating a worktree for any repo whose `paseo.json` invokes a bare devDependency CLI (e.g. `cross-env`) fails immediately on a packaged daemon — the setup command exits with `cross-env: command not found` (exit 127).

This is a dogfood bug: Paseo's own `paseo.json` calls `cross-env` in `worktree.setup` and in four service `scripts`. Creating a worktree for the Paseo repo using the **official desktop build** (or a global npm install / `dist` daemon) hits it every time.

## Root cause

The daemon spawns `paseo.json` commands directly, with a PATH that has no `node_modules/.bin`. Bare devDependency CLIs therefore don't resolve. The only reason it isn't constantly visible: the upstream dev workflow starts the daemon via `npm run dev`, and npm injects `node_modules/.bin` into PATH — masking the problem. Every non-npm-launched daemon (desktop app, global install, `dist`) is affected.

Setup reads the **worktree's own checked-out `paseo.json`**, so it can't be worked around by editing config — it has to be fixed daemon-side.

## Repro

1. On a daemon **not** started from an npm script (desktop / global install / `node dist`),
2. create a worktree for a repo whose `worktree.setup` runs a bare devDependency CLI such as `cross-env ...`.
3. Setup fails: `cross-env: command not found`, exit 127.

Mechanism (PATH is the only variable):

```sh
# no node_modules/.bin on PATH -> 127
env -i PATH="$(dirname $(command -v node)):/usr/bin:/bin" \
  cross-env FOO=bar node -e 'console.log(process.env.FOO)'   # cross-env: command not found

# prepend <cwd>/node_modules/.bin -> works
env -i PATH="$PWD/node_modules/.bin:$(dirname $(command -v node)):/usr/bin:/bin" \
  cross-env FOO=bar node -e 'console.log(process.env.FOO)'    # -> bar
```

## Expected

Repo-configured commands should resolve locally-installed CLIs the way `npm run` does.

## Prior art

#737 already does this exact `node_modules/.bin` → PATH prepend for the CLI test runner. This is the same fix for the daemon's runtime command path. I have a PR ready.

cleiter:
### Linked issue

Closes #1577

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

When running Paseo on the Paseo repo itself, the worktree setup step and the `paseo.json` service scripts (`daemon`/`app`/`desktop`/`ios-simulator`) fail with `cross-env: command not found` unless cross-env is installed globally.

The daemon runs `paseo.json` commands in a plain shell that doesn't have the project's `node_modules/.bin` on `PATH` the way `npm run` does, so the bare `cross-env` (a local devDependency) doesn't resolve. It only works when the daemon runs from the source checkout, which is why it hasn't come up.

This drops `cross-env` from those commands and sets the env vars inline. cross-env only exists for Windows `cmd` compatibility, and every one of these commands launches a unix `./scripts/*.sh`, so Windows support is already gone here — cross-env wasn't doing anything. The dependency stays: it's still used/needed by the `npm run` scripts in `package.json` / `packages/app` / `packages/server` (run via npm on Windows, invoke expo/tsx/node). Scoped to `paseo.json` only — removed `cross-env ` from 5 commands (setup + 4 services).

### How did you verify it

macOS, with cross-env NOT installed globally. Created a fresh Paseo workspace and confirmed the previously-failing commands now run without cross-env — the env-setting step in worktree setup and the daemon/app service scripts launch fine. Before this change they failed with `cross-env: command not found`.

After rebasing onto current `main`: `npm run typecheck`, `npm run lint`, and `npm run format` pass.

### Checklist

- [x] `npm run lint` passes
- [ ] UI changes include screenshots or video for every affected platform — n/a, no UI changes
- [ ] Tests added or updated where it made sense — n/a, config-only change

### DIFF
--- a/paseo.json
+++ b/paseo.json
@@ -3,7 +3,7 @@
     "setup": [
       "npm ci",
       "node ./scripts/seed-ios-native-cache.mjs",
-      "cross-env PASEO_DEV_MANAGED_HOME=1 PASEO_DEV_SEED_HOME=\"$PASEO_SOURCE_CHECKOUT_PATH/.dev/paseo-home\" PASEO_HOME=\"$PWD/.dev/paseo-home\" ./scripts/dev-home.sh",
+      "PASEO_DEV_MANAGED_HOME=1 PASEO_DEV_SEED_HOME=\"$PASEO_SOURCE_CHECKOUT_PATH/.dev/paseo-home\" PASEO_HOME=\"$PWD/.dev/paseo-home\" ./scripts/dev-home.sh",
       "npm run build:server",
       "npm run build --workspace=@getpaseo/expo-two-way-audio",
       "cp \"$PASEO_SOURCE_CHECKOUT_PATH/packages/server/.env\" \"$PWD/packages/server/.env\""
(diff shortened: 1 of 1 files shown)


boudra:
@paseo-bot rebase, baby sit to green and deliver

PR #1667, opened 2026-06-22
fix(server): no-op agent hooks when PASEO_TERMINAL_ID is unset

### LINKED ISSUE #1666: bug: Claude hooks fail outside Paseo terminals when PASEO_TERMINAL_ID is unset
## Summary

The shell command generated by `buildAgentHookShellCommand` in
[`packages/server/src/terminal/agent-hooks/agent-hook-installer.ts`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/agent-hooks/agent-hook-installer.ts#L136-L142)
short-circuits on `[ -n "$PASEO_TERMINAL_ID" ] && ...` and returns exit 1
when the env var is unset (Claude Code running outside a Paseo-managed
terminal).

Claude Code interprets the non-zero exit as a hook error and prints on
every configured event:

    UserPromptSubmit hook error
    Failed with non-blocking status code: No stderr output

It fires on every turn for the 5 Claude events declared in
[`agent-hooks/claude/claude.ts`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/agent-hooks/claude/claude.ts):
`UserPromptSubmit`, `Stop`, `StopFailure`, `SessionEnd`, `Notification`.
Same shape on the Codex provider via `buildAgentHookWindowsCommand` / POSIX.

## Reproduce

```bash
unset PASEO_TERMINAL_ID
bash -c '[ -n "$PASEO_TERMINAL_ID" ] && "${PASEO_HOOK_CLI:-paseo}" hooks claude UserPromptSubmit'
echo "exit=$?"
# → exit=1
```

The `[ -n "" ]` test fails and short-circuits the `&&`, so no CLI runs
but the conditional itself returns failure.

The existing tests pass because they only assert the literal command
string ([`claude.test.ts:144`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/agent-hooks/claude/claude.test.ts#L144),
[`codex.test.ts:69`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/agent-hooks/codex/codex.test.ts#L69))
without ever executing it.

## Expected

When `PASEO_TERMINAL_ID` is absent the hook should be a true no-op
(exit 0), and when it is present the underlying CLI's exit code should
propagate unchanged. The OpenCode provider already follows this shape
with `if (!process.env.PASEO_TERMINAL_ID) return;`
([`opencode-plugin.ts:18`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/agent-hooks/opencode/opencode-plugin.ts#L18)).

## Environment

- `@getpaseo/cli@0.1.98` (`@getpaseo/server@0.1.98` bundled)
- macOS 15, bash 5
- Claude Code (Opus 4.7)

## Notes

- `enableTerminalAgentHooks` defaults to `false` so this only affects
  users who opted in. For those users the error appears on every turn.
- Happy to send a PR that uses `if [ -n ... ]; then ...; fi` (POSIX) and
  adds an `else (exit /b 0)` branch on the Windows variant, plus a
  spawnSync-based regression that runs the generated command and asserts
  exit 0 across all Claude events.

everton-dgn:
Fixes #1666

## Problem

`buildAgentHookShellCommand` produced
`[ -n "$PASEO_TERMINAL_ID" ] && "\${PASEO_HOOK_CLI:-paseo}" hooks <agent> <event>`,
which exits 1 when the env var is unset because the `[ -n "" ]` test fails
and short-circuits the `&&`. Claude Code reads the non-zero exit as a hook
error and prints `Failed with non-blocking status code: No stderr output`
on every configured event (and the Codex provider was affected the same way).

## Fix

Replace the `&&` gate with an explicit `if [ -n ... ]; then ...; fi` so the
hook is a true no-op when there is no Paseo terminal, and the underlying
CLI's exit code propagates unchanged when it actually runs. Same shape as
`opencode-plugin.ts` (`if (!process.env.PASEO_TERMINAL_ID) return;`). The
Windows variant gets the matching `else (exit /b 0)` branch.

## Test plan

- [x] `npm run typecheck` (workspaces, all green)
- [x] `npm run lint` — 0 warnings, 0 errors across 2280 files
- [x] `npx oxfmt --check` on the three changed files — clean
- [x] `npm run test:unit --workspace=@getpaseo/server -- src/terminal/agent-hooks/` — **35/35** passing, including the new regression
- [x] Manual repro: `unset PASEO_TERMINAL_ID; bash -c '<old command>'` returns exit 1; the new command returns exit 0. With `PASEO_TERMINAL_ID=t PASEO_HOOK_CLI=false`, the new command still propagates exit 1 (the underlying CLI's failure is not masked).
- [x] Local install of the patched dist into `~/.local/share/.../@getpaseo/cli` confirmed Claude Code no longer reports the hook error across `UserPromptSubmit`, `Stop`, `StopFailure`, `SessionEnd`, `Notification`.

## Changes

- `agent-hook-installer.ts` — POSIX gate switched to `if … then … fi`; Windows variant gets `else (exit /b 0)`.
- `claude/claude.test.ts` — updates the literal-string assertions, adds an executable regression that runs the generated command through `/bin/sh` for all five Claude events and asserts exit 0.
- `codex/codex.test.ts` — updates the literal-string assertions for both POSIX and Windows commands.


everton-dgn:
@boudra when you have a moment, could you prioritize this one? It fixes a regression where every configured Claude Code and Codex hook event reports a spurious "Failed with non-blocking status code" error when `PASEO_TERMINAL_ID` is unset (the default outside a Paseo terminal). CI is green across all 15 checks and Greptile signed off 5/5 on the latest commit.

ru3tyYT (approved):

PR #1677, opened 2026-06-23
fix(server): increase Pi isAvailable() timeout for Windows cold starts

eunho:
## Problem

On Windows 11, the Paseo daemon always reports Pi as `unavailable` even though `pi` is installed and on PATH.

## Root Cause

Pi's `isAvailable()` method in `PiRpcAgentClient` wraps the entire availability check chain in `withTimeout(..., 2000)` (2 seconds):

1. Binary probe (`pi --version`): ~859ms
2. Start RPC session (`pi --mode rpc`): ~3.2s on cold start
3. Model fetch
4. Cleanup

The 2-second timeout is too tight for Windows, where `pi --mode rpc` cold start alone takes ~3.2 seconds.

## Fix

Increased the timeout from `2000` to `15_000` (15 seconds), which is well within the daemon's outer 30-second provider refresh timeout (`PASEO_PROVIDER_REFRESH_TIMEOUT_MS`).

## Verification

- Built and tested on Windows 11 + PowerShell
- `paseo daemon status` now shows Pi as `available` instead of `unavailable`
- `paseo provider models pi` successfully lists all available models

## Notes

- Only Pi has this pattern (binary check + session start + model fetch in its `isAvailable()`). All other providers (Claude, Codex, OpenCode, Copilot, etc.) do simple binary resolution without an inner timeout, so they are unaffected.
- The outer `ProviderSnapshotManager` timeout is already configurable via `PASEO_PROVIDER_REFRESH_TIMEOUT_MS`.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -163,6 +163,10 @@ jobs:
           build_args=(-- --publish never --mac --${{ matrix.electron_arch }})
           build_args+=("-c.publish.releaseType=$RELEASE_TYPE")
           build_args+=("-c.publish.channel=$RELEASE_CHANNEL")
+          if [[ "$SHOULD_PUBLISH" != "true" ]]; then
+            build_args+=("-c.mac.hardenedRuntime=false")
+            build_args+=("-c.mac.notarize=false")
+          fi
           npm run build:desktop "${build_args[@]}"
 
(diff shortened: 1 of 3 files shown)


boudra:
cosed in favour iof https://github.com/getpaseo/paseo/pull/1660

PR #1687, opened 2026-06-23
feat(server): add option to disable context usage probing

### LINKED ISSUE #1685: bug: Quota tracking calls getContextUsage after every message, doubling API requests
## What's broken

The quota tracking feature calls `getContextUsage()` after every successful message completion, which sends a probe API request (with max_tokens=1) to retrieve current context usage.

This doubles the API request count during active conversations:
- Each normal message: 1 API request
- Each quota check: 1 additional probe request (max_tokens=1)

In practice, if a user sends 10 messages, this results in:
- 10 normal API requests
- 10 quota tracking probe requests
- Total: 20 API requests

This not only increases API call count and latency, but may also trigger rate limits during high-frequency usage.

## Steps to reproduce

1. Start Paseo daemon
2. Connect a Claude Code agent
3. Send multiple messages (e.g., 5-10 messages)
4. Observe network requests or daemon.log
5. You'll see an additional `getContextUsage()` API request after each successful message

**Code location:**
`packages/server/src/server/agent/providers/claude/agent.ts:3459-3462`

```typescript
const currentContextUsage =
  message.type === "result" && message.subtype === "success"
    ? await this.queryCurrentContextUsage(activeQuery)
    : undefined;
```

`packages/server/src/server/agent/providers/claude/agent.ts:3486-3496`

```typescript
private async queryCurrentContextUsage(
  activeQuery: Query,
): Promise<ClaudeCurrentContextUsage | undefined> {
  try {
    const usage = await withTimeout(activeQuery.getContextUsage(), 3_000, "timeout");
    return readCurrentContextUsage(usage);
  } catch (error) {
    this.logger.debug({ err: error }, "Claude context usage query failed");
    return undefined;
  }
}
```

## Where did this happen

Daemon

## Paseo version

0.1.98

## OS version

Cross-platform issue (Linux / macOS / Windows)

## Agent provider

Claude Code

## Provider configuration

Claude Code with Anthropic API (affects all Claude API configurations)

## Logs

Relevant log entries (probe requests after each message):

```
[debug] Claude context usage query started
[debug] Claude context usage query completed
```

Check daemon.log for frequent `getContextUsage()` calls.

## Suggested solutions

**Option 1: Use existing usage data from responses**

Claude API responses already include usage information. Use that directly instead of making additional requests:

```typescript
// Read from message.usage directly instead of calling getContextUsage()
const currentContextUsage =
  message.type === "result" && message.subtype === "success" && message.usage
    ? readCurrentContextUsage(message.usage)
    : undefined;
```

**Option 2: Batch or defer queries**

If `getContextUsage()` is necessary:
- Don't query after every message
- Query every N messages or every T seconds instead
- Or only query when user actively views quota statistics

**Option 3: Cache**

Cache quota information and avoid re-querying within a short time window (e.g., 10 seconds).

---

**Impact:**

This affects all users with Claude API, especially:
- High-frequency conversation users
- Users with lower API rate limits
- Cost-conscious users

Recommended for priority fix as it significantly increases API call count and cost.

chyendongnhanh338:
### Linked issue

Closes #1685

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Adds `PASEO_DISABLE_CONTEXT_USAGE_PROBE` environment variable to allow disabling the `getContextUsage()` API probe calls introduced in v0.1.98.

**Background:** In v0.1.98 (commit `a924059da`), the Claude provider started calling `getContextUsage()` after every successful message to get precise context usage statistics. While this provides accurate data, it doubles the API request count - each message now triggers:
1. The actual message request
2. A probe request with `max_tokens=1` to query context usage

**This PR:** Adds an opt-out via environment variable. When `PASEO_DISABLE_CONTEXT_USAGE_PROBE=true`, the daemon skips the probe call and falls back to v0.1.97 behavior: calculating context usage from stream events and result usage data.

**Benefits when disabled:**
- 50% fewer API requests (no probe after each message)
- Reduced latency (no 3-second timeout for context query)
- Lower risk of hitting rate limits
- Lower API costs

**Context usage remains accurate** - the fallback mechanism (`buildResultUsage()`) uses:
1. Stream events (`message_start`, `message_delta`) for incremental token counts
2. Result usage data from the final API response

This is the same mechanism used successfully in v0.1.97.

### How did you verify it

**1. Code review:**
- Verified the fallback logic in `buildResultUsage()` (lines 1826-1827) properly handles `undefined` currentContextUsage
- Confirmed environment variable is checked before calling `queryCurrentContextUsage()`
- Reviewed the three-tier fallback: `currentContextUsage?.totalTokens ?? this.streamUsedTokens() ?? activeResultUsageTokens`

**2. Pre-commit hooks passed:**
- ✅ `npm run lint` - 0 warnings, 0 errors
- ✅ `npm run format:check` - all files correctly formatted  
- ✅ `npm run typecheck` - all workspaces pass (23s)

**3. Tested behavior (manual):**
```bash
# Without the env var (default behavior - probe enabled)
npm run cli -- daemon restart
# Observed: getContextUsage() called after each message (doubles API requests)

# With the env var (probe disabled)
export PASEO_DISABLE_CONTEXT_USAGE_PROBE=true
npm run cli -- daemon restart
# Observed: no additional probe requests, context usage still reported correctly
```

**4. Verified backward compatibility:**
- When env var is not set or set to anything other than `"true"`, behavior is unchanged (probe enabled)
- Existing deployments unaffected

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A - server-side only)
- [x] Tests added or updated where it made sense (existing test mocks already handle `getContextUsage` returning undefined)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,6 +46,7 @@ At the start of non-trivial work, list `docs/` and skim anything relevant to the
 | [docs/android.md](docs/android.md)                             | App variants, local/cloud builds, EAS workflows                                                                                |
 | [docs/release.md](docs/release.md)                             | Release playbook, draft releases, completion checklist                                                                         |
 | [docs/terminal-activity.md](docs/terminal-activity.md)         | Terminal activity indicators — source-agnostic tracker, agent hook reporting, adding a new hook provider                       |
+| [docs/context-usage-probing.md](docs/context-usage-probing.md) | Context usage probing — how it works, disabling to reduce API calls, fallback behavior                                         |
 | [SECURITY.md](SECURITY.md)                                     | Relay threat model, E2E encryption, DNS rebinding, agent auth                                                                  |
 
 ## Quick start
--- a/docs/context-usage-probing.md
+++ b/docs/context-usage-probing.md
@@ -1,0 +1,93 @@
+# Context Usage Probing
+
+## Overview
+
+Starting in v0.1.98, Paseo's Claude provider queries the current context window usage after each successful message completion by calling `getContextUsage()`. This provides precise context usage statistics but doubles the API request count.
+
+**Prior to v0.1.98**, context usage was calculated from stream events and result usage data without additional API calls. This configuration allows you to restore that behavior.
+
+## How it works
+
+After each successful message, the daemon calls `getContextUsage()` on the Claude SDK, which sends a probe API request (with `max_tokens=1`) to retrieve:
+
+- `totalTokens`: Current context window usage
+- `maxTokens`: Maximum context window size for the current model
+
+This information powers the context usage meter in the Paseo UI.
+
+## Disabling context usage probing
+
+If you want to reduce API request count and avoid the extra probe requests, you can disable this feature using an environment variable.
+
+### Option 1: Environment variable (recommended)
+
+Set `PASEO_DISABLE_CONTEXT_USAGE_PROBE=true` before starting the daemon:
+
+```bash
+export PASEO_DISABLE_CONTEXT_USAGE_PROBE=true
+paseo daemon restart
+```
+
+Or add it to your shell profile (`~/.bashrc`, `~/.zshrc`, etc.):
+
+```bash
+echo 'export PASEO_DISABLE_CONTEXT_USAGE_PROBE=true' >> ~/.bashrc
+source ~/.bashrc
+```
+
+### Option 2: Per-provider configuration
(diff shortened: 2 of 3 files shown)


boudra:
Superseded by #1701, which removes the per-turn Claude context probe instead of adding a toggle.

PR #1691, opened 2026-06-23
(feat): Default Windows terminals to PowerShell instead of cmd.exe 

Justmalhar:
## What

New terminals on Windows now launch **Windows PowerShell** instead of `cmd.exe`.

## Why

`resolveDefaultTerminalShell` (the single source of truth for a new terminal's shell — `terminal-manager.ts` never passes an explicit `shell`) read `env.ComSpec` on `win32`. Windows always points `ComSpec` at `…\System32\cmd.exe`, so every new terminal opened cmd.

## How

The `win32` branch now returns `…\System32\WindowsPowerShell\v1.0\powershell.exe`, built from `%SystemRoot%` (honoring a relocated Windows install) with a `C:\Windows` fallback. Windows PowerShell 5.1 ships in System32 on every supported Windows, so the path is always valid and the resolver stays pure/synchronous (no PATH probing).

Untouched and verified safe:
- `resolveTerminalSpawnCommand` (routes agent `.cmd`/`.bat` profiles like `claude` through `cmd.exe /c`) — unchanged.
- Shell integration only ever existed for zsh, so there's no Windows integration to lose.

## Tests

- Rewrote the contract unit test to expect PowerShell and assert `ComSpec` no longer pulls it back to cmd.
- Added a test for a relocated `%SystemRoot%`.

> Note: this is a hardcoded switch with no opt-out. If a per-user shell knob is wanted later, the clean spot is a `PASEO_TERMINAL_SHELL` check at the top of the same `win32` branch.


xdewx:
非常影响使用体验，在终端这方面不如orca

gstamp:
I'm not a maintainer but I have one concern with the proposed default path.

This switches the default from cmd.exe to Windows PowerShell 5.1, but it will ignore users who have PowerShell 7 or newer installed. PowerShell 7 uses pwsh.exe, while powershell.exe is the legacy Windows PowerShell executable.

For example, on my machine:

    pwsh.exe        7.6.3
    powershell.exe  5.1

Would it be possible to look for pwsh.exe on PATH first and use the resolved path when it is available? If it is not found, the resolver could fall back to the Windows PowerShell 5.1 path proposed here.

That would support modern PowerShell installations, including per-user installs, while preserving compatibility on machines that only have Windows PowerShell. If the goal is to select the absolute newest installed version, Paseo could enumerate candidates and query each with -NoProfile and -NonInteractive, but PATH-first seems more predictable because it respects the user's existing shell preference.

boudra:

Issue #1231, opened 2026-05-29
feat:


zhugc9:
### Prior search

- [x] I searched existing issues and discussions, and this isn't already proposed.

### What's the problem

Hello!!

## Problem 1: MCP tools are all-or-nothing

Currently `daemon.mcp.injectIntoAgents` is a single Boolean toggle. 
All 32 tools are injected when enabled. In most cases only 3-5 tools 
(create_agent, send_agent_prompt, list_agents, etc.) are needed. 
Injecting all 32 costs ~6000-7000 tokens in every agent's context window.

Suggestion: allow per-tool or per-category (agents / terminals / 
schedules / providers / worktrees) granularity, similar to how Claude 
Code handles MCP server configs.

## Problem 2: Agent defaults to Paseo MCP for subagent calls

When an agent needs to spawn a subagent, it currently has no way to 
distinguish "one-off quick task" from "long-running reusable agent." 
It often defaults to `create_agent` via Paseo MCP, producing 
persistent agents that accumulate and bloat the agent list.

Suggestion: 
- Agents should default to their native subagent mechanism 
  (e.g. Codex task tool) for one-off tasks.
- Paseo `create_agent` should only be used when the user or skill 
  explicitly requests a long-running, resumable agent.
- Consider a lightweight "wake up by prompt" mechanism so long-running 
  Paseo agents can be resumed naturally.

### What would solve it

-

### Alternatives you considered

I just have the agent distinguish between native sbuagents calls and PASEO calls.

PR #1707, opened 2026-06-24
Prepare providers for direct Paseo tools

boudra:
### Linked issue

Refs #1231

### Type of change

- [ ] Bug fix
- [x] Docs

### What does this PR do

Paseo's agent control tools now live in a shared runtime catalog instead of being defined inside the MCP server. The MCP endpoint stays as a thin adapter, so existing agent integrations continue to receive the same tools through MCP, while providers that opt in later can receive the catalog directly without duplicate tool injection.

The daemon now builds the catalog from shared host dependencies, passes it through launch context only for clients that advertise direct tool support, and strips the internal MCP fallback from those launches. It also removes an unused per-session self-MCP client path.

### How did you verify it

- `npm run format`
- `npm run format:check`
- `npm run typecheck`
- `npm run lint`
- `npx vitest run packages/server/src/server/agent/mcp-server.test.ts --bail=1`
- `npx vitest run packages/server/src/server/agent/agent-manager.test.ts --bail=1`
- `npx vitest run packages/server/src/server/agent/runtime-mcp-config.test.ts --bail=1`
- `npx vitest run packages/server/src/server/agent/agent-mcp.e2e.test.ts --bail=1`

All passed locally. The pre-commit hook also reran file-scoped lint, file-scoped format check, and full typecheck.

### Risk surface

The main risk is provider launch config routing: integrations that only support MCP should still receive the fallback, while direct-tool integrations should not receive duplicate internal tool injection. Tests cover existing MCP injection, disabled injection, and the direct-tool capability path. There is no UI or platform-specific surface in this change.

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A, no UI changes)


boudra:
Follow-up on the stale protocol COMPAT review: addressed in f0af46ad6 by removing the field from the shared protocol entirely. Native Paseo tool support is server-internal only now.

boudra:
Follow-up on Greptile's create_schedule error-message note: addressed in 2c30ab4ed by making the schema allow omission at the transport boundary and preserving the explicit handler error: provider is required when target is new-agent.

PR #1718, opened 2026-06-25
Link worktrees to PRs from differently named tracked branches

### LINKED ISSUE #144: Prefer upstream branch for GitHub PR lookup
## Problem

When a local worktree branch tracks a different remote branch name, plain `gh pr view` fails to find the PR even though the checkout is clearly attached to that PR branch.

Example from this repo:

- local branch: `refactor/manager-launch-context`
- upstream: `origin/codex/expose-paseo-agent-id`
- actual PR head branch: `codex/expose-paseo-agent-id`

Observed behavior:

- `gh pr view` fails with `no pull requests found for branch "refactor/manager-launch-context"`
- `gh pr view 143` works
- `gh pr view codex/expose-paseo-agent-id` works

## Why this matters

For worktrees and local alias branches, the local branch name is not always the best source of truth for PR association. The tracked upstream branch is usually a better signal.

## Proposed behavior

When we need to resolve “current PR”, prefer:

1. upstream tracking branch, if configured
2. otherwise local branch name

So the lookup order should be:

- `git rev-parse --abbrev-ref --symbolic-full-name @{upstream}`
- fallback to `git branch --show-current`

Then resolve the PR from that branch name.

## Node implementation sketch

```ts
import { execFile } from "node:child_process";
import { promisify } from "node:util";

const execFileAsync = promisify(execFile);

async function readGitRef(args: string[], cwd: string): Promise<string | null> {
  try {
    const { stdout } = await execFileAsync("git", args, { cwd });
    const value = stdout.trim();
    return value.length > 0 ? value : null;
  } catch {
    return null;
  }
}

async function resolvePreferredPrBranch(cwd: string): Promise<string | null> {
  const upstream = await readGitRef(
    ["rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{upstream}"],
    cwd,
  );
  if (upstream) {
    const slashIndex = upstream.indexOf("/");
    return slashIndex >= 0 ? upstream.slice(slashIndex + 1) : upstream;
  }

  return await readGitRef(["branch", "--show-current"], cwd);
}
```

## Scope

This issue is about Paseo’s own PR lookup behavior when shelling out to GitHub CLI. It is not proposing a workaround in bash; the fix should live in Node.

boudra:
### Linked issue

Closes #144

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Paseo now links a worktree to its PR when the local branch tracks a differently named remote branch. The broken case was a local branch such as `tender-parrot` tracking `origin/refactor/workspace-scripts`: PR derivation handed the GitHub executor `tender-parrot`, and the executor correctly discarded the real PR because its head branch was `refactor/workspace-scripts`.

This widens the existing tracking-aware derivation rule instead of adding a second lookup path. The on-demand path now routes through the same builder as snapshot facts; the builder uses the tracked head ref only when it differs from the local branch, reads the tracked remote URL for any tracking remote, and attaches `headRepositoryOwner` only when the tracked remote is cross-repo or cannot be proven same-repo. Same-repo origin tracking supplies only the corrected `headRef`.

No-tracking pushes are intentionally not covered. If `git push` was run without `-u`, Git does not write `branch.*` tracking config, so there is no reliable signal for this path to read.

The base-branch risk is guarded: same-repo tracking to the resolved base branch, such as accidentally tracking `main`, keeps the local branch lookup. Scoped fork lookups like `owner/main` still keep their owner and continue to resolve.

### How did you verify it

- `npm run format`
- `npm run typecheck`
- `npm run lint`
- `npx vitest run packages/server/src/utils/checkout-git.test.ts --bail=1` — 111 passed
- `npx vitest run packages/server/src/services/github-service.test.ts --bail=1` — 75 passed
- Pre-commit hook reran lint, format check, and typecheck successfully

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A: daemon-only)

### DIFF
--- a/packages/server/src/utils/checkout-git.test.ts
+++ b/packages/server/src/utils/checkout-git.test.ts
@@ -149,6 +149,44 @@ function createPullRequestStatus(overrides?: Partial<GitHubCurrentPullRequestSta
   };
 }
 
+interface RequestedPullRequestTarget {
+  headRef: string;
+  headRepositoryOwner?: string;
+}
+
+interface RecordingPullRequestTargetsOptions {
+  requestedTargets: RequestedPullRequestTarget[];
+  statusOverrides?: Partial<GitHubCurrentPullRequestStatus>;
+}
+
+function createGitHubServiceRecordingPullRequestTargets(
+  options: RecordingPullRequestTargetsOptions,
+): GitHubService {
+  const github = createGitHubServiceForStatus(null);
+  github.getCurrentPullRequestStatus = async (request) => {
+    options.requestedTargets.push({
+      headRef: request.headRef,
+      ...(request.headRepositoryOwner ? { headRepositoryOwner: request.headRepositoryOwner } : {}),
+    });
+    return createPullRequestStatus({
+      ...options.statusOverrides,
+      headRefName: request.headRef,
+    });
+  };
+  return github;
+}
+
+async function readPullRequestLookupTargetFromFacts(
+  repoDir: string,
+  paseoHome: string,
+): Promise<RequestedPullRequestTarget | null> {
+  const facts = await getCheckoutSnapshotFacts(repoDir, { paseoHome });
+  if (!facts.isGit) {
+    throw new Error("Expected git checkout facts");
+  }
+  return facts.pullRequestLookupTarget;
+}
+
 function setupRemoteTrackingMain(
   repoDir: string,
   tempDir: string,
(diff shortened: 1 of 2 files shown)


boudra:
Addressed the automated review note in 4fad9012e: the on-demand path now asks the shared lookup-target builder for the local-branch fast path before reading remote.origin.url or resolving the base branch.

Issue #1475, opened 2026-06-11
bug: Increase timeout or make it configurable for opencode models / provider load time


pmist:
### What's broken

The Windows-based OpenCode instance can take 2–3 minutes to start up and become fully responsive. Despite the startup delay, it is an extremely useful tool for running OpenCode on Windows.

On a typical laptop, the setup works well without issues. However, when the instance is hosted on a remote VM, the longer startup time can cause connection timeouts, particularly for clients such as Paseo that expect a faster response during initialization.

<img width="1172" height="502" alt="Image" src="https://github.com/user-attachments/assets/b4069521-2fff-4d7c-9c0e-c8429153a482" />

### Steps to reproduce

1. Setup the Opencode and Paseo on windows server
2. Access it using paseo windows or mobile app with opencode provider
3. Not able to even initialize the Opencode properly ( opencode serve - without any credential works well on normal browser)

### Where did this happen

Desktop (Electron)

### Paseo version

v0.1.93

### OS version

Windows 11 ( Client), Windows Server (daemon)

### Agent provider

OpenCode

### Logs

```text
Daeamon Logs 

[2026-06-09 16:13:05.974] [warn]  [expo-notifications] Listening to push token changes is not yet fully supported on web. Adding a listener will have no effect.
[2026-06-09 16:13:08.682] [warn]  [Layout children]: No route named "h/[serverId]/workspace/[workspaceId]" exists in nested children: host-runtime-bootstrap,index,pair-scan,welcome,h/[serverId]/index,h/[serverId]/new,h/[serverId]/open-project,h/[serverId]/sessions,h/[serverId]/settings,h/[serverId]/agent/[agentId],h/[serverId]/workspace/[workspaceId]/index,settings/[section],settings/index,settings/hosts/[serverId]/[hostSection],settings/hosts/[serverId]/index,settings/projects/[projectKey],settings/projects/index
[2026-06-09 16:13:09.148] [info]  [open-project] renderer requested pending path: { webContentsId: 1, pendingPath: null }
[2026-06-09 16:14:23.107] [info]  [desktop daemon] initial status check before start {
  pid: null,
  status: 'stopped',
  listen: '127.0.0.1:6767',
  serverId: 'srv_Z_j5tEz2BPat',
  error: null,
  desktopManaged: false
}
[2026-06-09 16:14:23.132] [info]  [desktop daemon] starting detached daemon {
  pid: 3336,
  appIsPackaged: true,
[...]

PR #1723, opened 2026-06-25
Keep liveness pings responsive during slow requests

boudra:
### Linked issue

Refs #1475

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Long-running session requests could hold the WebSocket's inbound message path, delaying top-level liveness pongs and making connected clients look dead.

This removes the socket-wide request queue so the WebSocket server routes inbound frames immediately. Ordering-sensitive work stays scoped to the subsystem that needs it; file uploads keep their per-upload queue.

### How did you verify it

Reproduced the bug with a regression test: a provider diagnostic request is held open, then a top-level ping is sent on the same socket. Before the fix the socket produced no pong; after the fix it responds immediately.

Local verification:

- `npx vitest run packages/server/src/server/websocket-server.relay-reconnect.test.ts --bail=1`
- `npx vitest run packages/server/src/server/file-upload/index.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A, server-only)

### DIFF
--- a/packages/server/src/server/websocket-server.relay-reconnect.test.ts
+++ b/packages/server/src/server/websocket-server.relay-reconnect.test.ts
@@ -681,6 +681,49 @@ describe("relay external socket reconnect behavior", () => {
     await server.close();
   });
 
+  test("sends rpc_error when an async session request fails", async () => {
+    const server = createServer();
+    const socket = new MockSocket();
+    await attachRelayAndHello({
+      server,
+      socket,
+      clientId: "cid-session-request-failure",
+    });
+
+    const session = sessionMock.instances[0];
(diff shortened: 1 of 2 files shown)


boudra:
Addressed review follow-ups in dd7826ff7: removed the test-body try/finally, moved the held request resolver into a helper, and added a regression test proving a second session request reaches the session while the first request is still running.

boudra:
Addressed the async error-response review in 3d1fc5a6c: fire-and-forget session and binary-frame failures now go through the existing client error responder, preserving rpc_error/status responses without restoring socket-wide serialization. Added regressions for both paths.

PR #1736, opened 2026-06-26
fix(agent-hooks): stop POSIX hook guard from leaking exit 1

### LINKED ISSUE #1666: bug: Claude hooks fail outside Paseo terminals when PASEO_TERMINAL_ID is unset

ehsan:
### Linked issue

Closes #1666

### Type of change

- [X] Bug fix
- [ ] Docs

### What does this PR do

`buildAgentHookShellCommand` emitted the POSIX hook command as:

    [ -n "$PASEO_TERMINAL_ID" ] && "${PASEO_HOOK_CLI:-paseo}" hooks <agent> <event>

When `PASEO_TERMINAL_ID` is unset (i.e. the terminal isn't being driven by paseo), the `[ -n "" ]` test itself returns exit 1 and the `&&` short-circuits, so the whole hook command exits 1. Agents treat any non-zero hook exit as a failure: Claude Code surfaces it on every prompt as

    UserPromptSubmit hook error
    Failed with non-blocking status code: No stderr output

and Codex reports the same for its events. It's noise — the hook is meant to be a no-op when paseo isn't attached.

Emit a guard that no-ops cleanly instead:

    [ -z "$PASEO_TERMINAL_ID" ] || "${PASEO_HOOK_CLI:-paseo}" hooks <agent> <event>

When the var is unset the guard is true and short-circuits with exit 0; when it is set, the hook runs and its real exit code is preserved. Same active behavior, no spurious failure when paseo isn't driving the terminal.

Affects both the claude and codex providers (both call buildAgentHookShellCommand). The Windows variant uses `if defined ... (...)`, which leaves errorlevel intact when the var is undefined, so it is unchanged.

Updates the four test assertions that pin the exact generated command string.

### How did you verify it

Tested locally

### Checklist

- [X] One focused change. Unrelated cleanups split out.
- [X] `npm run typecheck` passes
- [X] `npm run lint` passes
- [X] `npm run format` ran (Biome)

### DIFF
--- a/packages/server/src/terminal/agent-hooks/agent-hook-installer.ts
+++ b/packages/server/src/terminal/agent-hooks/agent-hook-installer.ts
@@ -138,7 +138,7 @@ export function buildAgentHookShellCommand<TConfig>(
   event: AgentHookEventDefinition,
 ): string {
   const hookCommand = `"\${PASEO_HOOK_CLI:-paseo}" hooks ${shellToken(provider.id)} ${shellToken(event.event)}`;
-  return `[ -n "$PASEO_TERMINAL_ID" ] && ${hookCommand}`;
+  return `[ -z "$PASEO_TERMINAL_ID" ] || ${hookCommand}`;
 }
 
 export function buildAgentHookWindowsCommand<TConfig>(
--- a/packages/server/src/terminal/agent-hooks/claude/claude.test.ts
+++ b/packages/server/src/terminal/agent-hooks/claude/claude.test.ts
@@ -81,7 +81,7 @@ describe("Claude terminal agent hooks", () => {
       );
       expect(paseoCommands).toHaveLength(1);
       expect(paseoCommands[0]).toBe(
-        `[ -n "$PASEO_TERMINAL_ID" ] && "\${PASEO_HOOK_CLI:-paseo}" hooks ${provider.id} ${event.event}`,
+        `[ -z "$PASEO_TERMINAL_ID" ] || "\${PASEO_HOOK_CLI:-paseo}" hooks ${provider.id} ${event.event}`,
       );
     }
     expect(registeredAgentHooksAreInstalled({ configDir })).toBe(true);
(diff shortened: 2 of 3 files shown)


boudra:
fixed in https://github.com/getpaseo/paseo/pull/1667

PR #1743, opened 2026-06-26
Fix stale Ask Question permission responses

### LINKED ISSUE #1742: bug: Stale Ask Question cards keep loading after the request was answered elsewhere
### What's broken

When an Ask Question (`request_user_input`) card is shown in the desktop app and the user answers the same request through another route (for example, by typing into the terminal that the agent is running in, or by responding from the Codex/agent side rather than clicking in the Paseo Ask Question view), the Paseo Ask Question card stays on screen and is treated as still pending.

If the user then tries to clear it from the Paseo UI — Dismiss, or submit an answer in the card — the card flips into a loading state and never resolves. The pending card never goes away unless the agent is killed or the app is restarted.

### Steps to reproduce

1. Run a Codex agent in Paseo so it issues a `request_user_input` (Ask Question) — for example, ask it something that will trigger a clarifying question.
2. Instead of answering inside the Paseo Ask Question view, answer through a different route. For example, type the answer in the terminal that the same Codex session is reading from, or otherwise let the underlying agent receive the answer outside Paseo.
3. Wait for the agent to acknowledge the answer and continue.
4. Return to the Paseo desktop app and click Dismiss (or submit a value) in the still-visible Ask Question card.

### Expected

Once the underlying agent has resolved the permission/question request, the Paseo Ask Question card for it should clear automatically. Dismissing or submitting from the card after the fact should resolve immediately without entering a loading state.

### Actual

The card stays visible after the agent has moved on, and Dismiss / submit puts the card into a permanent loading state.

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.101

### OS version

macOS 26.3.2

### Agent provider

Codex

### Notes

The pattern in the daemon logs is `respondToPermission` failing on the provider side with `No pending permission request with id '<id>'` (because the underlying agent has already moved past it). The Paseo `AgentManager` then never dispatches `permission_resolved` for that request, so the desktop UI's `respondToPermissionAndWait` never sees a resolution and the card sticks in loading.

I have only investigated this against the current `main` of this repo and the daemon's `AgentManager` code path. I have not done an end-to-end manual reproduction in a fresh desktop build, so the repro above is described in terms of behavior I expect from reading the code path; if there is a more reliable way to reproduce, happy to update.

soichiyo:
### Linked issue

Closes #1742

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

`AgentManager.respondToPermission` assumed that if a pending Ask Question
permission existed in `pendingPermissions`, the underlying provider session
still had it too. When the request had already been resolved on the
provider side (e.g. the user answered the same request through the
terminal or directly via the agent), the provider would throw
`No pending permission request with id '<id>'`, the error would bubble out
of `respondToPermission`, and no `permission_resolved` stream event was
ever dispatched. The desktop Ask Question card then stayed in its loading
state forever.

This change makes `respondToPermission` idempotent for that case:

- If the manager itself has no `pendingPermissions` entry for the request
  anymore, treat it as a stale response, dispatch `permission_resolved`,
  and return without calling the provider.
- If the provider call throws with the `No pending permission request`
  shape, treat it the same way: clear the pending entry (if any),
  re-emit state, and dispatch `permission_resolved` so the UI can release
  the card.

No other code paths are changed.

### How did you verify it

- `./node_modules/.bin/vitest run packages/server/src/server/agent/agent-manager.test.ts` — 117/117 pass, including a new regression test (`respondToPermission resolves stale permission cards when provider request is already gone`) that fails on `main` and passes with this change.
- `npm run typecheck` — passes (via the lefthook `pre-commit` hook).
- `npm run lint` and `npm run format:check:files` on the two changed files — clean.
- `git diff --check` — clean.

I have **not** independently reproduced the original symptom end-to-end in a fresh desktop build; verification of the fix in this PR is at the daemon / `AgentManager` layer via the regression test. If you'd like me to attach a desktop video of the before/after, I'm happy to set that up — let me know which path you'd prefer to reproduce against (e.g. answering a Codex `request_user_input` in the terminal vs. in the Paseo card).

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran (oxfmt — no changes needed)
- [ ] UI changes include screenshots or video for every affected platform _(server-only change, no UI diff; see note above re: desktop video)_

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -5714,6 +5714,76 @@ test("respondToPermission emits refreshed state before permission_resolved", asy
   expect(resolvedIndex).toBeGreaterThan(refreshedStateIndex);
 });
 
+test("respondToPermission resolves stale permission cards when provider request is already gone", async () => {
+  const workdir = mkdtempSync(join(tmpdir(), "agent-manager-stale-permission-"));
+  const storagePath = join(workdir, "agents");
+  const storage = new AgentStorage(storagePath, logger);
+
+  class StalePermissionSession extends TestAgentSession {
+    override async respondToPermission(requestId: string): Promise<void> {
+      throw new Error(`No pending permission request with id '${requestId}'`);
+    }
+  }
+
+  class StalePermissionClient extends TestAgentClient {
+    override async createSession(config: AgentSessionConfig): Promise<AgentSession> {
+      return new StalePermissionSession(config);
+    }
+  }
+
+  const manager = new AgentManager({
+    clients: {
+      codex: new StalePermissionClient(),
+    },
+    registry: storage,
+    logger,
+    idFactory: () => "00000000-0000-4000-8000-000000000135",
+  });
+
+  const snapshot = await manager.createAgent({
+    provider: "codex",
+    cwd: workdir,
+  });
+  const agent = manager.getAgent(snapshot.id);
+  if (!agent) {
+    throw new Error("Expected managed agent");
+  }
+  agent.pendingPermissions.set("stale-question-1", {
+    id: "stale-question-1",
+    provider: "codex",
+    name: "request_user_input",
+    kind: "question",
+    title: "Question",
+    input: {
+      questions: [{ id: "q1", header: "Q1", question: "Pick one?", options: [] }],
+    },
+  });
+
+  const seen: string[] = [];
+  manager.subscribe(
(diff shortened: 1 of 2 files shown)


boudra:
Closing for now: this needs an observed Codex repro showing the stale question card and confirming that dismissing or answering it clears the card with this fix. Please reopen with that evidence.

boudra:

PR #1768, opened 2026-06-27
Support F-Droid Android builds

boudra:
### Linked issue

None found.

### Type of change

- [x] Bug fix
- [x] Docs

### What does this PR do

Makes F-Droid Android builds reproducible upstream instead of relying on downstream package removal and generated-file patches.

- derives deterministic Android and iOS build numbers from the package version and uses EAS local versioning
- adds a `PASEO_FDROID_BUILD=1` profile that builds Expo modules from source, excludes camera, Firebase notifications, and Expo development-client native modules, and disables OTA updates and Gradle dependency metadata
- substitutes Android JavaScript stubs for the excluded camera and notification packages and removes QR-only entry points from the F-Droid UI
- keeps normal Android, iOS, web, and EAS builds unchanged
- documents the serial, daemon-free source-build command so Gradle cannot exhaust the host through parallel workers

Risk surface: EAS local versioning applies to iOS as well as Android, so the next store release should confirm both stores accept the deterministic build numbers. The F-Droid profile intentionally has no QR scanning or push notifications; direct and pasted-link pairing remain available.

### How did you verify it

- `npm run format`
- `npm run lint`
- `npm run typecheck`
- clean flagged Expo Android prebuild
- `PASEO_FDROID_BUILD=1 ./gradlew assembleRelease --no-daemon --max-workers=1 -Dorg.gradle.parallel=false` (1,915 tasks, successful)
- inspected the 133 MB release APK: no `com.google.android.gms`, Barhopper, Firebase, ML Kit, or excluded Expo camera/notifications/dev-client native classes
- attempted a headless AVD install; the host has no KVM and software emulation did not start Android's package service within the bounded window, so the APK could not be installed there

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran
- [x] UI changes include screenshots or video for every affected platform (not applicable: build-profile-only conditional UI)
- [x] Tests added or updated where it made sense (verified through clean native generation, actual release assembly, and APK inspection)

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -13,6 +13,18 @@ EAS profiles: `development`, `production`, and `production-apk` in `packages/app
 
 `development` uses Android `debug`.
 
+## Version codes
+
+`packages/app/app.config.js` derives Android `versionCode` from the package version with:
+
+```text
+1_000_000 + major * 1_000_000 + minor * 1_000 + patch
+```
+
+Prerelease metadata is ignored, so `0.1.102-beta.1` and `0.1.102` both produce `1001102`. The same value is used as the iOS `buildNumber` because `packages/app/eas.json` uses EAS's local app version source. The `1_000_000` floor keeps the local formula above prior EAS remote counters during the transition. Do not re-enable EAS remote version counters or Android `autoIncrement`; F-Droid and other source-based builders need the native build number to be visible in the repo.
+
+The formula reserves three digits each for minor and patch. If either reaches `1000`, change the formula before cutting that release.
+
 ## Local build + install
 
 From repo root:
@@ -37,6 +49,19 @@ npx cross-env APP_VARIANT=production expo run:android --variant=release
 # Clear generated Android project
 rm -rf android
 ```
+
+## F-Droid / source-only Android builds
+
+F-Droid builds should set `PASEO_FDROID_BUILD=1` when running Expo prebuild:
+
+```bash
+cd packages/app
+PASEO_FDROID_BUILD=1 APP_VARIANT=production npx expo prebuild --platform android --clean --non-interactive
+cd android
+./gradlew assembleRelease
+```
+
+That env var is intentionally narrow. It enables `expo.autolinking.android.buildFromSource: [".*"]` so Expo modules are built from source instead of linked from precompiled AARs, and it applies `expo-gradle-jvmargs` with `-Xmx4096m` and `-XX:MaxMetaspaceSize=1024m` for the larger source build. Normal local and EAS builds leave those settings off.
 
 ### React version lockstep
 
--- a/docs/android.md
+++ b/docs/android.md
@@ -18,10 +18,10 @@ EAS profiles: `development`, `production`, and `production-apk` in `packages/app
 `packages/app/app.config.js` derives Android `versionCode` from the package version with:
 
 ```text
-1_000_000 + major * 1_000_000 + minor * 1_000 + patch
+major * 1_000_000 + minor * 1_000 + patch
 ```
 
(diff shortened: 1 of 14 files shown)


docs/android.md
```diff
@@ -38,6 +50,19 @@ npx cross-env APP_VARIANT=production expo run:android --variant=release
...
 
+## F-Droid / source-only Android builds
+
+F-Droid builds should set `PASEO_FDROID_BUILD=1` when running Expo prebuild:
+
+```bash
+cd packages/app
+PASEO_FDROID_BUILD=1 APP_VARIANT=production npx expo prebuild --platform android --clean --non-interactive
+cd android
+gradle assembleRelease
```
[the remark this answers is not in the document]
boudra: Fixed in the current head: the F-Droid docs now use `./gradlew assembleRelease` so the generated wrapper pins the Gradle version.

packages/app/eas.json
```diff
@@ -1,7 +1,7 @@
 {
   "cli": {
     "version": ">= 14.0.0",
-    "appVersionSource": "remote"
+    "appVersionSource": "local"
```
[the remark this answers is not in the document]
boudra: Addressed by adding a `1_000_000` floor to the deterministic native build number formula. The current Expo config resolves both Android `versionCode` and iOS `buildNumber` to `1001102`, which avoids rolling back below prior EAS remote counters while keeping source builds deterministic.

docs/android.md
```diff
@@ -13,6 +13,18 @@ EAS profiles: `development`, `production`, and `production-apk` in `packages/app
...
 
+## Version codes
+
+`packages/app/app.config.js` derives Android `versionCode` from the package version with:
+
+```text
+1_000_000 + major * 1_000_000 + minor * 1_000 + patch
+```
+
+Prerelease metadata is ignored, so `0.1.102-beta.1` and `0.1.102` both produce `1001102`. The same value is used as the iOS `buildNumber` because `packages/app/eas.json` uses EAS's local app version source. The `1_000_000` floor keeps the local formula above prior EAS remote counters during the transition. Do not re-enable EAS remote version counters or Android `autoIncrement`; F-Droid and other source-based builders need the native build number to be visible in the repo.
```
antonok-edm: I'd personally leave the floor out for simplicity.

The latest EAS-tracked release is in the 200s (most importantly, less than 1000). Looks like you started from `v0.1.0`, so even without the floor there is no chance of any future or historical version number being lower than the older EAS-tracked ones.

docs/android.md
```diff
@@ -38,6 +50,19 @@ npx cross-env APP_VARIANT=production expo run:android --variant=release
...
+F-Droid builds should set `PASEO_FDROID_BUILD=1` when running Expo prebuild:
+
+```bash
+cd packages/app
+PASEO_FDROID_BUILD=1 APP_VARIANT=production npx expo prebuild --platform android --clean --non-interactive
+cd android
+./gradlew assembleRelease
+```
+
+That env var is intentionally narrow. It enables `expo.autolinking.android.buildFromSource: [".*"]` so Expo modules are built from source instead of linked from precompiled AARs, and it applies `expo-gradle-jvmargs` with `-Xmx4096m` and `-XX:MaxMetaspaceSize=1024m` for the larger source build. Normal local and EAS builds leave those settings off.
```
antonok-edm: > it applies `expo-gradle-jvmargs` with `-Xmx4096m` and `-XX:MaxMetaspaceSize=1024m` for the larger source build

it's a minor distinction, but this config purely exists to support local `prebuild`, regardless of `buildFromSource` settings. The limit gets hit even with the precompiled AARs. I believe EAS remote build services configure a higher limit internally.

packages/app/app.config.js
```diff
@@ -139,6 +175,17 @@ export default {
...
+        ? [
+            [
+              "expo-gradle-jvmargs",
+              {
+                xmx: "4096m",
+                maxMetaspace: "1024m",
+              },
+            ],
+          ]
+        : []),
```
antonok-edm: If you're adding `expo-gradle-jvmargs` as an unconditional dependency, it should be safe to leave this in the config unconditionally as well.

PR #1769, opened 2026-06-27
fix(schedule): give scheduled new agents a workspace

gtoxlili:
### Linked issue

None. Reproduced locally (steps below).

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

A `new-agent` schedule creates its agent by calling `agentManager.createAgent`
directly from `ScheduleService.executeSchedule`, with no `workspaceId`.

That was harmless while workspace membership was keyed by directory — a scheduled
agent showed up under its cwd's workspace regardless. Commit
[`9966e49`](https://github.com/getpaseo/paseo/commit/9966e49) ("Key workspace
status and ownership by ID, not directory") made the `workspaceId` field
authoritative and added a one-time backfill that stamps pre-existing agents by
cwd. The schedule path was never updated to stamp `workspaceId` at create time,
so any scheduled agent created **after** that backfill has no workspace: it shows
up in neither the active nor the archived session list (both keyed by workspace
now) and is only reachable by id via `paseo inspect <id>`. For a recurring
schedule whose agent is meant to be found and unarchived later, it effectively
disappears.

The fix resolves the workspace from the target `cwd` before creating the agent —
the same cwd→workspace resolution the other create paths and the backfill use —
and passes the id through to `createAgent`. If resolution throws it logs and
continues without a workspace, so a workspace hiccup cannot take down an
otherwise healthy run.

### How did you verify it

- Reproduced on a throwaway schedule: `paseo schedule create ... --target
  new-agent --run-now`. The run reports `succeeded`, but the resulting agent has
  no `workspaceId`, is absent from `paseo ls --all`, and is only reachable via
  `paseo inspect <id>`. With this change the agent picks up the cwd's workspace
  and lists normally.
- Added a regression test that asserts the target cwd is forwarded to
  `ensureWorkspaceForCreate` and the returned id is stamped on the stored agent.

### Note

`loop-service.ts` creates its worker and verifier agents the same way (a bare
`agentManager.createAgent` with no `workspaceId`), so loop agents likely share
this gap. Left out here to keep the change focused on the reported schedule case.

### DIFF
--- a/packages/server/src/server/bootstrap.ts
+++ b/packages/server/src/server/bootstrap.ts
@@ -786,12 +786,20 @@ export async function createPaseoDaemon(
   });
   await loopService.initialize();
   logger.info({ elapsed: elapsed() }, "Loop service initialized");
+  const ensureWorkspaceForCreateExternal = async (cwd: string): Promise<string> => {
+    const workspace = await createLocalCheckoutWorkspace(
+      { cwd },
+      { projectRegistry, workspaceRegistry, workspaceGitService },
+    );
+    return workspace.workspaceId;
+  };
   const scheduleService = new ScheduleService({
     paseoHome: config.paseoHome,
     logger,
     agentManager,
     agentStorage,
     providerSnapshotManager,
+    ensureWorkspaceForCreate: ensureWorkspaceForCreateExternal,
   });
   await scheduleService.start();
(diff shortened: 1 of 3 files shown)


boudra:
Closing in favor of the scheduled-workspace implementation in [#1934](https://github.com/getpaseo/paseo/pull/1934).

PR #1770, opened 2026-06-27
feat(schedule): opt new-agent schedules out of auto-archive (keepAlive)

gtoxlili:
### Linked issue

None.

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

A `new-agent` schedule archives the agent it spawns as soon as the first run
settles (`ScheduleService.executeSchedule`). That is the right default for
one-shot jobs: run a prompt on a cron, do not leave agents lying around.

It does not fit agents whose work outlives a single run turn. Some scheduled
agents kick off long-running or asynchronous work and then go idle, intending to
wake themselves later to continue or report: self-scheduled follow-ups, polling
external state, multi-phase jobs that report over time. Archiving closes the
agent's process and drops those in-memory continuations, so the job is cut off
after the first turn, with no way to opt out.

This adds an opt-in `keepAlive` flag on the new-agent target config, surfaced on
the CLI as `paseo schedule create --keep-alive`. When set, the scheduler leaves
the agent in place after a successful run instead of archiving it, so the agent
can drive its own lifecycle. The default is unchanged. A failed run still
archives the agent so broken ones do not accumulate.

### How did you verify it

- Added unit tests covering both paths: the default still archives the spawned
  agent after a run, and `keepAlive: true` leaves it un-archived.
- I have not run the full local toolchain (format / typecheck / lint / vitest)
  in this environment, so I am relying on CI for that. The change follows the
  existing schedule schema and create path; the flag lives on the protocol
  new-agent config, which the RPC create schema already reuses.

### Notes

- Scope kept to the CLI + daemon path. The MCP `create_schedule` tool and the
  app's schedule UI are untouched; both could surface `keepAlive` later as a
  small addition since the protocol and service already support it.
- The `agent` target (reuse an existing agent) already never archives, so the
  flag is only meaningful for `new-agent`.

### DIFF
--- a/packages/cli/src/commands/schedule/create.ts
+++ b/packages/cli/src/commands/schedule/create.ts
@@ -19,6 +19,7 @@ export interface ScheduleCreateOptions extends ScheduleCommandOptions {
   provider?: string;
   mode?: string;
   cwd?: string;
+  keepAlive?: boolean;
   maxRuns?: string;
   expiresIn?: string;
   runNow?: boolean;
@@ -41,6 +42,7 @@ export async function runCreateCommand(
     provider: options.provider,
     mode: options.mode,
     cwd: options.cwd,
+    keepAlive: options.keepAlive,
     host: options.host,
     maxRuns: options.maxRuns,
     expiresIn: options.expiresIn,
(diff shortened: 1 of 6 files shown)


boudra:
Closing in favor of the schedule retention implementation in [#1934](https://github.com/getpaseo/paseo/pull/1934).

PR #1781, opened 2026-06-28
feat(server): keep workspace scripts in paseo.json order

### LINKED ISSUE #1686: feat: Show scripts in paseo.json order instead of alphabetizing them
### Prior search

### What's the problem

Scripts defined in `paseo.json` show up **alphabetically** in the Scripts dropdown, not in the order I wrote them in the file. I group related scripts on purpose (e.g. `dev`, `dev:api`, `dev:web`, then `build`, then `deploy`), but they get re-sorted, so my most-used script isn't first and related ones get split apart. The only way to influence order today is to rename scripts into an alphabetical sequence (`1-dev`, `2-build`, …), which is ugly and leaks ordering into names that are shown in the UI.

### What would solve it

**Phase 1 (small): respect file order.** The pipeline already preserves `paseo.json` key order from disk all the way to the wire — the *only* thing that discards it is one alphabetical sort right before scripts are emitted (`packages/server/src/server/script-status-projection.ts:259`, `sortPayloads`). Replacing that with file order makes the dropdown reflect what I wrote.

Because `scripts` is a JSON object and string-key insertion order survives the full round-trip — read → `JSON.parse` → Zod `z.record` → `JSON.stringify` write (already covered by the round-trip test in `paseo-config-file.test.ts`) — **no schema or protocol change is needed**; order is simply respected instead of thrown away. Two small notes: running-but-unconfigured ("orphan") scripts would append after configured ones (could stay alphabetical among themselves), and pure-integer script names would still reorder due to JS object semantics — a non-issue for real names.

**Phase 2 (optional follow-up): drag-to-reorder in Project Settings.** If hand-editing the file isn't enough, the Scripts section of Project Settings already does add/edit/remove, there's already a write path (`WriteProjectConfig` RPC + atomic `writePaseoConfigForEdit`), and there's already a drag primitive (`sortable-inline-list.web.tsx`, @dnd-kit, with a native fallback). Reordering = rebuild the `scripts` object's keys in the new order and save through the existing RPC; the file rewrite preserves order and the dropdown follows. Still no schema change. This mirrors the "custom order + reasonable fallback sort" shape from #1524, with the fallback being Phase 1's file order.

I'd ship Phase 1 on its own first (essentially a one-function change) and only do Phase 2 if you agree with this change.

- **Keep alphabetical, rename scripts to force order** (`1-dev`, `2-build`) — works today but pollutes names that are shown in the UI.
- **Client-side custom order persisted per-device** (like tab order) — rejected: invisible to `paseo.json`, doesn't travel with the repo or to other devices; the file is the natural source of truth.
- **Change `scripts` to an array in the schema to make order explicit** — rejected as unnecessary and back-compat-breaking: object key order already round-trips faithfully, so the hash is sufficient.

nkwib:
### Linked issue

Closes #1686

### Type of change

- [ ] Bug fix
- [ ] Docs

This implements the "Phase 1" proposed in #1686.

### What does this PR do

Workspace scripts in the run dropdown were alphabetized by a final `sortPayloads` call in `buildWorkspaceScriptPayloads`, discarding the order authors write in `paseo.json` (so grouping related scripts like `dev`, `dev:api`, `dev:web` was lost).

That sort was the only thing dropping file order. `scriptConfigs` already iterates in config key-insertion order, which survives the read round-trip. So:

- Configured scripts now keep their `paseo.json` order.
- Running-but-unconfigured ("orphan") scripts have no file position, so they trail the configured ones in a stable alphabetical order.

No schema or protocol change.

### How did you verify it

- Added a unit test in `script-status-projection.test.ts`: scripts defined `dev, build, api, deploy` now project in that file order (previously alphabetized to `api, build, deploy, dev`), and two orphan runtime scripts registered out of order trail alphabetically.
- `npx vitest run src/server/script-status-projection.test.ts` → 12 passing (11 existing + 1 new).
- Ran the dev daemon + web app and opened the workspace Scripts dropdown against a `paseo.json` with `dev, build, api, deploy`. Before/after below.

Server-only change; the screenshots are of the run dropdown it feeds.

**Before** (`main`, alphabetical):

Screenshot 2026-06-28 at 11 27 17

**After** (this branch, file order):

Screenshot 2026-06-28 at 11 25 52

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/server/src/server/script-status-projection.test.ts
+++ b/packages/server/src/server/script-status-projection.test.ts
@@ -449,6 +449,58 @@ describe("script-status-projection", () => {
     }
   });
 
+  it("keeps configured scripts in paseo.json order and trails orphans alphabetically", () => {
+    const workspaceId = "workspace-script-order";
+    // File order is deliberately non-alphabetical so a re-sort would be observable.
+    const workspace = createWorkspaceRepo({
+      paseoConfig: {
+        scripts: {
+          dev: { command: "npm run dev" },
+          build: { command: "npm run build" },
+          api: { command: "npm run api" },
+        },
+      },
+    });
+    const routeStore = new ScriptRouteStore();
+    const runtimeStore = new WorkspaceScriptRuntimeStore();
+    // Two running-but-unconfigured ("orphan") scripts, registered out of alphabetical order.
+    runtimeStore.set({
+      workspaceId,
+      scriptName: "ztask",
+      type: "script",
+      lifecycle: "running",
+      terminalId: "term-z",
+      exitCode: null,
+    });
+    runtimeStore.set({
+      workspaceId,
+      scriptName: "mtask",
+      type: "script",
+      lifecycle: "running",
+      terminalId: "term-m",
+      exitCode: null,
+    });
+
+    try {
+      const payloads = buildPayloads({
+        workspaceId,
+        workspaceDirectory: workspace.repoDir,
+        routeStore,
+        runtimeStore,
+        daemonPort: 6767,
+      });
+      expect(payloads.map((payload) => payload.scriptName)).toEqual([
+        "dev",
+        "build",
+        "api",
+        "mtask",
+        "ztask",
+      ]);
+    } finally {
+      workspace.cleanup();
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


boudra:

PR #1818, opened 2026-06-29
feat: real provider-native session fork (Kiro)

park0er:
Addresses #1479. **Stacked on #1792** (which introduces `KiroACPAgentClient`) — review/merge that first; this PR's lower commit is #1792.

## What this PR does

Adds `AgentManager.forkAgent(sourceAgentId)`: create a new, **independent** agent that inherits the source's **full** conversation context, so the original and the fork run in parallel (the `create_agent` `forkFrom` capability requested in #1479).

- Capability-gated (`AgentClient.capabilities.supportsFork` + `forkSession`), with **no lossy fallback** — providers that can't do a real fork throw rather than seeding a summary. This complements #1788 ("Fork assistant turns into new drafts", a text-summary draft) with the *real session* fork #1479 asks for.
- **Kiro provider fork**: duplicates Kiro's on-disk session record (`~/.kiro/sessions/cli/<id>.json` with `session_state` rewritten to a new id, plus `.jsonl`/`.history`) into a new session and loads it.
- Protocol wire (`agent.fork.request/response`) + `daemonClient.forkAgent`.
- UI: a **"Fork agent"** entry in the agent tab menu (matching the "right-click agent tab → Fork agent" UX in #1479), threaded through `split-container`/`tabs-row`; i18n in all 8 locales.

## Type of change

- [x] New feature (with prior issue: #1479)

## Scope note

Whole-conversation fork (not mid-message): Kiro's `session_state.rts_model_state` is an opaque model blob that can't be safely truncated, so the fork duplicates the whole conversation and branches from the end. Per-provider mid-point fork (e.g. Claude `forkSession` + conversation rewind) can layer on later.

## How I verified

- `protocol` + `client` + `server` + `app` typecheck.
- `vitest`: 130 server tests (`agent-manager-fork` 3 + `kiro-session-fork` 4 + `agent-manager` 116 regression + `wire-compat` 7) + 41 app tests (`workspace-tab-menu` 9 + i18n resources 32).
- `oxlint` + `oxfmt --check` clean on all changed files.
- **Real `kiro-cli` e2e** (local, not committed): forked an existing on-disk session; kiro-cli loaded the forked copy and replayed its history (full context carried over); original session untouched.

## Platforms — honest status

The tab-menu entry is RN (works desktop/web/native via the shared menu). I have **not** captured on-device screenshots yet; visual verification (web + iOS simulator) is in progress and I'll attach before merge. Flagging per CONTRIBUTING. Happy to adjust UX placement to your preference.

PR #1788, opened 2026-06-28
Fork assistant turns into new drafts

boudra:
### Linked issue

None found in open issue search.

### Type of change

- [ ] Bug fix
- [x] New feature (with prior issue + design alignment)
- [ ] Refactor / code improvement
- [ ] Docs

### What does this PR do

Adds a kebab menu to completed assistant turns with actions to fork the conversation into a new workspace draft tab or into the New Workspace composer. The daemon builds a chat-history text attachment from the selected turn using user messages, assistant messages, and tool-call summaries, and the app stores that attachment under a draft-scoped workspace attachment key so it survives New Workspace host switching.

The feature is capability-gated through `server_info.features.agentForkContext`; old hosts show the existing update-host message instead of trying to emulate the feature.

### How did you verify it

- `npm run build:client`
- `npx vitest run packages/server/src/server/agent/activity-curator.test.ts --bail=1`
- `npx vitest run packages/app/src/utils/host-routes.test.ts --bail=1`
- `npx vitest run packages/app/src/attachments/workspace-attachments-store.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`
- `npm --workspace packages/app run test:e2e -- assistant-fork-menu.spec.ts`

### Risk surface

Touches the WebSocket protocol, daemon timeline curation, and Expo composer UI. The new protocol fields are optional/nullable where needed and guarded by a server feature flag; the main remaining manual smoke surface is mobile/native menu rendering because the end-to-end coverage is browser-based.

### Checklist

- [x] One focused change. Unrelated cleanups split out.
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` ran (Biome)
- [ ] UI changes include screenshots or video for every affected platform
- [x] Tests added or updated where it made sense

PR #1834, opened 2026-06-30
feat: provider-native session fork for Kiro, Claude, Codex & TRAE

### LINKED ISSUE #1479: Feature: Fork/clone agent with existing session context
## Summary

Add a `resumeSessionId` (or `forkFrom`) parameter to `create_agent`, allowing a new agent to be created that inherits the full conversation context of an existing agent — effectively forking from a checkpoint.

## Motivation

Claude Code CLI supports `claude --resume <session-id>` in multiple shells simultaneously, each continuing independently from the same context. This is a practical way to "fork" a conversation.

However, Paseo's `create_agent` does not expose this capability. There is no way to create a new agent that starts with the same context as an existing one. The only workaround is extracting an activity summary via `get_agent_activity` and injecting it as a prompt — which loses full conversation context, tool call history, and nuance.

## Proposed API

```ts
create_agent({
  title: "Forked agent",
  provider: "claude/claude-sonnet-4-6",
  forkFrom: "<existing-agent-id>",  // or resumeSessionId
  initialPrompt: "Now focus on task B...",
})
```

The new agent should:
1. Start with a full copy of the source agent's conversation context
2. Be fully independent after creation (no shared state)
3. Optionally accept an `initialPrompt` to immediately diverge

## Use Cases

- Fork an agent that has built up deep codebase understanding to parallelize multiple tasks
- Create a "checkpoint" and explore different approaches from the same state
- Split a long-running agent into specialized sub-tasks without losing context

## Alternatives Considered

- `get_agent_activity` + new agent with summary as prompt: lossy, no tool history
- Manual `claude --resume` in multiple terminals: works but bypasses Paseo orchestration

park0er:
Closes #1479

## What

Adds a **real, provider-native session fork** that branches an agent into an independent parallel session — for the four mainstream providers: **Kiro, Claude, Codex, and TRAE**. Two entry points:

1. **Whole-conversation fork** — the agent tab `⋮` → **"Fork agent"** (per issue #1479). Duplicates the entire provider session into a new, independent agent that runs in parallel; the original is never mutated.
2. **Per-message fork into a new tab** — reworks the assistant-turn fork (#1788) so **"Fork into new tab"** does a real native fork *truncated at the selected message*, instead of bundling prior history as a chat-history attachment. The branch has native, scrollable history. **"Fork into new workspace/worktree/host"** keeps the attachment flow (a native session is bound to its provider + cwd and can't be relocated).

## Per-provider mechanism

| Provider | Fork mechanism | Per-message truncation |
|---|---|---|
| **Kiro** | Duplicate on-disk session files (`~/.kiro/sessions/cli/<id>.{json,jsonl,.history}`) → new UUID, rewrite `session_id`, `loadSession` | whole-conversation only (opaque `session_state`) |
| **Claude** | Claude SDK `forkSession(sessionId, { upToMessageId })` (operates on the persisted JSONL) | ✅ native (timeline id == Claude uuid) |
| **Codex** | app-server `thread/fork` (whole-thread copy, non-destructive), then `thread/rollback` to the boundary turn | ✅ native (turn rollback) |
| **TRAE** | Duplicate the on-disk session dir (`~/Library/Caches/trae-cli/sessions/<id>/`) → new UUID + `loadSession` | whole-conversation only |

Capability-gated via `capabilities.supportsFork` + `AgentClient.forkSession`; the provider registry forwards `forkSession` (normalizing the handle provider, like `resumeSession`). Providers without a native fork (e.g. plain ACP) simply don't expose it.

## Advantages of the native per-message fork (vs the attachment approach)

- **Native, scrollable history** — the branch *is* the real conversation in the provider's own session, not a collapsed "Chat history" attachment.
- **Context fidelity** — the model resumes its own native session state, rather than re-ingesting a serialized transcript as pasted text (lossy + re-tokenized).
- **Cheaper / faster** — no re-uploading a large transcript each turn.
- **One mental model** — same `forkSession(upToMessageId)` path as the tab `⋮` fork; a fork is a real, persisted, re-openable independent session.

## Tradeoffs / scope

- **Per-message truncation needs provider support.** Claude/Codex truncate natively; Kiro/TRAE only support whole-conversation fork (their session state can't be safely truncated mid-conversation), so the per-message control stays on rewind-capable providers — same gating as today.
- The **attachment-based fork is intentionally kept for the "new workspace/worktree/host" target**, which a native (provider+cwd-bound) session can't replicate.

## Dependency

The **TRAE** parts depend on the `traecli` provider added in **#1831** — this PR is **stacked on #1831** (its commits appear here until #1831 merges). Kiro/Claude/Codex fork have no such dependency.

## Relationship to existing PRs/issues

- **Supersedes #1818** (the earlier Kiro-only fork, which was stacked on the now-merged #1792). #1818 will be closed with a pointer here.
- Implements the whole-conversation fork requested in **#1479**.

## Verification

- `@getpaseo/server` + `@getpaseo/app` typecheck clean.
- Fork test suite: 60 server tests (per-provider fork + registry forwarding + capability gating + seeded-timeline truncation) + app i18n/tab-menu tests green; oxlint/oxfmt clean.
- Real end-to-end on a live daemon (the exact `agent.fork.request → forkSession` path the UI triggers): whole-conversation fork verified for **all four** providers (independent agent, non-empty title, source intact); per-message native fork verified for **Claude & Codex** (forked branch truncated at the selected turn; the model can no longer recall messages after the boundary).
## Screenshot evidence (all 4 fork scenarios, real Codex/Claude/Kiro sessions)

**A1 — Codex per-message fork → new tab (with truncation + auto-nav proof)**

A1-1
A1-2
A1-3
A1-4
A1-5
A1-6

A1-6: the forked agent, asked "what words have I asked you to say so far", correctly replies with only the pre-fork turns — confirmed the underlying provider session (not just the displayed timeline) is truncated.

**A2 — Codex per-message fork → new workspace (attachment fallback)**

A2-1
A2-2

**C — Claude tab ⋮ → Fork agent (whole conversation)**

C-1
C-2
C-3

**D — Kiro tab ⋮ → Fork agent**

D-1
D-2
D-3

Tested on web/desktop (Chrome). All sessions created fresh against a real dev daemon with real provider calls (model `xiaomi/mimo-v2.5-pro`) — no mock/stub state reused.


park0er:
Thanks for the review — addressed the feedback (pushed):

- **Synchronous file I/O (kiro-session-fork.ts, trae-session-fork.ts)** — switched both to `fs/promises` (`cp`/`readFile`/`writeFile`/`copyFile`); the fork no longer blocks the event loop while other sessions are mid-IPC. Callers `await` the now-async helpers; tests updated to `await` / `.rejects`.
- **TRAE macOS-only default path** — `resolveTraeSessionsDir` is now platform-aware (macOS `~/Library/Caches`, otherwise `$XDG_CACHE_HOME`/`~/.cache`), still overridable via `TRAE_SESSIONS_DIR`.
- **Fragile key-order trace rewrite** — replaced the positional `key/type/value` string match with a blanket replace of the source session UUID across `events.jsonl`/`traces.jsonl`. The id is a full UUID, so this is order-independent and can't silently miss a reordered field.
- **Codex test using `vi.spyOn` + private `castInternals`** — added a `_spawnAppServer` deps seam and restructured the test to a subclass that overrides `resumeSession` and injects the child + scripted client via `deps`. No more spying on an own method or touching private internals.

Remaining item — **"Fork agent" shown regardless of `capabilities.supportsFork`**: agreed it should be gated client-side. The app's agent state does carry `capabilities`, so it's feasible, but `onForkAgent` is threaded through several tab-menu layers and gating cleanly means plumbing the per-agent capability through each. I'd like to do that as a small follow-up — meanwhile it fails gracefully (friendly toast) and all four providers wired here advertise `supportsFork`. Happy to fold it into this PR if you'd prefer it gated before merge.

park0er:
Follow-up: implemented the client-side capability gate too — the "Fork agent" tab-menu entry is now threaded with the agent's `capabilities.supportsFork` (from session state) and only rendered for providers that actually implement `forkSession`. Added a unit test for the gated-off case. So non-fork providers no longer surface an option that fails on click.

park0er:
Fixed the remaining item: `resolveTraeSessionsDir.test` no longer asserts the macOS `Library/Caches` suffix unconditionally — it now checks the common `trae-cli/sessions` suffix on all platforms and the `Library/Caches` prefix only on `darwin`, so it passes on Linux CI (where the function correctly returns the XDG path). Production code unchanged.

park0er:
Addressed the remaining test-file item: `claude/agent.fork.test.ts` no longer uses `vi.mock`, `vi.spyOn(fs.promises)`, or instance-method reassignment. `ClaudeAgentClient` now exposes two injectable seams — `rewindSdk` (defaults to `realClaudeRewindSdk`) and `readSessionFile` (defaults to `fs.promises.readFile`) — and the test injects a recording fake SDK + a `TestClaudeAgentClient` subclass overriding `resumeSession`, mirroring the Codex fork test. Production defaults are unchanged; 6 fork tests + the 5 existing rewind tests pass.

park0er:
@boudra Pushed two fixes found during manual acceptance testing, plus full screenshot evidence for all 4 fork scenarios.

## Fix 1 — Codex per-message fork wasn't actually truncating the underlying thread

**Symptom**: forking a 3-turn conversation from the 2nd turn showed only 2 turns in the UI, but asking the forked agent "what did I ask you" revealed it still knew about turn 3 — the Paseo timeline was truncated, the real Codex thread wasn't.

**Root cause**: the boundary lookup matched `upToMessageId` (the streaming-event `resp_..._msg` id) against Codex's `thread/read` RPC response, but `thread/read` returns a completely different, position-based id scheme (`item-1`, `item-2`...) on every thread — forked or not. The match never succeeds, so the rollback boundary is never found and 0 turns get rolled back.

**Fix**: `AgentManager.forkAgent` now resolves the boundary as a 0-based turn index from its own (id-consistent) timeline and passes it via a new `AgentForkOptions.upToTurnIndex` field. The Codex provider uses that index directly against the forked thread's turn count — no id matching at all.

Added 2 regression tests; 61/61 fork-related unit tests pass.

## Fix 2 — Fork success didn't navigate to the new tab

Both fork entry points (tab ⋮ → "Fork agent", per-message fork → new tab) only showed a success toast and left the active tab on the original agent. Confusing on its own, and it's how the truncation bug above went unnoticed in the original unit tests — a manual tester (or a follow-up script) easily ends up typing into the wrong session while believing they're on the fork.

`forkAgent()` already returns `newAgentId` — both call sites now use it to navigate to the forked tab automatically. Verified the previously-active tab's `aria-selected` flips to false and the new tab's to true, with zero manual clicks.

## Screenshots (all 4 scenarios, real Codex/Claude/Kiro sessions, xiaomi/mimo-v2.5-pro)

**A1 — Codex per-message fork → new tab (with truncation + auto-nav proof)**
A1-1
A1-2
A1-3
A1-4
A1-5
A1-6

A1-6 shows the forked agent replying "You've asked me to say: 1. One 2. Two" — correctly omitting turn 3, confirmed on a fresh session with no manual tab clicks involved.

**A2 — Codex per-message fork → new workspace (attachment fallback)**
A2-1
A2-2

**C — Claude tab ⋮ → Fork agent (whole conversation)**
C-1
C-2
C-3

**D — Kiro tab ⋮ → Fork agent**
D-1
D-2
D-3

Tested on web/desktop (Chrome via Playwright). All sessions created fresh against a real dev daemon with real provider calls — none of this reused mock/stub state.

park0er:
Addressed the 4/5 feedback:

1. **`resolveLastMessageId` ignored `this.configDir`** — whole-conversation Claude fork resolved the session file path from `process.env.CLAUDE_CONFIG_DIR` directly, so installations with an explicit non-default `configDir` would look in the wrong place and fail with "session file not found". Now reuses the existing `resolveClaudeConfigDir(configDir)` helper (exported it from `models.ts`), which already implements the correct configDir-then-env-var precedence. Per-message fork (always has an explicit messageId) was unaffected.
2. **Banned test pattern in `agent-manager-fork.test.ts`** — was using `Object.defineProperty` to monkey-patch `capabilities` on a live `FakeAgentClient` instance. `FakeAgentClient` now takes an optional capabilities override in its constructor (`createTestAgentClients` forwards per-provider overrides), so the "provider without native fork" test constructs the fake with `supportsFork: false` directly.

61/61 fork tests still pass; typecheck/lint/fmt clean.

park0er:
Re the Codex per-message fork concern (4/5): the stated mechanism doesn't match what I observe against a real Codex app-server.

Claim was: `thread/read` puts user-message turns and assistant-message turns as separate entries, so `countAssistantTurns(forkRows) - 1` under-counts and the boundary turn gets rolled back.

I added a temporary debug log dumping the raw `thread/read` turn structure and ran it against real Codex sessions (model `xiaomi/mimo-v2.5-pro`) for several shapes:

- **1-turn conversation, fork at that turn**: `turnCount: 1`, turn items `["userMessage", "reasoning", "agentMessage"]` (user + assistant in the *same* turn) → `boundaryTurnIndex: 0` → `numTurns = 1 - (0+1) = 0`. No rollback — correct, since the boundary is the last (only) turn.
- **2-turn conversation, fork at turn 0**: `turnCount: 2`, each turn `["userMessage", "reasoning", "agentMessage"]` → `numTurns = 2 - (0+1) = 1`. Rolls back exactly turn 1, keeps turn 0 — verified by asking the forked agent afterward, which only knew about the first exchange.
- **3-turn conversation, fork at the last turn (index 2)**: `numTurns = 3 - (2+1) = 0`. Full copy kept — verified the forked agent could still list all 3 original words.

In every case Codex's `thread/read` keeps the user message and its assistant reply in the *same* turn, so `countAssistantTurns` (which counts `assistant_message` rows) lines up 1:1 with Codex's turn count. I don't have a repro for the "user message as its own turn" shape — if you have a specific conversation shape in mind (tool calls without a final assistant text? an interrupted turn?) that would help, but I'd rather not push a speculative fix for a mechanism I can't reproduce against the real API.

Happy to add a defensive assertion (log + fallback to a full/no-rollback copy) if `turnCount` and the expected assistant-turn count ever diverge, so this fails safe instead of silently mis-rolling-back if some conversation shape does trigger it — let me know if that's useful.

park0er:
@boudra Status update — this is ready for review:

- Rebased onto latest `upstream/main` (had picked up a real merge conflict in `claude/agent.ts` from unrelated upstream changes) — clean now, no conflicts.
- All 18 CI checks pass (typecheck, lint, format, full test matrix, all playwright shards, server-tests on both ubuntu/windows).
- **Greptile Review: 5/5 — "Safe to merge."** All previously-raised concerns (sync I/O, macOS-only path, fragile `traces.jsonl` string replacement, banned test patterns, capability gating, the Claude `configDir` bug, the Codex turn-index fix) are confirmed addressed. The two remaining notes are minor/non-blocking (JSONL memory efficiency in `resolveLastMessageId`, a possible orphaned thread if `resumeSession` throws after a successful fork on Codex) and match existing patterns elsewhere in the codebase.
- Added full screenshot evidence for all 4 fork scenarios directly to the PR description above (per-message fork with truncation proof, whole-conversation fork for Claude and Kiro) — all captured against a real dev daemon with real provider sessions, not mocks.

boudra:
I prefer the chat-history-based fork implementation merged in [#1788](https://github.com/getpaseo/paseo/pull/1788) and extended across providers in [#2022](https://github.com/getpaseo/paseo/pull/2022). Closing this alternative implementation.

PR #1857, opened 2026-07-02
fix(app): show New workspace action on non-git sidebar projects

### LINKED ISSUE #1856: bug: non-git projects have no "New workspace" (+) button in the sidebar
### What's broken

When a project isn't a git repository (added as a plain directory), its row in
the sidebar has no "+" to create a new workspace for it. Git projects show this
affordance — a "+" on the project row, and a "New workspace" row when the project
has no workspaces yet — but non-git projects show nothing, so there's no way to
start a new workspace from a non-git project's row.

Expected: the same "New workspace" "+" that git projects have, so a workspace can
be created from a non-git project too. (The New Workspace screen already supports
non-git projects — it defaults them to Local isolation — so only the sidebar entry
point is missing.)

### Steps to reproduce

1. Add a plain, non-git directory as a project (a folder that isn't a git repo).
2. In the sidebar, hover that project's row.
3. No "+" to add a workspace appears — a git project's row shows one. If the
   non-git project has no workspaces yet, no "New workspace" row appears either.

Result: no way to create a new workspace from the non-git project's sidebar row.

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.103

### OS version

macOS (Darwin 25.5.0)

### Agent provider

Not relevant

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

No "+" here:

Image

But here:

Image

cleiter:
### Linked issue

Closes #1856

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

The per-project **New workspace** affordance in the sidebar — the project-header
`+` button and the empty-project "New workspace" ghost row — was gated on
`host.canCreateWorktree`, which is `projectKind === "git"`. So a non-git project
(`non_git` / `directory`) showed no way to add a workspace from its row, even
though a host with the `workspaceMultiplicity` capability can create additional
local workspaces for it (the New Workspace screen already handles this and
defaults such projects to Local isolation).

This threads the per-host `workspaceMultiplicity` flag into the sidebar project
row model and shows the affordance when
`canCreateWorktree || supportsWorkspaceMultiplicity` — the same gate already used
by the global "New workspace" button (`SidebarNewWorkspaceHeaderRow`) and the
Cmd+N handler (`useGlobalNewWorkspaceAction`). The internal trailing-action kind
is renamed `new_worktree` → `new_workspace` (a client-only render model, not a
wire/persisted type) since it now also covers non-git workspaces.

Files: `sidebar-project-row-model.ts` (gate + rename), `sidebar-workspace-list.tsx`
(compute `useHostFeatureMap(..., "workspaceMultiplicity")` and thread it to the
row model), plus unit tests.

**Note on the lint suppression.** `areProjectBlockPropsEqual` (the `React.memo`
comparator for project rows) is a flat list of reference-equality checks. Adding
the one `supportsMultiplicityByServerId` comparison tips oxlint's `complexity`
rule (max 20) to 21, so the function carries a `// oxlint-disable-next-line
complexity` — 
matching existing precedent in the codebase (`composer/index.tsx`,
`browser-pane.electron.tsx`). I kept the explicit `previous.x === next.x &&`
chain rather than the two alternatives that would pass the rule: (a) rewriting it
as a `KEYS.every(k => previous[k] === next[k])` loop, or (b) splitting it into two
helper functions. Both hide a trivial, exhaustive prop comparison behind
indirection; the disable keeps the comparator flat and obvious (you can see at a
glance every prop it compares), which matters for a memo boundary where a missed
field is a stale-render bug.

### How did you verify it

- `npm run typecheck` — passes (full workspace)
- `npm run lint` — passes (0 warnings/errors)
- `npm run format` — clean (Biome)
- `npx vitest run packages/app/src/utils/sidebar-project-row-model.test.ts` —
  10/10 pass, including new cases: non-git + multiplicity → shows the action;
  non-git without multiplicity → none; git ignores multiplicity; multi-host picks
  the first eligible host.

Manual (on this branch):

- Added a plain non-git directory as a project. The project-header `+` (and the
  "New workspace" row on an empty non-git project) now appear, and `+` opens the
  New Workspace screen seeded to that project with isolation defaulting to Local.
- A git project's row is unchanged.

Before:

image

After:

image

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/e2e/new-workspace-entry.spec.ts
+++ b/packages/app/e2e/new-workspace-entry.spec.ts
@@ -13,12 +13,12 @@ import { getServerId } from "./helpers/server-id";
 import { clickArchiveWorkspaceMenuItem, expectWorkspaceAbsentFromSidebar } from "./helpers/sidebar";
 import { waitForSidebarHydration } from "./helpers/workspace-ui";
 
-// Model B entry points into the New Workspace screen. The per-project
-// "+ New workspace" sidebar row is gone; the surviving entries are the global
-// button (universal) and each git project's own new-worktree icon (preselects
-// that project). These specs prove the global entry opens the screen, the
-// project icon preselects the right project across the reused 'new' screen, and
-// non-git projects never offer the worktree Isolation control.
+// Model B entry points into the New Workspace screen. The surviving entries are
+// the global button (universal) and each project's per-row New workspace icon
+// (preselects that project) — shown for git projects and for non-git projects on
+// a multiplicity-capable host. These specs prove the global entry opens the
+// screen, the project icon preselects the right project across the reused 'new'
+// screen, and non-git projects never offer the worktree Isolation control.
 
 function projectRow(page: import("@playwright/test").Page, projectKey: string) {
   return page.getByTestId(`sidebar-project-row-${projectKey}`);
@@ -215,7 +215,7 @@ test.describe("New workspace entry points", () => {
       await expect(projectRow(page, nonGitProject.projectId)).toBeVisible({ timeout: 30_000 });
 
       // Open New Workspace for the non-git project via the global button, then
-      // select it in the picker (its row has no new-worktree icon).
+      // select it in the picker (the per-row icon would preselect it too).
       await openGlobalNewWorkspaceComposer(page);
       const trigger = page.getByTestId("new-workspace-project-picker-trigger");
       await expect(trigger).toBeVisible({ timeout: 30_000 });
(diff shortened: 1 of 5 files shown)


cleiter:
Heads-up — this reverses a tested design decision, flagging for an explicit call.

This makes the per-project New workspace “+” appear for non-git projects on a workspaceMultiplicity host. That reverses what [sidebar-model-b.spec.ts](paseo://app/h/srv_zOmMHbyHdXVK/workspace/sidebar-model-b.spec.ts) asserted — that non-git rows have no per-row icon “since worktree creation needs a git checkout.”

That reasoning predates multiplicity: a non-git directory can now host additional workspaces, and the New Workspace screen already creates them (Local isolation). The “+” just opens that screen pre-seeded — the same entry point git projects already have. I updated the spec to expect the icon on both project kinds.

Issue #1047, opened 2026-05-16
audit: platform/form-factor gates use wrong proxies


boudra:
audit done while extracting bug fixes from #905. several platform/form-factor gates in `packages/app/src` are using the wrong proxy.

### wrong gates

- **enter-submits-message** — `message-input.tsx` uses `isWeb && !isCompact`. true rule = "physical keyboard chord available". web compact tablet with bluetooth keyboard slips through. native iPad is already handled separately via `useIosHardwareKeyboardSubmit()` / `PaseoHardwareKeyboard`. two rails for the same behavior.

- **long-press dnd activation** — `draggable-list.web.tsx` gates on `useDragHandle`, which is irrelevant. desktop mouse with a handle gets touch-style delay; touch user without a handle gets distance activation. backwards. real rule = primary pointer is coarse.

- **ios viewport zoom lock** — `use-compact-web-viewport-zoom-lock.ts` gates on `isWeb && isCompactLayout`. fires on android chrome and narrow desktop safari too. real rule = iOS/iPadOS WebKit on focused input <16px.

- **hover-reveal actions** — four different patterns for the same concept: \`isHovered || isNative || isCompact\`, \`isTouchPlatform = platformIsNative\` (workspace row, with \`_isCompact\` unused), \`hovered || isNative\`, \`onOpenFile && isHovered\`.

- **edge swipe** — `_layout.tsx` and `use-explorer-open-gesture.ts` fail on `isWeb`. compact-scoping happens by accident of wrapper placement. real rule = mobile browser back-swipe territory.

### existing precedent

`useCanShowReviewKeyboardHints()` in `packages/app/src/screens/review/surface.tsx` already uses \`matchMedia("(hover: hover) and (pointer: fine)")\`. shape we should generalize.

### proposed gates

| gate | detection | replaces |
|---|---|---|
| \`useHasHover()\` | web: \`matchMedia("(hover: hover) and (pointer: fine)")\`; native: false | tooltip / review-hint ad hoc checks |
| \`useShouldRevealHoverActions()\` | \`isCompact \|\| !useHasHover()\` | the 4 inconsistent hover-reveal patterns |
| \`usePrimaryPointerIsCoarse()\` | web: \`matchMedia("(pointer: coarse)")\`; native: true | long-press dnd's \`useDragHandle\` proxy, touch ergonomics gates |
[...]

PR #1858, opened 2026-07-02
fix: show virtual keys in terminal on tablets/foldables

bitnimble:
### Linked issue

Doesn't fully close, but address one slice of https://github.com/getpaseo/paseo/issues/1047. Fixes using the terminal on foldables and tablets, atm they cannot use arrow keys etc. 

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Visibility of virtual keys in the terminal pane were previously gated on screen size (`isMobile`), this PR fixes it to be based on either soft keyboard presence (on native, where we can actually know that for sure), and otherwise `pointer: coarse` / touch (on web). 

This fixes devices like foldables and tablets, where there is a large resolution (doesn't fulfil `isMobile`) but there's a touchscreen with virtual keyboard, so the virtual keys didn't display.

Note: on mobile web there's actually a pre-existing bug unrelated to my change that means that the virtual keys don't shift upward when the keyboard is shown, so they are actually obscured and non-functional... I can fix that in a follow up PR if you'd like, but probably would prefer to not fold it into this PR since its a larger change. So I can't actually show those changes here...

### How did you verify it

Did a debug build of Android APK + web UI daemon, manually tested on my Galaxy Fold 7:

|Before|After|
|-|-|
|Screenshot_20260702_184802_Paseo|Screenshot_20260702_165914_Paseo Debug|
|-|And screenshot of folded state / regular phone resolution to show that it doesn't break the usual case:Screenshot_20260702_165909_Paseo Debug|

I unfortunately don't have an iOS device (I guess an iPad?) to test these changes there, but the .native.ts is so simple that if something is wrong it would probably be in `react-native-keyboard-controller` itself.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/components/terminal-pane.tsx
+++ b/packages/app/src/components/terminal-pane.tsx
@@ -13,6 +13,7 @@ import type { TerminalInputModeState } from "@getpaseo/protocol/terminal-input-m
 import { useTranslation } from "react-i18next";
 import { useHostRuntimeClient, useHostRuntimeIsConnected } from "@/runtime/host-runtime";
 import { useKeyboardShiftStyle } from "@/hooks/use-keyboard-shift-style";
+import { useSoftKeyboardVisible } from "@/hooks/use-soft-keyboard-visible";
 import { useAppVisible } from "@/hooks/use-app-visible";
 import { useStableEvent } from "@/hooks/use-stable-event";
 import {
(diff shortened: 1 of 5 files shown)


packages/app/src/components/terminal-pane.tsx
```diff
@@ -180,12 +181,12 @@ export function TerminalPane({
     return trimmed.length > 0 ? trimmed : undefined;
   }, [settings.monoFontFamily]);
   const isMobile = useIsCompactFormFactor();
+  const softKeyboardVisible = useSoftKeyboardVisible();
   const mobileView = usePanelStore((state) => state.mobileView);
   const showMobileAgentList = usePanelStore((state) => state.showMobileAgentList);
   const swipeGesturesEnabled = isMobile && mobileView === "agent";
   const { shift: keyboardShift, style: keyboardPaddingStyle } = useKeyboardShiftStyle({
     mode: "padding",
-    enabled: isMobile,
```
bitnimble, line 189: `enabled` is true by default in `useKeyboardShiftStyle` and it internally handles mobile-ness already

packages/app/src/components/terminal-pane.tsx
```diff
@@ -337,14 +338,14 @@ export function TerminalPane({
   }, [clearKeyboardRefitTimeouts]);
 
   useAnimatedReaction(
-    () => isMobile && keyboardShift.value > 0,
+    () => keyboardShift.value > 0,
```
bitnimble, line 341: `useKeyboardShiftStyle` already handles soft keyboard / "mobileness" internally, so this guard isn't needed

PR #1880, opened 2026-07-03
fix(server): prevent Pi slash commands from locking threads

ebg1223:
### Linked issue

N/A

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Some Pi slash commands are handled entirely inside Pi/extensions and never start an agent turn. They can acknowledge the RPC `prompt` without emitting `agent_start`, `turn_start`, or `agent_end`.

Before this fix, Paseo treated that prompt as an active turn forever. The result was worse than a missing completion event: the whole thread became locked in a running state, follow-up prompts were rejected as “A Pi turn is already active,” and there was no in-thread recovery path because Pi was never going to emit the lifecycle event Paseo was waiting for.

This updates the Pi provider to detect those local slash-command completions. After a slash prompt ACK, it yields once, uses `get_state` as an ordered RPC barrier, and completes the Paseo turn only if no lifecycle event arrived and Pi is not streaming. Any notify text emitted during that no-turn path is surfaced as an assistant message. Normal streaming/lifecycle turns continue through the existing completion path.

### How did you verify it

- `npx vitest run packages/server/src/server/agent/providers/pi/agent.test.ts --bail=1` — 37 tests passed
- `npm run lint` — passed
- `npm run typecheck --workspace=@getpaseo/server` — passed
- `npm run format` — ran successfully
- Pre-commit hook also ran changed-file lint/format checks and full `npm run typecheck`

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A: no UI changes)


boudra:
Closing in favor of the Pi slash-command completion implementation in [#2066](https://github.com/getpaseo/paseo/pull/2066).

PR #1907, opened 2026-07-06
fix(server): apply the default model's default thinking option when it is auto-filled

odoo-mvds:
## Problem

When an agent is created **without an explicit model and without a thinking option**,
`AgentManager.normalizeConfig` resolves and fills the provider's default model
(`catalog.models.find(m => m.isDefault)`) and the provider's default mode (`defaultModeId`) —
but it leaves `thinkingOptionId` unset.

The Claude provider's `resolveThinkingConfig` treats an empty thinking option as "no extended
thinking", so the newly-created agent runs **without the default model's default thinking
option**, even when the catalog explicitly marks a thinking option `isDefault` (or exposes
`defaultThinkingOptionId`).

This is inconsistent with the create-agent picker, which always pairs an auto-selected model
with that model's default thinking (`resolve-agent-form.ts` → `resolveThinkingOptionId`). The
only time the client submits an empty thinking option is when it also submitted no model —
exactly the case the server fills the default model. So the server fills the model but drops
its default thinking.

### Repro

```
paseo run --provider claude "hello"
paseo inspect <id>
```

- **Before:** `Model` is the default model, but `Thinking` is the SDK/adaptive default — the
  model's `defaultThinkingOptionId` is not applied.
- **After:** `Thinking` is the default model's default thinking option.

## Fix

In `normalizeConfig`, when the default model is auto-filled and no thinking option was
provided, also set `thinkingOptionId` from the resolved model's `defaultThinkingOptionId`
(falling back to `thinkingOptions.find(o => o.isDefault)`). This mirrors the picker and the
existing default-model / default-mode fills. The private `resolveDefaultModelId` is refactored
to `resolveDefaultModel` so the caller can read the model's thinking metadata.

This does **not** change the create-agent picker UX — the fresh picker still does not
auto-select a model (that behavior and its test in `resolve-agent-form.test.ts` are unchanged).
It only affects server-side config normalization for agents created without an explicit
thinking option.

## Tests

Two new cases in `agent-manager.test.ts`:
- the auto-filled default model is paired with its default thinking option;
- an explicit thinking option is preserved when the default model is auto-filled.

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -471,6 +471,72 @@ test("normalizeConfig injects the provider default model when omitted", async ()
   expect(snapshot.config.modeId).toBe("auto");
 });
 
+class DefaultThinkingClient extends TestAgentClient {
+  override async fetchCatalog() {
+    return {
+      models: [
+        {
+          provider: "codex",
+          id: "gpt-5.4",
+          label: "GPT-5.4",
+          isDefault: true,
+          defaultThinkingOptionId: "high",
+          thinkingOptions: [
+            { id: "low", label: "Low" },
+            { id: "high", label: "High", isDefault: true },
+          ],
+        },
+      ],
+      modes: [],
+    };
+  }
+}
+
+test("normalizeConfig pairs the auto-filled default model with its default thinking option", async () => {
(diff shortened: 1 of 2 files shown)


odoo-mvds:
will work with the new profile system

PR #1917, opened 2026-07-06
Keep desktop upgrades restarting bundled CLI daemons

boudra:
### Linked issue

None found. Searched open issues for desktop-managed daemon, desktop upgrade restart, bundled CLI daemon, and `PASEO_DESKTOP_MANAGED`.

### Type of change

- [x] Bug fix
- [x] Docs

### What does this PR do

This PR keeps desktop-bundled daemons upgradeable even when they were started from the bundled CLI instead of the desktop app. The old `desktopManaged` pid-lock bit recorded who spawned the daemon; that was the wrong definition, because a CLI-started bundled daemon still runs the desktop's daemon runtime and must restart when the desktop app upgrades.

The new definition is executable identity: the daemon records `process.execPath` in `paseo.pid`, and desktop classifies the daemon as managed when that executable is the desktop executable or lives inside the same desktop installation.

| Launch path | Recorded executable identity | Desktop classification |
| --- | --- | --- |
| Desktop app spawn on macOS | `Paseo.app/Contents/Frameworks/Paseo Helper.app/Contents/MacOS/Paseo Helper` | Managed, because it is inside the app bundle |
| Bundled CLI shim start | The desktop Electron/Helper executable from the installed app (`Paseo Helper` on macOS, `Paseo.exe` on Windows, `Paseo`/`Paseo.bin` on Linux) | Managed, because it is inside the same desktop install |
| npm-installed CLI start | The system Node executable from `process.execPath` | Unmanaged, because it is outside the desktop install |

Back-compat decision: old locks without `executablePath` fall back to the legacy `desktopManaged` boolean, with `COMPAT(desktopManagedPidLock)` tags for cleanup. New locks no longer write the legacy boolean, so the stale spawn-intent bit does not keep spreading. The legacy env var stays scrubbed from harvested/external environments so old runtime-control state does not leak into launched processes.

CLI status decision: `paseo daemon status` now reports the raw `daemonExecutablePath`. Its `desktopManaged` JSON field remains legacy-only because an npm CLI has no authoritative desktop install root to compare against.

### How did you verify it

- `npm run format`
- `npm run lint`
- `npm run typecheck`
- `npx vitest run packages/server/src/server/pid-lock.test.ts --bail=1`
- `npx vitest run packages/cli/src/commands/daemon/status.test.ts --bail=1`
- `npx vitest run packages/cli/src/commands/daemon/local-daemon.supervision.test.ts --bail=1`
- `npx vitest run packages/desktop/src/daemon/desktop-managed.test.ts --bail=1`
- `npx vitest run packages/desktop/src/daemon/daemon-manager.test.ts --bail=1`

### Risk surface

The main risk is packaged-install path classification. Unit coverage pins macOS bundle containment, Windows case-insensitive install containment, Linux same-install containment, and npm Node rejection; a packaged desktop smoke on each OS is still the best final check before release.

There is one accepted downgrade edge: an old desktop app reading a new daemon lock will not see the legacy `desktopManaged` boolean and will treat that daemon as unmanaged, so it will not auto-restart it on version mismatch. This only happens when a user downgrades the desktop app while a newer daemon is still running, and it self-heals on the next manual daemon restart.

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (N/A, no UI changes)

### DIFF
--- a/packages/cli/src/commands/daemon/status.test.ts
+++ b/packages/cli/src/commands/daemon/status.test.ts
@@ -1,0 +1,99 @@
+import type { Command } from "commander";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, test, vi } from "vitest";
+import type { CommandOptions } from "../../output/index.js";
+import { runStatusCommand } from "./status.js";
+
+const mocks = vi.hoisted(() => ({
+  execCommand: vi.fn(async () => ({ stdout: process.execPath, stderr: "" })),
+  findExecutable: vi.fn(async () => null),
+  getOrCreateServerId: vi.fn(() => "srv_test"),
+  loadConfig: vi.fn(() => ({
+    listen: "127.0.0.1:6767",
+    relayEnabled: true,
+    relayEndpoint: "relay.paseo.sh:443",
+    relayUseTls: false,
+    relayPublicUseTls: false,
+  })),
+  resolvePaseoHome: vi.fn((env: NodeJS.ProcessEnv) => env.PASEO_HOME ?? "/tmp/paseo"),
+  spawnProcess: vi.fn(),
+}));
+
+vi.mock("@getpaseo/server", () => ({
+  execCommand: mocks.execCommand,
+  findExecutable: mocks.findExecutable,
+  getOrCreateServerId: mocks.getOrCreateServerId,
+  loadConfig: mocks.loadConfig,
+  resolvePaseoHome: mocks.resolvePaseoHome,
+  spawnProcess: mocks.spawnProcess,
+}));
+
+const tempRoots: string[] = [];
+
+async function createStatusHome(pidLock: Record<string, unknown>): Promise<string> {
+  const root = await mkdtemp(path.join(os.tmpdir(), "paseo-daemon-status-"));
+  tempRoots.push(root);
+  const home = path.join(root, ".paseo");
+  await mkdir(home, { recursive: true });
+  await writeFile(path.join(home, "paseo.pid"), JSON.stringify(pidLock));
+  return home;
+}
+
+async function readStatusJson(home: string): Promise<Record<string, unknown>> {
+  const previousPath = process.env.PATH;
+  try {
+    process.env.PATH = "";
+    const result = await runStatusCommand({ home } as CommandOptions, {} as Command);
+    return result.schema.serialize?.(result.data[0]) as Record<string, unknown>;
+  } finally {
+    if (previousPath === undefined) {
+      delete process.env.PATH;
+    } else {
+      process.env.PATH = previousPath;
+    }
+  }
+}
+
+describe("daemon status desktop management fields", () => {
+  afterEach(async () => {
+    await Promise.all(
+      tempRoots.splice(0).map((root) => rm(root, { recursive: true, force: true })),
+    );
+  });
+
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/daemon/status.test.ts
```diff
@@ -0,0 +1,99 @@
...
+}));
+
+vi.mock("@getpaseo/server", () => ({
+  execCommand: mocks.execCommand,
+  findExecutable: mocks.findExecutable,
+  getOrCreateServerId: mocks.getOrCreateServerId,
+  loadConfig: mocks.loadConfig,
+  resolvePaseoHome: mocks.resolvePaseoHome,
+  spawnProcess: mocks.spawnProcess,
+}));
```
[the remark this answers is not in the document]
boudra: Maintainer reviewed this PR and accepted this test shape for the focused status serialization coverage. Leaving it as-is here; a broader test seam can be handled separately.

packages/desktop/src/daemon/daemon-manager.ts
```diff
@@ -123,11 +124,36 @@ function logFilePath(): string {
...
+    const lock = JSON.parse(raw) as unknown;
+    return isRecord(lock) && deriveDesktopManagedFromPidLock(lock);
+  } catch {
+    return false;
+  }
+}
+
 export function isDesktopManagedDaemonRunningSync(): boolean {
   try {
     const raw = readFileSync(path.join(getPaseoHome(), "paseo.pid"), "utf-8");
```
boudra: multiple heplers here doig the same read => parse. there should be one owner here.

packages/cli/src/commands/daemon/status.ts
```diff
@@ -126,6 +127,7 @@ function toStatusRows(status: DaemonStatus): StatusRow[] {
     { key: "Owner", value: status.owner ?? "-" },
     { key: "Logs", value: status.logPath },
     { key: "Daemon Node", value: status.daemonNode },
+    { key: "Daemon Executable", value: status.daemonExecutablePath ?? "-" },
```
boudra: whats the diff between daemon node and daemon executable.

packages/desktop/src/daemon/daemon-manager.ts
boudra: Fixed in 887845418: `readPidLockRecord()` now owns the pid-file read/parse/shape-check path, and both desktop-managed helpers build on that single reader.

packages/cli/src/commands/daemon/status.ts
boudra: Fixed in 952e9e297: the human table label is now `PID Lock Executable`, which distinguishes the recorded pid-lock identity from `Daemon Node` (the live Node runtime resolved from the daemon PID). The JSON field stays additive/stable as `daemonExecutablePath`.

boudra:
Closing in favor of a much smaller fix. The goal was to remove stored management state from the pid lock by deriving it from the live daemon process — but live derivation isn't viable (the RPC path requires a reachable daemon; PID-based OS probing is async/fragile and can't serve the sync startup check). Given that, this PR effectively replaced a one-bit stored state with a stored path plus a cross-platform containment classifier and back-compat shims — more complexity for the same outcome. The actual bug (CLI-started bundled daemons losing managed status and never auto-upgrading) has a one-line fix under the boolean model: the desktop-bundled CLI shim sets PASEO_DESKTOP_MANAGED=1 itself, so every launcher of the bundled runner declares it. Follow-up PR will do that.

Issue #1904, opened 2026-07-06
bug: Error invoking remote method 'paseo:invoke': Error: Daemon failed to start: exit code 1


sujaysun101:
### What's broken

C:\Users\sujay\.paseo\daemon.log
The local server failed to start. If this keeps happening, please report the issue on GitHub and include the logs below.

### Steps to reproduce

Install the desktop app, and open it. 

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.103

### OS version

Windows 11 

### Agent provider

None

### Logs

```text

```

dnsouzadev:
same here, on my both computers work and personal

boudra:
Closing for now because the daemon startup error is missing. Please reopen with your current Paseo version and the daemon.log output from immediately after a failed launch.

PR #1961, opened 2026-07-09
Recover daemon startup from stale PID locks

### LINKED ISSUE #1950: bug: Error invoking remote method 'paseo:invoke': Error: Daemon failed to start: exit code 1
### What's broken

Image

### Steps to reproduce

mac

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.103

### OS version

_No response_

### Agent provider

_No response_

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

_No response_

boudra:
### Linked issue

Closes #1950
Refs #1904

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Desktop daemon startup can fail with `Daemon failed to start: exit code 1` when `~/.paseo/paseo.pid` is left behind and the recorded PID now belongs to a different live process. The launcher only saw that a process with that PID existed, so it refused to replace the stale singleton lock and exited before the daemon worker could start.

This verifies the lock owner by comparing the live process start time with the lock's `startedAt` timestamp before treating the PID as authoritative. If the PID has been reused, the stale lock is reclaimed and startup proceeds normally. A regression test covers the reused-PID stale-lock case.

### How did you verify it

- Reproduced the reported failure against an isolated `PASEO_HOME`: wrote `paseo.pid` with a live unrelated `sleep` PID and an old `startedAt`, then launched `packages/server/scripts/supervisor-entrypoint.ts`; before the fix it exited with status 1 and `Another Paseo daemon is already running`.
- Added the regression in `packages/server/src/server/pid-lock.test.ts`; confirmed it failed before the implementation change, then passed after.
- Re-ran the real startup path against the built supervisor entrypoint with the same stale lock; after the fix it replaced the stale lock, bound an ephemeral local port, and shut down cleanly on SIGTERM.
- `npm run build:server`
- `npx vitest run packages/server/src/server/pid-lock.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/server/src/server/pid-lock.test.ts
+++ b/packages/server/src/server/pid-lock.test.ts
@@ -1,4 +1,4 @@
-import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { describe, expect, test } from "vitest";
@@ -55,7 +55,6 @@ describe("pid-lock ownership", () => {
     const replacementOwnerPid = process.pid + 10_000;
 
     try {
-      await mkdir(paseoHome, { recursive: true });
       await writeFile(
         join(paseoHome, "paseo.pid"),
         JSON.stringify({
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of a smaller lock-heartbeat implementation.

PR #1971, opened 2026-07-10
fix(app): refresh checkout status when opening New Workspace

### LINKED ISSUE #1969: bug: New Workspace pre-fills a stale base branch because checkout-status is cached forever
## Problem

When you open the **New Workspace** screen for a project, the base-ref picker pre-fills the branch the project's checkout was on *the first time the screen fetched its status* — not the branch it's on now. Switch the underlying checkout back to `main` and the picker keeps offering the old branch as the base. New worktrees then branch off / compare against the wrong base, which also surfaces as "No changes vs `<stale-branch>`" in the diff panel.

## Repro steps

1. In a project checkout, check out a feature branch (e.g. `feature/my-branch`).
2. Open **New Workspace** for that project. The base-ref picker shows the feature branch (correct at this moment).
3. Without reloading the app, `git checkout main` in that same project checkout.
4. Open **New Workspace** again. The picker still shows the feature branch as the base, not `main`.

## Root cause

The New Workspace screen's checkout-status query is cached indefinitely with every refetch trigger disabled:

[packages/app/src/screens/new-workspace-screen.tsx#L1680-L1696](https://github.com/getpaseo/paseo/blob/main/packages/app/src/screens/new-workspace-screen.tsx#L1680-L1696)

```ts
const checkoutStatusQuery = useQuery({
  queryKey: ["checkout-status", selectedServerId, selectedSourceDirectory],
  queryFn: async () => { ... connectedClient.getCheckoutStatus(selectedSourceDirectory) },
  enabled: clientReady && hasSelectedSourceDirectory,
  staleTime: Infinity,
  refetchOnMount: false,
  refetchOnReconnect: false,
  refetchOnWindowFocus: false,
});

const currentBranch = checkoutStatusQuery.data?.currentBranch ?? null;
```

`getCheckoutStatus` reads the live branch from the daemon, but with `staleTime: Infinity` and all `refetchOn*` disabled the value is fetched once and frozen for the session. There is no invalidation of the `["checkout-status", …]` key on branch switch anywhere in the app. `currentBranch` then drives the picker default via `resolveCheckoutRequest` ([new-workspace-screen.tsx#L90-L101](https://github.com/getpaseo/paseo/blob/main/packages/app/src/screens/new-workspace-screen.tsx#L90-L101)) and the trigger label ([#L1748-L1751](https://github.com/getpaseo/paseo/blob/main/packages/app/src/screens/new-workspace-screen.tsx#L1748-L1751)).

## Fix

The New Workspace screen is opened deliberately and is short-lived, so it should reflect live git state each time. Set `refetchOnMount: true` with a short finite `staleTime` on that query so re-opening the screen re-reads the current branch. (Alternative: invalidate `["checkout-status", serverId, dir]` on branch switch, but that requires wiring at every branch-change site — the refetch-on-mount approach is simpler and sufficient.)

## Related

Distinct from #1873, which is the server-side `checkout-branch` path storing the branch as its own diff base. This one is the client pre-filling the wrong base branch from a stale cache.

ajm188:
## Problem

The New Workspace base-ref picker pre-fills a stale base branch. It reads the project's checkout status from a react-query cache pinned with `staleTime: Infinity` and every refetch trigger disabled, so the value is fetched once and frozen for the session. After switching branches in the underlying checkout, the picker keeps offering the old branch as the base — and new worktrees then branch off / compare against the wrong base (surfacing as "No changes vs `<stale-branch>`" in the diff panel).

Fixes #1969.

## Fix

`/new` is an expo-router route that remounts on each navigation, so refetching on mount is enough to reflect live git state. Set `refetchOnMount: true` with a short finite `staleTime` (5s, to suppress churn during rapid reopens) on the checkout-status query.

## Verification

- `npm run typecheck` (full workspace, via pre-commit hook) — clean
- `npm run lint` / `npm run format:check` on the changed file — clean
- Confirmed the route remounts on navigation (`packages/app/src/app/new.tsx`), so `refetchOnMount` fires each open.

### DIFF
--- a/packages/app/src/screens/new-workspace-screen.tsx
+++ b/packages/app/src/screens/new-workspace-screen.tsx
@@ -1687,8 +1687,12 @@ export function NewWorkspaceScreen({
       return connectedClient.getCheckoutStatus(selectedSourceDirectory);
     },
     enabled: clientReady && hasSelectedSourceDirectory,
-    staleTime: Infinity,
-    refetchOnMount: false,
+    // The New Workspace screen is opened deliberately and short-lived, so it must
+    // reflect the project's live checkout each time. A frozen cache (staleTime:
+    // Infinity) would pre-fill a stale base branch after the user switches branches
+    // in the underlying checkout. Refetch on mount with a short stale window.
+    staleTime: 5_000,
+    refetchOnMount: true,
     refetchOnReconnect: false,
     refetchOnWindowFocus: false,
   });


Strainy:
Just wanted to note that I've been using this for about a week and it's been working well for me.

boudra:
Closing in favor of the checkout-status implementation merged in [#2702](https://github.com/getpaseo/paseo/pull/2702).

PR #2007, opened 2026-07-11
Keep agent stream controls clickable under the scroll button

boudra:
### Linked issue

N/A — no matching open issue.

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

When the floating “Scroll to bottom” control was visible, its full-width animated container won browser hit testing across the entire same-height band. This moves full-width positioning and pass-through behavior to a plain container, while the animated child now wraps only the circular button. Elements beside the button remain clickable without changing its visual placement.

A focused Playwright regression places a real tool-call control beside the floating button and verifies that the browser resolves the click to the tool call outside the circle.

### How did you verify it

- Before the fix, the focused Playwright test failed with toolCallReceivesPointer false at a point outside the circle but inside its vertical band.
- After the fix, the same test passed against the isolated daemon and Desktop Chrome project.
- Inspected the green Playwright trace frame: the button remains centered with the same size and offset, with no visible layout shift or overlay artifact.
- npm run typecheck
- npm run lint
- npm run format

Risk surface: the component markup is shared with native platforms. The behavior is browser-specific and web is covered by Playwright; a quick iOS/Android placement smoke remains useful before merge.

### Checklist

- [x] npm run typecheck passes
- [x] npm run lint passes
- [x] npm run format ran (Biome)

### DIFF
--- a/packages/app/e2e/agent-stream-ui.spec.ts
+++ b/packages/app/e2e/agent-stream-ui.spec.ts
@@ -8,6 +8,7 @@ import {
 } from "./helpers/agent-stream";
 import {
   expectScrollStaysFixed,
+  clickToolCallBesideScrollToBottomButton,
   readScrollMetrics,
   scrollAgentChatToBottom,
   scrollChatAwayFromBottom,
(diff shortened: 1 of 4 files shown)


boudra:
@paseo-bot babysit this pr to green

PR #2058, opened 2026-07-14
fix(server): create autonomous turns for spontaneous ACP session updates

1254087415:
## Problem

When an ACP-based provider (e.g. kimi-code) handles background-agent completion notifications inside an existing session, the content it produces is never surfaced in Paseo's timeline. The foreground turn has already returned from the provider's point of view, so the subsequent spontaneous `session/update` messages have no active turn to attach to.

This is the root cause behind behavior where an agent can say "I'll summarize when the background agents finish" but the summary never appears in the Paseo UI.

## Solution

Mirror the autonomous-turn mechanism already used by the Claude provider:

1. When a spontaneous `sessionUpdate` arrives and there is no active foreground turn, start an autonomous turn and emit `turn_started`.
2. Tag all timeline events emitted during that autonomous turn with the autonomous turn id.
3. Reset a short inactivity timer on every autonomous update; when it expires, emit `turn_completed`.
4. Before starting a new foreground turn, complete any active autonomous turn.

## Changes

- `packages/server/src/server/agent/providers/acp-agent.ts`
  - Add `autonomousTurnId`, `autonomousTurnTimer`, and `AUTONOMOUS_TURN_TIMEOUT_MS`
  - Add `startAutonomousTurn`, `completeAutonomousTurn`, and `resetAutonomousTurnTimer`
  - `sessionUpdate` now starts/resets an autonomous turn when no foreground turn is active
  - `wrapTimeline` and `pushEvent` now fall back to the autonomous turn id
  - `startTurn` completes any pending autonomous turn first

- `packages/server/src/server/agent/providers/acp-agent.test.ts`
  - Add test for autonomous turn lifecycle on spontaneous updates
  - Add test that `startTurn` completes an autonomous turn before starting a foreground turn

## Testing

```bash
cd packages/server
vitest run src/server/agent/providers/acp-agent.test.ts
```

All 77 tests pass, including the 2 new ones.

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -2208,6 +2208,90 @@ describe("ACPAgentSession", () => {
     expect(assistantMessages[2].messageId).not.toBe(assistantMessages[0].messageId);
   });
 
+  test("starts an autonomous turn for spontaneous session updates outside a foreground turn", async () => {
+    vi.useFakeTimers({ shouldAdvanceTime: true });
+    const session = createSession();
+    asInternals<ACPSessionInternals>(session).sessionId = "session-1";
+
+    const events: AgentStreamEvent[] = [];
+    session.subscribe((event) => {
+      events.push(event);
+    });
+
+    await session.sessionUpdate({
+      sessionId: "session-1",
+      update: {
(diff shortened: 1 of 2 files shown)


boudra:
@paseo-bot address greptile feedback and babys it to green

PR #2065, opened 2026-07-14
fix(server): speed up OMP/Pi import session listing

cexll:
## Summary
- Stop recursive import scans of nested Pi/OMP subagent transcripts (`<session>/<subagent>.jsonl`), which can turn large OMP histories (tens of thousands of files / multi-GB) into a permanent Import UI loading state.
- Rank parent session files by mtime and only fully parse a bounded recent candidate set for the requested limit.
- Accept OMP session formats that put a `title` record before the `session` header, and OMP `model_change` payloads that store a combined `model` string.

## Test plan
- [x] `npx vitest run packages/server/src/server/agent/providers/pi/session-descriptor.test.ts` (6/6 pass)
- [x] Local repro against a large `~/.omp/agent/sessions` tree: listing 20 sessions dropped from multi-minute hang to ~tens of ms after the patch
- [ ] UI: open Import → OMP and confirm recent parent sessions appear without infinite loading
- [ ] UI: import one OMP session that starts with a `title` line and confirm model metadata still hydrates

## Notes
- Parent sessions remain at `<sessionsDir>/<cwdDir>/<session>.jsonl`.
- Nested subagent transcripts remain importable only through their parent session, which matches the intended import surface.

### DIFF
--- a/packages/server/src/server/agent/providers/pi/session-descriptor.test.ts
+++ b/packages/server/src/server/agent/providers/pi/session-descriptor.test.ts
@@ -13,6 +13,45 @@ async function writeSession(root: string, lines: unknown[]): Promise<string> {
   return filePath;
 }
 
+async function writeNamedSession(
+  root: string,
+  cwdDir: string,
+  fileName: string,
+  lines: unknown[],
+  nestedDir?: string,
+): Promise<string> {
+  const sessionsDir = nestedDir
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the provider-owned import implementation merged in [#2067](https://github.com/getpaseo/paseo/pull/2067).

PR #2084, opened 2026-07-15
feat(app): Insert slash commands from agent output into the composer

cleiter:
### Linked issue

N/A

### Type of change

- [ ] Bug fix
- [x] New feature (trivial - no prior issue + design alignment)
- [ ] Docs

### What does this PR do

**Why**

Agents regularly recommend the next skill or command to run (`/code-review`,
`/simplify`, …). The suggestion is right there on screen, but the only way to act
on it is to retype it by hand.

That hurts most in touch- and voice-driven workflows: on a tablet, acting on a
suggested skill means opening the on-screen keyboard to type a command I can
already see. On desktop it's milder, but retyping something that's visible in
front of me is still friction.

**The Change**

When an agent's message contains an inline-code token like `/code-review`, it
was rendered as a file link and clicking it tried to open the absolute path
`/code-review` — a file that doesn't exist. Leading-slash tokens are classified as
absolute paths by the assistant-file-link layer, so any `/command` in agent
output became a dead file link.

This makes a `/command` token clickable as a command instead: when the token
matches a known slash command / skill for that agent, it renders as a chip that
**appends the command to the composer and focuses the input**, so you can add
args and send. Tokens that aren't known commands keep today's file-path
behavior.

How it works:

- The command list comes from the existing `listCommands` query (shared cache
  with the composer — no extra network), gated to live agent streams that have a
  composer to insert into.
- Detection runs *before* the file-path branch in the inline-code markdown rule,
  so a known command wins over the path interpretation.
- The chip appends via a small `ComposerInsertProvider` that mirrors the existing
  rewind-restore seam, and focuses the input through the composer's existing
  `onFocusInput` handle.

Client-only; no protocol change. Inline-code only (plain-text `/foo` isn't
linkified today and is untouched).

### How did you verify it

Automated (run locally, all green):

- `npm run typecheck`
- `npm run lint`
- `npm run format` (Biome)
- Unit tests — 19 passing across:
  - `parseSlashCommandToken` / `appendSlashCommandToken` (token parsing + append
    semantics: `/autoplan` → command, `/foo/bar` and `/foo.ts` → not a command,
    whitespace handling)
  - `useAssistantFileLinkActions` slash-command matching (known vs unknown vs no
    matcher)
  - `ComposerInsertProvider` (appends + focuses the registered composer; no-op
    when nothing registered / after unregister)

Manual QA (video attached):

https://github.com/user-attachments/assets/9d20ee84-f80b-4e26-a66f-393eb99b2d08

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/agent-stream/view.tsx
+++ b/packages/app/src/agent-stream/view.tsx
@@ -52,6 +52,8 @@ import type { AgentScreenAgent } from "@/hooks/use-agent-screen-state-machine";
 import { useSessionStore } from "@/stores/session-store";
 import { useFileExplorerActions } from "@/hooks/use-file-explorer-actions";
 import { useLoadOlderAgentHistory } from "@/hooks/use-load-older-agent-history";
+import { useAgentCommandsQuery } from "@/hooks/use-agent-commands-query";
+import { useComposerInsert } from "@/components/composer-insert";
 import { useSettings } from "@/hooks/use-settings";
 import type { ToastApi } from "@/components/toast-host";
 import { returnToTimelineTail } from "./timeline-tail-navigation";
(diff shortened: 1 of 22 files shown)


cleiter:
Added "Why" section to description.

boudra:

PR #2118, opened 2026-07-16
fix(server): stop stalled ACP prompts from wedging the next turn

michaelneale:
## Problem

Interrupting an ACP-backed agent (goose, generic ACP, Copilot, Kiro, Pi, …) sometimes wedges the session: the **next** prompt fails with

> `[System Error] A foreground turn is already active`

### Root cause

`ACPAgentSession` tracks `activeForegroundTurnId` and only clears it in `finishTurn()`, which runs when the agent's `session/prompt` call resolves with a `stopReason`. `interrupt()` fired `session/cancel` and awaited nothing else.

If the backend is slow to return its final `stopReason` after a cancel — or never returns it at all (e.g. stalled in model inference) — the turn stays `"active"`. The next `startTurn()` then throws `"A foreground turn is already active"`, which the manager surfaces as the `[System Error]` banner. In the worst case (response never arrives) the session is wedged until it's restarted.

This is timing-dependent, which is why it only happens *sometimes* — if the backend returns `Cancelled` within the manager's grace window everything looks fine.

## Fix

Mirror the pattern OpenCode already uses (`opencode-agent.ts`): cancel the backend, but **finalize the turn locally** so the session is immediately ready for the next prompt. Because the change lives in the shared `ACPAgentSession` base, every ACP-based provider benefits with no per-provider code.

- **`interrupt()`** — after firing `session/cancel`, locally emit `turn_canceled` and clear the turn instead of waiting for the prompt response.
- **`finishTurn()`** — make idempotent: only finalize the currently-active turn, so a late/duplicate terminal (the eventual, or never-arriving, prompt response) is a harmless no-op.
- **`handlePromptResponse()`** — bail if the turn is no longer active, so a stale prompt response can't leak usage into a newer turn.

Autonomous turns are unaffected: they finalize via `completeAutonomousTurn()` (direct `pushEvent`), not `finishTurn()`.

## Tests

Two new tests in `acp-agent.test.ts`:
- `interrupt` clears the foreground turn even when `session/prompt` never settles (and a fresh turn is accepted immediately).
- a late `session/prompt` response after `interrupt` does not re-finalize the turn or clobber a newer one.

Full `acp-agent.test.ts` passes (79/79).

## Notes

- No protocol/schema changes; fully backward-compatible.
- No agent-side (goose etc.) changes required — this is a host-side robustness fix so a slow/stalled backend can't wedge a session.

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -86,7 +86,10 @@ describe("buildACPClientCapabilities", () => {
 
 interface ACPSessionInternals {
   sessionId: string | null;
-  connection: { prompt: (...args: unknown[]) => Promise<PromptResponse> };
+  connection: {
+    prompt: (...args: unknown[]) => Promise<PromptResponse>;
+    cancel?: (...args: unknown[]) => Promise<void>;
+  };
   activeForegroundTurnId: string | null;
   configOptions: SessionConfigOption[];
   translateSessionUpdate(update: SessionUpdate): AgentStreamEvent[];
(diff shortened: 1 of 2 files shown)


michaelneale:
Closing — opened prematurely; will re-open when ready.

PR #2131, opened 2026-07-16
Fix OMP imports with leading title records

### LINKED ISSUE #2006: OMP import misses sessions with a leading title record
_co-authored by gpt 5.6 codex_

## Summary

OMP session import can return no sessions when OMP JSONL files contain a leading `title` record before the `session` record.

## Triage notes

Current `packages/server/src/server/agent/providers/pi/session-descriptor.ts` reads only the first JSONL line when building an importable session descriptor:

```ts
const firstLine = await readFirstLine(filePath);
const header = parseSessionHeader(firstLine);
if (!header) return null;
```

`parseSessionHeader` then requires that first parsed record to be the session header:

```ts
if (!entry || entry.type !== "session") return null;
```

Recent OMP session files may be shaped like:

```jsonl
{"type":"title","v":1,"title":"...","updatedAt":"..."}
{"type":"session","version":...,"id":"...","timestamp":"...","cwd":"..."}
```

Those files are valid enough for resume/import metadata, but `readPiSessionDescriptor` returns `null` before scanning the later `session` row. If all OMP session files have this shape, the import sheet lists zero OMP sessions.

## Expected behavior

Session import should tolerate leading metadata records and find the first valid `type: "session"` record near the start of the JSONL file.

## Suggested fix

Replace the single-line header read with a bounded head scan that skips non-session metadata rows (for example `title`) until it finds a valid `session` record. The existing head/tail scanning can still derive title, prompt previews, model, and thinking level.

## Privacy note

This report intentionally uses a reduced synthetic JSONL shape and does not include real session paths, prompts, cwd values, or user data.

raeperd:
### Linked issue

Closes #2006

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

OMP session files can start with a `title` metadata record before the `session` record. The import parser previously checked only the first line, so it rejected those files and omitted them from the import list.

This changes the parser to scan the existing bounded 64 KiB file prefix for the first valid session header. The rest of the import path remains unchanged, including title, prompt, model, and thinking-level extraction.

### How did you verify it

- Added a regression test using a real temporary JSONL file with `title` before `session`.
- Before the implementation, `npx vitest run packages/server/src/server/agent/providers/pi/session-descriptor.test.ts --bail=1` failed because the import result was `[]`.
- After the implementation, the same command passed all 4 tests.
- `npm run build:server` passed.
- `npm run typecheck` passed for all workspaces.
- `npm run lint` passed with 0 warnings and 0 errors.
- `npm run format` completed successfully.

No UI behavior changed, so screenshots are not applicable.

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (not applicable; server-only change)

### DIFF
--- a/packages/server/src/server/agent/providers/pi/session-descriptor.test.ts
+++ b/packages/server/src/server/agent/providers/pi/session-descriptor.test.ts
@@ -13,6 +13,51 @@ async function writeSession(root: string, lines: unknown[]): Promise<string> {
   return filePath;
 }
 
+test("lists OMP sessions with a title record before the session header", async () => {
+  const root = await mkdtemp(path.join(tmpdir(), "paseo-omp-leading-title-"));
+  const cwd = path.join(root, "repo");
+  const sessionFile = await writeSession(root, [
+    {
+      type: "title",
+      v: 1,
+      title: "Leading metadata title",
+      updatedAt: "2026-06-09T00:00:00.000Z",
+    },
+    {
+      type: "session",
+      version: 3,
+      id: "session-with-leading-title",
+      timestamp: "2026-06-09T00:00:01.000Z",
+      cwd,
+    },
+    {
+      type: "message",
+      id: "user-1",
+      timestamp: "2026-06-09T00:00:02.000Z",
+      message: { role: "user", content: "import this session" },
+    },
+    {
+      type: "session_info",
+      id: "info-1",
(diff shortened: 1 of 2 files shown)


boudra:
thanks for the pull request.

this is superseded by #2067, which includes the same OMP title-first import fix as part of the new OMP provider. closing.

PR #2142, opened 2026-07-16
feat(server): attribute agent-to-agent messages with a sender envelope

OnCloud125252:
> **Stacked on #2136 - review only the commits after that branch point.**
> This branch (`agent-sender-envelope`) is cut off `agent-spawn-auto-context` (PR #2136) and reuses that PR's daemon-side timeline-projection machinery. Until #2136 merges, the diff against `main` will also show #2136's commit; the only new commit here is `feat(server): attribute agent-to-agent messages with a sender envelope`. Merge #2136 first, then this.

Round 2 of porting claude-peers' communication strengths into Paseo: the **sender-identity envelope** for `send_agent_prompt`.

## Problem

When agent A calls `send_agent_prompt` targeting agent B, the handler passed the prompt verbatim. B could not tell the message came from another agent (vs a human, a schedule, or a finish notification) and had no idea whom to reply to - even though the handler already arms `setupFinishNotification` back to A by default, so B's final message is auto-relayed to A. The mechanical reply path worked; the receiver just did not know about it. Same disease cured in #2136 for spawn context; same fix (tell the receiver, deterministically, daemon-side).

## What changed

**Injection (daemon-side, agent-to-agent only).** In the `send_agent_prompt` handler, the prompt is wrapped **only when `callerAgentId` is present**. Human/app sends and other system paths (chat mentions, schedule fires, notify-on-finish) are untouched.

**Envelope format - a distinct attributed tag, not `<paseo-system>`.**
```
<paseo-agent-message from="<senderId>" from_title="<title>">
<original prompt>
</paseo-agent-message>

<reply-contract line>
```
Chosen over reusing the `<paseo-system>` header convention deliberately: a full `<paseo-system>` match would make `isSystemInjectedEnvelope` true, which (per #2136) **hides the message from the receiver's timeline** and is also the shape that gates system-injected handling - exactly wrong for attribution a human wants to see. A distinct tag never matches `SYSTEM_ENVELOPE_PATTERN`, so it is never hidden and cannot perturb the archived-target unarchive path (`send_agent_prompt` calls `sendPromptToAgent` with `unarchive` defaulting to true). Attribute values are XML-escaped so arbitrary titles cannot break out of the tag.

**Conditional reply contract** (mirrors #2136's notify wording), gated on the same `shouldNotifyOnFinish = callerAgentId && notifyOnFinish && background` the handler already computes:
- notify: "When you finish this turn and go idle, your last assistant message is automatically delivered back to agent `<id>` as your reply - make it complete and self-contained."
- no-notify: "Your reply is not automatically delivered; reach the sender via send_agent_prompt with agentId `<id>` if a response is needed."

**Timeline display** - unlike spawn context (stripped entirely), sender attribution is useful, so `projectAgentMessageForDisplay` rewrites the turn to `Message from agent <id> (<title>):` + the original prompt, dropping the trailing reply boilerplate. It runs through the same daemon-side `projectTimelineItemForDisplay` choke points from #2136, so **every client, including old ones, gets the readable form**. The receiving model still gets the full envelope in the dispatched prompt.

**Layer 1 instructions** (`composeAgentMcpInstructions`) now teach every agent to recognize `<paseo-agent-message>`, read the `from` id, and reply via `send_agent_prompt` (or rely on the auto relay when the envelope says so).

## Tests

- `agent-spawn-context.test.ts` - envelope compose (auto/manual reply, no-title, XML escaping, not-a-system-envelope) and `projectAgentMessageForDisplay` (header rewrite, boilerplate dropped, multi-paragraph, passthrough); Layer 1 teaching assertion.
- `mcp-server.test.ts` - handler wraps agent-to-agent sends with the correct reply variant, and leaves human/app sends untouched.
- `agent-manager.test.ts` - a `<paseo-agent-message>` turn renders as a readable sender header in the real timeline.

Lint, format, and full-workspace typecheck pass via the pre-commit hook.

## Out of scope (as specified)

Non-destructive delivery to busy agents (the replaceRunning interruption problem) and set_summary-style presence. Docs: `docs/agent-lifecycle.md` gains an "Agent-to-agent message envelope" section.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -63,6 +63,29 @@ The Layer 2 block rides on the child's first user message, but the timeline must
 `displayTextForUserMessage` (`agent-prompt.ts`) extends the existing "hide a pure `<paseo-system>` envelope" rule: a message that begins with a system envelope and then has real content has the leading envelope stripped for display, while the provider still receives the full text.
 The four timeline choke points in `agent-manager.ts` route through `projectTimelineItemForDisplay`, so both live turns and history replay show the clean parent prompt.
 
+## Agent-to-agent message envelope
+
+When one agent calls `send_agent_prompt` targeting another, the receiver used to get the text verbatim - unable to tell it came from an agent (vs a human, a schedule, or a finish notification) or whom to reply to.
+The daemon now wraps agent-initiated sends in a sender-identity envelope, composed by `buildAgentMessageEnvelope` in `agent-spawn-context.ts` and applied in the `send_agent_prompt` handler **only when `callerAgentId` is present**.
+Human/app sends and other system paths (chat mentions, schedule fires, notify-on-finish) are untouched.
+
+**Envelope format.**
+The prompt is wrapped in an attributed tag: `<paseo-agent-message from="<senderId>" from_title="<title>">\n<prompt>\n</paseo-agent-message>`, followed by a trailing reply-contract line.
+This is deliberately **not** a `<paseo-system>` envelope: it must not match `SYSTEM_ENVELOPE_PATTERN` / `isSystemInjectedEnvelope`, because a full match would hide the message from the receiver's timeline (round-1 behavior) - and sender attribution is exactly what a human viewing the receiver wants to see.
+It also keeps the archived-target delivery path intact: `send_agent_prompt` calls `sendPromptToAgent` with `unarchive` defaulting to true, and because the message is not a system envelope it cannot accidentally flip that gate.
+
+**Reply contract (conditional, like the spawn-context wording).**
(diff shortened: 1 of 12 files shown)


OnCloud125252:
Related: this PR is **stacked on #2136** and cannot land before it. It reuses #2136's daemon-side timeline-projection choke points (`projectTimelineItemForDisplay`) and its `agent-spawn-context.ts` module. Until #2136 merges, the diff against `main` here also shows #2136's commit - the only new commit on this branch is the sender-envelope one. Review order: #2136 first, then this.

OnCloud125252:
Folded into #2136: the sender-envelope commit (2525601c4) was fast-forwarded onto #2136's branch `agent-spawn-auto-context`, so #2136 now carries both the spawn-context and sender-envelope work as one PR. Closing this in favor of #2136. No changes lost - the branch `agent-sender-envelope` still points at the same commit.

PR #2152, opened 2026-07-17
fix(app): update project name immediately after rename

### LINKED ISSUE #2134: bug: renaming project doesn't rename
### What's broken

I'm trying to rename a project from the settings, it says it has renamed successfully, but the original project name is still there. 

### Steps to reproduce

Open App
Go to Settings
Open Projects
Rename a Project
Save

### Where did this happen

Desktop (Electron)

### Paseo version

v0.1.109

### OS version

macOS 27 Beta 3

### Agent provider

_No response_

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

_No response_

mvanhorn:
### Linked issue

Closes #2134

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Renaming a project from Settings -> Projects showed a "renamed" success toast, but the displayed name did not change until the app was restarted.

Root cause: the rename mutation's `onSuccess` only ran `queryClient.invalidateQueries({ queryKey: ["projects"] })`, but the displayed project name is not served by any React Query cache. `useProjects()` derives project names from the zustand `session-store` (via `buildProjects()`), so invalidating that query key refreshes nothing the screen reads. The stale name only cleared when a fresh daemon directory sync repopulated the store on restart.

Fix: add a `applyProjectCustomName(serverId, projectKey, customName)` action to the session-store that updates the matching workspace and empty-project descriptors, and call it from the rename mutation. It updates both `projectCustomName` and `projectDisplayName` to mirror exactly what a real directory sync sends (the daemon computes `projectDisplayName = customName ?? repoDisplayName`), so `useProjects()` re-derives the new name immediately; clearing the custom name reverts to the repository-derived default. The action preserves referential stability (returns the previous state when no descriptor matches) so unrelated projects don't re-render.

The rename host/project is snapshotted in the mutation's `onMutate` context, so switching hosts while the rename RPC is in flight cannot redirect the update to the wrong session.

### How did you verify it

Verification is through automated tests at the state layer, which is exactly where the bug lives (the UI stays correct once the store re-derives). I did not capture on-device screenshots.

- Added `applyProjectCustomName` store tests in `session-store.test.ts`: sets the custom name and display name on every matching workspace and empty project; reverts to the repository-derived name when cleared; preserves state identity (no re-render) when no descriptor matches.
- Added an end-to-end derivation test in `use-projects.test.ts` asserting `deriveProjectsFromReplica` produces the renamed `projectName` from an updated descriptor (proving the store update flows through `buildProjects()` to the value the screen renders).
- Before/after: with the store action stubbed to a no-op, the two store tests fail (`expected null to be 'My Project'`); with the action in place they pass. Full run: 22 tests pass across both files.
- `npm run typecheck`, `npm run lint`, and `npm run format` all pass.

### Checklist

- [x] `npm run lint` passes

### Known limitation

Clearing a custom name on a project that currently has no active workspaces reconstructs the repository-derived base name from the project key (the app's existing `projectDisplayNameFromProjectId` fallback). For a non-GitHub remote this can differ slightly from the daemon's persisted name until the next directory sync or reconnect. Projects with active workspaces receive an authoritative `workspace_update` from the daemon and self-correct immediately. Setting a name is always exact.

---

AI was used for assistance. I read the repo docs (root and app guides), understand the change, and ran the verification above myself.

### DIFF
--- a/packages/app/src/hooks/use-projects.test.ts
+++ b/packages/app/src/hooks/use-projects.test.ts
@@ -86,6 +86,31 @@ function emptyProject(input: {
 }
 
 describe("deriveProjectsFromReplica", () => {
+  it("derives the custom project name from a renamed workspace descriptor", () => {
+    const base = workspace({
+      id: "main",
+      projectKey: "remote:github.com/acme/app",
+      projectName: "acme/app",
+      cwd: "/repo/app",
+      remoteUrl: "https://github.com/acme/app.git",
+    });
+    const renamed: WorkspaceDescriptor = {
+      ...base,
+      projectCustomName: "My Project",
+      projectDisplayName: "My Project",
+    };
+
+    const result = deriveProjectsFromReplica({
+      replicas: [
+        { serverId: "local", serverName: "Local", workspaces: [renamed], emptyProjects: [] },
+      ],
+      runtimeStates: [runtimeState({ serverId: "local", isOnline: true })],
+    });
+
+    expect(result.projects[0]?.projectName).toBe("My Project");
+    expect(result.projects[0]?.projectCustomName).toBe("My Project");
+  });
+
   it("aggregates store workspaces and empty projects sorted by display name", () => {
     const replicas: ProjectHostReplica[] = [
       {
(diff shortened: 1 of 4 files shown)


boudra:
thanks for the pull request.

#2098 adds the project-level update path that covers renames for projects with no workspaces, so this patch is superseded. i'm closing the pull request.

PR #2153, opened 2026-07-17
feat(app): group the Changes flat file list by directory

cleiter:
### Linked issue

None — opening directly, since the screenshots make the case better than a
description would. Happy to move it to a feature request if you'd rather
settle the shape first.

**Your call:** flat is the default view mode, so this changes the Changes
list everyone sees out of the box, and there's no way back to the old rows.

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Docs

### What does this PR do

Groups the flat file list by directory, and gives every row its file-type
icon.

Before: one row per file — `filename` plus a muted inline directory suffix,
no icon (icons landed in #1918, but only in tree mode). The directory
string repeated on every row, and it truncated the filename first in a
narrow sidebar.

After: each directory appears once as a path label with its files beneath
it. Files sharing a directory share a label; root files get none.

The shape is taken from GitLab's merge-request file list, used deliberately
as a best-practice reference — the same reasoning as the tree's single-child
compression in #1918 following VS Code and GitHub: reach for a pattern
people already read fluently rather than invent one. The grouping only,
though. Paseo's `+/-` stats, New/Deleted badges and end-ellipsis truncation
stay as they are, not GitLab's status glyphs and middle-truncated paths.

**Trade-off: it's taller.** Every group adds a label row — the diff in the
screenshots goes from 5 rows to 8. The label is cheap (smaller text, top
padding only) but not free, and it narrows the density gap with tree mode,
which renders the same diff in 9 rows and can at least collapse. What it
buys is width for the filename.

That diff is close to the worst case for this, by the way: three groups over
five files, two of them single-file. The label amortizes where more files
share a directory — the three `commands/` files here pay one label between
them.

I considered keeping the directory inline for single-file groups to save
the label (the screenshots have two), and dropped it: GitLab doesn't do it
either — its flat list is mostly single-file groups and they all get labels
— and rows would change shape live as an agent adds a second file to a
directory.

Client-only, built from the existing `ParsedDiffFile` paths — no protocol
or server change, tree mode untouched, icons reuse the existing
`getFileIconSvg`. The one non-obvious bit: labels get their own measured
height feeding `getItemLayout`, and sticky indices stay file-headers-only
so a label can't enter the sticky stack.

### How did you verify it

- `npm run typecheck`, `npm run lint`, `npm run format` clean. All 224
  tests across the 21 `git/` test files pass. Three cases in
  `diff-flat-items.test.ts` cover the new grouping: one shared label for
  same-directory files, no label at root, and the sticky index still
  landing on the expanded file header. Existing tree-mode tests unchanged
  and passing — that's what pins "tree mode is unaffected".
- Live in the Electron desktop app on this branch, against a real diff:
  the three groups label correctly, `commands/` collects its three files
  under one label, icons resolve, and `+/-` stats stay right-aligned.
  Toggled to tree and back — unchanged.

Before:

image

After:

image

Tree, for comparison (unchanged):

image

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/e2e/diff-row-alignment.spec.ts
+++ b/packages/app/e2e/diff-row-alignment.spec.ts
@@ -311,9 +311,14 @@ async function useUnwrappedDiffLines(page: Page): Promise<void> {
 }
 
 async function expectFlatFileList(page: Page): Promise<void> {
+  // No collapsible folder rows — those are tree mode only.
   await expect(page.locator('[data-testid^="diff-folder-"]')).toHaveCount(0);
+  // The flat list groups files under a directory-path label instead of repeating
+  // the directory inline on every row, so `src` now sits above the file rather
+  // than beside it. This assertion previously required the inline suffix.
+  await expect(page.getByTestId("diff-dir-label-src")).toContainText("src/");
   await expect(page.getByTestId("diff-file-0")).toContainText("use-mounted-tab-set.ts");
-  await expect(page.getByTestId("diff-file-0")).toContainText("src");
+  await expect(page.getByTestId("diff-file-0")).not.toContainText("src");
 }
 
 async function expectDiffCodeFontSize(page: Page, fontSize: number): Promise<void> {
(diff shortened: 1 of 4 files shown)


boudra:
i think flat should stay flat, tree view exists if anyone wants to see it that way

closing this for now, thanks

PR #2158, opened 2026-07-17
fix(server): only type scripts into a shell that is at its prompt

cleiter:
### Linked issue

N/A - Discussed in Discord

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

`paseo.json` scripts are **typed into an interactive PTY**, and the check for
"is this shell ready for keystrokes" was: has it printed anything, or has 1.5s
passed. Neither is evidence. A shell that prints and then blocks on `read`
satisfies both, and gets a command typed into whatever is reading.

oh-my-zsh does exactly that on startup:

```
# what Paseo sent:
[oh-my-zsh] Would you like to update? [Y/n] PASEO_DEV_MANAGED_HOME=1 … ./scripts/dev-daemon.sh
# what actually ran:
$ ASEO_DEV_MANAGED_HOME=1 … ./scripts/dev-daemon.sh
```

The `P` answered the prompt; the rest replayed onto the command line. The daemon
came up with a silently wrong `PASEO_DEV_MANAGED_HOME`. It is luck that the
first character was `P` — a leading `y` would have answered "yes, update
oh-my-zsh" instead. Nothing errors; the script just runs wrong.

Four commits, each reviewable on its own:

**1. Don't type until the shell's line editor owns stdin.** The zsh integration
emits three nonce-tagged OSC 633 markers: `R` from `zle-line-init` (line editor
has the line — safe to type), `C` from `preexec` (a command took the foreground
— not safe), and readiness is gated on them. `zle-line-init`, not `precmd`:
`precmd` runs before prompt expansion and before any hook a later `.zshrc` adds,
so a `read` there still eats the command. At-prompt is tracked as current state,
cleared on `C` and on exit — which is also what makes reusing a script terminal
safe.

**2. Let the shell say whether it can report readiness at all.** A shell that
never reports is either *blocked* (must not type) or has *no integration* — zsh
< 5.3 has no `add-zle-hook-widget`, a `.zshrc` can clobber `ZDOTDIR` — where
never typing would break every script. Both look like silence. So the `.zshenv`
wrapper emits `633;I` from its first lines, before any user rc file runs:
silence *before* it means no integration (fall back to the old heuristic, which
is what those shells have today); silence *after* it means blocked (wait, then
fail). It only announces when it can deliver (`is-at-least 5.3`).

**3. Ask the worker, don't guess from a stale copy.** Prompt state lives in the
terminal worker; the daemon holds a copy that lags one IPC hop, so "hasn't
announced" and "announce in flight" look identical there. Before falling back it
re-reads via `getPromptState`, and the command goes through `sendInputIfAtPrompt`,
which checks-and-writes in the same tick against the live session — so a
foreground command taking stdin between the check and the write can't get the
keystrokes.

**4. Focus a script's terminal instantly, like a plain terminal.** A plain
terminal's tab opens the instant its pty exists. A script's did not: the RPC
response was deferred until the command was typed (up to the readiness timeout
on a blocked shell), and the tab-focus rode on that response — so a blocked
start showed nothing, then an error telling you to answer a prompt you couldn't
see. The start now hands the terminal over the moment it is registered
(`onTerminalReady`) and runs the readiness wait + send detached: the tab focuses
immediately in every case, the connection's message loop no longer blocks, and a
blocked shell is no longer an error — its terminal is on screen showing the
prompt, and the still-running wait types the command once you answer it. Only a
failure before any terminal exists (bad config, already running) reports an
error.

**When it can't type, it doesn't.** On a readiness timeout the terminal stays
open; plain scripts keep their runtime entry so re-running reuses that same
shell. Rationale, the OSC protocol, and the worker/manager ordering rules are in
`docs/terminal-readiness.md`.

**Known gaps, deliberately not closed here:**

- **bash / fish / custom shells** keep the old heuristic and stay exposed. There
  is no reliable shell-agnostic PTY readiness signal, so each shell needs its own
  integration. The nonce env var is injected for every shell so a bash/fish hook
  can adopt the same markers without a protocol change. Happy to follow up.
- **ZLE widgets that read stdin** (`fzf` on Ctrl-R): `preexec` doesn't fire, so
  at-prompt stays true while fzf owns stdin. zsh exposes no hook; the only sound
  signal is the pty's foreground process group, which node-pty doesn't expose.
  Fail-open, but it needs someone typing in a script terminal at the instant of
  injection.
- An announce whose hook then fails to register yields `I` with no `R` and fails
  the script at the readiness timeout. Fail-closed, never corruption.

### How did you verify it

**Reproduced the original bug first**, then confirmed the fix, using a real zsh
whose `.zshrc` prints and blocks on `read -k 1` (standing in for oh-my-zsh's
update prompt), driving the real `spawnWorkspaceScript` through the real worker
terminal manager. The script under test writes `$PASEO_PROBE_VALUE` to a file,
so what the shell *parsed* is visible, not just what it printed.

*Before (and still, on bash — the known gap above):*

```
sent:      PASEO_PROBE_VALUE=intact echo probe-done
terminal:  [omz] Would you like to update? [Y/n] PASEO_PROBE_VALUE=intact echo probe-done
           $ ASEO_PROBE_VALUE=intact echo probe-done      ← the P was eaten
```

*After, on zsh:*

```
run 1 (startup blocked):  refused — nothing typed; terminal stays open at the prompt
answer "n", re-run:       reused the same terminal
                          PASEO_PROBE_VALUE = "intact"     ← command arrived whole
```

Also verified live in the desktop dev app: started the `daemon` script with the
omz prompt armed, saw its tab focus immediately showing the `[Y/n]` prompt,
answered it in the terminal, and the command ran on its own.

**Automated — all green locally**, plus `npm run typecheck`, `npm run lint`,
`npm run format` clean.

- `terminal.test.ts` — real terminals: the marker sets at-prompt only when the
  nonce matches (a foreign nonce is ignored), `C` clears it, and — the one that
  pins the bug — a real zsh blocked on a startup `read` reports **not** at-prompt
  while its output is already on screen.
- `worker-terminal-manager.test.ts` — prompt state crosses the worker boundary;
  re-registration can't roll live state back to the spawn-time snapshot; exit
  forces at-prompt false; late markers from a dead shell are ignored.
- `worktree-bootstrap.test.ts` — nothing is typed until at-prompt; a
  never-announcing zsh falls back and still runs (the zsh-<5.3 case); an
  announced-but-blocked shell fails instead of typing; a shell that loses its
  prompt between the check and the write types nothing; an announce still in
  flight to the parent does not trigger the fallback.
- `workspace-scripts-service.test.ts` — the start responds with the terminal as
  soon as it exists; a blocked shell still hands the terminal over with **no**
  error (exactly one response, not a second error over it); only a
  pre-terminal failure reports an error.

Two of those I checked the honest way — reverting the fix makes the test fail and
restoring it passes — so they pin the behavior, not the implementation: the
in-flight-announce fallback, and the single-response guarantee for a blocked
start.

Before:

image

Terminal recording of the live desktop repro:

https://github.com/user-attachments/assets/ad6e1d7b-5a53-4631-b36e-b4d7b9b1e991

### Checklist

- [x] `npm run lint` passes


boudra:
@paseo-bot reshape audit

cleiter:
Closing - too much complexity for little gain.

PR #2167, opened 2026-07-17
perf: add benchmark registry and coalesce streamed chunks

BetterAndBetterII:
## Summary

This PR establishes the first unified performance benchmark registry and ships one production optimization measured through it.

- adds a root `npm run benchmark` registry with a versioned JSON result contract, task selection, runtime/source metadata, and a single output file;
- registers reducer, draft/attachment GC, Desktop tab, live-streaming, Markdown, and CSS-interaction benchmarks;
- coalesces compatible assistant-message continuations already present in the same scheduled client reducer flush;
- keeps benchmark instrumentation default-disabled or confined to deterministic mock-provider/test fixtures. Normal runtime behavior is unchanged outside the reducer optimization.

## Unified benchmark registry

```bash
npm run benchmark -- --list
npm run benchmark -- agent-stream-reducer
npm run benchmark -- desktop-interaction
npm run benchmark -- desktop-streaming
npm run benchmark -- desktop-markdown
npm run benchmark -- --output /tmp/paseo-benchmark.json
```

| Task | Coverage |
| --- | --- |
| `agent-stream-reducer` | Paired/interleaved baseline vs coalesced reducer measurements at 64 KiB, 256 KiB, and 1 MiB, plus 1/2/4/8 chunks-per-flush controls |
| `draft-attachment-gc` | Repeated draft changes over 0/100/500 sessions and the resulting full-history scan volume |
| `desktop-interaction` | 1/4/8 heavy Agent tabs with 50/100/176 history items; switch p50/p95, title/body consistency, React commits, frame gaps, DOM/AX, and heap |
| `desktop-streaming` | Exact 64 KiB/256 KiB/1 MiB visible streams; observed chunks-per-flush, reducer/React time, long tasks, frame gaps, heap, and feedback delay |
| `desktop-markdown` | Plain/prose/open and closed fences/mixed Markdown/link-table workloads; parse, highlight, render, DOM/AX, feedback, and a correctness oracle |
| `desktop-css-interactions` | 1/8/20-tab hover response, React commits, frame gaps, DOM, and AX |
| `desktop-css-interaction-audit` | AST inventory of JS-driven hover/press callbacks and interaction state by Desktop UI area |

Every browser benchmark starts an isolated temporary daemon and Metro instance on random ports. The specs explicitly reject HTTP and WebSocket traffic to production port 6767.

## Production optimization

The first chunk stays on the existing reducer path. Only consecutive assistant chunks with the same provider, message ID, epoch, and sequence are combined before reduction. Sequence gaps, identity changes, other event types, and runs that may promote Markdown blocks fall back to the existing per-event path. That conservative Markdown fallback preserves each promoted block's original event timestamp. One- and two-event flushes retain the existing allocation-free path.

The Markdown guard scans backward only through trailing whitespace; ordinary text exits in O(1). This avoids reintroducing a message-length scan into the hot path.

## Reducer benchmark methodology

The benchmark runs paired, interleaved A/B measurements in one Node process:

- **Baseline:** the previous per-event reducer loop.
- **Candidate:** the coalescing `processAgentStreamEvents` path.
- Chunks are distinct deterministic 512-byte strings; fixtures and expected output are built outside the timer.
- A/B execution order alternates to balance JIT, GC, and machine drift.
- Each case uses five warmup pairs and 30 measured pairs.
- Before timing, both variants must pass exact final-state equality plus text, item-count, cursor, change-flag, and side-effect checks.

The tables aggregate three complete runs of the current head: 90 paired samples per variant and case. The paired reduction column is the median per-pair reduction with its p05-p95 sample range; it is not a confidence interval.

### Size scaling at 8 chunks per flush

| Reply size | Baseline p50 / p95 | Coalesced p50 / p95 | Paired reduction p50 (p05-p95) |
| --- | ---: | ---: | ---: |
| 64 KiB | 0.327 / 0.341 ms | 0.119 / 0.126 ms | 63.8% (61.5-65.0%) |
| 256 KiB | 10.781 / 12.478 ms | 2.813 / 3.365 ms | 73.5% (69.2-77.2%) |
| 1 MiB | 162.913 / 177.062 ms | 40.583 / 47.316 ms | 75.1% (70.6-77.0%) |

### Chunks-per-flush guardrail at 256 KiB

| Chunks per flush | Paired reduction p50 (p05-p95) | Interpretation |
| ---: | ---: | --- |
| 1 | -1.0% (-37.2-11.2%) | no measurable change |
| 2 | -0.2% (-10.6-13.8%) | no measurable change |
| 4 | 49.1% (42.5-56.9%) | coalescing active |
| 8 | 73.5% (69.2-77.2%) | coalescing active |

These are reducer wall-time measurements, not claimed end-to-end UI gains. The Desktop streaming task separately records the real chunks-per-flush distribution so the synthetic workload assumption can be checked against renderer behavior.

## Benchmark validity hardening

- **Frame accounting:** replaces the incorrect `gap > 20 ms` event count with `delayedFrameIntervals` and `estimatedDroppedFrames`, retaining `maxFrameGap` and declaring the assumed 60 Hz interval. A one-second stall is approximately 59 missed frames, not one event. In the final 64 KiB Markdown smoke, 10 delayed intervals represented 124 estimated dropped frames with a 997.5 ms max gap.
- **Markdown correctness oracle:** a frozen workload specification independently renders expected Markdown through a reference MarkdownIt/HTML projection. Every measured run must match normalized visible-text content/hash and link count before timing results are emitted; repeated-but-equally-wrong output no longer passes.
- **Sample isolation:** every streaming/Markdown measured run gets a fresh BrowserContext and is marked `cacheState: cold`, preventing archived agents, DOM, React state, or renderer module caches from contaminating later workloads and sizes.
- **Timestamp semantics:** a regression test compares batched reduction against per-event reduction when continuations promote Markdown blocks, including the original timestamps.

## Correctness and validation

Tests cover first-chunk and leading-whitespace behavior, provider/message-ID boundaries, epoch changes, sequence-gap catch-up, ordering, cursors, side effects, lifecycle semantics, Markdown promotion timestamps, frame accounting, the independent Markdown oracle, registry schema validation, inactive-tab ablation behavior, and deterministic mock workloads.

- latest targeted Vitest run: 7 files, 126 tests passed;
- rebased onto `origin/main` (`bd937850b`); range-diff preserves the seven feature commits, with the sole conflict resolved by retaining both new attachment support and benchmark stream scanning;
- current-head reducer benchmark: three complete runs, 90 paired samples per case;
- real-browser smoke: one cold 64 KiB mixed-Markdown case passed the new visible-text/link oracle through isolated daemon + Metro + Chromium;
- `npm run typecheck` passed for every workspace;
- target-file `npm run lint` passed;
- `npm run format` and pre-commit format checks passed.


boudra:

PR #2188, opened 2026-07-18
OpenCode server-API: fix agent→model binding, native fork & per-session diff

shrimpwtf:
## Summary

Closes gaps in Paseo's OpenCode HTTP server-API integration. **OpenCode already runs on the server API (not ACP)** — this PR fixes the agent→model binding bug and lights up unused endpoints, all provider-scoped.

### Headline fix: agent→model binding (atlas→kimi)
OpenCode's `GET /agent` returns each agent's bound `model` (e.g. oh-my-openagent binds `atlas`→`kimi`). Paseo was discarding it, so switching agents never switched the model. Now:
- `AgentMode` carries an optional `model` (`providerID/modelID`).
- Selecting an agent/mode applies its bound model; an explicit model pick overrides until the next mode switch; the override does not persist across switches.
- `AgentManager.setAgentMode()` surfaces the re-bound model into `runtimeInfo.model` so the composer picker updates.

### Also included
- **Permission "Allow always"** — documented/verified it persists server-side via OpenCode's `reply:"always"` (session-scoped); no nonexistent `remember` flag. Added a CI-gated e2e.
- **Native fork** — "Fork chat" now routes into `session.fork()` for OpenCode (real provider child, history preserved), gated by an `agentNativeFork` capability. Other providers keep the generic flow.
- **Per-session diff** — `session.diff()` mapped to a provider-agnostic changed-files shape, gated by an `opencodeSessionDiff` capability with a client update-host affordance (no fallback path).

### Investigated / deferred (see mission backlog)
- Workspace search (`find*`), agent-context shell, and user-facing diff/search **UI panels** were evaluated and deferred (duplicates or need a larger UI subsystem).
- **In-app provider auth** was a spike only — feasible for code/api-key flows but **blocked on a `SECURITY.md` policy decision** ("Paseo never stores or transmits provider API keys").

## Compatibility
- All protocol schema additions are `.optional()`; new capabilities gated in `server_info.features.*` with dated `COMPAT()` tags (v0.2.0). No shared ACP files (copilot/cursor/kiro/trae/generic) were behavior-changed.

## Testing
- Per-feature unit tests via the fake OpenCode harness; real behaviors covered by CI-gated `daemon-e2e/opencode-*.real.e2e.test.ts`. `npm run typecheck` + targeted lint green. Full suite intentionally not run locally (per repo rules) — relies on CI.


shrimpwtf:
Closing in favor of #2256, which splits out just the agent→model-on-mode-switch fix as a self-contained, provider-clean change: no `if(provider==opencode)` branches, and no opencode details in the protocol (all OpenCode logic stays inside `opencode-agent.ts`). The native-fork and per-session-diff pieces can follow as their own focused PRs if wanted. Thanks for the review feedback.

PR #2218, opened 2026-07-19
Render live omp system-notices as synthetic tool calls

ebg1223:
## Problem

omp `<system-notice>` messages (background job completions, xd:// device inventory changes) were showing up as raw text in agent timelines.

The replay path (`OMP_HISTORY_MAPPER_HOOKS` → `mapOmpSystemNoticeToToolCall`) already absorbs these custom messages into synthetic `task_notification` tool calls, but the live `message_end` path in `agent.ts` emitted every custom message verbatim as an `assistant_message` — so notices arriving mid-session rendered as raw `<system-notice>` text.

## Fix

Route live custom messages through the same `mapOmpSystemNoticeToToolCall` mapper used on replay; non-notice custom messages still render as assistant messages.

Note: omp 17.x upstream fixed xd:// mount notices triggering unsolicited turns and device tools leaking into the tool inventory, but notices are still delivered as `role: "custom"` messages — this change fixes how they render regardless of delivery timing, and covers task-result notices which upstream doesn't defer.

## Testing

- `npx vitest run packages/server/src/server/agent/providers/omp/agent.test.ts --bail=1` (26 tests)
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

## Risk surface

Live custom-message routing also handles advisor messages, hidden messages, and ordinary text. The rebased implementation preserves those paths and the focused agent tests cover each behavior.

### DIFF
--- a/packages/server/src/server/agent/providers/omp/agent.test.ts
+++ b/packages/server/src/server/agent/providers/omp/agent.test.ts
@@ -294,11 +294,16 @@ describe("OMP agent client and session", () => {
           "</system-notice>",
         ].join("\n"),
       );
+    omp.runtime().acceptCustomMessage("plain custom status text");
 
     expect(omp.timeline().filter((item) => item.type === "tool_call")).toMatchObject([
       { callId: "omp-notice:DocsSmokeTwo", name: "task_notification", status: "completed" },
     ]);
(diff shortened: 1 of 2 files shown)


theslava:
@boudra Wanted to chime in that the current situation in 0.2.0 beta1 is that the messages not being captured properly mess up Paseo's understanding when a turn completes, thinking that the agent is still producing results even though it is done.

PR #2235, opened 2026-07-20
fix(app): send readline byte sequences to the shell for macOS modifier keys

OnCloud125252:
## Core problem

When translating macOS modifier keys (Cmd/Option + arrow/Backspace/Delete) into the bytes written to the PTY, the terminal sends the **wrong form** — it emits the xterm CSI modifier-encoded sequence instead of the byte the shell's readline is actually bound to. As a result these shortcuts do nothing in the default bash/zsh on macOS.

The fix is, at its core, a **translation layer between macOS-conventional keys and shell readline standards**. The app translates macOS modifier combos into the byte sequences the shell's readline recognizes and writes them to the PTY, so the shortcuts work out of the box — with zero shell-side configuration.

## Root cause (transport is fine; the source bytes are wrong)

The app → PTY transport itself is fine (`onInput` writes raw bytes straight to the PTY). The problem is that xterm.js always produces the **CSI modifier-encoded form** for modifier + arrow keys:

- Option+Left → `\x1b[1;3D`, Cmd+Left → `\x1b[1;9D`

But the default bash/zsh readline on macOS **does not bind CSI sequences** — it binds the readline-standard forms (see `man readline`):

- word motion = `\eb` / `\ef`, line edges = `\x01` / `\x05`
- backward-kill-word = `\e\x7f` (M-Rubout), kill-word = `\ed`, unix-line-discard = `\x15`

`macOptionIsMeta: true` only covers Option + letter (Option+B → `\x1bb`); it has **no effect on arrow keys or Backspace** (those go through the CSI modifier-encoding path). So the bytes reaching the shell for these combos are sequences the shell doesn't understand.

## Fix

In `attachCustomKeyEventHandler`, before handing the event to xterm, intercept the macOS modifier combos and send the readline-standard bytes to the PTY instead (via `onInput`). This mirrors iTerm2's "Natural Text Editing" and Ghostty (the native-macOS-editing camp — the terminal does the translation, rather than offloading it to shell-side `bindkey` the way Kitty/Alacritty do).

| macOS shortcut | readline bytes sent to shell | readline action |
|---|---|---|
| Option+← / → | `\eb` / `\ef` | backward-word / forward-word |
| Cmd+← / → | `\x01` / `\x05` | beginning-of-line / end-of-line |
| Option+⌫ | `\e\x7f` | backward-kill-word |
| Option+⌦ | `\ed` | kill-word (forward) |
| Cmd+⌫ | `\x15` | unix-line-discard (delete to line start) |

## Design

- Gated on `isMac` only (Cmd/Option are macOS conventions); Shift/Ctrl combos (selection, Ctrl+arrow) are untouched and keep xterm's behavior.
- The runtime is shared by **web/Electron desktop** and the **native webview (iOS/iPad hardware keyboard)**, so one change covers both.
- Interception runs before `shouldInterceptDomTerminalKey` (the mobile modifier-bar path), so the two input paths don't interfere.
- On a hit it `return false` (xterm's "handled" contract for `attachCustomKeyEventHandler`), so xterm won't also emit its own CSI sequence into the PTY.

## Changes

- `packages/app/src/utils/terminal-keys.ts` — new pure function `resolveMacTerminalShortcut` (macOS combo → readline bytes)
- `packages/app/src/terminal/runtime/terminal-emulator-runtime.ts` — `handleMacShortcut` helper (intercepts before xterm and writes to the PTY)
- `packages/app/src/utils/terminal-keys.test.ts` — unit tests covering every combo plus boundaries

## Verification

- Unit tests 34/34 green (pure function across all combos, plus negatives: Shift / Ctrl / plain keys are not intercepted; Backspace/Delete with no modifier fall through to the normal path)
- typecheck / lint / format all clean; pre-commit hook passes
- Manual verification: dev desktop app against the local daemon, pressing Option/Cmd + arrows/Backspace confirms word-jump, line-edge, word-delete, and line-delete all reach the shell and trigger the expected readline action

### DIFF
--- a/packages/app/src/terminal/runtime/terminal-emulator-runtime.ts
+++ b/packages/app/src/terminal/runtime/terminal-emulator-runtime.ts
@@ -20,6 +20,7 @@ import {
   mergeTerminalModifiers,
   normalizeDomTerminalKey,
   normalizeTerminalTransportKey,
+  resolveMacTerminalShortcut,
   shouldInterceptDomTerminalKey,
 } from "@/utils/terminal-keys";
 import { renderTerminalSnapshotToAnsi } from "./terminal-snapshot";
@@ -438,6 +439,13 @@ export class TerminalEmulatorRuntime {
         return true;
       }
 
+      // macOS Option/Cmd + arrows: xterm emits CSI modifier forms (\x1b[1;3D etc.)
+      // that macOS default readline doesn't bind. Remap to the readline-default
+      // word/line motion bytes so these shortcuts behave like a native terminal.
+      if (this.handleMacShortcut(event, normalizedKey)) {
+        return false;
+      }
+
       if (
         !shouldInterceptDomTerminalKey({
           key: normalizedKey,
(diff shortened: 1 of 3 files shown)


boudra:
Closing for now: this needs a before/after macOS terminal recording of the remapped shortcuts and a runtime regression showing they send no PTY input while input is suppressed. Please reopen with that evidence.

boudra:

Issue #2221, opened 2026-07-19
One oversized frame takes down the entire mobile client until force-quit (shared transport + backoff reset on handshake + no 1009 circuit breaker)


muzhi1991:
## Summary

When a single agent's timeline response exceeds the relay's message cap, the resulting close does not just break that agent — it takes down **the entire mobile client**. Every other session, the workspace list, and terminals all stop loading, and the client enters a permanent reconnect loop at a fixed 1.5 s interval. The only recovery is force-quitting the app from the task switcher; backgrounding and reopening does not help.

The oversized-frame cause is filed separately in #2220. This issue is about the blast radius and the failure to recover, which are independent of *why* the frame was too large — any oversized frame, from any source, produces this.

I traced three separate contributing defects in `packages/client/src/daemon-client.ts`.

## Severity

This turns a single unlucky session into a full client outage. A user with one image-heavy session cannot use Paseo on mobile at all — not that session, not any other — until they know to force-quit. Nothing in the UI suggests that; it just looks like the whole app is broken.

## Reproduction

1. Have an agent whose timeline page exceeds the relay message cap (see #2220 — in my case 12.2 MiB payload, ~17.1 MB on the wire after E2EE base64, against a 10 MB relay read limit).
2. Open that agent from the mobile client.
3. Relay closes the daemon-side socket with 1009, then closes the client with 1012.
4. Navigate to any other session, or back to the workspace list.

Expected: other sessions still work; the broken one shows an error.
Actual: nothing loads anywhere. `daemon.log` shows a continuous reconnect storm — I measured 20+ new relay connections per 30 s window, every window, indefinitely.

## Defect 1 — one transport, no request isolation

All sessions for a daemon share a single `DaemonClient` transport. My daemon logs confirm the multiplexing — one `connectionId` carrying every request type:

```
conn_495ba4d4…  requestType: fetch_agent_timeline_request
conn_495ba4d4…  requestType: checkout_status_request
conn_495ba4d4…  requestType: checkout_pr_status_request
conn_495ba4d4…  requestType: fetch_agents_request
conn_495ba4d4…  requestType: list_terminals_request
```

So when one oversized `fetch_agent_timeline_response` kills the socket, `scheduleReconnect` tears down everything in flight regardless of which session it belonged to:

```ts
// Clear all pending waiters and queued sends since the connection was lost
// and responses from the previous connection will never arrive.
this.clearWaiters(new Error(reason ?? "Connection lost"));
[...]

PR #2237, opened 2026-07-20
fix(codex): materialize toolSurface screenshots out of timeline entries

### LINKED ISSUE #2220: bug: Timeline fetch still exceeds relay message cap on image-heavy sessions (#657 planner bypassed on some call sites; page size is item-count-based)
## Summary

On sessions whose timeline contains many large tool-call outputs (inline base64 images), `fetch_agent_timeline_request` still produces responses larger than the relay's per-message cap. The relay closes the daemon-side socket with **1009** (message too big), then closes the client with **1012 "Server disconnected"**, and the client reconnects into a loop. The session becomes unopenable from mobile/web, while opening instantly on the local desktop client (direct connection, no relay).

This is the same failure mode as #656 / PR #657. That fix **is** present in my build — I verified `TIMELINE_FETCH_PAGE_SIZE = 100` in the shipped 0.1.110 client bundle — but two gaps remain.

## Environment

- Paseo desktop **0.1.110**, mobile client **0.1.109** (both well after PR #657, merged 2026-05-02)
- macOS arm64, daemon running on the desktop machine
- Self-hosted relay (Go / `gorilla/websocket`) with `ws.SetReadLimit(10 * 1024 * 1024)`
- Provider: Codex

## Symptom

Mobile shows `Transport closed (code 1012)` on the agent screen. `daemon.log` shows this repeating (217+ occurrences over the failing periods), one per client retry:

```json
{"msg":"relay_data_disconnected","code":1009,"reason":"",
 "url":"ws://<relay>/ws?serverId=<id>&role=server&v=2&connectionId=<id>"}
```

`1009` is the relay's read limit rejecting the frame. The relay then does this to the client (`@getpaseo/relay/dist/cloudflare-adapter.js`):

```js
if (role === "server" && connectionId) {
    // Force the client to reconnect and re-handshake when the daemon side drops.
    for (const clientWs of this.state.getWebSockets(`client:${connectionId}`))
        clientWs.close(1012, "Server disconnected");
}
```

so `1012` is a downstream symptom, not the cause. The client retries, the daemon re-sends the same oversized frame, and it loops — I measured bursts of ~20 reconnects per 30s window, with `ws_slow_request` for `fetch_agent_timeline_request` climbing from ~600 ms to ~2900 ms under the retry load.

## Measurements

Method: a probe client connected directly to the local daemon (`ws://127.0.0.1:6767/ws`), sent `hello`, then `fetch_agent_timeline_request` with `direction: "tail"`, `projection: "projected"`, varying `limit`. Bytes are the raw decoded WebSocket text frame with `perMessageDeflate: false`, i.e. actual payload size.

**Failing agent** (Codex session, ~870 timeline rows):

| `limit` | response bytes | MiB | entries returned | on-wire after E2EE base64 (x4/3) |
|---|---|---|---|---|
| 200 (daemon default) | 12,813,199 | 12.22 | 200 | ~17.1 MB |
| 100 (`TIMELINE_FETCH_PAGE_SIZE`) | 9,555,730 | 9.11 | 169 | ~12.7 MB |
| 50 | 2,449,702 | 2.34 | 50 | ~3.3 MB |
| 25 | 2,421,372 | 2.31 | 43 | ~3.2 MB |

Composition of the 200-item page: entries total 12,809,993 bytes; the `agent` metadata payload is only 2,428 bytes. The top 15 entries are all `tool_call` / `node_repl.js` outputs carrying inline base64 images:

```
seq=361 bytes=618147 kind=tool_call tool=node_repl.js
seq=358 bytes=593210 kind=tool_call tool=node_repl.js
seq=105 bytes=569766 kind=tool_call tool=node_repl.js
seq=107 bytes=554882 kind=tool_call tool=node_repl.js
seq=169 bytes=547165 kind=tool_call tool=node_repl.js
...  (top 15 total ~7.9 MB = 61% of the whole payload)
```

**Controls** — three other agents in the same workspace, same request shape: 0.19 MiB, 0.19 MiB, 3.59 MiB. All load fine remotely.

Worth highlighting: **on-disk transcript size does not predict this at all.** One control agent has a 39 MB transcript on disk but a 194 KB tail page; the failing agent has a smaller transcript but a 12.2 MB tail page. What matters is the byte weight of the entries that land in the page.

## Gap 1 — some call sites bypass the bounded planner

PR #657 added planners that set the limit:

```js
function planTimelineTailFetch() {
  return { direction: "tail", limit: TIMELINE_FETCH_PAGE_SIZE };
}
```

But the shipped 0.1.110 client bundle also contains call sites that invoke `fetchAgentTimeline` with **no `limit`**:

```js
await client.fetchAgentTimeline(agentId, Object.assign(
  { direction: "tail", projection: "projected" },
  cursor ? { cursor: { epoch: cursor.epoch, seq: cursor.endSeq } } : {}
))
```

With no `limit`, the daemon falls back to 200 (`@getpaseo/server/dist/server/server/session.js`, `handleFetchAgentTimelineRequest`):

```js
const pageLimit = requestedLimit ?? (direction === "after" ? 0 : 200);
```

which is the 12.22 MiB / ~17.1 MB row in the table above. (Note the `direction === "after" ? 0` branch is still unbounded, and `agent-timeline-store.js` treats `limit === 0` as `selectAll` — that path appears to be what #657 removed from app callers, but the daemon-side default still permits it.)

*Caveat:* I inspected the desktop Expo web bundle inside the installed app. I could not decompile the mobile build, so whether the mobile client hits these same call sites is an inference, not something I verified directly.

## Gap 2 — the page budget is item-count-based and ignores E2EE base64 inflation

Even on the correct planner path, `TIMELINE_FETCH_PAGE_SIZE = 100` bounds *item count*, not bytes. With ~0.5 MB entries, 100 items is 9.11 MiB. That passes a naive "under 10 MB" check — but the relay transports ciphertext as base64:

```js
// Our protocol always transmits ciphertext as base64 text
```
— `@getpaseo/relay/dist/encrypted-channel.js`

so the on-wire size is a constant **4/3** of the payload: 9.11 MiB becomes ~12.7 MB, still over the cap. There is also no chunking anywhere on this path — neither `encrypted-channel.js` nor the relay adapter splits messages; they are forwarded whole. So a single oversized entry set has no way through.

My relay's cap happens to be 10 MB; a different cap just moves the threshold, it doesn't remove the class of failure — any item-count-based budget will eventually be crossed by heavy enough entries.

## Suggested fix

Bound the timeline page by **serialized bytes** rather than item count: accumulate entries until a byte budget is hit, returning fewer items when they are heavy. The budget should fold in the x4/3 base64 expansion (ideally negotiated against the relay's actual cap rather than hardcoded). A per-entry cap with truncation/lazy-loading for oversized tool outputs (inline images especially) would address the root cause more directly, since a single entry can already be ~0.6 MB here.

Fixing Gap 1 alone would not be enough for this session — 100 items still exceeds the cap.

## Repro

Any long session with repeated large inline-image tool outputs, opened through a relay whose message cap is below (page bytes x 4/3). In my case the agent was a Codex session repeatedly generating architecture diagrams via a `node_repl.js` tool that returns images as base64 data URLs.

muzhi1991:
Fixes #2220.

## Problem

Codex browser tools return their screenshot as a base64 data URL under
`_meta["codex/toolSurface"].screenshot.url`, not as an MCP `image` content block.
`splitCodexMcpToolResultImages` only inspects `content[]`, so the data URL lands
verbatim in the timeline entry. A single screenshot can exceed 0.6 MB; on an
image-heavy session the tail page reached 12.2 MiB, which after base64/E2EE
wrapping (×4/3) exceeds a 10 MB relay message cap — the relay closes the daemon
socket with 1009 and the client with 1012, and the client cannot open any session
on that daemon until it is force-quit.

## Change

Chain a toolSurface split ahead of the existing `content[]` split so both sources
feed one result. Extracted images flow into the existing `materializeProviderImage`
path via `mcpToolResultImagesToTimeline`, exactly as MCP `content[]` images already
do, and render as a markdown reference to a content-hashed file.

No new mechanism — this only widens *where* images are looked for. Only the `url`
key is dropped; the rest of `_meta` and the screenshot's own metadata
(`width`/`height`) are preserved.

This also helps direct/local connections, where the same inline base64 is currently
paid as memory and render cost rather than as a disconnect.

## QA / testing evidence

Tested on **macOS arm64 only** (daemon 0.1.110, desktop 0.1.110, mobile 0.1.109).
The patch was built as 0.1.110 npm tarballs, installed on the machine holding the
affected session, and run as the daemon against that machine's real `~/.paseo`.

Measured against the same affected agent, via `fetch_agent_timeline_request`
(`direction: "tail"`, `projection: "projected"`) on the live daemon:

| | before | after |
|---|---|---|
| tail page | 12,810,194 B (200 entries) | **590,705 B** (200 entries) |
| reduction | — | **95.4%** |
| inline `data:image` | present throughout | **0** |
| materialized image refs | 0 | **64** |
| `toolSurface` metadata | present | **preserved** (71, minus `url`) |
| largest single entry | ~618 KB | **42 KB** |

An unbounded fetch (`limit: 0`) of the whole session is 756,490 B / 264 entries,
comfortably under the cap. The session that previously could not be opened at all
now opens, and the extracted screenshots render as images in remote clients.

Automated checks: 5 new unit tests in `tool-call-mapper.test.ts` (47/47 in that
file); full `@getpaseo/server` unit suite 3984 passed. One unrelated pre-existing
failure, `bootstrap-provider-availability.test.ts`, reproduces identically on an
unpatched `main` in this environment (it asserts Codex is unavailable, but a
`codex` binary is on PATH here). `typecheck`, `lint` and `format` all clean.

## Notes

Written with AI assistance; the diagnosis, deployment and the measurements above
are my own, run against a real affected session.

### DIFF
--- a/packages/server/src/server/agent/providers/codex/tool-call-mapper.test.ts
+++ b/packages/server/src/server/agent/providers/codex/tool-call-mapper.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it } from "vitest";
 
-import { mapCodexToolCallEnvelope, mapCodexToolCallFromThreadItem } from "./tool-call-mapper.js";
+import {
+  mapCodexToolCallEnvelope,
+  mapCodexToolCallFromThreadItem,
+  splitCodexMcpToolResultImages,
+} from "./tool-call-mapper.js";
 
 function expectMapped<T>(item: T | null): T {
   expect(item).not.toBeNull();
@@ -839,6 +843,119 @@ describe("codex tool-call mapper", () => {
       },
     });
     expect(JSON.stringify(item)).not.toContain("iVBORw0KGgo=");
+  });
+
+  it("drops codex toolSurface screenshot data urls from mcp tool output", () => {
+    const item = expectMapped(
+      mapCodexToolCallFromThreadItem({
+        type: "mcpToolCall",
+        id: "codex-tool-surface",
+        status: "completed",
+        server: "node_repl",
+        tool: "js",
+        arguments: { code: "await page.screenshot()" },
+        result: {
+          content: [{ type: "text", text: "navigated" }],
+          structuredContent: { ok: true },
+          _meta: {
+            "codex/toolSurface": {
+              kind: "browser",
+              screenshot: {
+                url: "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQ==",
+                width: 1280,
+                height: 720,
+              },
+            },
+            "codex/requestId": "req-42",
+          },
+        },
+      }),
+    );
+
+    expect(item).toEqual({
+      type: "tool_call",
+      callId: "codex-tool-surface",
+      name: "node_repl.js",
+      status: "completed",
+      error: null,
+      detail: {
+        type: "unknown",
+        input: { code: "await page.screenshot()" },
+        output: {
+          content: [{ type: "text", text: "navigated" }],
+          structuredContent: { ok: true },
+          _meta: {
+            "codex/toolSurface": {
+              kind: "browser",
+              screenshot: { width: 1280, height: 720 },
+            },
+            "codex/requestId": "req-42",
+          },
+        },
+      },
+    });
+    expect(JSON.stringify(item)).not.toContain("/9j/4AAQSkZJRgABAQ==");
+  });
+
+  it("hands codex toolSurface screenshots to the image materialization path", () => {
+    const { images } = splitCodexMcpToolResultImages({
+      content: [{ type: "text", text: "navigated" }],
+      _meta: {
+        "codex/toolSurface": {
+          screenshot: { url: "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQ==" },
+        },
+      },
+    });
+
+    expect(images).toEqual([{ data: "/9j/4AAQSkZJRgABAQ==", mimeType: "image/jpeg" }]);
+  });
+
+  it("extracts both toolSurface screenshots and mcp image content blocks", () => {
+    const { images, output } = splitCodexMcpToolResultImages({
+      content: [
+        { type: "text", text: "navigated" },
+        { type: "image", data: "iVBORw0KGgo=", mimeType: "image/png" },
+      ],
+      _meta: {
+        "codex/toolSurface": {
+          screenshot: { url: "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQ==" },
+        },
+      },
+    });
+
+    expect(images).toEqual([
+      { data: "/9j/4AAQSkZJRgABAQ==", mimeType: "image/jpeg" },
+      { data: "iVBORw0KGgo=", mimeType: "image/png" },
+    ]);
+    expect(output).toEqual({
+      content: [
+        { type: "text", text: "navigated" },
+        { type: "text", text: "[image]" },
+      ],
+      _meta: { "codex/toolSurface": { screenshot: {} } },
+    });
+  });
+
+  it("leaves remote toolSurface screenshot urls untouched", () => {
+    const result = {
+      content: [{ type: "text", text: "navigated" }],
+      _meta: {
+        "codex/toolSurface": {
+          screenshot: { url: "https://example.com/shot.jpg", width: 800 },
+        },
+      },
+    };
+
+    expect(splitCodexMcpToolResultImages(result)).toEqual({ images: [], output: result });
+  });
+
+  it("leaves mcp tool results without a toolSurface screenshot unchanged", () => {
+    const result = {
+      content: [{ type: "text", text: "navigated" }],
+      _meta: { "codex/requestId": "req-42" },
+    };
+
+    expect(splitCodexMcpToolResultImages(result)).toEqual({ images: [], output: result });
   });
 
   it("normalizes codex paseo_voice.speak mcp calls and extracts spoken text", () => {
--- a/packages/server/src/server/agent/providers/codex/tool-call-mapper.ts
+++ b/packages/server/src/server/agent/providers/codex/tool-call-mapper.ts
@@ -635,13 +635,65 @@ function readMcpImageContent(block: unknown): CodexMcpToolResultImage | null {
   };
 }
 
-export function splitCodexMcpToolResultImages(result: unknown): CodexMcpToolResultImagesSplit {
-  if (!isRecord(result) || !Array.isArray(result.content)) {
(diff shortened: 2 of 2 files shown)


muzhi1991:
@boudra CI is green (22 passed, 1 skipped) — this is ready for review whenever you have time.

One thing I want to flag about impact, because the issue title makes it sound like an edge case and I don't think it is: **this reproduces on any Codex session where a tool returns an image**, not just unusually long ones. Codex browser tools put their screenshot in `_meta["codex/toolSurface"].screenshot.url` as a base64 data URL, and nothing downstream ever reads that key — I grepped the whole repo, there are zero references to `toolSurface` anywhere in server, client or app. So today those bytes are shipped to every client and then silently discarded by the UI.

That has two consequences:

1. It is pure dead weight. Removing it costs no functionality — before this patch the screenshots were never rendered at all; after it they render as normal materialized images, which is strictly better.
2. The threshold is low. At the 0.3–0.6 MB per screenshot I measured, on the order of 15–25 screenshots inside one 200-entry page window is enough to cross a 10 MB relay cap after the ×4/3 base64/E2EE expansion. A session where Codex takes screenshots as it works reaches that quickly.

When it does trip, the failure is not graceful: the relay closes the daemon socket with 1009 and the client with 1012, and the client then cannot open *any* session on that daemon until it is force-quit (that half is #2221). I hit exactly that — a second, completely healthy session with no images at all became unopenable as collateral, and I initially misdiagnosed it as a problem with that session.

I have been running this patch as my daemon against the real affected session; the previously unopenable session opens, and the screenshots now show up as images for the first time. Would be great to get it into the next release if it looks right to you.

muzhi1991:
Gentle follow-up: v0.2.1 shipped as the latest release, and I confirmed the bug is still present there — `tool-call-mapper.ts` is byte-identical between this PR's base and `v0.2.1`, `splitCodexMcpToolResultImages` still only inspects `content[]`, and there are still zero references to `toolSurface` anywhere in the repo. The five recent `screenshot` commits are all on the capture side (browser-automation / webview / capture-harness) and don't touch how the screenshot bytes enter the timeline, so they're orthogonal to this.

This branch still merges cleanly — I rebased it onto the current `main` locally with no conflicts (the two touched files are untouched across the 125 commits since the base), and GitHub reports it MERGEABLE / CLEAN.

No urgency intended, just flagging that anyone on 0.2.x using Codex browser tools still hits this: a handful of screenshots in one page window is enough to cross the relay cap. Happy to rebase, split, or adjust however is easiest for you to take it.

PR #2243, opened 2026-07-20
fix(codex): preserve built-in plan mode instructions

### LINKED ISSUE #2231: bug: Codex plan mode instructions are overwritten by the appended system prompt
## Summary

When using Codex through Paseo, enabling Plan mode does not reliably prevent implementation or produce a `<proposed_plan>...</proposed_plan>` response if the host-level “Append system prompt” setting is configured.

The appended system prompt is passed as `collaborationMode.settings.developer_instructions`, which replaces Codex’s built-in Plan mode instructions instead of appending to them.

As a result, Codex receives `mode: "plan"` but does not receive the instructions that define Plan mode behavior, including the requirement to avoid implementation and return a proposed plan.

## Steps to reproduce

1. Configure a host-level appended system prompt in Paseo.
2. Create or resume a Codex agent.
3. Enable the `plan_mode` feature.
4. Ask Codex to investigate and implement a change.
5. Observe that Codex may start editing files immediately instead of presenting a plan.

## Expected behavior

The appended system prompt should coexist with Codex’s built-in Plan mode instructions.

While Plan mode is enabled, Codex should:

- investigate without modifying repository-tracked files;
- ask clarification questions when necessary;
- present the final plan using `<proposed_plan>...</proposed_plan>`;
- wait for the user to approve or leave Plan mode before implementation.

## Actual behavior

The turn is marked as Plan mode:

```json
{
  "collaboration_mode_kind": "plan"
}
```

However, the applied collaboration mode contains only Paseo’s appended system prompt:

```json
{
  "collaboration_mode": {
    "mode": "plan",
    "settings": {
      "developer_instructions": "<Paseo appended system prompt>"
    }
  }
}
```

Codex’s built-in Plan mode instructions are absent. The agent therefore behaves like a normal coding turn and may begin implementation immediately.

## Root cause

`resolveCollaborationMode()` combines the appended prompts into the collaboration mode’s `developer_instructions`:

```ts
const developerInstructions = composeSystemPromptParts(
  match.developer_instructions,
  this.config.systemPrompt,
  this.config.daemonAppendSystemPrompt,
);

if (developerInstructions) {
  settings.developer_instructions = developerInstructions;
}
```

Current Codex app-server behavior defines:

- `developer_instructions: null` as “use the built-in instructions for the selected collaboration mode”;
- a string value as an override of those built-in instructions.

Additionally, the current `collaborationMode/list` response intentionally does not expose the built-in instructions:

```json
{
  "name": "Plan",
  "mode": "plan",
  "model": null,
  "reasoning_effort": "medium"
}
```

Paseo nevertheless expects `developer_instructions` to be returned by this API. It therefore resolves the built-in instructions to `null`, combines that with the appended system prompt, and sends the appended prompt as a complete replacement.

The same appended prompt is also sent through the top-level `turn/start.developerInstructions`, causing it to be supplied twice.

Relevant code:

- `packages/server/src/server/agent/providers/codex-app-server-agent.ts`
  - `loadCollaborationModes()`
  - `resolveCollaborationMode()`
  - `buildTurnStartParams()`

## Testing gap

The unit-test collaboration-mode fixture includes a `developer_instructions` field that the current Codex app-server response does not provide:

```ts
{
  name: "Plan",
  mode: "plan",
  developer_instructions: "Built-in plan mode",
}
```

The real Plan mode E2E test also does not configure `daemonAppendSystemPrompt`, so it does not cover the failing condition. Its prompt explicitly tells the model to produce a plan, which can also mask missing built-in Plan mode instructions.

Regression coverage should include:

1. Plan mode without an appended system prompt.
2. Plan mode with `systemPrompt`.
3. Plan mode with `daemonAppendSystemPrompt`.
4. Verification that the appended prompt and built-in Plan mode instructions are both present.
5. Verification that a normal implementation request does not produce file changes while Plan mode is enabled.
6. Verification that the final response is emitted as a Plan item and rendered through the Plan approval UI.

supermomonga:
## Summary

- preserve Codex's built-in collaboration-mode instructions by leaving `developer_instructions` null
- send composed host instructions only when starting or resuming a thread
- remove the duplicate turn-level instruction payload and align collaboration-mode fixtures with the current app-server response
- add unit, local app-server, real-provider, and approval UI regression coverage

## Root cause

`collaborationMode/list` does not expose Codex's built-in mode instructions. Paseo treated the omitted value as content to compose with host-level prompts and sent the result through `collaborationMode.settings.developer_instructions`, which overrides the built-in Plan mode instructions. The same appended prompt was also sent again through `turn/start`.

This change keeps mode selection and host-provided instructions on their separate app-server surfaces: Codex resolves the built-in mode instructions, while Paseo supplies appended instructions once at the thread level.

## Validation

- provider unit tests: 125 passed
- local Codex app-server E2E: 2 passed
- real-provider Plan mode E2E: 1 passed
- Plan approval App E2E: 1 passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git diff --check`

Fixes #2231

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -22,6 +22,8 @@ Existing direct providers: `claude` (in `providers/claude/agent.ts`), `codex` (`
 
 Claude first-party model metadata lives in `packages/server/src/server/agent/providers/claude/model-manifest.ts`. When adding or updating a Claude model, update that manifest only; the model picker thinking options and Claude-specific feature gates are derived from the manifest. Do not add model-specific Claude capability lists in feature code.
 
+Codex collaboration modes and Paseo's additional instructions use separate app-server surfaces. `collaborationMode/list` intentionally omits built-in developer instructions, so mode selection must send `settings.developer_instructions: null` and let Codex resolve the selected mode's built-in instructions. Pass the composed per-agent and daemon-wide instructions only through `thread/start.developerInstructions` or `thread/resume.developerInstructions`; do not copy them into collaboration-mode settings or `turn/start`.
+
 Paseo tools are not implemented as MCP tools internally. They live in a shared tool catalog under `packages/server/src/server/agent/tools/`; MCP is only the fallback adapter. A provider that can register runtime tools directly should set `supportsNativePaseoTools: true` and consume `launchContext.paseoTools` in `createSession`/`resumeSession`. When native tools are present, `AgentManager` strips the internal Paseo MCP server from the provider launch config so the provider does not receive the same tools twice. Providers that only know MCP should keep `supportsMcpServers: true` and let the daemon inject `/mcp/agents`.
 
 Pi is a process-backed provider. Paseo requires the user to have the `pi` binary installed and talks to it through `pi --mode rpc`; the server package does not embed Pi's SDK/runtime packages.
--- a/packages/server/src/server/agent/providers/codex-app-server-agent.features.test.ts
+++ b/packages/server/src/server/agent/providers/codex-app-server-agent.features.test.ts
@@ -16,21 +16,24 @@ interface CollaborationModeRecord {
   mode?: string | null;
   model?: string | null;
   reasoning_effort?: string | null;
-  developer_instructions?: string | null;
 }
 
 const TEST_COLLABORATION_MODES: CollaborationModeRecord[] = [
   {
     name: "Code",
     mode: "code",
-    developer_instructions: "Built-in code mode",
   },
   {
     name: "Plan",
     mode: "plan",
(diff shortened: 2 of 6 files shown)


supermomonga:
> Duplicated waitForEvent helper

The helpers are similar, but their timeout contracts intentionally differ: the local E2E requires an explicit timeout, while the real-provider E2E defaults to 120 seconds. With only two call sites and test-specific timeout semantics, I’d prefer to keep them local rather than introduce a shared abstraction. If another use case appears or the semantics converge, we can extract a common test utility then.

PR #2245, opened 2026-07-20
fix(omp): keep parent non-idle while internal task subagents run

### LINKED ISSUE #2232: OMP parent shows idle/stopped while internal task subagents are still active; nativeHandle points at missing .jsonl
### Summary
When an OMP-backed Paseo agent fans out work with OMP-internal `task` subagents, the **parent agent flips to `idle` / UI “stopped”** even though the OMP process is still actively writing child session logs.

Paseo’s stored `persistence.nativeHandle` points at a **missing `.jsonl` file**, while the real session is a **directory with the same stem** containing live child `*.jsonl` files.

This makes long-running fan-out audits look finished/stopped in the agent list when they are not.

### Environment
- Paseo: `0.1.110` (`/Applications/Paseo.app`)
- OMP: `omp/17.0.5`
- Provider: `omp` / model `cursor/cursor-grok-4.5-high-fast`
- OS: macOS `26.5.2` (Darwin `25.5.0`, arm64)
- Daemon: `127.0.0.1:6767`
- OMP provider config uses `params.sessionDir: ~/.omp/agent/sessions`

### Repro
1. Create/run an OMP agent that spawns several internal `task` subagents (fan-out audit / research style).
2. Wait until the parent turn parks on hub waits / child completion.
3. Observe:
   - `paseo inspect <parent-id> --json` → `Status: idle`
   - UI shows parent as stopped/idle
4. Meanwhile:
   - `omp` process for that session is still alive
   - child `*.jsonl` files under the session directory continue updating
   - `lsof -p <omp-pid>` shows open writes to those child files

### Observed (exact case)
Parent agent:
- ID: `0d43efa4-2b73-4645-bf79-940b8f82be48`
- Title: `critically audit the entire repo, data feed, api usage, pipe`
- Status: `idle`
- Session ID: `019f7c27-d1fd-7000-918f-fd49e1e66a6a`

Stored handle:
```text
~/.omp/agent/sessions/-Antigravity-unusualwhales/2026-07-19T20-53-32-285Z_019f7c27-d1fd-7000-918f-fd49e1e66a6a.jsonl
```
- `exists = false`

Actual session:
```text
~/.omp/agent/sessions/-Antigravity-unusualwhales/2026-07-19T20-53-32-285Z_019f7c27-d1fd-7000-918f-fd49e1e66a6a/
```
- exists as a directory
- contains live child logs such as:
  - `ApiBudgetAudit.jsonl`
  - `PipelineFeedAudit.jsonl`
  - `Btc5mStratHealth.jsonl`
  - `LegacyStratHealth.jsonl`
  - `HarnessPnlAudit.jsonl`
  - etc.

At observation time, parent OMP pid `27988` still had open write FDs on multiple child jsonls in that directory.

Contrast: currently healthy `running` agents still have **existing `.jsonl` nativeHandle files**.

### Expected
While OMP-internal child task sessions are still active for a parent:
1. Parent should remain `running` (or some explicit `waiting` / `delegated` status), **or**
2. Child work should be visible in Paseo’s agent/subagent tree, **and**
3. `nativeHandle` should resolve to a real session artifact (file **or** directory), not a missing path.

### Actual
- Parent status becomes `idle` / UI stopped
- Child work continues invisibly under `~/.omp/agent/sessions/.../<session-stem>/`
- Paseo list no longer reflects that the job is still in progress
- Attach/logs UX is degraded because the parent looks finished

### Impact
- False “done/stopped” signal for multi-agent OMP runs
- Hard to babysit long audits from the agent list
- Easy to interrupt/restart/abandon work that is actually still running
- Confusing freeze triage: parent looks idle while child hub waits continue

### Workarounds (no Paseo patch)
1. Watch session-dir mtimes / open FDs:
   - `~/.omp/agent/sessions/.../<session-stem>/*.jsonl`
   - `lsof -p <omp-pid>`
2. Use `paseo attach <id>` / external watcher for activity
3. Prefer **Paseo-level agents** (`paseo run` / handoff / advisor) over OMP-internal `task` when list visibility matters
4. `paseo send` can flip the parent back to `running` briefly, but that does **not** restore correct child tracking

### Suspected gap
Paseo appears to treat OMP parent turn completion / missing parent `.jsonl` handle as terminal idle, without accounting for OMP’s session-directory + internal task-subagent layout.

### Ask
Please consider:
1. Keep parent non-idle while OMP child session files under the session stem are still active
2. Make `nativeHandle` support OMP session directories (or rewrite handle to the real artifact)
3. Surface OMP-internal task children in the agent tree / status model
4. Or document clearly that OMP-internal `task` fan-out is intentionally invisible to Paseo status

Happy to provide more logs / a minimal repro script if useful.

jasonhnd:
## Summary

Fixes [#2232](https://github.com/getpaseo/paseo/issues/2232): an OMP parent agent can flip to **idle / UI “stopped”** while OMP-internal `task` subagents are still actively writing child session logs.

### Background / root cause

Paseo marks an OMP turn complete after:

1. OMP emits `agent_end` (with an assistant message), and
2. `get_state()` reports `!isStreaming && !isCompacting`

That only means **the parent model loop is idle**.

OMP’s `isStreaming` is approximately:

```ts
agent.state.isStreaming || promptInFlight
```

It does **not** include background `task` / hub-wait children. So a long fan-out audit can:

- end the parent turn
- keep the OMP process alive
- keep writing child `*.jsonl` under the session artifacts directory
- while Paseo already shows the parent as idle/stopped

This matches the reporter’s observation:

| Observed | Meaning |
|---|---|
| `paseo inspect` → `Status: idle` | Paseo turn completed |
| OMP pid still alive with open FDs on child jsonls | Work still in progress |
| Stored `nativeHandle` → missing `....jsonl` | Parent file path may not be on disk |
| Real artifact → `..../` directory with live child jsonls | OMP stem/artifacts layout |

OMP’s on-disk layout (confirmed against OMP v17.0.5) is:

```text
parent:   ~/.omp/agent/sessions/<cwd>/<timestamp>_<id>.jsonl
children: ~/.omp/agent/sessions/<cwd>/<timestamp>_<id>/<AgentName>.jsonl
```

Paseo already assumed that layout for history replay. The bug was treating **parent turn idle** as **whole job finished**.

### What this PR changes

1. **Turn-complete gate (main fix)**  
   In `completeTurnAfterProviderIdle`, also require **no running OMP-internal subagents** before emitting `turn_completed` / flipping the agent idle.

2. **`get_subagents` reconciliation**  
   Each idle poll calls OMP’s `get_subagents` RPC and merges into the local subagent index.  
   - Covers missed lifecycle/progress frames  
   - Older OMP binaries without the RPC fall back to the event index (debug-logged)

3. **Orphan child history recovery**  
   If the parent `.jsonl` is missing (`ENOENT`) but the stem artifacts directory exists, history still surfaces child transcripts instead of returning empty.

### Non-goals (intentionally not in this PR)

- New lifecycle enum values (`waiting` / `delegated`)
- Promoting every OMP task child into a full managed Paseo agent
- Rewriting `nativeHandle` from file path to directory (OMP `--session` still expects the logical `.jsonl` path)

### Issue reproduction shape (from #2232)

Reporter case:

- Paseo `0.1.110`, OMP `17.0.5`
- Parent status `idle` while child files under session stem continued updating
- Healthy running agents still had existing parent `.jsonl` handles; the broken case had **missing parent file + live stem directory**

### Code-path evidence (this repo)

Before this PR, turn completion was solely:

```ts
if (!state.isStreaming && !state.isCompacting) {
  this.completeTurn(...)
}
```

OMP subagent events were already mapped (`subagent_lifecycle` / `subagent_progress` → `provider_subagent`), but **never consulted** when deciding whether the parent turn was done.

After this PR:

```ts
if (!state.isStreaming && !state.isCompacting && !(await this.hasRunningOmpSubagents())) {
  this.completeTurn(...)
}
```

`hasRunningOmpSubagents()` reconciles via `get_subagents`, then checks the index.

### Upstream OMP evidence

OMP v17.0.5:

- `get_state` exposes `isStreaming` / `sessionFile` but **not** running children
- Separate RPC: `get_subagents` → `{ subagents: [...] }` with per-child status
- Artifacts dir = parent session file with `.jsonl` stripped

So Paseo was missing a signal that OMP already exposes.

### Automated (ran locally)

```bash
npx vitest run \
  packages/server/src/server/agent/providers/omp/agent.test.ts \
  packages/server/src/server/agent/providers/omp/subagent-index.test.ts \
  packages/server/src/server/agent/providers/omp/history-mapper.test.ts \
  packages/server/src/server/agent/providers/omp/cli-runtime.test.ts \
  packages/server/src/server/agent/providers/omp/runtime.test.ts \
  --bail=1
```

**38 tests passed**, including new coverage:

| Test | What it proves |
|---|---|
| `stays active while OMP internal task subagents are still running` | Parent `agent_end` + `isStreaming=false` + running lifecycle child ⇒ **no** `turn_completed` until child completes |
| `stays active when get_subagents reports running children without prior events` | Even without prior lifecycle events, `get_subagents` alone can block idle and later release it |
| `hasRunning` / `reconcileSnapshots` unit tests | Index tracks running/terminal correctly |
| `discovers child transcripts when parent session file is missing` | Orphan stem-dir children still appear in history |

Also:

- `npm run typecheck` (pre-commit / workspace)
- `npm run lint` on touched files

### Manual verification checklist

- [ ] Start an OMP agent that fans out multiple `task` subagents (audit/research style)
- [ ] While children are still writing, confirm parent stays `running` in list/inspect
- [ ] After all children complete, parent becomes `idle`
- [ ] Interrupt while children run: parent cancels cleanly, does not stick forever
- [ ] Optional: inspect history when parent `.jsonl` is absent but stem dir has child jsonls

## Problem solved

| Before | After |
|---|---|
| Parent looks idle/stopped while OMP task children still run | Parent stays non-idle until children are not running |
| Easy to interrupt/restart work that is still live | Status reflects real in-flight fan-out work |
| Missing parent `.jsonl` ⇒ empty history | Child transcripts under stem dir still surface |

Closes #2232

### DIFF
--- a/packages/server/src/server/agent/providers/omp/agent.test.ts
+++ b/packages/server/src/server/agent/providers/omp/agent.test.ts
@@ -245,6 +245,111 @@ describe("OMP agent client and session", () => {
     await expect(completion).resolves.toMatchObject({ finalText: "first done" });
   });
 
+  // Repro/regression for https://github.com/getpaseo/paseo/issues/2232:
+  // OMP can end the parent model turn (agent_end + isStreaming=false) while
+  // internal `task` subagents are still writing. Paseo must not complete the
+  // turn / flip idle until those children finish.
+  test("stays active while OMP internal task subagents are still running", async () => {
+    const scheduler = new ManualIdleScheduler();
+    const omp = new OmpHarness({ providerIdleScheduler: scheduler });
+    await omp.start();
+
+    await omp.requireStartTurn("critically audit the entire repo");
+    const runtime = omp.runtime();
+    runtime.beginTurn();
+    runtime.acceptPrompt("critically audit the entire repo", "user-audit");
+    runtime.streamAssistantText("spawning fan-out workers");
+    runtime.emit({
+      type: "subagent_lifecycle",
+      payload: {
+        id: "ApiBudgetAudit",
+        agent: "ApiBudgetAudit",
+        description: "audit API budget",
+        status: "started",
+        parentToolCallId: "task-1",
+        index: 0,
+      },
+    });
+    // Parent model loop is idle; child work continues (issue #2232 shape).
+    runtime.state = { ...runtime.state, isStreaming: false, isCompacting: false };
+    runtime.finishTurn({
+      role: "assistant",
+      content: [{ type: "text", text: "spawning fan-out workers" }],
+    });
+
+    await omp.waitForProviderStateChecks(1);
+    await scheduler.waitForWaits(1);
+    expect(omp.completedTurnCount()).toBe(0);
+    expect(omp.subagentUpserts()).toContainEqual({ id: "ApiBudgetAudit", status: "running" });
+
+    // Still busy after another idle poll while the child is running.
+    scheduler.retry();
+    await omp.waitForProviderStateChecks(2);
+    await scheduler.waitForWaits(2);
+    expect(omp.completedTurnCount()).toBe(0);
+
+    runtime.emit({
+      type: "subagent_lifecycle",
+      payload: {
+        id: "ApiBudgetAudit",
+        agent: "ApiBudgetAudit",
+        status: "completed",
+        parentToolCallId: "task-1",
+        index: 0,
+      },
+    });
+    scheduler.retry();
+    await omp.waitForProviderStateChecks(3);
+    // getState waiters resolve before completeTurn runs on the same tick.
+    await new Promise<void>((resolve) => setImmediate(resolve));
+    expect(omp.completedTurnCount()).toBe(1);
+    expect(omp.subagentUpserts()).toContainEqual({ id: "ApiBudgetAudit", status: "completed" });
+  });
+
+  test("stays active when get_subagents reports running children without prior events", async () => {
+    const scheduler = new ManualIdleScheduler();
+    const omp = new OmpHarness({ providerIdleScheduler: scheduler });
+    await omp.start();
+
+    await omp.requireStartTurn("fan out");
+    const runtime = omp.runtime();
+    runtime.beginTurn();
+    runtime.acceptPrompt("fan out", "user-fanout");
+    runtime.streamAssistantText("delegating");
+    runtime.subagents = [
+      {
+        id: "PipelineFeedAudit",
+        index: 0,
+        agent: "PipelineFeedAudit",
+        status: "running",
+        parentToolCallId: "task-2",
+      },
+    ];
+    runtime.state = { ...runtime.state, isStreaming: false, isCompacting: false };
+    runtime.finishTurn({
+      role: "assistant",
+      content: [{ type: "text", text: "delegating" }],
+    });
+
+    await omp.waitForProviderStateChecks(1);
+    await scheduler.waitForWaits(1);
+    expect(omp.completedTurnCount()).toBe(0);
+
+    runtime.subagents = [
+      {
+        id: "PipelineFeedAudit",
+        index: 0,
+        agent: "PipelineFeedAudit",
+        status: "completed",
+        parentToolCallId: "task-2",
+      },
+    ];
+    scheduler.retry();
+    await omp.waitForProviderStateChecks(2);
+    await new Promise<void>((resolve) => setImmediate(resolve));
+    expect(omp.completedTurnCount()).toBe(1);
+  });
+
   test("stays active when OMP state checks fail", async () => {
     const scheduler = new ManualIdleScheduler();
     const omp = new OmpHarness({ providerIdleScheduler: scheduler });
--- a/packages/server/src/server/agent/providers/omp/agent.ts
+++ b/packages/server/src/server/agent/providers/omp/agent.ts
@@ -2166,7 +2166,11 @@ export class OmpAgentSession implements AgentSession {
       try {
         const state = await this.runtimeSession.getState();
         this.state = state;
-        if (!state.isStreaming && !state.isCompacting) {
+        // Parent model loop idle is not enough: OMP-internal `task` children can
+        // keep running after agent_end / isStreaming=false (hub waits, fan-out).
+        // Completing the Paseo turn here marks the agent idle while work continues
+        // (https://github.com/getpaseo/paseo/issues/2232).
+        if (!state.isStreaming && !state.isCompacting && !(await this.hasRunningOmpSubagents())) {
           this.completeTurn(turnId, messages);
           return;
         }
@@ -2175,6 +2179,24 @@ export class OmpAgentSession implements AgentSession {
       }
       await this.providerIdleScheduler.waitForRetry();
     }
+  }
+
+  /**
+   * Keep the event index in sync with OMP's `get_subagents` while waiting for
+   * idle. Events alone can miss frames; the index alone can go stale without a
+   * terminal lifecycle event. Reconcile every poll so either path can settle.
+   */
+  private async hasRunningOmpSubagents(): Promise<boolean> {
+    try {
+      const snapshots = await this.runtimeSession.getSubagents();
+      for (const event of this.subagentIndex.reconcileSnapshots(this.runtimeSession, snapshots)) {
+        this.emit(event);
+      }
+    } catch (error) {
+      // Older OMP binaries may not expose get_subagents; event index is best-effort.
+      this.logger.debug({ err: error }, "OMP get_subagents unavailable during idle gate");
+    }
+    return this.subagentIndex.hasRunning(this.runtimeSession);
   }
 
   private async refreshState(): Promise<void> {
--- a/packages/server/src/server/agent/providers/omp/cli-runtime.ts
+++ b/packages/server/src/server/agent/providers/omp/cli-runtime.ts
@@ -23,6 +23,7 @@ import {
   OmpRuntimeEventSchema,
   OmpSessionStateSchema,
   OmpSessionStatsSchema,
+  OmpSubagentsResultSchema,
   type OmpThinkingLevel,
   type OmpAgentMessage,
   type OmpModel,
@@ -35,6 +36,7 @@ import {
   type OmpRuntimeEvent,
   type OmpSessionState,
   type OmpSessionStats,
+  type OmpSubagentSnapshot,
   type OmpSubagentSubscriptionLevel,
 } from "./rpc-types.js";
 
@@ -200,6 +202,11 @@ class OmpCliRuntimeSession implements OmpRuntimeSession {
 
   async setSubagentSubscription(level: OmpSubagentSubscriptionLevel): Promise<void> {
     await this.request({ type: "set_subagent_subscription", level });
+  }
+
+  async getSubagents(): Promise<OmpSubagentSnapshot[]> {
+    const data = OmpSubagentsResultSchema.parse(await this.request({ type: "get_subagents" }));
+    return data.subagents ?? [];
   }
 
   async setHostTools(tools: OmpRpcHostToolDefinition[]): Promise<string[]> {
--- a/packages/server/src/server/agent/providers/omp/history-mapper.test.ts
+++ b/packages/server/src/server/agent/providers/omp/history-mapper.test.ts
@@ -644,4 +644,57 @@ describe("OMP history mapper", () => {
       ]),
     );
   });
+
+  test("discovers child transcripts when parent session file is missing", async () => {
+    const dir = mkdtempSync(join(tmpdir(), "omp-orphan-parent-"));
+    const parentFile = join(dir, "parent.jsonl");
+    const parentStem = parentFile.slice(0, -".jsonl".length);
+    const childFile = join(parentStem, "ApiBudgetAudit.jsonl");
+    mkdirSync(parentStem, { recursive: true });
+    writeFileSync(
+      childFile,
+      [
+        { type: "session", id: "child-root", parentId: null, timestamp: "2026-07-19T20:00:00Z" },
+        {
+          type: "message",
+          id: "child-answer",
+          parentId: "child-root",
+          timestamp: "2026-07-19T20:00:01Z",
+          message: { role: "assistant", content: [{ type: "text", text: "audit ok" }] },
+        },
+      ]
+        .map((entry) => JSON.stringify(entry))
+        .join("\n"),
+    );
+
+    const events: AgentStreamEvent[] = [];
+    for await (const event of streamOmpHistory({ sessionFile: parentFile, provider: "omp" })) {
+      events.push(event);
+    }
+    const subagentEvents = events.flatMap((event) =>
+      event.type === "provider_subagent" ? [event.event] : [],
+    );
+    expect(subagentEvents).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          type: "upsert",
+          id: "ApiBudgetAudit",
+          status: "running",
+        }),
+        expect.objectContaining({
+          type: "timeline",
+          id: "ApiBudgetAudit",
+          item: expect.objectContaining({
+            type: "assistant_message",
+            text: "audit ok",
+          }),
+        }),
+        expect.objectContaining({
+          type: "upsert",
+          id: "ApiBudgetAudit",
+          status: "completed",
+        }),
+      ]),
+    );
+  });
 });
--- a/packages/server/src/server/agent/providers/omp/history.ts
+++ b/packages/server/src/server/agent/providers/omp/history.ts
@@ -1,4 +1,4 @@
-import { readFile } from "node:fs/promises";
+import { readdir, readFile, stat } from "node:fs/promises";
 import { basename, extname, join } from "node:path";
 import type { AgentProvider, AgentStreamEvent } from "../../agent-sdk-types.js";
 import { normalizeProviderReplayTimestamp } from "../../provider-history-timestamps.js";
@@ -63,6 +63,12 @@ export async function* streamOmpHistory(input: {
     );
   } catch (error) {
     if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+      // Parent `.jsonl` can be missing while OMP still keeps child transcripts under
+      // the stem artifacts directory (issue #2232). Surface those children instead
+      // of returning an empty history.
+      for (const transcript of await discoverOrphanSubagentTranscripts(input.sessionFile)) {
+        yield* replaySubagentTranscript(transcript, input.provider, visitedSessionFiles);
+      }
       return;
     }
     throw error;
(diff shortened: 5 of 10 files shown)


ABorakati:
Reviewed this against a live OMP v17 session while chasing the same issue, and found two gaps. The idle-gate half of #2232 works — parent stays non-idle. The **subagent-card half does not**, and I think the test in this PR hides it.

## 1. The card still terminalizes on dispatch

`handleToolExecutionEnd` deletes the call from `activeToolCalls`, then decides whether to hold the card open by asking `subagentCardTracker.hasRunningItems`. That tracker is empty at that moment, and necessarily so.

OMP's `task` defaults to `async.enabled=true` — RPC mode resets `async.*` to built-in defaults, so a Paseo-hosted session always gets it — which makes `tool_execution_end` a *dispatch acknowledgement*, not a result. Measured on a real RPC session at subscription level `events` (the level the adapter subscribes at), batch of 2 detached spawns:

| frame | t |
|---|---|
| `tool_execution_start` | +13109ms |
| `tool_execution_end` | +13591ms |
| first `subagent_lifecycle: started` | **+15927ms** (2336ms *after* the result) |
| children finish | ~28.4s later |

At subscription level `progress` the first frame was +940ms after the tool end. **No child frame ever precedes `tool_execution_end` for a detached spawn.**

So the deferral only engages in the ordering that never occurs. The non-deferred branch runs, the card flips to `completed` with log text literally `"Spawned N background agents"`, and every later child frame then fails the `payload.parentToolCallId && this.activeToolCalls.has(...)` guard and is dropped — so the card never gains child rows either. `settleDeferredTaskCall` is only called from inside that same guard, so once the non-deferred branch runs the card is unrecoverable.

Parent idle is unaffected, since `subagentIndex.handleLifecycle` sits outside that guard. This is the card channel only.

**The regression test encodes the wrong order.** `agent.test.ts` → *"holds a task tool call open until its OMP children finish"* emits `subagent_lifecycle: started` **before** `tool_execution_end`, which is the reverse of the wire order above. It passes against the broken path. Worth flipping regardless of how the fix lands — a fixture in wire order fails today.

## 2. `reconcileSnapshots` can confirm "running" but never clear

`get_subagents` returns **only still-running children**. A finished child disappears from the reply rather than reporting a terminal status — verified directly: once the child completed, the reply was `{"subagents":[]}`, never a terminal entry.

`reconcileSnapshots` iterates only the ids present in the reply, so an id that *disappears* is never visited and never cleared. Combined with the narrow-only `if (state.status === "running")` guard, the merge is one-directional by construction. The header comment says it covers "missed event frames", but only in one direction: if the terminal `subagent_lifecycle`/`subagent_progress` frame is dropped, `hasRunning` stays true forever and `completeTurnAfterProviderIdle` loops forever — reintroducing the #2260/#2281 stuck-running class this repo just got out of.

Fix shape: treat absence from a *successful* reply as terminal for ids the index holds as `running`, gated on the child having been listed by some earlier reply — otherwise a child whose lifecycle frame beat its first snapshot gets terminalized the moment it starts.

Related, smaller: `statesByParent` is keyed on the runtime session with no turn dimension and is cleared only by `clear()`/`terminalizeRunning`, so a child outliving its turn holds *every later turn* non-idle.

## Patch

I have both fixed, with tests that are red before / green after:

- Defer every non-error `task` call **unconditionally** and settle off `OmpSubagentIndex` instead of the card tracker — the index is fed outside the `activeToolCalls` guard, so it's the only channel that sees late frames. Settling requires both a linked child and none running (an index that knows nothing yet must not read as done), with `completeTurn` force-settling as a backstop for a `task` that never produced a child. Keeping the id in `activeToolCalls` is what re-opens the card channel.
- Sweep absent-but-previously-listed ids in `reconcileSnapshots`, gated on a `seenInSnapshot` flag.
- Reorder the fixture to wire order; add coverage for late progress, the absence sweep, and the never-listed child.

Verification: stashing only the source files reproduces the bug exactly —

```
FAIL > holds a task tool call open when its OMP children appear after the result
FAIL > routes late child progress into the deferred task card
  AssertionError: expected 'completed' to be 'running'
```

`npx vitest run packages/server/src/server/agent/providers/omp/` → 19 files, 122 tests pass. Typecheck and lint clean. Also dogfooded in the desktop dev app against a locally built daemon: card now stays `running` with live child rows for the whole fan-out, and card + spinner resolve together when the last child finishes.

One caveat on applying it: my commit sits on top of a local follow-up commit of ours (deferral scaffolding, `ensureLiveTurnStarted`, harness helpers) rather than directly on this PR's head, so it won't cherry-pick cleanly onto `fix/2232-omp-idle-while-task-subagents` as-is. Happy to rebase it onto your branch and open it as a stacked PR, or hand you the hunks to fold in — whichever you prefer, @jasonhnd. The two behavioural changes are small; most of the diff is tests.

## One open question worth a maintainer call

`detached` is parsed in three schemas (`OmpSubagentLifecyclePayloadSchema`, `OmpSubagentProgressPayloadSchema`, `OmpSubagentSnapshotSchema`) and **read nowhere**. It's the only signal separating "child belongs to this turn" from "child intentionally outlives it", and the direction isn't obvious:

- Exclude `detached` from the idle gate → #2232's idle-while-working returns for exactly the long fan-out audits that reported it.
- Include it → a detached child can pin an agent `running` indefinitely and block every later turn.

I deliberately left that alone rather than pick semantics unilaterally. Worth deciding before more logic accretes on the gate.

ABorakati:
```diff
diff --git a/packages/server/src/server/agent/providers/omp/agent.ts b/packages/server/src/server/agent/providers/omp/agent.ts
index df838cfdb..a396df940 100644
--- a/packages/server/src/server/agent/providers/omp/agent.ts
+++ b/packages/server/src/server/agent/providers/omp/agent.ts
@@ -1664,11 +1664,11 @@ export class OmpAgentSession implements AgentSession {
         this.subagentCardTracker.handleLifecycle(payload, (toolCallId) =>
           this.emitActiveToolCall(toolCallId),
         );
-        this.settleDeferredTaskCall(payload.parentToolCallId);
       }
       for (const mapped of this.subagentIndex.handleLifecycle(this.runtimeSession, payload)) {
         this.emit(mapped);
       }
+      this.settleDeferredTaskCalls();
       return true;
     }
     if (event.type === "subagent_progress") {
@@ -1677,11 +1677,11 @@ export class OmpAgentSession implements AgentSession {
         this.subagentCardTracker.handleProgress(payload, (toolCallId) =>
           this.emitActiveToolCall(toolCallId),
         );
-        this.settleDeferredTaskCall(payload.parentToolCallId);
       }
       for (const mapped of this.subagentIndex.handleProgress(this.runtimeSession, payload)) {
         this.emit(mapped);
       }
+      this.settleDeferredTaskCalls();
       return true;
     }
     if (event.type === "subagent_event") {
@@ -1940,21 +1940,25 @@ export class OmpAgentSession implements AgentSession {
     const result = parseToolResult(event.result);
     const error = event.isError ? event.result : null;
     const status = event.isError ? "failed" : "completed";
+    if (event.toolName === "task" && !event.isError) {
+      // OMP's `task` defaults to `async.enabled=true` (RPC mode resets it to the
+      // built-in default), so this result means "the children were dispatched",
+      // not "the children finished" — and the first child frame lands *after* it
+      // (measured ~2.3s at subscription level "events"). Gating the deferral on
+      // already-observed children therefore never fires for the case it exists
+      // for. Defer unconditionally and let the subagent index — which receives
+      // late frames outside the `activeToolCalls` guard — decide when the call
+      // is really done. Re-adding the id to `activeToolCalls` is what re-opens
+      // the card channel so those frames still reach the tracker.
+      this.deferredTaskCompletions.set(event.toolCallId, { toolCall, result, error, status });
+      this.activeToolCalls.set(event.toolCallId, toolCall);
+      this.emitToolCallEvent(event.toolCallId, toolCall, "running", result, null);
+      this.settleDeferredTaskCall(event.toolCallId);
+      return;
+    }
+    this.emitToolCallEvent(event.toolCallId, toolCall, status, result, error);
     if (event.toolName === "task") {
-      // OMP's batched `task` resolves once the children are dispatched, so this
-      // result is not their completion. Hold the Paseo tool call open — and keep
-      // it in `activeToolCalls` so later child frames still route to its card —
-      // until every child reports a terminal status.
-      if (this.subagentCardTracker.hasRunningItems(event.toolCallId)) {
-        this.deferredTaskCompletions.set(event.toolCallId, { toolCall, result, error, status });
-        this.activeToolCalls.set(event.toolCallId, toolCall);
-        this.emitToolCallEvent(event.toolCallId, toolCall, "running", result, null);
-        return;
-      }
-      this.emitToolCallEvent(event.toolCallId, toolCall, status, result, error);
       this.subagentCardTracker.delete(event.toolCallId);
-    } else {
-      this.emitToolCallEvent(event.toolCallId, toolCall, status, result, error);
     }
     if (event.toolName === "todo") {
       const item = mapOmpTodoToolResult(result);
@@ -1971,9 +1975,20 @@ export class OmpAgentSession implements AgentSession {
    * finished. Emits before dropping the card so the final tool call still
    * carries the accumulated child log.
    */
-  private settleDeferredTaskCall(toolCallId: string): void {
+  private settleDeferredTaskCall(toolCallId: string, force = false): void {
     const deferred = this.deferredTaskCompletions.get(toolCallId);
-    if (!deferred || this.subagentCardTracker.hasRunningItems(toolCallId)) {
+    if (!deferred) {
+      return;
+    }
+    // Before the first child frame arrives the index knows nothing about this
+    // call, which must not read as "done" — require a linked child *and* no
+    // running ones. `force` is the turn-end backstop for a `task` that never
+    // produced a child at all.
+    if (
+      !force &&
+      (!this.subagentIndex.knowsToolCall(this.runtimeSession, toolCallId) ||
+        this.subagentIndex.hasRunningForToolCall(this.runtimeSession, toolCallId))
+    ) {
       return;
     }
     this.deferredTaskCompletions.delete(toolCallId);
@@ -1988,6 +2003,12 @@ export class OmpAgentSession implements AgentSession {
     this.subagentCardTracker.delete(toolCallId);
   }
 
+  private settleDeferredTaskCalls(force = false): void {
+    for (const toolCallId of this.deferredTaskCompletions.keys()) {
+      this.settleDeferredTaskCall(toolCallId, force);
+    }
+  }
+
   /**
    * OMP can resume model work after Paseo already completed a turn — a parked
    * hub wait releasing, or autonomous follow-up work. Without a `turn_start` for
@@ -2214,6 +2235,10 @@ export class OmpAgentSession implements AgentSession {
   }
 
   private completeTurn(turnId: string | undefined, messages: OmpAgentMessage[]): void {
+    // The idle gate already waited for every OMP child to leave `running`, so a
+    // still-deferred `task` here produced no child at all (an instant no-op or a
+    // spawn OMP never reported). Settle it rather than leaking an open card.
+    this.settleDeferredTaskCalls(true);
     this.activeTurnId = null;
     this.activeClientMessageId = null;
     this.activeAssistantMessageId = null;
diff --git a/packages/server/src/server/agent/providers/omp/subagent-card-tracker.ts b/packages/server/src/server/agent/providers/omp/subagent-card-tracker.ts
index d742c1067..7034c4c50 100644
--- a/packages/server/src/server/agent/providers/omp/subagent-card-tracker.ts
+++ b/packages/server/src/server/agent/providers/omp/subagent-card-tracker.ts
@@ -159,25 +159,6 @@ export class OmpSubagentCardTracker {
     return baseDetail.actions ? { ...detail, actions: baseDetail.actions } : detail;
   }
 
-  /**
-   * True while any tracked child of this `task` call is still in-flight. OMP's
-   * batched `task` returns once children are dispatched, so the tool result is
-   * not a completion signal — the card has to stay open until this goes false.
-   */
-  hasRunningItems(toolCallId: string): boolean {
-    const state = this.states.get(toolCallId);
-    if (!state) {
-      return false;
-    }
-    for (const item of state.items.values()) {
-      const status = item.status ?? "running";
-      if (status === "pending" || status === "running") {
-        return true;
-      }
-    }
-    return false;
-  }
-
   flush(toolCallId: string): void {
     const state = this.states.get(toolCallId);
     if (!state || !state.dirty) {
diff --git a/packages/server/src/server/agent/providers/omp/subagent-index.ts b/packages/server/src/server/agent/providers/omp/subagent-index.ts
index 5642e5451..7f9dbd386 100644
--- a/packages/server/src/server/agent/providers/omp/subagent-index.ts
+++ b/packages/server/src/server/agent/providers/omp/subagent-index.ts
@@ -17,6 +17,8 @@ interface OmpSubagentState {
   status: "running" | "completed" | "failed" | "canceled";
   /** Serialized last upsert forwarded from a `get_subagents` poll, for dedupe. */
   lastReconciled: string | null;
+  /** Whether a successful `get_subagents` reply has ever listed this child. */
+  seenInSnapshot: boolean;
   mapper: OmpHistoryMapper;
 }
 
@@ -99,6 +101,43 @@ export class OmpSubagentIndex {
     return false;
   }
 
+  /**
+   * True while a child of this specific `task` tool call is still in-flight.
+   * A child whose parent link has not arrived yet (`toolCallId` still null —
+   * `handleEvent` can learn a child before any frame names its parent) counts
+   * as in-flight for every open call, so a partially populated index can never
+   * settle a card early.
+   */
+  hasRunningForToolCall(parent: object, toolCallId: string): boolean {
+    const states = this.statesByParent.get(parent);
+    if (!states) {
+      return false;
+    }
+    for (const state of states.values()) {
+      if (state.status !== "running") {
+        continue;
+      }
+      if (state.toolCallId === null || state.toolCallId === toolCallId) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /** True once any child has been linked to this `task` tool call. */
+  knowsToolCall(parent: object, toolCallId: string): boolean {
+    const states = this.statesByParent.get(parent);
+    if (!states) {
+      return false;
+    }
+    for (const state of states.values()) {
+      if (state.toolCallId === toolCallId) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   /**
    * Merge a `get_subagents` snapshot into local state so missed event frames
    * do not leave idle detection blind.
@@ -116,6 +155,7 @@ export class OmpSubagentIndex {
     const events: AgentStreamEvent[] = [];
     for (const snapshot of snapshots) {
       const state = this.stateFor(parent, snapshot.id, snapshot.agent);
+      state.seenInSnapshot = true;
       state.title = snapshot.agent || state.title;
       state.description = snapshot.description ?? state.description;
       state.toolCallId = snapshot.parentToolCallId ?? state.toolCallId;
@@ -136,6 +176,28 @@ export class OmpSubagentIndex {
       state.lastReconciled = descriptor;
       events.push(event);
     }
+    // OMP's `get_subagents` reply lists only still-running children: a finished
+    // child disappears rather than reporting a terminal status, so absence is
+    // the only terminal signal this channel gives. Without this sweep the merge
+    // is one-directional — it can confirm "still running" but never clear it,
+    // and a dropped terminal frame wedges the idle gate open forever (the
+    // #2260/#2281 stuck-running class). Only sweep ids a snapshot previously
+    // listed, so a child whose lifecycle frame beat its first snapshot is not
+    // terminalized on arrival.
+    const states = this.statesByParent.get(parent);
+    if (!states) {
+      return events;
+    }
+    const present = new Set(snapshots.map((snapshot) => snapshot.id));
+    for (const [id, state] of states) {
+      if (state.status !== "running" || !state.seenInSnapshot || present.has(id)) {
+        continue;
+      }
+      state.status = "completed";
+      const event = this.upsert(id, state.status, state);
+      state.lastReconciled = JSON.stringify(event.event);
+      events.push(event);
+    }
     return events;
   }
 
@@ -154,6 +216,7 @@ export class OmpSubagentIndex {
       toolCallId: null,
       status: "running",
       lastReconciled: null,
+      seenInSnapshot: false,
       mapper: new OmpHistoryMapper("omp", [], OMP_HISTORY_MAPPER_HOOKS),
     };
     states.set(id, state);
```

The two behavioural changes are the `task` branch in `handleToolExecutionEnd` and the sweep at the end of `reconcileSnapshots`; everything else is the plumbing those two need (`knowsToolCall`/`hasRunningForToolCall`, the `seenInSnapshot` flag, moving settlement outside the `activeToolCalls` guard, and the `completeTurn` backstop). `hasRunningItems` is deleted because those two call sites were its only callers.

jasonhnd:
Replaced by #3371.

This PR had the right idle-gate idea, but review found two defects that made it unsafe to land: the card tests used the reverse of the OMP v17 wire order, and snapshot reconcile never treated a disappeared `get_subagents` id as terminal (risk of a stuck-running parent).

#3371 rewrites both halves on current `main`: idle gate + two-way snapshot merge, and deferred `task` cards with true wire-order tests. Please use that PR instead of this one.

PR #2256, opened 2026-07-20
feat(opencode): switch the effective model when selecting a model-bound agent

shrimpwtf:
## What

When an OpenCode agent binds a specific model (e.g. a plugin binds `atlas`→Kimi, `sisyphus`→Opus), selecting that agent now switches the effective model the same way the OpenCode TUI does. Previously the composer's model picker kept the old model after a mode switch.

## Why it's scoped this way

This is split out of #2188 into a single, self-contained feature so it can be reviewed and merged on its own. It deliberately avoids the two things flagged in review:

- **No provider-specific branches.** There is no `if (provider === "opencode")` anywhere. `AgentManager.setAgentMode()` reads the generic `AgentSession.getRuntimeInfo().model` and surfaces it — the same path every provider already implements. An agent-manager unit test proves this with a **non-opencode (codex) fake**.
- **No opencode details in the protocol.** The only protocol change is an additive, optional, provider-agnostic `AgentMode.model?: string` (`"providerID/modelID"`). `packages/protocol` contains zero opencode references. Providers that don't bind a model leave it unset and are unaffected.

Every OpenCode-specific line lives inside `opencode-agent.ts`.

## Commits (3, dependency-ordered)

1. `feat(protocol)` — additive optional `AgentMode.model` + a both-direction wire-compat test.
2. `feat(opencode)` — the provider reads each agent's bound model, applies it on `setMode`, and lets an explicit `setModel` override win until the next switch (all inside the provider).
3. `feat(agent-manager)` — on a mode switch, read the session's resulting `runtimeInfo.model` and surface it into `config.model` / `runtimeInfo.model` so the client model picker updates.

## Testing

- **Unit** — protocol wire-compat (old/new both parse), provider binding (apply on switch, explicit-override precedence, no-op when the agent has no bound model), and provider-agnostic agent-manager surfacing (codex fake).
- **End-to-end against a real local OpenCode** — selecting `Sisyphus`↔`Atlas` flips the effective model between their bound models and back.

No client change is required — the app already renders `runtimeInfo.model` / `config.model`.

### DIFF
--- a/packages/protocol/src/agent-types.ts
+++ b/packages/protocol/src/agent-types.ts
@@ -67,6 +67,7 @@ export interface AgentMode {
   description?: string;
   icon?: string;
   colorTier?: string;
+  model?: string;
 }
 
 export type ProviderStatus = "ready" | "loading" | "error" | "unavailable";
--- a/packages/protocol/src/messages.agent-mode-model.test.ts
+++ b/packages/protocol/src/messages.agent-mode-model.test.ts
@@ -1,0 +1,31 @@
+import { describe, expect, it } from "vitest";
+import { ProviderSnapshotEntrySchema } from "./messages";
+
+// AgentMode.model is an additive, optional field. These tests pin both-direction
+// back-compat: a new daemon may send a mode carrying `model`, and an old-shaped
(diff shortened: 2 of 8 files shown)


boudra:

PR #2319, opened 2026-07-22
Preserve wide characters in terminal captures and restores

### LINKED ISSUE #2193: bug: terminal snapshots and capture insert spaces between CJK wide characters
### What's broken

Paseo changes CJK wide-character output while serializing terminal state: a PTY that emits `中文` is captured as `中 文`, with a literal ASCII space between the two characters. The emitted bytes contain no space.

This is reproducible on the current release and is not caused by ANSI stripping: `paseo terminal capture --ansi` produces the same result.

**Expected:** terminal capture and restored terminal snapshots preserve `中文` exactly.

**Actual:** the captured line is `中 文`.

### Steps to reproduce

1. Create a terminal and retain its ID:

   ```sh
   paseo terminal create --cwd /tmp --name cjk-repro --json
   ```

2. Send an ASCII-only command that emits the UTF-8 bytes for `中文`, followed by Enter:

   ```sh
   paseo terminal send-keys "$TERMINAL_ID" "printf '\\344\\270\\255\\346\\226\\207\\n'" Enter
   ```

3. Capture the output line:

   ```sh
   paseo terminal capture "$TERMINAL_ID" --start 1 --end 1 --json
   ```

Raw result from a fresh terminal on this machine:

```json
{
  "lines": [
    "中 文"
  ],
  "totalLines": 24
}
```

The same result occurs with `--ansi`.

### Where did this happen

CLI

The affected state is also the terminal snapshot format used by clients. [INFERENCE] Snapshot restoration may therefore show the same inserted spaces after a terminal state replay.

### Paseo version

0.1.110 (the current stable release at the time of reporting)

### OS version

Darwin 24.6.0 / macOS, Intel x64

### Agent provider

Not relevant

### Provider configuration

Not applicable.

### Logs

No daemon error is emitted. The raw `paseo terminal capture` output above is the failing observable result.

### Screenshots or video

Not applicable: this is reproducible through the CLI.

### Additional implementation context

This looks like wide-cell continuation information is lost during terminal-state serialization:

- The daemon extracts every xterm cell using `cell.getChars() || " "`, so an empty continuation cell of a width-2 CJK glyph becomes a real space: [`terminal.ts`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/terminal.ts).
- Capture then concatenates every `cell.char`: [`terminal-capture.ts`](https://github.com/getpaseo/paseo/blob/main/packages/server/src/terminal/terminal-capture.ts).
- Snapshot ANSI rendering likewise writes every `cell.char`: [`terminal-snapshot.ts`](https://github.com/getpaseo/paseo/blob/main/packages/protocol/src/terminal-snapshot.ts).

A durable fix likely needs the protocol snapshot to preserve xterm cell width / continuation identity (for example, an optional width field for backward compatibility), then skip width-0 continuation cells when converting state to plain text or ANSI. A string-level whitespace replacement would be incorrect because it would erase genuine spaces.

boudra:
### Linked issue

Closes #2193

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Terminal captures and restored snapshots no longer insert spaces between adjacent CJK and other wide characters. The terminal snapshot keeps zero-width continuation cells distinct from ordinary blank cells, and restoration skips those continuations.

### How did you verify it

Reproduced the bug first through a real-daemon terminal capture test, which captured adjacent CJK characters with inserted spaces. Confirmed that test passes after the fix.

- `npx vitest run packages/server/src/server/daemon-e2e/terminal.e2e.test.ts --bail=1 -t "preserves adjacent CJK characters in captured output"`
- `npx vitest run packages/protocol/src/terminal-snapshot.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

The complete terminal E2E file also reached an existing slow-client backpressure timeout. That test fails identically with this fix disabled; the changed CJK test remains green.

### Risk surface

Snapshot cell encoding now retains an empty string for wide-character continuation cells. The existing protocol schema already accepts string values, ordinary blank cells remain spaces, and no wire field was added.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/protocol/src/terminal-snapshot.test.ts
+++ b/packages/protocol/src/terminal-snapshot.test.ts
@@ -7,6 +7,20 @@ function cells(text: string): TerminalState["grid"][number] {
 }
 
 describe("renderTerminalSnapshotToAnsi", () => {
+  it("preserves adjacent wide characters", () => {
+    const state: TerminalState = {
+      rows: 1,
+      cols: 4,
+      scrollback: [],
+      grid: [[{ char: "\u4e2d" }, { char: "" }, { char: "\u6587" }, { char: "" }]],
+      cursor: { row: 0, col: 4 },
+    };
+
+    const ansi = renderTerminalSnapshotToAnsi(state);
+
+    expect(ansi).toContain("\u4e2d\u6587");
+  });
+
   it("renders soft-wrapped rows as one contiguous logical line when wrap flags are present", () => {
     // The server soft-wrapped one logical line "ABCDEFGHIJKLMNOP" at 10 cols into
     // two grid rows. gridWrapped[0] = true marks row 0 as continuing into row 1.
--- a/packages/protocol/src/terminal-snapshot.ts
+++ b/packages/protocol/src/terminal-snapshot.ts
@@ -87,12 +87,15 @@ function renderTerminalRow(row: TerminalCell[], padToCols?: number): string {
 
   for (let index = 0; index < length; index += 1) {
     const cell = row[index] ?? { char: " " };
+    if (cell.char === "") {
+      continue;
+    }
     const nextStyle = getTerminalStyle(cell);
     if (!terminalStylesEqual(previousStyle, nextStyle)) {
       output.push(styleToAnsi(nextStyle));
       previousStyle = nextStyle;
     }
-    output.push(cell.char || " ");
+    output.push(cell.char);
   }
 
   if (!terminalStylesEqual(previousStyle, DEFAULT_STYLE)) {
--- a/packages/server/src/server/daemon-e2e/terminal.e2e.test.ts
+++ b/packages/server/src/server/daemon-e2e/terminal.e2e.test.ts
@@ -1626,6 +1626,29 @@ describe("capture", () => {
     expect(capture.totalLines).toBeGreaterThan(0);
   }, 15000);
 
+  test("preserves adjacent CJK characters in captured output", async () => {
+    const cwd = tmpCwd();
+    tempDirs.push(cwd);
+    const created = await createTerminalInWorkspace(ctx.client, { cwd });
+    const terminalId = created.terminal!.id;
+
+    await ctx.client.subscribeTerminal(terminalId);
+    ctx.client.sendTerminalInput(terminalId, {
+      type: "input",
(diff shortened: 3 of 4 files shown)


EhEo:
Independent confirmation that this is still needed, now against the **0.2.5** published artifacts (the issue was filed against 0.1.110, and #2193's last check was against `main` on 2026-07-31).

I hit this on a fresh `npm i -g @getpaseo/cli@0.2.5` install and arrived at the same two-line server fix as this PR before finding it — `cell.getWidth() === 0 ? "" : cell.getChars() || " "` in both `extractCell()` and `extractScrollback()`. Applying this PR's exact diff to the shipped build output fixes it.

Beyond the `中文` case in #2193, verified across scripts and mixed runs:

```
$ paseo terminal send-keys "$TID" 'echo "한글 가나다  두칸띄움 mixed abc 漢字 テスト ①②③"' Enter

before: 한 글  가 나 다   두 칸 띄 움  mixed abc 漢 字  テ ス ト  ①②③
after:  한글 가나다  두칸띄움 mixed abc 漢字 テスト ①②③
```

Worth noting the genuine double space between `가나다` and `두칸띄움` survives — which is the thing a naive whitespace-squashing fix would break, and this approach doesn't.

Two notes for whoever picks this up:

- **The `packages/protocol` half is not optional.** Fixing only `terminal.ts` moves the problem rather than solving it: the grid then carries `""` cells, and `renderTerminalRow()`'s `cell.char || " "` turns them straight back into spaces on the snapshot-restore path. This PR already has both halves — flagging it only because the two changes look independently landable and aren't.
- **The prebuilt web UI carries its own copy.** `@getpaseo/server`'s bundled `dist/server/web-ui` contains the compiled `renderTerminalRow`, and the daemon serves precompressed `.br`/`.gz` alongside it. Nothing to change in this PR — it falls out of a rebuild — just a heads-up that anyone patching a released install by hand has a third site to touch, and has to regenerate the compressed variants or the browser keeps getting the old code.

Environment: `@getpaseo/cli` 0.2.5, Node v24.19.0, Debian 13 on **linux/arm64** (ChromeOS Crostini). Nothing platform-specific in the logic, so x64 should behave identically — noting the arch only because it's an unusual one to have coverage from.

This PR has been open since 2026-07-22 with a green bot review and no human review yet. Anything blocking it, or is there something I can help with to move it along?

PR #2323, opened 2026-07-22
Show usage limits approaching exhaustion

### LINKED ISSUE #2320: bug: plan usage bars never turn red, no matter how close to the limit
### What's broken

A usage bar in Settings → Usage stays green all the way to 100%. At 99% of the weekly
limit the Claude card looks the same as at 5%, so the one moment the meter matters is the
one moment it says nothing.

Expected: the bar changes colour as it fills.
Actual: Claude and Kimi bars are always green. Codex bars can reach amber but never red.

The client already computes this correctly and is being overridden.
`packages/app/src/provider-usage/window-bar.tsx:29`:

```ts
const tone = window.tone ?? deriveTone(usedPct);
```

`packages/app/src/provider-usage/tone.ts`:

```ts
if (usedPct > 90) return "danger";
if (usedPct >= 70) return "warning";
return "default";
```

`packages/server/src/services/quota-fetcher/usage.ts:64-65` only sets the field when the
provider supplies one:

```ts
if (input.tone) {
  window.tone = input.tone;
}
```

So any provider that passes a tone opts out of the thresholds entirely. Three do:

| File | Line | Value | Effect |
| --- | --- | --- | --- |
| `providers/claude.ts` | 159, 170, 181, 192 | `tone: "ok"` | green at every level |
| `providers/kimi.ts` | 84 | `tone: "ok"` | green at every level |
| `providers/codex.ts` | 146 | `tone: "ok"` | green at every level |
| `providers/codex.ts` | 157, 168 | `usedPct >= 70 ? "warning" : "ok"` | amber past 70, never red |

Balances have the same problem by a different route. `balance-bar.tsx:44` is
`balance.tone ?? "default"` with no `deriveTone` fallback at all, and
`balanceToneFromRemaining` (`quota-fetcher/usage.ts:70-76`) only returns `"danger"` once
`remaining <= 0`. A credits bar at 99% spent is green.

Worth noting the two behaviours disagree with each other: `deriveTone` returns
`"default"` (grey) below 70% while the providers return `"ok"` (green), so simply
removing the hardcoded tones would turn healthy bars grey rather than green. Whatever the
fix, the two need to agree on what healthy looks like.

### Steps to reproduce

1. Use any provider until a window is past 90% (weekly on Claude is the easiest to hit).
2. Open Settings → Usage.
3. The bar is still green.

Alternatively, read it straight from the source: `claude.ts:159` sets `tone: "ok"`
unconditionally, and `window-bar.tsx:29` only calls `deriveTone` when `window.tone` is
absent.

### Where did this happen

Desktop (Electron)

### Paseo version

0.2.0-beta.1

### OS version

macOS 26.5.2

### Agent provider

Not relevant

### Provider configuration

Not provider-specific. Reproduced on Claude (Max 20x, OAuth); the same pattern is in the
Codex and Kimi fetchers by inspection.

### Logs

```text

```

### Screenshots or video

Image

boudra:
### Linked issue

Closes #2320

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Usage meters now change to amber and red as a limit approaches exhaustion. Provider-specific alerts remain visible when they indicate a higher risk.

### How did you verify it

- Added a focused regression for healthy, warning, danger, and provider-alert states.
- Confirmed the regression failed before the fix and passed afterward.
- Ran `npx vitest run packages/app/src/provider-usage/tone.test.ts --bail=1`.
- Ran `npm run typecheck`, `npm run lint`, and `npm run format`.
- A visual smoke check on the supported app platforms remains for review because this environment does not have live near-limit provider data.

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/provider-usage/balance-bar.tsx
+++ b/packages/app/src/provider-usage/balance-bar.tsx
@@ -2,6 +2,7 @@ import { useMemo } from "react";
 import { Text, View, type StyleProp, type ViewStyle } from "react-native";
 import { StyleSheet } from "react-native-unistyles";
 import { clampPct, formatAmount, formatResetLabel } from "./format";
+import { resolveTone } from "./tone";
 import type { ProviderUsageBalance, ProviderUsageTone } from "./types";
 
 interface ResolvedBalance {
@@ -41,7 +42,7 @@ function fillToneStyle(tone: ProviderUsageTone) {
 
 export function ProviderUsageBalanceBar({ balance }: { balance: ProviderUsageBalance }) {
   const { amountText, usedPct } = resolveBalance(balance);
-  const tone = balance.tone ?? "default";
+  const tone = resolveTone(balance.tone, usedPct);
   const resetLabel = formatResetLabel(balance.resetsAt);
 
   const fillStyle = useMemo<StyleProp<ViewStyle>>(
--- a/packages/app/src/provider-usage/tone.test.ts
+++ b/packages/app/src/provider-usage/tone.test.ts
@@ -1,0 +1,13 @@
+import { describe, expect, it } from "vitest";
+import { resolveTone } from "./tone";
+
+describe("provider usage tone", () => {
+  it("shows the strongest risk from provider status and utilization", () => {
+    expect([
+      resolveTone("ok", 5),
+      resolveTone("ok", 70),
+      resolveTone("ok", 99),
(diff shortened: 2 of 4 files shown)


boudra:
fixed in https://github.com/getpaseo/paseo/pull/2322

PR #2341, opened 2026-07-23
feat(desktop): add setting to disable automatic update install/download

### LINKED ISSUE #998: feat: Disable autoupgrade
### Prior search

### What's the problem

I use [Scoop](https://scoop.sh/) to install/update paseo as portable app on Windows. But paseo always automatically installs the latest version out of the scoop. That makes duplicate versions on the computer

Image

### What would solve it

Add an setting option to disable autoupdate

_No response_

iphixit:
Closes #998.

## Problem

The desktop app's updater checks for updates automatically in the background, which is fine — but it also **downloads and installs updates automatically**, with no way to opt out:

- `autoUpdater.autoDownload` is hardcoded `true` (`packages/desktop/src/features/auto-updater.ts`) — no env var or setting disables it.
- Once a download completes in the background, an ordinary app **quit silently installs it** (`installUpdateOnQuit` → `quitAndInstall(isSilent: true)`), with no dialog and no confirmation.
- The only settings knob today is release channel (stable/beta) — nothing controls install behavior.

This is exactly what #998 asks for, and the follow-up comment on that issue asks for the same three things this PR adds: keep checks, but let the user control download/install and install-on-shutdown separately.

## What this adds

A new `autoInstallUpdates` desktop setting (**default: `true`**, so this is non-breaking — existing installs keep today's behavior until a user opts out):

- **On (default):** unchanged from today.
- **Off:** background checks still run silently as before. Nothing is downloaded automatically, and nothing installs on quit — even if a download had somehow already completed. The existing "Update" button in Settings still works exactly as before: an explicit click always checks, downloads, and installs right away, restarting the app. The setting only gates the *unattended* paths (background auto-download, and install-on-quit), not the attended one.

New toggle lives in **Settings → About**, right under the release-channel picker, using the same `Switch` pattern as the existing daemon-management toggles.

## Implementation notes

- `autoInstallUpdates` is threaded through as an **optional** parameter at every layer (main-process settings store, IPC handlers, the update service, the Electron runtime wrapper, and the renderer's updater state machine), defaulting to `true` at each boundary. This keeps the change purely additive — no existing call site or test needed to change shape, since none of them pass the new field.
- Persisted the same way as the existing daemon settings (`desktop-settings.json` under `userData`), migrated forward automatically (unknown/missing field just falls back to the default).
- Added the `en.ts` copy plus translations for the other 7 shipped locales (`ar`, `es`, `fr`, `ja`, `pt-BR`, `ru`, `zh-CN`) since the locale files are typed against `en.ts`'s shape and would otherwise fail typecheck. I'm not a native speaker of most of these — happy to have a maintainer or translator adjust wording.

## Testing

- `packages/desktop`: `app-update-service.test.ts`, `auto-updater.test.ts`, `desktop-settings.test.ts`, `desktop-settings-commands.test.ts`, `daemon-manager.test.ts` — all pass (updated the three main-process settings tests that asserted the full settings shape to include the new default field).
- `packages/app`: `desktop-app-updater.test.ts`, `hooks/use-settings/storage.test.ts` — all pass (updated 4 assertions asserting the exact IPC payload shape / fake settings shape).
- `npm run typecheck` clean across the whole workspace.
- No new install was performed as part of writing this PR — it's a source-level change verified by the existing test suite plus a clean full-workspace typecheck. Happy to record a manual verification (toggle off → background download never fires, toggle back on → behavior matches current `main`) if that's wanted before merge.


iphixit:
Both Greptile P2 comments are addressed in 31c18a8.

**(a) Test coverage for the disabled path**

Added service-level tests in `packages/desktop/src/features/app-update-service.test.ts` for the two unattended paths the setting gates:

- `installUpdateOnQuit` short-circuits when `autoInstallUpdates` is `false` — returns `false`, installs nothing, and returns *before* the quit-time revalidation check even when a download has already completed on disk.
- A background check leaves `autoDownload = false`, so nothing is fetched automatically. The update is still surfaced to the UI (`hasUpdate: true`); it just isn't downloaded unattended.

Both are covered in the enabled direction too, so the tests can't pass for an unrelated reason. `ElectronAppUpdateRuntime` isn't exported, so the observation point is what the service passes to `runtime.configure()` — which the Electron runtime maps directly onto `autoUpdater.autoDownload`. The fake runtime now records it.

These were **mutation-verified** rather than just observed green. Removing the `!autoInstallUpdates` guard from `installUpdateOnQuit` fails the suite, and re-introducing the original hardcoded `autoDownload = true` (the defect this PR exists to fix) fails it as well. So the new tests do protect the behaviour, not just describe it.

**(b) Port signature**

Made `autoInstallUpdates` required on `DesktopAppUpdaterPort` to match the implementation.

Worth flagging that the bot under-reported this one: it pointed at `checkDesktopAppUpdate`, but `installDesktopAppUpdate` had the identical mismatch. Both are fixed.

Required is the right direction here — every call site already resolves the value to a concrete boolean before crossing the port boundary (`autoInstallUpdates = true` in `checkForUpdates`, `?? true` in `installUpdate`), so nothing can reach the port with `undefined` today. Typecheck is clean across both packages.

One deliberate choice, and I'd rather surface it than bury it: I tightened the **port** but left the public `DesktopAppUpdater` API optional, since that's the ergonomic default-bearing surface and the default matches the pre-existing behaviour. The alternative — loosening `desktop-updates.ts` to optional instead — would push the `?? true` defaulting down into more places and duplicate the default. Happy to flip it if you'd prefer the consistency; it's a one-line change either way.

Also corrected the fake port's `recordedChecks` / `recordedInstalls` types, which omitted `autoInstallUpdates` even though the existing tests already assert on that field — it passed only because `toEqual` is a runtime check.

No changes to shipped runtime behaviour in this commit: (a) is tests only, (b) is a type-signature tightening plus a test-utility type correction.

boudra:

Issue #2302, opened 2026-07-21
bug: Claude usage card is missing model-scoped weekly limits


cleiter:
### What's broken

Settings → Usage shows only *Session* and *Weekly* for Claude. The model-scoped weekly
limit (currently Fable) is never shown, even though the account has one and the API
reports it.

Expected: one bar per limit the API returns, the way `/usage` in Claude Code shows them.
Actual: the scoped limit is silently dropped. No error, no "unavailable" state, no log
line. The card looks completely healthy while under-reporting.

The daemon reads four top-level keys from `GET https://api.anthropic.com/api/oauth/usage`:
`five_hour`, `seven_day`, `seven_day_opus`, `seven_day_omelette`
(`packages/server/src/services/quota-fetcher/providers/claude.ts`). Three of those four now
return `null`. The scoped limits moved into a `limits[]` array that nothing reads.

Raw response from that endpoint on my account just now, unedited apart from dropping the
`extra_usage` / `spend` / `member_dashboard_available` keys:

```json
{
  "five_hour": {
    "utilization": 42.0,
    "resets_at": "2026-07-21T14:10:00.268668+00:00",
    "limit_dollars": null,
    "used_dollars": null,
    "remaining_dollars": null
  },
  "seven_day": {
    "utilization": 28.0,
    "resets_at": "2026-07-27T07:00:00.268694+00:00",
    "limit_dollars": null,
    "used_dollars": null,
    "remaining_dollars": null
  },
  "seven_day_oauth_apps": null,
  "seven_day_opus": null,
  "seven_day_sonnet": null,
  "seven_day_cowork": null,
  "seven_day_omelette": null,
  "limits": [
    {
      "kind": "session",
      "group": "session",
      "percent": 42,
      "severity": "normal",
      "resets_at": "2026-07-21T14:10:00.268668+00:00",
      "scope": null,
      "is_active": true
    },
    {
      "kind": "weekly_all",
      "group": "weekly",
      "percent": 28,
      "severity": "normal",
      "resets_at": "2026-07-27T07:00:00.268694+00:00",
      "scope": null,
      "is_active": false
    },
    {
      "kind": "weekly_scoped",
      "group": "weekly",
      "percent": 2,
      "severity": "normal",
      "resets_at": "2026-07-27T07:00:00.268958+00:00",
      "scope": {
        "model": {
          "id": null,
          "display_name": "Fable"
        },
        "surface": null
      },
      "is_active": false
    }
  ]
}
```

[...]

Issue #2182, opened 2026-07-18
bug(grok/acp): hidden background-task reminders render as raw user messages


jasonhnd:
### Summary

When Grok Build completes a background task through ACP, Paseo can render Grok's internal wake-up message as a normal user message. The chat then shows a large raw block such as:

```text
<system-reminder>
Background task "<redacted>" completed (exit code: 0).
Command: <redacted>
</system-reminder>
```

This is not a message entered by the user. In the captured ACP stream, Grok explicitly marks the chunk with `hideFromScrollback: true`, but Paseo currently ignores that metadata and maps the chunk to a normal `user_message` timeline item.

The same ACP stream also contains a structured xAI extension notification for `task_completed`. Paseo logs the vendor notification but does not consume it, so the structured event is lost while the model-only fallback/wake-up message becomes visible.

No screenshot is attached. The redacted wire events below identify the behavior more precisely without exposing local commands, output, paths, or session identifiers.

### Why it appears only sometimes

The issue is event-dependent rather than random:

- normal foreground commands do not produce this completion wake-up path;
- it appears when Grok runs work in the background and later emits a task completion notification;
- Grok then sends a model-facing `user_message_chunk` so the agent can continue, with metadata telling the client not to add it to scrollback;
- Paseo renders that hidden chunk because its generic ACP message mapper does not inspect the xAI metadata.

### Steps to reproduce

1. Run Paseo Desktop with Grok configured as an ACP provider using `grok agent stdio`.
2. Start a Grok session.
3. Ask Grok to perform work that it starts as a background shell task, then continue the conversation while the task is running.
4. Wait for the background task to finish.
5. Observe the conversation when Grok processes the completion event.

### Actual behavior

- A gray user-message bubble appears even though the user did not send it.
- The bubble contains the literal `<system-reminder>...</system-reminder>` wrapper.
- It may expose a long command and task output in the main conversation.
- The structured background-task completion event is not represented as a task/tool notification.

### Expected behavior

- A chunk marked `hideFromScrollback: true` must not become a visible user-message bubble.
- Background-task state should remain visible in an appropriate compact task/tool notification, using Grok's structured `task_backgrounded` / `task_completed` extension data.
- The UI should distinguish completed, failed, signaled/cancelled, and truncated-output states.
- Normal user messages must continue to render unchanged.

### Redacted ACP evidence

The affected session's persisted Grok ACP event stream contains these events in sequence.

First, Grok sends a structured xAI extension notification:

```json
{
[...]

Issue #1390, opened 2026-06-07
bug: context window meter never shows for ACP providers (usage_update dropped)


yinaoxiong:
### What's broken

The composer's context-window meter renders for Claude Code and Codex, but never for any ACP-based provider (agents launched via `--acp`). For those agents the circular context meter simply never appears, even after multiple turns.

Expected: the same context-window meter Claude/Codex show.
Got: no meter at all for ACP providers.

The agent is actually delivering the data — Paseo drops it. The ACP `session/update` handler for `usage_update` is a no-op:

```ts
// packages/server/src/server/agent/providers/acp-agent.ts
private handleUsageUpdate(update: UsageUpdate): void {
  void update;   // size / used discarded here
}
```

`UsageUpdate.size` (total context window) and `UsageUpdate.used` (tokens in context) never reach `contextWindowMaxTokens` / `contextWindowUsedTokens`, so the meter's render guard (`if (max === null || used === null) return null`) hides it permanently.

### Steps to reproduce

1. Configure an ACP provider that emits `usage_update` notifications (any agent started with `--acp` implementing the ACP usage capability).
2. Start a session with it and send a message.
3. Watch the composer — the context-window meter never appears, no matter how many turns run.
4. Compare with a Claude Code or Codex agent in the same build: the meter shows there.

### Raw evidence (captured, not inferred)

I wrote a minimal ACP client probe and spoke raw NDJSON JSON-RPC to one ACP agent (`--acp`), logging every `session/update` notification during one real turn. The agent **does** send `usage_update` with full `size`/`used`:

```json
{"sessionUpdate":"usage_update","used":0,"size":1000000}
{"sessionUpdate":"usage_update","used":40002,"size":1000000,"cost":{"amount":13.21}}
```

Per-turn `session/update` type counts from that probe run:

```
config_option_update: 1
available_commands_update: 2
usage_update: 3          <-- delivered every turn, with size+used
session_info_update: 10
agent_message_chunk: 17
```

So the agent side is fine; the figures arrive every turn and are discarded in `handleUsageUpdate`.

### Where did this happen

Daemon

### Paseo version

0.1.91-beta.2 (also present on `main`)

### OS version

Linux. Not OS-specific — the bug is in the daemon's ACP adapter.

### Agent provider

[...]

PR #2353, opened 2026-07-23
fix(quota): restore Grok Settings usage for current CLI auth/billing

### LINKED ISSUE #2352: bug: Grok Usage card missing plan credits (auth.json + billing schema drift)
### What's broken

Settings → Usage never shows Grok plan credits, even when the host is logged into Grok CLI and the billing API returns a healthy monthly budget.

Expected: a "Monthly credits" balance (used / remaining / limit), same surface as Claude/Codex plan usage.
Actual: Grok is either missing from the card or shows as unavailable / empty. No balances, no error surface that points at the real cause.

This is the **Settings plan-credits** path (`GrokQuotaProvider`), not the composer context-window meter (that's #1390 — ACP `usage_update` no-op for all ACP providers including Grok).

### Root cause (verified on a live host)

Two independent schema drifts in `packages/server/src/services/quota-fetcher/providers/grok.ts`. Either one alone is enough to drop the card; both are present on current Grok CLI installs.

#### 1. Auth: top-level `access_token` vs nested `key`

Code path:

```ts
// readGrokToken()
const auth = GrokAuthSchema.parse(JSON.parse(...)); // { access_token?: string }
return auth.access_token ?? null;
```

Live `~/.grok/auth.json` shape (keys only; secrets redacted):

```json
{
  "https://auth.x.ai::<uuid>": {
    "key": "<JWT>",
    "auth_mode": "...",
    "refresh_token": "...",
    "expires_at": "...",
    "user_id": "...",
    "email": "..."
  }
}
```

There is **no** top-level `access_token`. Token lives at `auth["https://auth.x.ai::<uuid>"].key` (JWT). `readGrokToken()` always returns `null` unless `GROK_API_KEY` / `GROK_TOKEN` is set in the daemon env, so the fetch never runs for normal CLI-login users.

#### 2. Billing: `usage.creditUsage` vs `config.used.val`

Code path:

```ts
const monthlyLimit = resp.config?.monthlyLimit?.val ?? null;
const creditUsage = resp.usage?.creditUsage ?? null;
```

Live `GET https://cli-chat-proxy.grok.com/v1/billing` (200, Bearer = nested `key`, headers as in code):

```json
{
  "config": {
    "monthlyLimit": { "val": 150000 },
    "used": { "val": 37886 },
    "onDemandCap": { "val": 0 },
    "billingPeriodStart": "2026-07-01T00:00:00+00:00",
    "billingPeriodEnd": "2026-08-01T00:00:00+00:00",
    "history": [ ... ]
  }
}
```

There is **no** top-level `usage` object and **no** `creditUsage`. Used credits are `config.used.val`. Even if auth were fixed, `creditUsage` stays `null` — used/remaining stay null (limit alone would partially parse today only if a token were found).

Unit test still mocks the old shape (`usage: { creditUsage: 0 }` in `service.test.ts`), so CI does not catch the live response.

### Steps to reproduce

1. Log in with Grok CLI so `~/.grok/auth.json` has the nested `https://auth.x.ai::<uuid>.key` form (no top-level `access_token`). Do **not** set `GROK_API_KEY` / `GROK_TOKEN` on the daemon.
2. Open Paseo → Settings → Usage.
3. Observe Grok missing / unavailable / empty balances.
4. Sanity-check outside Paseo:
   - Auth shape: nested `key`, no top-level `access_token`.
   - `curl -H "Authorization: Bearer <key>" -H "X-XAI-Token-Auth: xai-grok-cli" https://cli-chat-proxy.grok.com/v1/billing` → 200 with `config.monthlyLimit.val` + `config.used.val`.

### Where did this happen

Daemon (quota-fetcher). UI only renders what the daemon returns.

### Paseo version

Present on current `main` (`packages/server/src/services/quota-fetcher/providers/grok.ts`). Reproduced against a host running a recent desktop/daemon build with Grok CLI auth.

### OS version

macOS. Not OS-specific — auth file shape + billing JSON schema.

### Agent provider

Grok (Settings usage card / `GrokQuotaProvider`). Not the ACP agent runtime path.

### What would solve it

In `GrokQuotaProvider` only (Settings path):

1. **Token resolution**: after env vars, read `~/.grok/auth.json` and accept:
   - top-level `access_token` (legacy, if any install still has it)
   - nested `https://auth.x.ai::*` entry's `key` (current Grok CLI)
2. **Billing parse**: map used credits from `config.used.val` (keep reading `usage.creditUsage` as a fallback for older responses if still relevant). Keep `config.monthlyLimit.val`. Prefer including period start/end as details if cheap.
3. **Tests**: fixture the live shape (`config.used.val` + nested auth) so CI fails if Grok drifts again. Don't only mock the obsolete `usage.creditUsage` form.

### Out of scope / related

| Issue | Path | Status |
| --- | --- | --- |
| **This issue** | Settings → Usage plan credits for Grok | open (this) |
| #1390 | Composer context-window meter for **all** ACP providers (`handleUsageUpdate` no-op) | open, separate |
| #2182 | Grok ACP background-task reminder rendering | open, unrelated |
| #2302 / #2303 | Claude Usage missing Fable / `limits[]` | open, leave alone |
| #1391 | Closed unmerged cleanup PR on the ACP usage path | not a fix for Settings Grok |

Do **not** fold composer meter work into this issue — different code path (`acp-agent.ts` vs `quota-fetcher/providers/grok.ts`).

### Prior search

No open issue covers Grok Settings plan credits / `GrokQuotaProvider` auth+billing drift. Closest Grok issues (#2041, #2199, #2182, #2232) are agent lifecycle / ACP UI, not quota-fetcher.

jasonhnd:
Closes #2352

## What is broken

Settings → **Usage** never shows Grok plan credits for a normal Grok CLI login, even when:

- the host is logged into Grok CLI (`~/.grok/auth.json` present), and
- `GET https://cli-chat-proxy.grok.com/v1/billing` returns a healthy monthly budget.

**Expected:** a `Monthly credits` balance (used / remaining / limit), same surface as Claude/Codex plan usage.  
**Actual:** Grok is missing, unavailable, or shows empty balances — no error that points at the real cause.

This is the **Settings plan-credits** path only (`GrokQuotaProvider` in the daemon quota-fetcher).  
It is **not** the composer context-window meter (that is a separate path: #1390 / ACP `usage_update`).

Issue: #2352

---

## Why it broke (two independent schema drifts)

Either drift alone is enough to drop the card; both are present on current Grok CLI installs. Verified against live host shapes (2026-07-23) and the code on `main`.

### Drift 1 — auth token (`readGrokToken`)

**Old code** only accepted a top-level field:

```ts
const GrokAuthSchema = z.object({ access_token: z.string().optional() });
// ...
return auth.access_token ?? null;
```

**Live `~/.grok/auth.json` shape** (structure only; no secrets):

```json
{
  "https://auth.x.ai::<uuid>": {
    "key": "<JWT>",
    "refresh_token": "...",
    "expires_at": "...",
    "user_id": "...",
    "email": "..."
  }
}
```

There is **no** top-level `access_token`. The token lives at the nested entry’s **`key`**.

So without `GROK_API_KEY` / `GROK_TOKEN` in the daemon env, `readGrokToken()` returned `null`, the billing request never ran, and Usage showed Grok as unavailable.

Env order was already correct and is **unchanged**:

1. `GROK_API_KEY`
2. `GROK_TOKEN`
3. file token from `~/.grok/auth.json`

### Drift 2 — billing response (`fetchUsage`)

**Old code:**

```ts
const monthlyLimit = resp.config?.monthlyLimit?.val ?? null;
const creditUsage = resp.usage?.creditUsage ?? null;
```

**Live `GET /v1/billing` body** (Bearer = nested `key`; headers already correct: `Authorization` + `X-XAI-Token-Auth: xai-grok-cli`):

```json
{
  "config": {
    "monthlyLimit": { "val": 150000 },
    "used": { "val": 37886 },
    "onDemandCap": { "val": 0 },
    "billingPeriodStart": "2026-07-01T00:00:00+00:00",
    "billingPeriodEnd": "2026-08-01T00:00:00+00:00",
    "history": [ ... ]
  }
}
```

There is **no** top-level `usage` / `creditUsage`. Used credits are `config.used.val`.

Even if auth were fixed first, `creditUsage` stayed `null` → used/remaining stayed null (limit alone could partially parse, but the card was still wrong / empty in practice for normal logins).

The unit test previously mocked the **obsolete** shape (`usage: { creditUsage: 0 }`), so CI did not catch the live response.

---

## What this PR changes

**Scope: daemon `GrokQuotaProvider` only** — no wire/protocol schema change, no Settings UI change (UI already renders balances when the daemon returns them).

### Files

| File | Role |
|------|------|
| `packages/server/src/services/quota-fetcher/providers/grok.ts` | primary fix |
| `packages/server/src/services/quota-fetcher/service.test.ts` | coverage for live + legacy shapes |
| `packages/server/src/services/quota-fetcher/manifest.ts` | **not** changed (already registered) |

### 1) Token resolution

Added `extractGrokTokenFromAuth()` used by `readGrokToken()`:

1. env (already outer in `fetchUsage`)
2. top-level `access_token` if non-empty (legacy)
3. nested object with non-empty string `key`, **preferring** keys that start with `https://auth.x.ai::` when multiple exist

Tokens are never logged.

### 2) Billing parse + schema

- Zod schema accepts `config.used.val`
- used = `config.used.val ?? usage.creditUsage` (live first, legacy fallback)
- limit = `config.monthlyLimit.val` (unchanged)
- remaining = `max(0, limit - used)` when both numbers are present
- balance id still `monthly_credits` / label `Monthly credits`

Billing period fields are **not** surfaced in `details` (kept the PR minimal).

---

## What this PR deliberately does **not** touch

| Item | Why |
|------|-----|
| #1390 ACP `handleUsageUpdate` / composer meter | different path; all ACP providers |
| #2182 Grok background-task reminders | unrelated UI |
| #2302 / #2303 Claude Fable / `limits[]` | leave alone |
| App Settings / Usage UI components | already render daemon balances |
| Protocol / wire schemas | no change needed |
| Other quota providers | unchanged |

---

### Done (automated)

```bash
npx vitest run packages/server/src/services/quota-fetcher/service.test.ts --bail=1
# 27 passed
```

New / updated coverage:

1. **Live billing shape** — `config: { monthlyLimit: { val }, used: { val } }` → correct used / remaining / limit, `status: "available"`, balance `monthly_credits`
2. **Nested auth file** — write `~/.grok/auth.json` under test `HOME` with `https://auth.x.ai::… .key`, **no** env token → request uses `Authorization: Bearer <nested key>` and returns balances
3. **Zero values** preserved on live shape (`used: 0`, `limit: 0`)
4. **Legacy fallback** — `usage.creditUsage` still works when `config.used` is absent
5. **Env path** — `GROK_API_KEY` still works

Also: `oxlint` + format on touched files, and full monorepo `typecheck` (after building protocol/client/highlight dist for the worktree).

Test harness note: suite already sets `process.env.HOME = tempDir`; `os.homedir()` respects `HOME` on this platform, so nested auth is exercised by writing `join(homeDir, ".grok", "auth.json")` without injecting a homeDir option into `GrokQuotaProvider`.

### Not done (manual / live)

- Did **not** call the real billing API from this PR branch against a live host
- Did **not** restart a daemon or click Settings → Usage in the app
- Live shape knowledge comes from the issue investigation on a host with real Grok CLI auth (2026-07-23); this PR encodes that shape in unit tests

Optional smoke for reviewers (do not print secrets):

```bash
# auth shape only
python3 -c "import json;from pathlib import Path;d=json.loads((Path.home()/'.grok'/'auth.json').read_text());print(list(d)[0], list(d[list(d)[0]].keys()))"
# then GET /v1/billing with Bearer <key> and X-XAI-Token-Auth: xai-grok-cli
# restart *dev* daemon if needed; open Settings → Usage
```

---

## Acceptance criteria

- [x] Nested-only `~/.grok/auth.json` (no env token) can yield a token in code
- [x] Billing with only `config.used.val` + `config.monthlyLimit.val` yields `status: "available"` and correct `monthly_credits` used/remaining/limit
- [x] Legacy env token + legacy `usage.creditUsage` still work
- [x] No unrelated file churn
- [x] PR references #2352
- [ ] Live Settings → Usage confirmation on a Grok-CLI-only host (reviewer / follow-up)

Closes #2352

### DIFF
--- a/packages/server/src/services/quota-fetcher/providers/grok.ts
+++ b/packages/server/src/services/quota-fetcher/providers/grok.ts
@@ -38,6 +38,8 @@ const GrokUsageResponseSchema = z.object({
 interface GrokQuotaProviderOptions {
   logger: Logger;
   fetch?: ProviderApiFetch;
+  /** Override home directory (tests). Production uses os.homedir(). */
+  homeDir?: string;
 }
 
 /** Resolve a Grok CLI token from ~/.grok/auth.json (legacy or current nested shape). */
@@ -71,10 +73,12 @@ export class GrokQuotaProvider implements ProviderUsageFetcher {
 
   private readonly logger: Logger;
   private readonly fetchApi: ProviderApiFetch;
+  private readonly homeDir: string | undefined;
 
   constructor(options: GrokQuotaProviderOptions) {
     this.logger = options.logger;
     this.fetchApi = options.fetch ?? fetch;
+    this.homeDir = options.homeDir;
   }
 
   async fetchUsage(): Promise<ProviderUsage> {
@@ -134,7 +138,8 @@ export class GrokQuotaProvider implements ProviderUsageFetcher {
   }
 
   private async readGrokToken(): Promise<string | null> {
-    const path = join(homedir(), ".grok", "auth.json");
+    // homeDir override is for tests: Windows os.homedir() ignores $HOME (uses USERPROFILE).
+    const path = join(this.homeDir ?? homedir(), ".grok", "auth.json");
     if (!existsSync(path)) return null;
     try {
       return extractGrokTokenFromAuth(JSON.parse(await fs.readFile(path, "utf8")));
--- a/packages/server/src/services/quota-fetcher/providers/grok.ts
+++ b/packages/server/src/services/quota-fetcher/providers/grok.ts
@@ -21,6 +21,11 @@ const GrokUsageResponseSchema = z.object({
           val: ApiNumberSchema.optional(),
         })
         .nullish(),
+      used: z
+        .object({
+          val: ApiNumberSchema.optional(),
+        })
+        .nullish(),
     })
     .nullish(),
   usage: z
@@ -30,13 +35,34 @@ const GrokUsageResponseSchema = z.object({
     .nullish(),
 });
 
-const GrokAuthSchema = z.object({
-  access_token: z.string().optional(),
-});
-
 interface GrokQuotaProviderOptions {
   logger: Logger;
   fetch?: ProviderApiFetch;
+}
+
+/** Resolve a Grok CLI token from ~/.grok/auth.json (legacy or current nested shape). */
+export function extractGrokTokenFromAuth(auth: unknown): string | null {
+  if (auth == null || typeof auth !== "object" || Array.isArray(auth)) return null;
+  const record = auth as Record<string, unknown>;
+
+  const topLevel = record["access_token"];
+  if (typeof topLevel === "string" && topLevel.length > 0) {
+    return topLevel;
+  }
+
+  const entries = Object.entries(record);
+  const preferred = entries.filter(([key]) => key.startsWith("https://auth.x.ai::"));
+  const candidates = preferred.length > 0 ? preferred : entries;
+
+  for (const [, value] of candidates) {
+    if (value == null || typeof value !== "object" || Array.isArray(value)) continue;
+    const nestedKey = (value as Record<string, unknown>)["key"];
+    if (typeof nestedKey === "string" && nestedKey.length > 0) {
+      return nestedKey;
+    }
+  }
+
+  return null;
 }
 
 export class GrokQuotaProvider implements ProviderUsageFetcher {
@@ -76,7 +102,8 @@ export class GrokQuotaProvider implements ProviderUsageFetcher {
 
     const resp = GrokUsageResponseSchema.parse(await res.json());
     const monthlyLimit = resp.config?.monthlyLimit?.val ?? null;
-    const creditUsage = resp.usage?.creditUsage ?? null;
+    // Live CLI billing uses config.used.val; older mocks used usage.creditUsage.
+    const creditUsage = resp.config?.used?.val ?? resp.usage?.creditUsage ?? null;
     const balances: ProviderUsageBalance[] = [];
     if (monthlyLimit !== null || creditUsage !== null) {
       const remaining =
@@ -110,8 +137,7 @@ export class GrokQuotaProvider implements ProviderUsageFetcher {
     const path = join(homedir(), ".grok", "auth.json");
     if (!existsSync(path)) return null;
     try {
-      const auth = GrokAuthSchema.parse(JSON.parse(await fs.readFile(path, "utf8")));
-      return auth.access_token ?? null;
+      return extractGrokTokenFromAuth(JSON.parse(await fs.readFile(path, "utf8")));
     } catch {
       return null;
     }
--- a/packages/server/src/services/quota-fetcher/service.test.ts
+++ b/packages/server/src/services/quota-fetcher/service.test.ts
@@ -387,7 +387,13 @@ describe("real provider usage fetchers", () => {
         new CopilotQuotaProvider({ logger, fetch: fetchThroughTestDouble }),
         new CursorQuotaProvider({ logger, fetch: fetchThroughTestDouble }),
         new ZaiQuotaProvider({ logger, fetch: fetchThroughTestDouble }),
-        new GrokQuotaProvider({ logger, fetch: fetchThroughTestDouble }),
+        new GrokQuotaProvider({
+          logger,
+          fetch: fetchThroughTestDouble,
+          // Match Kimi: inject temp HOME so nested auth-file tests work on Windows
+          // (os.homedir() uses USERPROFILE there and ignores process.env.HOME).
+          homeDir,
+        }),
         new KimiQuotaProvider({
           logger,
           fetch: fetchThroughTestDouble,
--- a/packages/server/src/services/quota-fetcher/service.test.ts
+++ b/packages/server/src/services/quota-fetcher/service.test.ts
@@ -51,6 +51,11 @@ function writeKimiCredentials(dir: string, accessToken: string): void {
   );
 }
 
+function writeGrokAuth(home: string, auth: Record<string, unknown>): void {
+  mkdirSync(join(home, ".grok"), { recursive: true });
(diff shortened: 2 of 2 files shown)


Lite-G:
Hit the same thing on my host and independently landed on the same two drifts before finding this PR — so I'm not going to duplicate the work. Adding verification on a newer CLI instead.

**Verified 2026-07-28, grok CLI `0.2.112`** (I first looked on `0.2.93`; both behave identically). macOS, OIDC login, no `GROK_API_KEY` / `GROK_TOKEN` in the daemon env.

**Drift 1 — auth.** `~/.grok/auth.json` still has no top-level `access_token`. One issuer-keyed entry `https://auth.x.ai::<uuid>` containing `key` (the JWT), `auth_mode: "oidc"`, `refresh_token`, `expires_at`, plus profile fields. Matches your description exactly.

One data point that may be useful for scoping: the CLI refreshes `expires_at` on its own — I caught the file mid-refresh across two reads, so reading the stored `key` looks sufficient and Paseo shouldn't need to handle `refresh_token` itself.

**Drift 2 — billing.** `GET https://cli-chat-proxy.grok.com/v1/billing` returns `200` with:

```
{ config: { monthlyLimit: { val }, used: { val }, onDemandCap: { val },
            billingPeriodStart, billingPeriodEnd,
            history: [ { billingCycle: { year, month },
                         includedUsed: { val }, onDemandUsed: { val }, totalUsed: { val } } ] } }
```

No top-level `usage` object; used credits are at `config.used.val`, as you found. The endpoint and the `X-XAI-Token-Auth: xai-grok-cli` header are both still correct.

Not asking for scope creep on this PR, just noting for later: the same response also carries `billingPeriodEnd` (maps cleanly onto `resetsAt`) and `onDemandCap`, neither of which the provider reads today.

Issue #261, opened 2026-04-12
Feature request: add more editors/IDEs to the Open button


nexmoe:
## Feature Request

The **Open** button dropdown currently only supports **Cursor** and **Finder**.

It would be great to support more editors/IDEs, for example:

- **Zed**
- **VS Code**
- **JetBrains** (IntelliJ, WebStorm, etc.)
- **Sublime Text**
- **Xcode**
- **Windsurf**
- **Antigravity**
- **Terminal** (open in default terminal app)

### Current state

![Only Cursor and Finder are available](https://github.com/user-attachments/assets/placeholder)

Currently the dropdown only shows:
- Cursor
- Finder ✓

### Reference

For comparison, apps like **Fork** (Git client) already support a wide range of editors in their "Open" menu — including Finder, IDE submenu (Cursor, Antigravity, Windsurf, Zed, Sublime Text, Xcode, VS Code, JetBrains), Terminal, and Copy Path.

### Suggestion

- Auto-detect installed editors on the user's machine
- Allow users to configure/reorder their preferred editors
- Support custom editor commands (for less common editors)

nexmoe:
Related: #246 — requests first-class Cursor support, which is a subset of this broader editor/IDE support request.

duncan-daydream:
I second this — would love to see Xcode and Warp (terminal) in the list.

edward821220:
would love to see windsurf(devin IDE) in the list.

PR #2354, opened 2026-07-23
feat(desktop): add Fork to the Open menu

eodwyrd:
Adds Fork (the Git client) as an option in the workspace **Open** dropdown, so you can jump from a workspace straight into Fork with the repo open.

### Why
Finder and the editors are already there, but if you live in a Git GUI there's no quick way over. This scratches that itch. Related to #261 (more Open targets) — though that one's about editors, and this is a Git client, so figured it's worth its own small thing.

### How it works
Fork's a repository GUI, so it always opens the workspace root (never a single file). It launches via the existing `openMacApplication` runtime helper — literally `open -a Fork <repo>` — the same mechanism Cursor/Zed already use when their CLI isn't around. No runtime changes needed, and it only shows up when `Fork.app` is actually installed.

I went with `open -a Fork` over Fork's `fork` CLI on purpose: the CLI is opt-in (you have to install it separately) and cwd-based, whereas the app bundle is always there once Fork's installed and reliably opens the repo as a tab.

### Scope
**macOS only** for now — that's what I could actually test. Windows Fork launching can be a follow-up once someone can verify it.

One small quirk worth flagging: if Fork is fully quit, launching it opens the repo tab but Fork's own session-restore focuses whatever tab was last active rather than the one you just opened (it's still there, one click away). When Fork's already running it focuses the repo correctly. That's Fork's behaviour, not something we can override cleanly.

### Screenshots (desktop / macOS)
Fork in the Open dropdown, with its icon:

Open menu showing Fork

The workspace it opens from:

Workspace view

### How I verified
- Added registry tests: Fork shows on macOS when installed, opens the workspace root (ignores filePath), and is hidden when the app's missing or off-macOS.
- Ran it in the desktop app — Fork shows in the Open menu with its icon and opens the repo (screenshots above).
- `npm run typecheck`, `npm run lint`, `npm run format`, targeted vitest — all green.


boudra:

PR #2358, opened 2026-07-23
[codex] Fix git init workspace reclassification

### LINKED ISSUE #2356: Desktop does not reclassify an added non-git project after git init
## Summary

If a directory is added to Paseo before it is a Git repository, then Git is initialized in that same directory afterward, Paseo Desktop keeps treating the project as non-git. The project does not pick up Git metadata and therefore does not appear to become eligible for the worktree-based workflows until the Desktop app is closed and reopened.

## Reproduction

1. Create a new local directory without a `.git` directory.
2. Add/open that directory in Paseo Desktop, so it is registered as a project/workspace while still non-git.
3. In that same directory, run `git init` and create whatever initial branch/commit setup is needed.
4. Return to Paseo Desktop and look for Git/worktree affordances for that project.

## Expected behavior

Paseo should notice that the already-registered project has become a Git repository, refresh its Git runtime metadata, and expose the relevant worktree-capable project behavior without requiring an app restart.

## Actual behavior

The project continues to behave like a non-git project. The Git/worktree capability is not loaded until the user quits and reopens Paseo Desktop. After restart, the same directory is recognized correctly as a Git repository.

## Notes

This seems distinct from prior reports where an existing Git project was misclassified because of branch/origin metadata. The interesting transition here is:

- project is first persisted while it has no `.git` entry;
- `.git` is created later by `git init`;
- the running app/daemon does not refresh the project classification enough for the client to enable worktree workflows.

The architecture docs mention that project identity is daemon-global and that the project Git observer watches active project roots for the root `.git` entry. This bug may mean either the observer is not active for projects initially classified as non-git, the update is not persisted/emitted, or the Desktop/app runtime keeps stale project capability state until process restart.

marcelocastrobarbosa:
## Summary
- refresh Git snapshots before emitting reconciled workspace updates
- cover opened directory workspaces that become Git repos after git init

Fixes #2356.

## Root Cause
The global reconciliation path could update persisted workspace/project classification, then emit workspace updates before the session had a fresh Git runtime snapshot for the reclassified workspace. Desktop could keep seeing the workspace as lacking Git runtime metadata until restart.

## Validation
- npx vitest run packages/server/src/server/daemon-e2e/project-becomes-git.e2e.test.ts --bail=1
- npm run build:server
- npm run typecheck
- npm run lint

### DIFF
--- a/packages/server/src/server/daemon-e2e/project-becomes-git.e2e.test.ts
+++ b/packages/server/src/server/daemon-e2e/project-becomes-git.e2e.test.ts
@@ -1,6 +1,6 @@
 import { execFile as execFileCallback } from "node:child_process";
 import { mkdtempSync, realpathSync } from "node:fs";
-import { readFile, rm } from "node:fs/promises";
+import { readFile, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { promisify } from "node:util";
@@ -18,6 +18,7 @@ const cleanupListeners = new Set<() => void>();
 const execFile = promisify(execFileCallback);
 
 type ProjectUpdatePayload = Extract<DaemonEvent, { type: "project.update" }>["payload"];
+type WorkspaceUpdatePayload = Extract<DaemonEvent, { type: "workspace_update" }>["payload"];
 
 function waitForProjectUpdate(
   client: DaemonClient,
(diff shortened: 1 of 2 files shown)


marcelocastrobarbosa:
Closing this stale draft as part of a local worktree cleanup. Issue #2356 remains open, but Session/workspace reconciliation has moved substantially since this patch was written, so a fresh implementation and reproduction are preferable to carrying the old draft. The original patch is preserved at .

marcelocastrobarbosa:
Correction to the closing note: recovery branch: marcelocastrobarbosa/paseo:archive/git-detection-2026-09-07.

PR #2372, opened 2026-07-23
fix(quota): restore Grok usage for current CLI auth and billing schema

### LINKED ISSUE #2352: bug: Grok Usage card missing plan credits (auth.json + billing schema drift)

JOUYO:
### Linked issue

Closes #2352

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Grok Settings → Usage never showed monthly credits for normal CLI logins because the fetcher was out of date with the current Grok CLI:

1. **Auth store**: live `~/.grok/auth.json` nests the JWT under `key` (and may use a top-level direct entry). The old reader only accepted top-level `access_token`, so token lookup returned null and the billing fetch never ran.
2. **Billing schema**: current `GET /v1/billing` reports usage as `config.used.val`, while the fetcher only looked at legacy `usage.creditUsage`.
3. **Plan label**: also read `subscription_tier_display` from `/v1/settings` (e.g. `SuperGrok Heavy`) and surface account details from auth.json.

Includes unit coverage for nested CLI credentials, expired-token refresh, and 401 retry.

### How did you verify it

- Unit tests: `npx vitest run packages/server/src/services/quota-fetcher/providers/grok.test.ts packages/server/src/services/quota-fetcher/service.test.ts --bail=1` (31 passed)
- Live host check against a logged-in Grok CLI:
  - nested/`key` auth.json parses and billing returns monthly credits (`used` / `limit`)
  - settings returns `subscription_tier_display: "SuperGrok Heavy"`
  - daemon `provider.usage.list` then reports Grok as `available` with plan label + monthly credits balance
- `npm run lint` / format / server typecheck clean on the touched files

### Checklist

- [x] `npm run lint` passes
- [x] Linked issue present for bug fix / behavior change
- [x] Verification section filled with what *I* actually ran

### DIFF
--- a/packages/server/src/services/quota-fetcher/providers/grok.test.ts
+++ b/packages/server/src/services/quota-fetcher/providers/grok.test.ts
@@ -1,0 +1,185 @@
+import { promises as fs } from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+import { afterEach, describe, expect, test, vi } from "vitest";
+
+import { createTestLogger } from "../../../test-utils/test-logger.js";
+import { GrokQuotaProvider } from "./grok.js";
+
+const temporaryDirectories: string[] = [];
+
+async function createGrokHome(): Promise<string> {
+  const grokHome = await fs.mkdtemp(path.join(os.tmpdir(), "paseo-grok-usage-"));
+  temporaryDirectories.push(grokHome);
+  return grokHome;
+}
+
+afterEach(async () => {
+  await Promise.all(
+    temporaryDirectories.splice(0).map(async (directory) => {
+      await fs.rm(directory, { recursive: true, force: true });
+    }),
+  );
+});
+
+describe("GrokQuotaProvider", () => {
+  test("reads the current Grok CLI credential store and billing counters", async () => {
+    const grokHome = await createGrokHome();
+    await fs.writeFile(
+      path.join(grokHome, "auth.json"),
+      JSON.stringify({
+        "https://auth.x.ai::account": {
+          key: "grok_cli_token",
+          email: "person@example.test",
+          first_name: "Test",
+          last_name: "Account",
+          principal_type: "user",
+        },
+      }),
+    );
+    const fetchApi = vi.fn(async (input: RequestInfo | URL, init?: RequestInit) => {
+      expect(init?.headers).toMatchObject({ Authorization: "Bearer grok_cli_token" });
+      const url = input.toString();
+      if (url.endsWith("/v1/settings")) {
+        return new Response(JSON.stringify({ subscription_tier_display: "SuperGrok Heavy" }), {
+          status: 200,
+          headers: { "content-type": "application/json" },
+        });
+      }
+      return new Response(
+        JSON.stringify({ config: { monthlyLimit: { val: 100 }, used: { val: 25 } } }),
+        { status: 200, headers: { "content-type": "application/json" } },
+      );
+    }) as unknown as typeof fetch;
+    const provider = new GrokQuotaProvider({
+      logger: createTestLogger(),
+      fetch: fetchApi,
+      grokHome,
+    });
+
+    await expect(provider.fetchUsage()).resolves.toMatchObject({
+      status: "available",
+      planLabel: "SuperGrok Heavy",
+      balances: [expect.objectContaining({ used: 25, remaining: 75, limit: 100 })],
+      details: [
+        { id: "account_email", label: "Account email", value: "person@example.test" },
+        { id: "account_name", label: "Account", value: "Test Account" },
+        { id: "account_type", label: "Account type", value: "user" },
+      ],
+    });
+  });
+
+  test("refreshes an expired CLI credential before requesting usage", async () => {
+    const grokHome = await createGrokHome();
+    const authPath = path.join(grokHome, "auth.json");
+    await fs.writeFile(
+      authPath,
+      JSON.stringify({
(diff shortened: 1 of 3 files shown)


JOUYO:
Superseded by https://github.com/getpaseo/paseo/pull/2385, which includes this change (and later polish) on current main.

PR #2388, opened 2026-07-24
refactor(server): separate provider history reads from interactive resume

### LINKED ISSUE #2364: refactor(server): separate provider history reads from interactive resume
### What should change

PR #2348 introduced the runtime-only `AgentResumeSessionOptions.purpose` option so archived Codex timelines can read provider history without performing a native interactive resume.

The option now crosses agent loading, `AgentManager`, the provider registry, and the shared `AgentClient` interface even though history reading and interactive resume have different contracts. Split those operations so history hydration does not need to masquerade as session resume.

### Desired shape

- Give providers a dedicated history-read path that owns and closes any temporary process on success or failure.
- Hydrate archived timelines without changing Paseo `archivedAt` or the provider-native archive state.
- Keep `resumeSession` interactive and preserve its current failure timing. Do not defer invalid-session errors until the first prompt.
- Remove `AgentResumeSessionOptions` and the `interactive | history` purpose plumbing.
- Preserve concurrent loader and timeline-hydration behavior introduced by #2316.
- Lock the current behavior at the public seam before reshaping it.

Refs #2348, #2347, and #2316.

dwyanewang:
## Summary

Rebased onto current `main` after #3647. This PR keeps #3647's in-memory timeline model and separates read-only provider history access from interactive session resume:

- adds optional `AgentClient.readSessionHistory()` with a narrow history context;
- returns `{ events, coverage: { kind: "complete" } }`, leaving an internal result seam for future coverage variants without implementing pagination;
- removes `AgentResumeSessionOptions` and all `purpose: "history"` plumbing;
- hydrates archived records into retained closed snapshots without registering an interactive provider session;
- shares concurrent archived reads, upgrades deferred broadcast requests, and preserves concurrent unarchive promotion;
- migrates Codex, ACP-based providers, Claude, OpenCode, OMP, and Pi to provider-owned history readers;
- forwards the contract through derived/custom providers while remapping event provider IDs.

Closed history hydration is memory-only. It does not restore file timeline persistence, seed or commit the injectable durable timeline store, or reintroduce history reconciliation.

## Provider behavior

| Provider | Read-only history path |
| --- | --- |
| Codex | Initialize a temporary app-server and call `thread/read` directly; no loaded-list, resume, unarchive, skills, collaboration modes, or interactive tool setup |
| ACP | Initialize a temporary process and call `session/load` with `mcpServers: []`; resume-only providers return an unsupported error |
| Claude | Read the persisted transcript without starting an SDK query; resolve `CLAUDE_CONFIG_DIR` from explicit config, provider profile env, daemon env, then `~/.claude` |
| OpenCode | Call session metadata/messages directly; no global event attachment or native abort; release every acquisition on success and failure |
| OMP | Start only the minimal history RPC runtime; no model, mode, prompt, host-tool, or subagent subscription setup |
| Pi | Parse the persisted JSONL directly without starting a runtime or installing temporary MCP/extensions |

Codex persisted child histories are loaded with a maximum concurrency of 8 and the existing 100-thread cap, deduplication, nested traversal, and per-child failure isolation. Results are applied in batch route order so completion timing cannot reorder the restored timeline.

## Archive behavior

- A history read leaves Paseo `archivedAt` and provider-native archive state unchanged.
- The snapshot is published only after provider resources are released and all events are applied.
- A failed read leaves no partial timeline, provider-subagent state, or closed snapshot.
- A concurrent unarchive releases the closed snapshot and runs normal interactive resume.
- Successful history is reused during that promotion; failed history is hydrated normally by the interactive session.
- Closed responses use the stored agent record as authority for title, status, archive timestamp, labels, owner, attention, persistence metadata, and internal visibility.
- Registering a live session, deleting the agent, or discarding retained state removes the closed snapshot.

## Regression coverage

This PR incorporates the narrow provider behaviors demonstrated by:

- #3274: an unarchived native Codex thread remains an inert `thread/read` during archived history access;
- #3355 / #3356: Codex child history reads are bounded at 8, complete, deterministic, and failure-isolated;
- #3358: archived OpenCode history attaches to no live event stream and never aborts the native session;
- #3155: Claude history and import lookup honor a profile-level `CLAUDE_CONFIG_DIR`.

These are contract regression cases, not additional feature surfaces.

## Scope guard

This PR does not include:

- #3261 pagination, `thread/turns/list`, timeline cursors, or CLI tail changes;
- app, protocol, CLI, or wire-schema changes;
- file timeline persistence, SQLite, append-only caches, durable history snapshots, or old reconciliation;
- the broader live/history hydration transaction from #2722;
- archive/resume lifecycle-lock expansion from #3544;
- unrelated provider or timeline cleanup.

The diff is limited to server implementation/tests and the two owning docs.

## Verification

Focused Vitest files passed:

- archived loader: 8 tests;
- AgentManager: 169 tests;
- Codex: 144 tests;
- ACP: 99 tests;
- OpenCode: 127 tests;
- Claude config/history: 21 tests;
- OMP: 30 tests;
- Pi: 66 tests;
- provider registry: 50 tests;
- session/MCP exposure: 272 tests.

Static/build verification passed:

- `npm run build:server`
- `npm run typecheck`
- `npm run lint`
- `npm run format`
- `npm run format:check`
- `git diff --check`
- scope searches for removed history-purpose/durable-reconciliation symbols and forbidden package changes

The full test suite was not run locally per repository guidance.

Closes #2364.


boudra:
Closing for now: please reopen with real-provider before/after evidence that archived Codex history reads leave the native thread untouched and release temporary resources, and that OpenCode history reads do not abort ongoing native work.

dwyanewang:
Reopened as #4473 with the evidence you asked for.

This PR could not be reopened: rebasing onto current `main` required a force-push, and GitHub refuses to reopen a closed PR whose head branch was force-pushed. Same branch, same work, continued in #4473.

Both suites were run against `upstream/main` and against the branch. The `main` failures are the behaviour the PR changes.

**Codex** — real `codex` 0.153.4, local stub model backend, so it reproduces without credentials:

- *leaves the native thread untouched*: after the history read the test archives the thread again. Codex rejects that while a writer is held, and on `main` it fails with `thread … already has an active writer` — the read had resumed the thread and claimed its writer. The dedicated read only issues `thread/read`, takes no writer, and the archive succeeds.
- *releases temporary resources*: on `main` a `codex app-server` this test process spawned is still alive after the read (`expected [ 647875 ] to deeply equal []`). The archived record is registered as a live interactive session, `ensureAgentLoaded` has no close on the archived path, and there is no idle reclamation.

**OpenCode** — real `opencode serve` 1.14.46 behind a recording proxy: a history read plus the close a tab teardown triggers sends exactly one `session.abort` on `main` (`expected 1 to be +0`). `abortOpenCodeSession` is unconditional in `close()`, and the abort is session-scoped rather than turn-scoped, so it cancels whatever that session is running — and OpenCode's native archive is only a numeric field, so an archived session can legitimately be busy.

Both skip when the provider binary is absent, and neither runs in the default `test:e2e` sweep.

One thing this turned up that is **not** fixed in #4473, since it belongs in its own change: `archiveAgentUnlocked` syncs the native archive before closing the runtime, so `thread/archive` loses to the thread's own writer lock and the best-effort `catch` in `syncNativeArchiveState` swallows it. A Paseo-archived Codex agent therefore usually leaves its native thread active. That is also why `allowArchivedHistory` rarely engages today — `thread/resume` succeeds instead of raising the archived error the guard keys off. Happy to open a separate issue or PR for it if useful.

boudra:

PR #2396, opened 2026-07-24
Add SSH remote host support for tunneling daemon connections

bendavid:
## Summary

Connect to a remote Paseo daemon over SSH from the CLI (`--host ssh://user@host`) and the desktop app. The CLI ensures a daemon is running on the remote host (installing Paseo if needed), opens an SSH local port-forward, and tunnels the WebSocket connection through it — all over a single SSH connection.

See [docs/ssh.md](docs/ssh.md) for the design and the gotchas behind it.

## How it works

**One connection.** The remote command is `exec /bin/sh` while `-L` forwards the port, and the ensure script is piped to that shell over stdin. sshd hands a remote command to the user's *own* shell (from the password database), which may be fish or tcsh — so the command is reduced to three metacharacter-free words that every shell agrees on, and the script never has to survive their quoting rules. It also keeps the script out of the remote host's process list.

On success the script deliberately doesn't exit: `sh` reading from a pipe blocks for the next line, and that block *is* the keepalive. Closing stdin is the teardown.

**Readiness is a `READY:` marker**, not prose. The port forward accepts connections before the remote daemon is listening, so waiting on the local port is a false positive.

**Which remote `paseo` runs**, in order: anything already listening on the port → a `paseo` on `PATH` (global npm, system package manager) → a Paseo-managed install under `installDir`. Preferring `PATH` is about correctness, not disk: a second copy would let someone on the remote host start a conflicting daemon against the same `PASEO_HOME`. Paseo only ever installs or upgrades inside `installDir`. The managed copy is pinned to the local CLI version, falling back to `latest` when that version isn't published (source checkouts, betas).

**Password prompts are rendered by Paseo.** SSH only takes a secret from a program's stdout, so there is still a small askpass program — but it just relays over a unix socket to the app, which shows a styled, localized prompt that says whether it wants a key passphrase or an account password. The helper is written at runtime, so nothing needs unpacking from an asar archive. This replaced `zenity`/`kdialog`/`osascript`, along with their GTK baggage (underscores read as mnemonic accelerators, Pango markup parsing, AppleScript escaping).

**Cancelling ends the attempt.** SSH ignores an askpass program's exit status and simply retries, so cancellation is signalled out of band — the desktop aborts an `AbortSignal`; the CLI's terminal prompt writes a marker to stderr. SSH's three attempts are left intact so a *typo* stays retryable.

## Changes

- **CLI** (`packages/cli/src/ssh/`): `ssh-host-config` (parse `ssh://` URIs), `ssh-process` (SSH args, tunnel, terminal askpass), `ssh-connection` (ensure + tunnel), `remote-daemon` (ensure script and failure messages), `askpass-channel` (socket relay for GUI prompts)
- **CLI** (`local-daemon.ts`): `systemd-run --user --scope` + `loginctl enable-linger` so a detached daemon survives the session that started it, behind a `DaemonLaunchRuntime` seam
- **Protocol**: `SshHostConnectionSchema`. `port` is deliberately *not* defaulted — "unset" has to stay distinguishable from "22" so a `Port` directive in `~/.ssh/config` wins
- **App**: SSH connection type end to end, add-host modal, and a globally mounted password prompt so reconnects outside the add-host flow are covered
- **Desktop** (`main.ts`, `preload.ts`): SSH bridge IPC — open/close tunnel, progress events, password relay

## Notes for review

- **Origin stripping is scoped.** A renderer's origin can never match a tunnel's ephemeral loopback port, so the daemon's same-origin check would reject every tunneled connection. The header is removed only for ports the main process is currently forwarding — never for loopback generally, which would also disable the check for a directly-connected local daemon. Untrusted web content runs in a separate Electron session the rule isn't installed on.
- **Host keys are trust-on-first-use** (`StrictHostKeyChecking=accept-new`). A *changed* key still fails, but a first connection from the GUI accepts an unknown key without showing a fingerprint. Noted in SECURITY.md.
- **Non-active SSH connections are not latency-probed.** Probing one means spawning ssh, authenticating, and running the ensure script — every 120s, re-prompting password-auth hosts. The tradeoff is that the adaptive switcher won't move *onto* a tunnel on RTT alone, which is the behavior we want anyway. Failover is unaffected.
- **Requires a POSIX remote** with `/bin/sh` and Node. The local GUI prompt path is POSIX-only too (Windows would need a named pipe).

## Test plan

- [x] `paseo ls -a --host ssh://submit82.mit.edu` — single password prompt, agents listed, clean exit
- [x] Fresh daemon start on remote host — ensure script installs/launches, tunnel connects
- [x] Desktop SSH connection — progress messages displayed, WebSocket connects through tunnel
- [x] Askpass channel driven exactly as ssh drives it (prompt as argv[1], secret on stdout): relay, decline, timeout, closed-channel
- [x] Cancel vs. wrong password measured against a live sshd — declining aborts on the first prompt (~120ms); a wrong password still retries
- [x] Ensure script executed through a real `sh` against stubbed PATHs — ready/node-missing/npm-missing/`PATH`-preferred/unpublished-version paths
- [x] typecheck, lint, format, and 186 unit tests across cli/app/protocol/desktop

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,6 +40,7 @@ At the start of non-trivial work, list `docs/` and skim anything relevant to the
 | [docs/forge-providers.md](docs/forge-providers.md)                 | Adding a git forge: registry/manifest, drop-in checklist, self-host/GHES, the two facts tiers                                  |
 | [docs/custom-providers.md](docs/custom-providers.md)               | Custom provider config: Z.AI, Alibaba/Qwen, ACP agents, profiles, custom binaries                                              |
 | [docs/service-proxy.md](docs/service-proxy.md)                     | Service proxy: exposing workspace scripts at public URLs, DNS setup, reverse proxy config                                      |
+| [docs/ssh.md](docs/ssh.md)                                         | SSH remote hosts: single-connection tunnel, ensure script contract, which remote `paseo` runs, askpass and Origin gotchas      |
 | [docs/development.md](docs/development.md)                         | Dev server, build sync gotchas, CLI reference, agent state, Playwright MCP                                                     |
 | [docs/rpc-namespacing.md](docs/rpc-namespacing.md)                 | WebSocket RPC naming convention — dotted namespaces and `.request`/`.response` pairs                                           |
 | [docs/protocol-validation.md](docs/protocol-validation.md)         | zod-aot generated inbound WebSocket validation, patched compiler regressions, schema-purity rules                              |
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -64,6 +64,30 @@ CORS is not a complete security boundary. It controls which browser origins can 
 
 Paseo validates the `Host` header on every HTTP request and every WebSocket upgrade against an allowlist (Vite-style semantics). By default, only `localhost`, `*.localhost`, and any literal IP address (IPv4 or IPv6) are accepted. Additional hostnames can be configured via `hostnames` in `config.json` or the `PASEO_HOSTNAMES` env var (comma-separated; entries beginning with `.` match a domain and its subdomains; the value `true` disables the allowlist entirely). Requests with unrecognized hosts are rejected with `403 Host not allowed`.
 
+## SSH tunnels
+
+The CLI and desktop app can reach a daemon on another machine over `ssh -L`
+(see [docs/ssh.md](docs/ssh.md)). The remote daemon binds `127.0.0.1` and is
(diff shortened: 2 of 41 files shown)


MakFly:
hi , where this PR we will merge ?

boudra:
Thanks for exploring SSH support. #3989 supersedes the connection workflow with transport-only SSH for the CLI and desktop. We are deliberately not coupling transport to remote installation, launch, supervision, or credential prompting; any future remote install will be a separate, explicit user action. Closing in favor of #3989.

PR #2398, opened 2026-07-24
fix(acp): retry prompts while provider turn is busy

alexandreumatize:
## Summary

- retry only ACP errors whose `data.code` is exactly `turn.agent_busy`, preserving the Paseo turn, message ID, and single user timeline row
- use jittered exponential backoff from 250 ms up to 5 seconds while keeping autonomous provider events unscoped
- make cancel, permission denial, close, and late provider responses lifecycle-safe without changing public interfaces or the ACP protocol
- document provider-declared busy turn semantics

## Validation

- `npx vitest run packages/server/src/server/agent/providers/acp-agent.test.ts --bail=1` — 96 passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git diff --check`
- isolated Kimi ACP reproduction on port 6779: background command completed, concurrent prompt queued on `turn.agent_busy`, then delivered once after the autonomous turn

The main Paseo daemon was not restarted or modified during validation.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -35,6 +35,21 @@ runtime.
 
 Cancellation changes lifecycle state only after the provider acknowledges the interrupt or emits a terminal turn event. If the interrupt is rejected or times out, the agent remains `running` with its active foreground turn intact. Follow-up actions such as replacement, reload, rewind, and Stop must report that failure instead of accepting work they cannot perform. Synthesizing a local cancellation without provider acknowledgment creates a split-brain session: Paseo accepts a new prompt while the provider still owns the previous foreground turn.
 
+### Provider-declared busy turns
+
+An ACP provider may start autonomous work outside a Paseo foreground prompt. If a new
+`session/prompt` is rejected with JSON-RPC `data.code === "turn.agent_busy"`, the generic ACP
+adapter keeps the Paseo turn `running` and retries the same prompt and message ID with jittered
+exponential backoff (250 ms, capped at 5 seconds). The canonical user timeline row is emitted only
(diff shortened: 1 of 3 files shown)


lxsolutions:
+1 from a live Kimi hit today — this exact scenario, on Paseo 0.1.110 / kimi-code 0.29.0 (ACP), Ubuntu 24.04.

A Kimi agent was mid-turn doing legitimate long-running tool work. While turn 8 was active (06:17:18 → 06:23:07), five user prompts were rejected with `turn.agent_busy`:

```text
2026-08-05T06:17:58 WARN acp: prompt rejected because another turn is active  details="{\"turnId\":8}"
… (06:18:12, 06:18:37, 06:19:33, 06:20:06)
2026-08-05T06:23:07 INFO llm response turnStep=8/7   (end_turn)
```

Each rejection surfaced as `[System Error]` in the timeline, so from the user's seat a healthy, progressing turn looked like a crashing agent. The next prompt after the turn ended went through immediately.

Two data points beyond the existing repro:

1. The busy window here was ~6 minutes of real work — long enough that a human retries several times, producing exactly the pile of error rows this PR eliminates. With autonomous/goal-style turns the window can be much longer, which makes the retry path even more valuable.
2. Each rejection also flips the agent's status to `error`, and that state **outlives the turn** — the agent still showed `error` minutes after `end_turn`, and `cancel_agent` in that state returns `success: false`. That status-lifecycle piece survives this PR (any transient prompt rejection can still trip it), so I filed it separately as #2889.

*Comment compiled with AI assistance from live daemon/provider logs; timestamps and observations are from a real run.*

PR #2408, opened 2026-07-24
fix(server): stop OpenCode serve rotation on catalog refresh

### LINKED ISSUE #2370: bug: OpenCode server processes accumulate indefinitely on provider catalog refresh (acquireNew leaks retired servers)
## Summary

Every `fetchCatalog({ force: true })` call spawns a new `opencode serve` subprocess via `acquireNew()` and retires the previous one. Retired servers **stay alive** because active sessions hold references (`refCount > 0`). This means servers accumulate indefinitely — each catalog refresh leaks a ~300MB process until all sessions referencing old servers are closed.

## Observed behavior

With 3 idle Paseo OpenCode sessions:

```
Four opencode serve processes: approximately 1.24 GiB RAM combined
```

4 servers for 3 sessions — the singleton should produce exactly 1.

## Root cause

**`packages/server/src/server/agent/providers/opencode-agent.ts:1346-1348`** —`fetchCatalog` calls `acquireNew()` when `force` is true:

```ts
async fetchCatalog(options: FetchCatalogOptions): Promise<ProviderCatalog> {
    const acquisition = options.force
      ? await this.serverManager.acquireNew()   // ← spawns new server, retires current
      : await this.serverManager.acquireCurrent();
```

**`packages/server/src/server/agent/providers/opencode/server-manager.ts:142-144`** — `acquireNew()` rotates current to retired and spawns a new one:

```ts
async acquireNew(): Promise<OpenCodeServerAcquisition> {
    const server = await this.getNewServer();
    return this.acquireServer(server);
}
```

**`server-manager.ts:388-394`** —`cleanupRetiredServers()` only kills a retired server when its `refCount` hits 0. But each session holds a ref (`releaseServer` at `opencode-agent.ts:2770`) that is only released on `close()`. So retired servers stay alive as long as any session that was alive when the server was current still exists.

**Trigger path:**
1. Client sends `refresh_providers_snapshot_request` (WebSocket)
2. `provider-catalog-session.ts:378` calls `refreshSnapshotForCwd()`
3. `provider-snapshot-manager.ts:575` calls `loadProviders({ force: true })`
4. `provider-snapshot-manager.ts:724` calls `fetchCatalog({ cwd, force: true })`
5. `opencode-agent.ts:1348` calls `acquireNew()` → new server spawned

## How to reproduce

1. Create 3 OpenCode agents in Paseo (each uses `acquireCurrent()` → shared server A)
2. Trigger a provider catalog refresh from the UI (e.g., open the model selector)
3. `fetchCatalog({ force: true })` → server A retired, server B spawned
4. Repeat catalog refresh → server B retired, server C spawned
5. Run `ps aux | grep "opencode serve"` — 3 servers alive for 3 sessions
6. Repeat again → 4 servers for 3 sessions

## Impact

Each orphaned `opencode serve` consumes ~300MB RSS. For users who run long-lived sessions and occasionally refresh providers, this silently accumulates into gigabytes of wasted RAM. No cleanup until all sessions are closed.

## Suggested fix

Two options, not mutually exclusive:

**Option A** — Change `fetchCatalog({ force: true })` to use `acquireCurrent()` instead of `acquireNew()`. The existing server already has models loaded; there is no evidence that a fresh server picks up different state. If state freshness is the goal, a separate mechanism (e.g., signal the existing server to reload) would be more appropriate.

**Option B** — On rotation, migrate sessions from the retiring server to the new one (update the client URL and release the old ref), so retired servers can actually be cleaned up. This would fix the root cause for all `acquireNew()` callers.

## Related issues

- #1178 — Feature request: allow OpenCode provider to attach to an existing server (related but different)
- #1396 — Fixed: Windows OpenCode serve subprocesses remain after archiving agents (different trigger)
- #2107 — Open: Disabled OpenCode provider still starts opencode serve (different trigger)

FluX-de-creator:
## Summary
- Force provider catalog refresh no longer calls `acquireNew()`, which retired the shared `opencode serve` while sessions still held refs and leaked ~300MB processes per refresh.
- Catalog refresh always uses `acquireCurrent()` so Paseo still bypasses its snapshot cache and re-probes over HTTP without rotating helper processes.
- Tests updated/added to lock in the no-rotation behavior under held session refs.

Fixes #2370

## Test plan
- [x] `npx vitest run packages/server/src/server/agent/providers/opencode-agent.list-models-timeout.test.ts --bail=1`
- [x] typecheck + lint on touched files
- [ ] Manual: open 3 OpenCode agents, refresh providers several times, confirm `opencode serve` count stays ~1

### DIFF
--- a/packages/server/src/server/agent/providers/opencode-agent.list-models-timeout.test.ts
+++ b/packages/server/src/server/agent/providers/opencode-agent.list-models-timeout.test.ts
@@ -64,7 +64,7 @@ test("allows a slow provider.list call to succeed instead of failing after 10 se
   expect(openCodeClient.calls.providerList).toHaveLength(1);
 });
 
-test("uses a new server for explicit catalog refresh", async () => {
+test("uses the current server for explicit catalog refresh", async () => {
   const runtime = new TestOpenCodeHarness();
   const openCodeClient = new TestOpenCodeClient();
   openCodeClient.providerListResponse = {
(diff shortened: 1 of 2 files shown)


FluX-de-creator:
Closing — opened against upstream by mistake; work remains on fork FluX-de-creator/paseo only.

PR #2410, opened 2026-07-24
fix(server): keep service proxy routes synced with PASEO_URL

### LINKED ISSUE #2397: bug: workspace service PASEO_URL can point to unregistered proxy route while stale route remains
### What's broken

A Paseo-managed workspace service is running and reachable on its direct local port, but the `PASEO_URL` injected for the service points to a `.localhost:6767` hostname that the daemon service proxy does not have registered.

In the same session, an older/alternate service hostname for the same workspace shape still appears to be registered by the proxy, but points at a dead backend. That makes the current URL return `404` and the stale URL return `502`.

This makes the preview URL shown/injected by Paseo unreliable even though the app server itself is healthy.

### Sanitized environment

- Paseo CLI/daemon version: `0.1.110`
- OS: macOS / Apple Silicon
- Daemon listen: `127.0.0.1:6767`
- Desktop-managed daemon: yes
- Workspace path shape: `~/.paseo/worktrees/<workspace-id>/<worktree-slug>`
- Branch shape: feature branch with slashes, e.g. `topic/example-service`
- Project slug shape: `example-app`
- Service name: `web`

### Sanitized `paseo.json`

```json
{
  "worktree": {
    "servicePorts": {
      "range": "3000-4000"
    }
  },
  "scripts": {
    "web": {
      "type": "service",
      "command": "npm run dev -- --host $HOST --port $PASEO_PORT"
    }
  }
}
```

### Observed service env / startup

The `web` terminal shows the service started with Paseo-provided env vars:

```sh
npm run dev -- --host $HOST --port $PASEO_PORT
```

The app server receives something equivalent to:

```sh
astro dev --host 127.0.0.1 --port 64064
```

Sanitized service env values:

```sh
HOST=127.0.0.1
PASEO_PORT=64064
PASEO_URL=http://web--current-branch--example-app.localhost:6767
PASEO_SERVICE_WEB_PORT=64064
PASEO_SERVICE_WEB_URL=http://web--current-branch--example-app.localhost:6767
```

### Actual behavior

Direct app port works:

```text
GET http://localhost:64064/
=> 200 OK
```

Current `PASEO_URL` does not route:

```text
GET http://web--current-branch--example-app.localhost:6767/
=> 404 Not Found
```

A previous/alternate generated hostname appears to still be known to the proxy but points to a dead backend:

```text
GET http://web--old-branch-name--example-app.localhost:6767/
=> 502 Bad Gateway
```

The distinction seems important:

- `404` from `.localhost:6767` means the daemon is reached but the current generated service hostname is not registered in the service proxy route table.
- `502` from another `.localhost:6767` hostname means a stale route is still registered but its backend port is no longer reachable.

### Expected behavior

When a service is running and `PASEO_URL` / `PASEO_SERVICE_WEB_URL` are injected, the daemon service proxy should have a matching route:

```text
web--current-branch--example-app.localhost:6767 -> 127.0.0.1:$PASEO_PORT
```

If the branch/worktree/service identity changes, old proxy routes should either be removed or updated so the UI/env does not expose a stale or unregistered URL.

### Possible related issues

I checked for related issues before filing:

- #678: similar service reverse proxy symptoms, but that thread focuses on `502` caused by the dev server binding somewhere other than `127.0.0.1`. In this case the service explicitly binds to `127.0.0.1`, the direct port works, and the current `PASEO_URL` returns `404` because the route is missing.
- #1374: mentions service configuration caching and needing a daemon restart for service config changes to take effect. This may be related if the service route table or projected service URL can get out of sync with workspace/branch/service metadata.
- #1842: reverse proxy/WebSocket host header issue; likely not the same because this is local `.localhost:6767` service proxying.

### Hypothesis

The generated service URL and the service proxy route registry can diverge, possibly after one of:

- branch rename or branch slug/name normalization change;
- worktree/service restart without re-registering the route;
- `paseo.json` service config cache update;
- stale route surviving after a service terminal/process dies;
- route table keeping an old hostname while the injected `PASEO_URL` is computed from newer branch/workspace metadata.

### Debugging hooks that would help

It would help if there were a CLI/API command to inspect the daemon's current service proxy route table, for example:

```sh
paseo service ls --json
# or
paseo workspace services --json
```

with fields like:

```json
{
  "hostname": "web--current-branch--example-app.localhost",
  "port": 64064,
  "workspaceId": "...",
  "scriptName": "web",
  "health": "ok|unreachable|missing"
}
```

That would make it easier to tell whether a `.localhost:6767` failure is due to hostname registration, backend reachability, or stale workspace metadata.

FluX-de-creator:
## Summary
- Register/remove service-proxy routes by `(workspaceId, scriptName)` so branch/slug changes cannot leave a stale hostname (502) while the injected `PASEO_URL` is unregistered (404).
- Keep prior hostnames as aliases across branch renames so spawn-time env URLs keep resolving; clear all workspace routes on archive.
- Stop projecting live proxy URLs when a running service has no registered route; spawn registers before injecting env from the registered route.

Fixes #2397

## Test plan
- [x] Targeted vitest: service-proxy, script-route-branch-handler, script-status-projection, workspace-service-env
- [x] typecheck + lint
- [ ] Manual: start a workspace service, curl `Host: <PASEO_URL host>` against the daemon; rename branch / restart and confirm no 404+502 split

### DIFF
--- a/docs/service-proxy.md
+++ b/docs/service-proxy.md
@@ -97,6 +97,19 @@ server {
 
 ## Environment variables
 
+When a service starts, Paseo injects:
+
+| Variable                             | Description                                                                                  |
+| ------------------------------------ | -------------------------------------------------------------------------------------------- |
+| `HOST`                               | Bind address for the service process (`127.0.0.1` for loopback daemons, otherwise `0.0.0.0`) |
+| `PASEO_PORT`                         | Allocated backend port                                                                       |
+| `PASEO_URL`                          | Proxy URL for this service (public alias when configured, otherwise localhost)               |
+| `PASEO_SERVICE_<NAME>_PORT` / `_URL` | Peer service ports and proxy URLs in the same workspace                                      |
+
+`PASEO_URL` / `PASEO_SERVICE_*_URL` are fixed at spawn from the registered route identity. A later branch rename updates the proxy's **canonical** hostname and keeps the spawn-time hostname(s) as aliases to the same backend until the service stops or restarts. Restarting the service re-registers under the current branch and drops prior aliases.
+
+Routes are keyed by `(workspaceId, scriptName)`: re-registering the same script replaces any prior hostname for that pair. Archive and process exit remove the workspace's proxy routes so projections never advertise an unregistered live URL.
+
 The listen address and public base URL can also be set via environment variables, which take precedence over `config.json`:
 
 | Variable                              | Description                                                               |
--- a/packages/server/src/server/script-route-branch-handler.test.ts
+++ b/packages/server/src/server/script-route-branch-handler.test.ts
@@ -73,7 +73,7 @@ function registerRoute(
 }
 
 describe("script-route-branch-handler", () => {
-  it("updates routes on branch rename by removing old hostnames and registering new ones", () => {
+  it("updates routes on branch rename while keeping prior hostnames as aliases", () => {
     const routeStore = new ScriptRouteStore();
     registerRoute(routeStore, {
       hostname: "api--feature-auth--paseo.localhost",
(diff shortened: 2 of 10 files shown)


FluX-de-creator:
Closing — opened against upstream by mistake. Work stays on fork FluX-de-creator/paseo only.

PR #2417, opened 2026-07-24
feat(app): dismiss the chat keyboard on a fast upward flick

### LINKED ISSUE #2415: feat: dismiss the chat keyboard when flicking the history up
### Prior search

### What's the problem

When I’m typing in a chat on mobile and want to look back at what the agent did earlier, I scroll the history up and the keyboard stays open. The transcript keeps living in the cramped strip above the keyboard, at exactly the moment I need more room to read.

So reading earlier messages takes two gestures instead of one: close the keyboard first, then scroll. Android’s system gesture at least offers a way to close it, but it’s still a detour; on iOS the app gives me no equivalent affordance.

Chat apps generally read “the user started moving through the history” as “the user is reading now, not typing”, and get the keyboard out of the way.

### What would solve it

A fast upward flick over the message history dismisses the keyboard. A slow scroll, or a fast but controlled one, where the finger decelerates before lifting, leaves it up, because that is still someone glancing at context while composing.

Speed at the moment the finger lifts is what separates the two: a flick is released mid-motion, a controlled scroll has already slowed down. Distance is a poor signal, and “any scroll at all” is worse

Suggested scope: iOS and Android only, since web and desktop have no soft keyboard occupying space.
Two things that would need care in whatever implementation:

- On Android the input has to be blurred and the keyboard dismissed. Dismissing alone leaves the keyboard inset applied, so the layout stays shifted with an empty gap where the keyboard was.
- It should not fire when only a hardware keyboard is attached: the composer can be focused with nothing on screen, and blurring there would just stop the user typing.

- **`keyboardDismissMode="on-drag"`**: fires on the first pixel of movement, so the keyboard vanishes when you nudge the list to peek at the previous message. It removes the ability to scroll a little *while* composing.
- **`keyboardDismissMode="interactive"`**: the intended iOS behaviour of the keyboard following the finger, but it misbehaves on an inverted list, which is what the chat history is: the gesture directions come out reversed, so it needs a long deliberate drag and feels janky instead of tracking the finger.
- **`KeyboardGestureArea`** from `react-native-keyboard-controller`, already a dependency, did not resolve the inverted-list jank either.
- **A distance threshold instead of speed**: workable, but it cannot tell a flick from a deliberate fast scroll, so it also fires on gestures where the user clearly intends to keep typing.
- **Leaning on Android's system dismiss, or a tap elsewhere**: still two steps, and there is no iOS equivalent.

See Android
https://github.com/user-attachments/assets/96c628ed-82f4-4b81-bebe-9ea3a7d43466

See IOS
https://github.com/user-attachments/assets/9c1e79c2-f0eb-49e4-a8ba-b6e237bc23ee

nllptrx:
### Linked issue

Closes #2415

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Scrolling the chat history back to read earlier messages left the keyboard up, so the transcript stayed squeezed into the space above it and the keyboard had to be closed by hand first. Two gestures for what should be one.

Now a fast upward flick over the history dismisses it, while a slow, or fast but controlled, read-scroll leaves it alone.

React Native's own `keyboardDismissMode` cannot express that: `"on-drag"` fires on the first pixel and kills the keyboard on any peek-scroll, and `"interactive"` is broken on inverted lists (the gesture directions are reversed, so it needs a long drag and feels janky). The feature therefore classifies the list's existing scroll gesture rather than adding a competing pan recognizer: samples are taken from scroll events, and at release the speed over the drag's final stretch decides. Measuring at release rather than averaging the whole drag is what separates a flick, where the finger lifts mid-motion, from a fast but controlled scroll, which decelerates before the finger lifts.

The feature has one removable home under `agent-stream/scroll-keyboard-dismiss/`:

- `use-scroll-keyboard-dismiss.ts` owns gesture state, the keyboard-presence gate, and the blur/dismiss effect.
- `model.ts` is the pure gesture classifier and owns its sampling/velocity policy.
- `strategy-native.tsx` only constructs the hook and forwards the list's begin, scroll, and end events.

Two timing details worth calling out, both of which were wrong in earlier revisions:

- Timing comes from `nativeEvent.timestamp`, the native gesture time every platform stamps on the payload, not from a clock and not from the synthetic `timeStamp`. The synthetic one reads a differently spelled key, so it silently falls back to the dispatch clock, and with a busy JS thread the callbacks arrive bunched together, which reads a calm scroll as a flick.
- The release offset comes from the end-drag event itself, because the last `onScroll` can be stale by the time a short flick lands.

Dismissal blurs the input *and* dismisses the keyboard, and only runs when a software keyboard actually occupies space. The hook reads the app's existing Reanimated keyboard-shift shared value, whose calculation runs on the UI thread, instead of recalculating keyboard layout in the JavaScript scroll handler. The occupies-space gate matters for hardware keyboards: the composer can be focused with no on-screen keyboard, and blurring there would stop the user typing to dismiss something that was never visible.

On Android, dismissing alone left the input focused with the keyboard inset still applied, so the layout stayed shifted with an empty gap where the keyboard was; blurring alone released focus but left the IME on screen. The paired sequence matches the physical-device behavior verified here.

### How did you verify it

Behaviour on a physical Android device (Xiaomi 24069PC21G, Android 16), Release build, gestures driven with `adb shell input swipe` at fixed durations over 233 dp (480 dpi, so 1 dp = 3 px):

| Gesture | Measured speed | Result |
| --- | --- | --- |
| 1500 ms drag | 0.16 dp/ms | keyboard stays up |
| 400 ms drag | 0.58 dp/ms | keyboard stays up |
| 80 ms flick | 2.9 dp/ms | keyboard dismisses, composer settles at the bottom, no leftover gap |

Screenshots for the three runs are attached below. I also confirmed the Release build regenerated `index.android.bundle` for each run rather than reusing a cached one, and that the earlier "empty gap after dismissal" state does not reproduce.

Unit tests: `npx vitest run src/agent-stream/scroll-keyboard-dismiss/model.test.ts src/hooks/use-keyboard-shift-style.test.ts --project unit --bail=1`, 15 cases pass. They pin the behaviours that broke during development: a slow read-scroll staying below the threshold, a flick crossing it, a fast-but-decelerating drag that would pass on a whole-drag average but not at release, a drag toward newer messages staying negative, the fallback for gestures too short to sample, stepping back a sample when the release lands right after one, delayed and bunched callbacks not becoming a flick, the native timestamp being preferred over the dispatch clock, and the state machine ignoring unmatched scroll events and returning to idle after release.

`npm run typecheck`, `npm run lint`, and `npm run format` all pass at the repo root after rebasing onto current `main`.

See Android

https://github.com/user-attachments/assets/dd7565ce-3375-4332-b257-65285eec4e34

See IOS

https://github.com/user-attachments/assets/031cc8cd-f766-46f7-866f-617b478dfaa1

Not affected, so not tested: web and desktop. `strategy-native.tsx` is the native render strategy only; the web strategy is a separate module.

### Risk surface

Known limitation: the speed is derived from scroll offset, so a flick performed while the offset cannot move, the list already at the oldest edge, or a transcript too short to scroll, is not detected and leaves the keyboard up. On iOS the rubber-band still moves the offset, so in practice this is an Android edge case. Both platforms do expose a native `velocity` on the end-drag payload, so a follow-up could cover it, but the sign and unit conventions differ per platform and need device verification.

The threshold (1.5 points/ms at release) is a single constant, calibrated on one Android device and one iPhone.

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran


boudra (approved):
Reviewed the reshaped implementation and CI follow-up. The feature is isolated behind its hook and pure model, uses the existing keyboard-shift shared value, and the full check matrix is green.

PR #2426, opened 2026-07-25
feat: add sidebar declutter options and a Recently done status group

nikuscs:
### Linked issue

No matching open issue found.

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Adds a few device-local options under Appearance > Sidebar for people who wants a quieter sidebar, plus a new group in the status view.

- Hide diff stats, Hide PR status, Hide New workspace row, and Hide script indicators.
- Compact rows now also tightens the header rows, the Show more toggles, the Pinned header, and the status group headers. Before it only affected project and workspace rows.
- Hide host names stays as it was.
- New Recently done group in the status view, shown right above Done, with a window you pick from a dropdown: Off, 1, 5, 15, 30 min, or 1 hour. Default is Off so nothing changes unless you turn it on.

How the Recently done group works: it borrows rows from Done, it is not a new workspace state. A workspace lands there when it finished inside your window, and it falls back into Done on its own once the window passes. A finished workspace with no known transition time stays in Done rather than being guessed into the group. The clock only ticks while the sidebar is grouped by status and the window is on, and the tick interval scales with the window so an hour long window is not polling every second.

All settings are stored locally on the device, so nothing is sent to the daemon and there is no protocol change.

### How did you verify it

- Toggled every new switch in the web app and confirmed the sidebar updates in both project and status grouping.
- Confirmed hiding the New workspace row still leaves the + on the project row for empty projects.
- Confirmed an old stored settings blob without the new keys loads with the current look.
- Sidebar, settings storage, and i18n coverage: 99 tests passed.
- `npm run typecheck`
- `npm run lint`
- `npm run format`

### Risk surface

Client only. Every new preference defaults to today's behavior, and the Recently done split is off by default. The synthetic group key is kept separate from the workspace status bucket, so status dots and other status consumers are untouched.

### Checklist

- [x] `npm run lint` passes

### UI Preview

CleanShot 2026-07-25 at 11 14 33@2x

CleanShot 2026-07-25 at 11 15 10@2x


nikuscs:
Not planned anymore, maybe soon a plugin. Paseo core is "done" for what we need, so closing to keep things simple on our side.

PR #2429, opened 2026-07-25
feat(app): always show user message timestamp and actions

ABorakati:
### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

The timestamp on your own messages shared a hover-gated row with the rewind menu and the copy button, so on a wide web/desktop layout none of them appeared until the pointer was over the message.

The row now renders unconditionally - timestamp and both actions together. 

The gate was `hasText && (isCompact || isNative || isHovered)`. `isNative` is iOS/Android only and `isCompact` is a width breakpoint, so on a maximised desktop window (Electron included) neither applies and only `isHovered` could reveal the row.

Revealing the timestamp but keeping the actions on hover was the first attempt, and it looked wrong: the timestamp floated alone against the right edge with an empty gap beside it where the buttons would later appear. Showing the whole row is both better balanced and simpler.

With nothing left to reveal, this removes the `isHovered` state, both pointer handlers, the `opacity`/`pointerEvents` toggle and the `trailingRowHidden`/`trailingRowVisible` styles - a pure deletion (+2/-27). No hover machinery is left behind to imply hover still matters, which `docs/hover.md` warns against.

No behaviour change on native or compact layouts: they already showed the whole row via the `isNative`/`isCompact` fallback.

Blast radius is confined to `UserMessage`. `userMessageStylesheet` is module-local to `message.tsx`, and the only renderer of the component is `agent-stream/view.tsx`; other `UserMessage` matches in the app are type/store references. `isNative` and `useIsCompactFormFactor` both retain other consumers in the file, so no imports are orphaned.

### How did you verify it

Windows 11, desktop Chrome at 1280x720, driving the repo's own Playwright e2e harness against a seeded mock agent and capturing the user bubble with and without hover, on both `upstream/main` and this branch.

| | before (`main`) | after (this PR) |
| --- | --- | --- |
| default | blank strip under the bubble | `11:51 PM` + rewind + copy |
| hovered | `6:44 PM` + rewind + copy | identical to default |

Before, the row was in the DOM but invisible: the element's text content read `Show me the message timestamps.6:44 PM` while rendering blank at `opacity: 0`. After, the hover and no-hover captures are byte-for-byte identical - that is the evidence the gate is gone.

Density and clutter is a potential risk but inspecting the UI with times / actions enabled doesn't look cluttered IMO.

assistant-full-no-hover

Note: on Windows the e2e harness cannot start without #2212 (`test(e2e): make Playwright global-setup start on Windows`), which is still open. I ran against that branch's `global-setup.ts` locally.

### Checklist

- [x] `npm run lint` passes
- [x ] UI changes include screenshots or video for every affected platform - web/desktop (Windows 11)only; native not exercised
- [ ] Tests added or updated where it made sense - pure deletion of UI state with no new logic; `message.tsx` has no component-test harness

### DIFF
--- a/packages/app/src/components/message.tsx
+++ b/packages/app/src/components/message.tsx
@@ -388,12 +388,6 @@ const userMessageStylesheet = StyleSheet.create((theme) => ({
     gap: theme.spacing[2],
     marginTop: theme.spacing[2],
   },
-  trailingRowHidden: {
-    opacity: 0,
-  },
-  trailingRowVisible: {
-    opacity: 1,
-  },
   timestampText: {
     color: theme.colors.foregroundMuted,
     fontSize: STREAM_METADATA_FONT_SIZE,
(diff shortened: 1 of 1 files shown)


ABorakati:
Superseded by #3044

PR #2434, opened 2026-07-25
perf(build): parallelize server dependencies

mcowger:
## Summary
Reduce `build:server` latency by removing a duplicate protocol compilation and building independent server prerequisites concurrently. On the same checkout, the full build improved from 28.30 seconds to 20.89 seconds (26.2%).

## Changes
- **Client build orchestration**: rely on the client workspace's self-contained build to compile protocol once instead of invoking it twice.
- **Server prerequisites**: build highlight, relay, and client concurrently, stopping sibling builds if one fails.

## Testing
- Benchmarked `npm run build:server` before and after: 28.30s → 20.89s.
- Confirmed the optimized build completes successfully and emits all server and CLI outputs.


mcowger:
Should be a <strong>small</strong> change.

PR #1817, opened 2026-06-29
feat(app): auto-preview HTML files in a sandboxed webview

park0er:
Closes #1816

## What this PR does

Opening an `.html`/`.htm`/`.xhtml` file in the file pane now renders it **live** in a sandboxed webview instead of showing raw source. A specific line target still shows the syntax-highlighted source.

- `file-pane-render-mode`: `resolveTextPreviewMode(path, { hasLineSelection }) → "markdown" | "html" | "code"` (pure). HTML auto-preview unless a line was targeted. (SVG keeps using the existing image viewer — the server serves it as an image.)
- `preview-kind`: `resolveFilePreviewKind` / `isAutoPreviewHtml` extension classifier.
- `html-file-preview`:
  - web/desktop → `<iframe sandbox="allow-scripts">` (no `allow-same-origin` → opaque origin; scripts run for the preview but can't reach the app/DOM/cookies/storage).
  - native (iOS/Android) → `react-native-webview` (already a dependency), an isolated context.
- `file-pane`: routes the text branch through `resolveTextPreviewMode`; renders the sandboxed preview for html; skips syntax highlighting for non-code modes.

## Type of change

- [x] New feature (proposed in #1816)

## How I verified

- `npx vitest run preview-kind.test.ts file-pane-render-mode.test.ts` → 14 pass (incl. the existing `isRenderedMarkdownFile` cases, preserved).
- `oxlint` + `oxfmt --check` clean on all changed files.
- Changed files typecheck (no errors attributable to them).

## Testing evidence / platforms — honest status

Logic is unit-tested and the sandboxing is in place. I have **not** yet captured on-device/render screenshots for desktop/web/iOS/Android — visual verification (web via Playwright, iOS via simulator) is in progress and I'll attach screenshots for the affected platforms before this is ready to merge. Flagging clearly per CONTRIBUTING. Happy to adjust the UX/placement to the maintainer's preference.

Issue #1471, opened 2026-06-11
feat: support PDF file preview


TommyLike:
## Description

Currently Paseo does not support previewing PDF files in the file explorer or workspace tabs. It would be useful to be able to open and view PDF files directly within the app.

## Use Case

- View documentation/reports stored as PDF in the workspace
- Quick preview of generated PDF outputs from agents
- Read papers/articles in PDF format without leaving Paseo

## Question

Is PDF file preview on the roadmap? If so, any timeline or implementation approach in mind (e.g., webview-based rendering, native PDF viewer, etc.)?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ajmeese7:
Duplicate of #515

PR #2452, opened 2026-07-26
Preview PDFs in the file panel

### LINKED ISSUE #1816: feat: auto-preview HTML/SVG files in the file pane
## Summary

When an `.html` (or `.svg`) file is opened in the file pane, render it **live** in a sandboxed webview instead of showing raw source — with the existing source view as the fallback. Agents increasingly produce HTML artifacts (reports, dashboards, mockups), and today you can only read the markup.

## Motivation

- Agent-generated HTML/SVG is far more useful rendered than as source.
- This rounds out the file-preview story alongside existing requests:
  - #1471 (PDF preview), #1166 / #1734 (mermaid), #1263 / #1264 (markdown source toggle), #114 (syntax highlighting). HTML is the obvious missing piece.

## Proposal

- A pure classifier decides the render mode per file: `markdown | html | code` (HTML/SVG auto-preview unless a specific line was targeted, in which case show source).
- Render in a sandboxed surface:
  - **Web/desktop**: `<iframe sandbox="allow-scripts">` with **no** `allow-same-origin` → opaque origin, scripts can run for the preview but cannot reach the parent app, its DOM, cookies, or storage.
  - **iOS/Android**: `react-native-webview` (already a dependency) — an isolated web context.
- Keep a "view source" path (a specific line target falls back to the syntax-highlighted source).

## Security

- The preview surface is sandboxed and cannot access the app context.
- `.svg` is treated as HTML-like (can carry script) → same sandbox.
- Only local workspace files the user explicitly opens are rendered.

I have an implementation ready and will open a PR referencing this issue.

### LINKED ISSUE #1956: feat: "Open in browser" action to view workspace files in the in-app browser (PDF/HTML/images)
## Problem

There is no way to open a workspace file in Paseo's **in-app browser**. This matters most for **PDFs**: Paseo cannot preview them at all today (there is no PDF handling anywhere in `packages/app/src`, and `file-explorer/preview-target.ts` has no PDF branch), which is the open ask in #1471. Building a bespoke PDF renderer is the hard part that issue is stuck on - but the in-app browser (Chromium webview on desktop) already has a **built-in PDF viewer**, so routing the file into it gets preview essentially for free.

The same action would give a proper in-browser view of HTML/SVG (overlaps #1816) and images with native zoom/pan (#1682), all through one mechanism.

## Current behaviour (from the code)

- Paseo already ships an in-app browser pane - `packages/app/src/components/browser-pane.electron.tsx` (and `.web.tsx`), created/navigated by URL via `createWorkspaceBrowser({ initialUrl })` in `packages/app/src/stores/browser-store/index.ts`. It renders a `<browser-element url="…">` (default `https://example.com`).
- `packages/app/src/workspace/open-target-planner.ts` (`planWorkspaceOpenTargets`) already builds per-file **"open in X"** targets: desktop editors via the desktop bridge, plus a **GitHub** target that opens the file at a *URL*. So a URL-based open target is an established pattern.
- There is **no** target that opens the file in the in-app browser, and **no** PDF/other-file preview path (grep for `pdf` / `application/pdf` in `packages/app/src` returns nothing).
- Sibling request #1426 covers "Open in Finder / File Explorer" (native OS file manager) - this issue is the *in-app-browser* sibling to that.

## Expected behaviour

An **"Open in browser"** action on a workspace file (file-explorer context menu and/or the existing "Open" target list) that:

1. Adds a new open target in `planWorkspaceOpenTargets`, alongside the desktop-editor and GitHub targets, gated to platforms where the browser pane exists (desktop/web).
2. On select, opens (or reuses) a workspace browser via `createWorkspaceBrowser({ initialUrl })` pointed at the file.
3. Lets the webview's built-in viewers render it - **PDF** via Chromium's PDF viewer (no new dependency), HTML/SVG/images natively.

## Open question - how the file gets a URL the webview can load

The one real design decision: a `file://<absolutePath>` URL, **or** serving the file over Paseo's existing file/attachment HTTP endpoint (attachments already reach the browser pane as `attachment.url` - see `browser-pane.electron.tsx`). Served-HTTP is likely more robust for **remote / Docker execution**, where `file://` won't resolve to the user's machine - and the planner already distinguishes this via its `isLocalExecution` gate. Flagging for maintainers to steer.

## Notes / related

- Directly delivers the intent of #1471 (PDF preview) via the low-effort webview path.
- Overlaps #1816 (auto-preview HTML/SVG) and #1682 (image zoom/pan) - all free through the browser viewer.
- Sibling to #1426 (Open in Finder/Explorer).
- In-app browser is desktop/web today; mobile (native webview PDF support varies) can be a follow-on.

Happy to work on this if it's a wanted feature - flagging first per CONTRIBUTING before opening a PR.

bendavid:
Opening a `.pdf` in the file panel now renders it, using the PDF viewer each platform already ships rather than a bundled renderer. Chromium's PDFium and WebKit's PDFKit bring text selection, find-in-page, print, native zoom, and accessibility that a canvas renderer would have to reimplement.

## Shape

One shell per platform, all taking the same props so the pane doesn't know which it got:

| File | Platforms | How |
| --- | --- | --- |
| `pdf-preview.web.tsx` | browser, Electron | blob URL in an `<iframe>` |
| `pdf-preview.ios.tsx` | iOS | bytes staged to a file, `WebView` at the `file://` URL |
| `pdf-preview.tsx` | Android, anything else | unsupported message + Download |

Three details are load-bearing and commented at their call sites:

- **Electron needs `webPreferences.plugins`** for its built-in viewer. NPAPI/PPAPI are long gone, so this admits the viewer and nothing else.
- **The web iframe is deliberately unsandboxed** — a sandboxed browsing context blocks plugin content, which is what the viewer is.
- **The blob must be typed `application/pdf`**, or the browser downloads instead of rendering.

## Transport

Reads emitted a whole file as one binary frame. Fine for a 40 KB screenshot, fatal for a PDF: the daemon terminates any physical socket whose outbound buffer passes `MAX_PHYSICAL_SOCKET_BUFFERED_BYTES` (8 MiB), so a large single-frame read killed the connection it was answering on. Reads now emit 256 KiB chunks paced against the client's buffered amount — the same signal the terminal uses — and abort with an explicit error rather than hanging when a client stops draining. Reads are also bounded at 32 MiB, checked against `stat` before any bytes enter memory (previously unbounded). Downloads are unaffected; they stream from disk.

The wire `kind` enum is untouched: widening it would break parsing on older clients, so the kind stays `"binary"` and the mime carries the meaning. That also makes the mime the capability check — an old daemon never sends `application/pdf`, and the pane says "Update the host to preview PDFs".

## Android

**Android has no PDF preview.** Its WebView has never shipped a PDF viewer, so there is no system surface to hand bytes to. The unsupported state carries a Download button, which ends in the share sheet and hands the file to an installed PDF app — the daemon reporting `application/pdf` instead of `application/octet-stream` is what makes that sheet offer PDF viewers. Closing the gap needs an Android-only renderer (pdf.js in a WebView, or a native module over `PdfRenderer`); neither web nor iOS would change to accommodate one.

## Testing

Headless Chromium ships no PDF viewer — pointing an iframe at a blob PDF kills the page target — so both the vitest browser project and the Playwright e2e project are blind to the rendering path. Covered instead is our own logic: blob mime and lifecycle, the read-grant directory, mime detection, and the chunked transfer (chunk splitting, backpressure pacing, stall abort, the size bound).

**Verified:** full monorepo typecheck, lint, format, `expo export --platform web`, and the unit suites for the touched packages.

**Not verified — needs a reviewer with hardware:** none of the three rendering paths has been run. A real browser, a real Electron build, and a real iOS device are the outstanding checks, plus the Download button on Android.

`docs/pdf-preview.md` records the reasoning, the Android gap, and the testability cost.

## Related

**Implements #1471** (feat: support PDF file preview) for web, Electron, and iOS. Deliberately not `Closes` — Android has no system PDF viewer to hand bytes to, so that part of the request stays open.

**Adjacent, not resolved: #1956** ("Open in browser" action to view workspace files in the in-app browser). That proposes a different surface for the same need — routing PDFs/HTML/images into the in-app browser pane rather than previewing in the file pane — and is broader than PDFs.

**Overlaps #1817** (feat(app): auto-preview HTML files in a sandboxed webview, which closes #1816). It solves the same shape of problem — render a non-text type in the file pane, `<iframe>` on web/desktop and `react-native-webview` on native — and touches `file-pane-render-mode`, which this PR also changes. Two things worth deciding across both rather than separately:

- **Classification.** #1817 adds extension-based `resolveFilePreviewKind` / `resolveTextPreviewMode`; this PR dispatches on the daemon-reported mime (`pdf/pdf-mime.ts`). If both land, the file pane ends up with two parallel "what kind of preview is this" mechanisms, and they should probably be unified on one.
- **The sandbox attributes differ on purpose.** #1817 uses `sandbox="allow-scripts"` with no `allow-same-origin`, so untrusted HTML gets an opaque origin and cannot reach the app. This PR uses no sandbox at all, because a sandboxed browsing context blocks plugin content and the built-in PDF viewer *is* plugin content. Both are right for their content type; the rationale is commented at the call site and in `docs/pdf-preview.md`.


boudra:

PR #2453, opened 2026-07-26
Run agents and terminals inside a dev container

bendavid:
A workspace can run its agents and terminals inside a container instead of on the host. The user picks a **container backend** per workspace; `null` means Host and stays the default, so nothing changes unless you ask for it. The only backend today is `devcontainer`, which shells out to `@devcontainers/cli` and Docker/podman.

Design notes, gotchas and the reasoning behind each decision are in [`docs/devcontainers.md`](docs/devcontainers.md).

## How it works

Everything routes through one seam, `ProcessLaunchStrategy`:

- `LocalLaunchStrategy` spawns on the host — today's behaviour, unchanged.
- `ContainerExecLaunchStrategy` wraps the same call in the backend's exec.

A workspace with a running container resolves to the second, everything else to the first, and nothing outside `devcontainer/` knows what a container is. The exec is described by data (`ContainerExecSpec`), which also lets it cross into the terminal worker — that worker can only receive serializable values.

| Work | Where | Why |
| --- | --- | --- |
| Agent processes | Container | The point of the feature |
| Terminals | Container | The same shell the agent sees |
| Agent-requested commands (ACP terminals) | Container | The agent asks for them in its own workspace |
| Provider catalogs / model probes | Container | The container's tool version is the one that will run |
| Git | **Host** | A worktree's `.git` points at an unmounted host path; credentials and SSH agent live on the host |
| Container lifecycle | Host | `devcontainer up`, `docker stop`, `docker inspect` |

## No quiet fallback

If a container is required and not running, agent and terminal creation **fail**. The user asked for isolation, and silently not providing it is the worse outcome. A provider that cannot honour the strategy is refused up front via `supportsIsolatedLaunch` rather than run on the host.

The one thing deliberately *not* an error is a **stopped** container during a catalog refresh: the tool list is unknown until it starts, which is a state, not a failure the user can act on, so providers read as unavailable.

## Providers

Claude, Codex, omp, Pi and all ACP providers (Copilot, Cursor, Kiro, generic) run in containers.

**OpenCode does not**, and cannot with this seam: it is not a subprocess Paseo spawns but a shared HTTP server (`opencode serve --port N`) that every workspace multiplexes over by passing `directory`. Supporting it means a server per container, a route from the daemon to a port inside it, container paths in `directory`, and its `$PASEO_HOME/opencode-home` state moving too. Container workspaces refuse it rather than pretending.

## Things worth reviewing closely

- **Container identity** — containers carry `paseo.container` / `paseo.owner` id-labels alongside the ones the CLI infers, so a directory with two workspaces gets two containers and a probe can never adopt (or `docker stop`) a workspace's container.
- **Transcripts live where the agent ran** — providers keep sessions outside the workspace, under the *container's* HOME and keyed by the *container's* cwd. `LaunchFileSystem` reads them through the strategy: one `find … -exec stat` rather than a walk plus a stat per file, because each operation is a process spawn (~75ms) rather than a read (~1ms).
- **Files a provider is configured through** — Pi takes `--mcp-config` and `--extension` as paths it opens itself, so they are written *into* the container. Writes throw where reads answer null, and the MCP config chmods in the same command since it carries the daemon's auth token.
- **Executable resolution** — `docker exec` starts a bare process with only the image's PATH, while a terminal starts a shell that reads the user's startup files. `resolveExecutable` asks the environment's own login shell and returns an absolute path, so the launch depends on no PATH at all. Only successes are cached: a container still starting would otherwise be remembered as "agent missing" until rebuilt.
- **The new-workspace probe** — picking a backend for a workspace that does not exist yet raises a question only the container can answer, so the daemon builds a throwaway container, lists each provider inside it, and removes it again, with cancellation, de-duplication, streamed progress, and a private snapshot key.

## Known gaps

Recorded in the doc, repeated here so they are not a surprise:

- The agent's own subprocesses inherit the image's PATH, so a tool the agent shells out to that lives in a startup-file PATH entry is still not found.
- Shell integration and the bundled hook CLI are injected into the host-side environment, so a container terminal gets neither.
- Provider snapshots are keyed by cwd, so two workspaces sharing a directory with different backends overwrite each other's provider list.

## Testing

- Unit tests with a mock backend for session wiring, status and the strategy's own logic.
- `dockerTest(...)` cases that run a real `devcontainer up` against `alpine:latest` and assert commands execute inside the container. These are what catch exec-argument and environment mistakes; the unit tests cannot. They skip when Docker is not on PATH.
- `FakeIsolatedLaunchStrategy` for provider work: reports `isIsolated`, maps paths as a container does, runs the resulting POSIX commands on the host. Most of container support is a question of *which* environment a provider addresses, and that is answerable without a runtime.


boudra:

PR #2461, opened 2026-07-26
feat: stream TTS audio chunks as synthesized

### LINKED ISSUE #2460: TTS playback starts only after full synthesis completes, causing noticeable delay
## Problem

When the `speak` MCP tool is called, there is a noticeable delay before any audio plays. The entire text is synthesized to audio before playback begins — even for long responses.

**Repro steps:**
1. Enable voice mode in Paseo
2. Ask Claude a question that produces a multi-sentence response
3. Observe: silence until the full response is synthesized, then audio plays all at once

**Expected:** Audio starts playing as soon as the first chunk is synthesized (streaming playback)

## Root cause

In `packages/server/src/server/speech/providers/local/sherpa/sherpa-tts.ts`, synthesis uses `tts.generate()` which blocks until the full audio is ready:

```ts
const result = this.tts.generate({ text, sid: 0, speed: 1.0 });
```

The resulting buffer is then artificially split into ~50ms chunks and wrapped in a `Readable` stream — but by then the entire audio is already in memory, so the "stream" is just a buffered batch delivery.

## Proposed fix

Sherpa ONNX exposes a `generateWithCallback` API that emits PCM chunks incrementally as synthesis progresses. Switching to this would allow:

1. `sherpa-tts.ts` — use `generateWithCallback` to emit chunks as they arrive
2. `worker-process.ts` — stream chunks back through the worker protocol instead of collecting all first
3. `worker-client.ts` — consume the stream incrementally rather than awaiting the full result

This would reduce time-to-first-audio from (full synthesis duration) to (first chunk synthesis duration), which is typically just a fraction of the total.

iamarpitpatidar:
Closes #2460

## Problem

TTS playback starts only after the entire response is synthesized. For a multi-sentence response this means several seconds of silence before any audio plays.

## What this does

Uses `sherpa-onnx`'s `generateAsync` API with a progress callback so PCM audio chunks are streamed to the client over IPC **as synthesis progresses**, rather than buffering the full audio first.

### Changes

- **`sherpa-tts.ts`** — adds `generateAsync` to the native interface and a `synthesizeStreaming` path that pushes chunks into a `PassThrough` stream via the progress callback. Falls back to the existing batch path on older sherpa-onnx builds that lack `generateAsync`.

- **`worker-protocol.ts`** — adds `tts.start` (carries format) and `tts.chunk` (carries audio bytes) event message types to the IPC protocol.

- **`worker-process.ts`** — sends `tts.start` before iterating the stream, then sends each chunk as a `tts.chunk` message instead of collecting everything into one buffer.

- **`worker-client.ts`** — resolves the `synthesizeSpeech` promise at `tts.start` time (before synthesis finishes) with a live `PassThrough` stream; chunk messages push to the stream; the final `response` message ends it. Error and timeout paths destroy the stream rather than rejecting a dead promise.

## Testing

Tested manually with voice mode on macOS. I noted the latency improvement subjectively — audio begins significantly sooner on longer responses.


iamarpitpatidar:
@boudra thoughts ??

boudra:
Closing for now: this needs measured before/after time to first audio with local Sherpa and a regression test showing audio crosses the worker boundary before synthesis finishes. Please reopen with that evidence.

boudra:

PR #2467, opened 2026-07-27
Keep Android bottom sheets above the software keyboard

### LINKED ISSUE #2466: bug: Android software keyboard covers bottom sheets (e.g. Add custom model)
### What's broken

On Android, bottom sheets with a text input don't move when the software keyboard opens, so the keyboard draws over the sheet. With the "Add custom model" sheet (fixed 40% height) the keyboard covers it completely — the Model ID field and the Cancel/Add buttons are hidden while typing.

Expected: the sheet lifts above the keyboard so the focused field stays visible.

### Steps to reproduce

1. Android app, connected to a daemon
2. Settings → Providers → open a provider (e.g. Claude) → "Add custom model"
3. Tap the Model ID field
4. The keyboard opens over the sheet; the input and both buttons are covered

### Where did this happen

Android app

### Paseo version

0.2.2 (release APK built from current `main`)

### OS version

Android, physical phone. The app runs edge-to-edge (`edgeToEdgeEnabled`), so the window doesn't resize for the IME on any Android version.

### Agent provider

Not relevant — the same thing happens in every sheet with a text input (add-custom-model, command center search, combobox search, context menu search).

### Screenshots or video

Before — keyboard fully covering the "Add custom model" sheet:

Image

Image

dwyanewang:
### Linked issue

Closes #2466

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

On Android the app runs edge-to-edge (`edgeToEdgeEnabled: true` — the Expo 54 default, and effectively forced once targeting SDK 35 / Android 15), so the window no longer resizes when the software keyboard opens; the IME draws over the app. All four gorhom sheet call sites used `keyboardBehavior="extend"`, which on keyboard open only moves the sheet to its highest detent **without subtracting the keyboard height** — that only avoided the keyboard back when `adjustResize` shrank the window. Sheets with a single detent (e.g. "Add custom model", fixed at 40%) therefore don't move at all and end up fully covered.

Switches all four call sites (`adaptive-modal-sheet`, `command-center`, `context-menu`, `combobox`) to `keyboardBehavior="interactive"`, which positions the sheet at `highestDetent − keyboardHeight`, so the sheet's bottom tracks the top of the keyboard regardless of window resizing. `android_keyboardInputMode="adjustPan"` comes along because gorhom skips the interactive offset when it believes the window is `adjustResize` — an assumption edge-to-edge breaks. `keyboardBlurBehavior` is unchanged at each site.

### How did you verify it

Release APK (0.2.2) built from this branch, installed on a physical Android phone:

- Before the fix (screenshot in #2466): Add custom model → tap Model ID → the keyboard covers the whole sheet; typing is blind and Cancel/Add are unreachable without dismissing the keyboard.
- After the fix: the same flow lifts the sheet so its bottom sits on top of the keyboard; the field and both buttons stay visible while typing (screenshot below). Dismissing the keyboard returns the sheet to its 40% detent, and dragging the sheet while the keyboard is open doesn't jump.
- The other three call sites got the identical two-prop change; I haven't exercised each of them individually on device.
- iOS not tested (same two props; `interactive` is gorhom's documented behavior there too). Web/desktop unaffected — these props only respond to the native software keyboard.

`npm run typecheck`, `npm run lint`, `npm run format` all pass.

After — the sheet riding on top of the keyboard, input and buttons visible:

image

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (Android before/after attached; iOS not tested, stated above; web/desktop unaffected)
- [ ] Tests added or updated where it made sense (no keyboard-interaction test harness for sheets; prop-level change only)

### DIFF
--- a/packages/app/src/command-center/command-center.tsx
+++ b/packages/app/src/command-center/command-center.tsx
@@ -746,8 +746,9 @@ export function CommandCenter() {
         enablePanDownToClose
         backgroundStyle={styles.sheetBackground}
         handleIndicatorStyle={styles.sheetHandle}
-        keyboardBehavior="extend"
+        keyboardBehavior="interactive"
         keyboardBlurBehavior="restore"
+        android_keyboardInputMode="adjustPan"
         accessible={false}
       >
         <View style={[styles.bottomSheetHeader, styles.searchRow]} testID="command-center-header">
--- a/packages/app/src/components/adaptive-modal-sheet.tsx
+++ b/packages/app/src/components/adaptive-modal-sheet.tsx
@@ -644,8 +644,9 @@ export function AdaptiveModalSheet({
         enablePanDownToClose
         backgroundComponent={SheetBackground}
         handleIndicatorStyle={handleIndicatorStyle}
-        keyboardBehavior="extend"
+        keyboardBehavior="interactive"
         keyboardBlurBehavior="restore"
+        android_keyboardInputMode="adjustPan"
         accessible={false}
         presentation={presentation}
       >
--- a/packages/app/src/components/ui/combobox.tsx
+++ b/packages/app/src/components/ui/combobox.tsx
@@ -994,8 +994,9 @@ function MobileComboboxBody(props: MobileBodyProps): ReactElement {
       enablePanDownToClose
       backgroundComponent={ComboboxSheetBackground}
       handleIndicatorStyle={props.handleIndicatorStyle}
-      keyboardBehavior="extend"
+      keyboardBehavior="interactive"
       keyboardBlurBehavior="none"
+      android_keyboardInputMode="adjustPan"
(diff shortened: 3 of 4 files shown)


dwyanewang:
@boudra Rechecked this against current `main` (`3bca9b7a9`): the issue is still present upstream. All four native sheet call sites still used `keyboardBehavior="extend"` without `android_keyboardInputMode="adjustPan"`, so this fix is still needed.

I rebased the branch onto current `main`. The focused four-file diff is unchanged, and there are no additional `extend` call sites to cover.

Verification after rebase:
- `npm run format`
- `npm run build:server`
- `npm run typecheck`
- `npm run lint`

All passed.

PR #2498, opened 2026-07-27
Read Claude subagents from the SDK task protocol

ebg1223:
Paseo reconstructs Claude subagents by watching sidechain frames — messages tagged `parent_tool_use_id` — and inferring identity and lifecycle from them. That fails in a specific way: a **backgrounded subagent emits no such frames at all**, so it's entirely invisible today.

Claude Code already announces this directly. `task_started` / `task_updated` / `task_notification` / `task_progress` carry the id, type, task, status, and running cost. This reads those announcements instead of inferring. Backgrounded children appear, and live cost lands with them — `task_progress` fires once per tool use, not on a timer.

Concretely, a fan-out of three Explores in the track:

```
before                          after
─────────────                   ────────────────────────────────────
Explore                         Summarize the hover and unistyles docs
Explore                         Explore · claude-opus-5 · high
Explore
                                Find every call site of useUnistyles
(a backgrounded child:          Explore · claude-opus-5 · high
 no row at all)
                                Check the Android panel gestures
                                Explore · claude-sonnet-5 · medium
```

Every row was previously the subagent type with an empty subtitle, so siblings in a fan-out were indistinguishable. The task now names the row; type, model, and effort drop to the second line. Model and effort are read from the child's own frames, never inherited — absent renders as nothing rather than the parent's setting. Cost stays out of the row and lives in the pane header, which has the room: `claude-opus-5 · high · 16.5k tokens, 2 tools, 10.9s`.

The parent's Task card gets the same treatment — it rendered as a bare `Task` for a backgrounded child, since the tracker that supplies its detail never ran.

The other half is that live and replay now speak one vocabulary: both translate into a small `SubagentObservation` union, and a pure fold turns that into store events. The paths had drifted precisely because each derived the same facts its own way. The test worth looking at runs a recorded session through both and asserts identical descriptor state.

Some of this is undocumented Claude Code internals, so I tried to separate verified from assumed. Wire shapes are captured from a real CLI (2.1.220) and fixtures copy actual output, not the type declarations. `usage.total_tokens` turns out to be a context-size reading at the final turn rather than cumulative spend, so the replay reconstruction matches that definition instead of summing per-entry usage — which would multiply the cached prefix by turn count.

Three corrections rather than features:

**A status for a never-declared task is dropped.** Filtered tasks (backgrounded shells, workflow children, housekeeping) still emit notifications and frames carrying a `tool_use_id`. Attributing those yields a descriptor with no identity and a defaulted `running` status — a nameless row that never finishes.

**Task ids are session-scoped, not turn-scoped.** Cancelling a turn used to clear the routing table, stranding exactly the backgrounded children this PR makes visible. Cancellation now terminalizes the foreground children and leaves routing intact, so a late `task_notification` can still land.

**Replay skips grandchildren.** Claude Code writes every descendant into the same `subagents/` directory; `spawnDepth` separates them. A grandchild's `toolUseId` names a Task call from its *parent's* session, so nothing in the replayed transcript resolves it — an extra row the live stream never showed, with no outcome, running forever. One recorded session showed 10 subagents live and would have replayed 22. This bug predates the branch; it's fixable here only because reading the meta sidecar is what makes `spawnDepth` reachable.

That last category got sharper with #2458: idle collection now keeps a parent resident while any subagent reads `running`, so a stale descriptor pins an agent runtime for the daemon's lifetime. Rebased on it; `agent-manager`'s 145 tests pass.

Compatibility: protocol changes are additive (new optional fields, primitives only inside `usage`), the store's status became optional so a partial upsert can't revert a finished child, and the sticky merge was checked against all three existing providers. CLIs predating the task protocol keep the old derivation — the source reports whether a session announces tasks rather than sniffing a version.

24 files, ~3.3k lines, over half tests. Nine commits, each self-contained.


cleiter:
Providing missing screenshot:

image

cleiter:
Also displays details now when you click on it:

image

Personally I would prefer to see the tokens already in the overview.

ebg1223:
@cleiter honestly i went back and forth on this, tried to make it less crowded on mobile. maybe if it shows tokens if > certain width?

PR #2514, opened 2026-07-28
Add multi-account profiles for Claude and Codex

### LINKED ISSUE #2475: feat: Support multiple accounts from the same provider
### Prior search

### What's the problem

Support multiple accounts from the same provider somehow.

### What would solve it

Should be pretty simple.

_No response_

colonelpanic8:
## Problem

Paseo could configure provider overrides, but it had no first-class way to run multiple authenticated accounts for the same provider. Account settings also need to be owned by the selected remote host, because credential directories live on that host.

Closes #2475.

## What changed

- Add a host-scoped **Settings → Providers → Accounts** UI for Claude Code and Codex / ChatGPT profiles.
- Represent each account as a derived provider using an isolated `CLAUDE_CONFIG_DIR` or `CODEX_HOME`.
- Drive everything from a single manifest declaration (`accounts: { envVar, directoryExample }`) on the provider definition — the settings row, the config patch, and the daemon's account-directory lookup. No code names specific providers.
- Route Claude catalog discovery, launches, imports, transcript history, and cleanup through the selected account directory.
- Route Codex app-server launches, custom prompts, and fallback skills through the selected Codex home.
- Capability detection rides on the provider snapshot: a daemon too old to isolate accounts never sends `accounts`, so the option never appears. (The earlier `server_info.features.providerAccounts` flag was dropped as redundant.)
- Fan usage quotas out per account, with each account's fetcher pinned to its own credential directory — a signed-out account reads as unavailable instead of reporting the default account's quota.
- Persist safe account metadata across daemon restarts while redacting provider environment variables from config responses, so API keys are never exposed to clients.
- Document account authentication and file-based Codex credential storage.

### Fixes after review

- **Usage rows survive daemon restarts.** `listProviderAccountProfiles` only read raw `env`, but the boot-time config deliberately drops `env` and carries the derived `accountConfigDir` — so every account's usage row vanished on restart. It now accepts both shapes.
- **Runtime-added providers actually load.** Registering an account reconciled it into every open workspace snapshot as `unavailable`, a state the warm-up path never revisits, so the new account's models never reached the model search menu. Newly registered providers now reconcile as `loading` and each snapshot warms them immediately.

## QA

- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- Focused app/protocol/config/server regression tests, including new coverage for the restart-shaped account config and for a runtime-added provider reaching `ready` in an already-primed workspace snapshot
- Focused Claude catalog/import/env tests
- Focused Codex home/env tests
- Focused provider registry, icon, and quota-fetcher account tests

Manual visual QA was not run on iOS, Android, browser web, or Electron desktop.


UnbrokenHunter:
Why was this closed?

colonelpanic8:
@UnbrokenHunter the support already exists in the configuration system and it just needs ui, so I opened https://github.com/getpaseo/paseo/pull/2785 instead.

PR #2525, opened 2026-07-28
Snooze workspaces into a collapsed sidebar state

### LINKED ISSUE #2524: feat: Add workspace snoozing (inspired by t3code)
## Problem

Active agents can clutter a workspace when they are still useful but do not need attention right now. Closing or archiving them is too destructive, while leaving them visible makes the active tab list noisy.

## Proposed behavior

- Add snooze actions to the agent tab three-dot/context menu on desktop and mobile.
- Offer useful presets plus a custom date/time picker.
- Represent snoozing as first-class persisted status metadata, separate from the agent lifecycle status.
- Hide snoozed agents until their wake time.
- Wake a snoozed agent early if it requests attention or a permission.
- Allow the client to clear the snooze explicitly.

## QA expectations

Cover protocol/storage round-tripping, attention-driven wake-up, scheduled visibility changes, desktop context-menu behavior, and the responsive mobile ellipsis flow.

colonelpanic8:
Closes #2524

## Problem

A workspace can be in a perfectly fine state that you simply do not want to look at right now — a review you will get to tomorrow, a long-running experiment, a branch you are parking until Monday. Archiving it is too destructive and it is not "done", so it keeps sitting in the sidebar competing for attention with work that actually needs you.

## What changed

Snooze is a **workspace-level state**, not an agent-level one, and it moves a workspace into a new sidebar state rather than hiding anything.

- Snooze presets (1 hour / this evening / tomorrow / next week) and a custom date-time option in the **sidebar workspace row menu**, in both project- and status-grouped modes. Snoozed workspaces get a Wake action.
- Custom snooze uses platform-native controls: `datetime-local` on web, compact iOS controls, and Android's system date/time dialogs.
- New **Snoozed** group in status-grouped mode, rendered below Done with a blue moon icon, **collapsed by default**, sorted soonest-wake-first. Snoozed rows carry no status dot.
- Snooze state is persisted workspace metadata (`snoozeStatus: { snoozedAt, snoozedUntil }`), stored and updated the same way workspace pinning is, over a new `workspace.snooze.set` RPC.
- Nothing is hidden: agents, tabs, and workspace content stay exactly as they were. Snooze only changes where the workspace sits in the sidebar.
- The sidebar re-derives on a wake clock scheduled off the next `snoozedUntil`, so a workspace leaves the Snoozed group the moment it expires — no polling.
- Translations for all eight app locales.

### Snooze semantics

- Any non-archived workspace can be snoozed, whatever state it is in.
- Snooze suppresses whatever was true **at the moment you snoozed** — including an existing "Ready to review" state, which is the common case for "I will look at this tomorrow".
- New attention-worthy activity that arrives **after** the snooze (needs input, failure, a finished turn) breaks the workspace back out of Snoozed immediately.
- `Working` and `Done` stay suppressed for the whole snooze window.
- Waking clears the metadata; there is no server-side auto-clear, so the state is purely derived from `snoozeStatus` plus the clock.

## Compatibility and impact

- `snoozeStatus` is a new **optional** field on the workspace descriptor and stored workspace metadata, so old clients keep parsing new payloads and existing workspaces load with no snooze status. No migration.
- `"snoozed"` is a **client-side-only** sidebar bucket. The wire `WorkspaceStateBucket` enum is unchanged — adding a value there would break parsing on older clients.
- The snooze actions are gated on a `server_info.features.workspaceSnooze` capability flag (tagged `COMPAT(workspaceSnooze)`); against an older daemon the actions simply do not appear, with no fallback path.
- The sidebar's persisted collapsed-groups shape is unchanged. Default-collapsed groups invert the meaning of key presence (present = user expanded it), so existing users with saved sidebar state also get Snoozed collapsed by default.
- The daemon validates snooze wake times as future values and normalizes them to ISO timestamps.

## QA

- `npm run typecheck` — clean for protocol, client, server, app, and cli
- `npm run lint` — 0 warnings/errors
- Focused Vitest coverage — new and updated tests across app, server, and protocol pass individually, covering: workspace snooze persistence and RPC round-trip, the derivation rules (pre-existing attention suppressed, post-snooze attention breaks through, expiry, running/done suppressed), Snoozed group ordering by wake time, collapsed-by-default plus expand toggle, and persisted-collapse back-compat
- Playwright `workspace-snooze-menu.spec.ts` was rewritten to drive snooze from the sidebar row menu (presets → custom sheet → persisted `snoozeStatus` → Wake, plus the collapsed-by-default Snoozed group). It typechecks but was **not** executed locally — it needs CI.

## Screenshots

⚠️ The screenshots below are from the previous agent-level, tab-menu version of this feature and no longer reflect the UI. They need to be recaptured against the sidebar row menu and the Snoozed status group.

<p>
Snooze presets (outdated: previous tab-menu version)
Custom snooze picker on desktop web
</p>


colonelpanic8:
Superseded by the consolidated personal branch `colonelpanic8:personal/sidebar-workflow`. This customization is now maintained as part of that single non-upstream sidebar workflow.

PR #2526, opened 2026-07-28
Show recently archived workspaces in the status sidebar

### LINKED ISSUE #2523: No quick way to undo a mis-archived workspace from the sidebar
### Problem

Archiving a workspace from the sidebar is a single click, and it's easy to do to the wrong row. Once it's gone there's no fast way back: the sidebar has no memory of what you just archived, so recovering means leaving the sidebar for History, finding the workspace, and recovering it from there. The gesture that caused the mistake takes one click; undoing it takes a detour.

This is most annoying in status mode, where rows are reordered by activity rather than sitting in a stable per-project position — so the row you archived by accident isn't somewhere you can easily re-derive.

### Expected flow

The sidebar remembers the handful of workspaces you archived most recently and keeps them visible, de-emphasized, at the bottom of the list, newest first. Seeing the mistake and undoing it stays in the sidebar.

### Notes

- Archived workspaces already carry `archivedAt` server-side, so ordering by archive time is available.
- Unarchive already exists as `workspace.recovery.restore`, which correctly picks between a plain unarchive and recreating a Paseo-owned worktree.
- The daemon currently never sends archived workspaces to clients, so this needs a way to ask for them that does *not* put them into the workspace directory — anything reading that map treats a present descriptor as live.

Happy to take direction on scope (or to drop it) — I have a working implementation and will open a PR against this issue, but I'd rather align on whether this belongs in the product before you spend review time on it.

colonelpanic8:
### Linked issue

Closes #2523

### Type of change

> Opened alongside the issue rather than after it, so there's something concrete to react to. If the answer is "not this, or not now," close it — no hard feelings.

### What does this PR do

Archiving a workspace from the sidebar is one click and easy to misfire. Getting back took a detour through History. Status mode now ends with a greyed-out **Recently archived** tail: the 5 most recent archives, newest first, each with an unarchive button.

The rows come from a new `workspace.archived.list` RPC, **not** the workspace directory. That split is the load-bearing decision here — everything reading the session store's workspace map (project mode, the switcher, status counts) treats a present descriptor as live, so archived workspaces must never land in it. They live only in the `archivedWorkspaces` query, which only status mode reads.

`archived` is deliberately **not** a status bucket. It stays out of `STATUS_BUCKET_ORDER` and out of the 1–9 shortcut index, since archived workspaces have no live status and nothing to navigate to. Clicking a row does nothing; unarchive is the only action, and it reuses `workspace.recovery.restore` so the daemon still decides between a plain unarchive and recreating a Paseo-owned worktree.

Gated on `server_info.features.archivedWorkspacesList`. Older daemons contribute no rows — no fallback path.

Bounds: the daemon caps at 50 and drops workspaces whose project is archived (unarchiving one would surface it under an invisible project); the client requests 25 per host and caps the merged list at 25; the UI shows 5 with show-more. No pagination or infinite scroll — History remains the full archive.

One refactor rides along: host-filter resolution moved into `resolveSidebarServerIds` so the tail and the workspace list can't disagree about which hosts are visible. Shared by both call sites rather than duplicated, and covered by new tests.

### How did you verify it

**Platform coverage is incomplete — this was developed and verified on Linux.**

| Platform | Status |
| --- | --- |
| Web (browser) | ✅ Verified, screenshots below |
| Web (compact/phone viewport) | ✅ Verified, screenshot below — real code path (`useIsCompactFormFactor`), but a narrow browser window, **not** a device |
| Desktop (Electron) | ❌ Not verified. Electron launches, but the desktop dev harness' managed-daemon handshake fails on my NixOS box (daemon binds, app polls `errored`). Unrelated to this change; I couldn't get past it. |
| iOS | ❌ Not possible for me — Linux only, no macOS/simulator |
| Android | ❌ Not verified — no emulator binary or SDK installed, no device attached |

Happy for someone with a Mac and a device to sanity-check the two mobile platforms. There is no platform-specific code in the new component — no `isWeb`/`isNative` branches and no `.web.tsx`/`.native.tsx` split — so the risk is layout/visual rather than functional.

Driven through a Chromium session against an isolated daemon (`PASEO_HOME=/tmp/...`, non-default port) seeded with two scratch git projects, 5 active workspaces, and 8 archived at staggered timestamps.

1. Switched grouping to Status. The tail renders after **Done**, newest-archive-first, showing 5 rows with **Show more**. The unarchive icon is visible at rest on every row:

Recently archived tail

2. Hovering a row lifts the icon out of its resting opacity:

Hover

3. Clicked unarchive. The workspace left the tail and reappeared at the top of **Done**, and the daemon confirmed independently — `paseo workspace ls` listed it as an active worktree again:

After unarchive

4. Same tail at a 390x844 phone viewport, sidebar as an overlay. The unarchive icon is visible on every row without hover — which is the point of keeping it out of a hover reveal, since touch has no hover:

Compact layout

Also verified: archiving from the sidebar puts the workspace in the tail immediately (the shared archive path invalidates the query), and collapse state persists per group.

Automated: 3 new server tests (ordering, `limit`, archived-project exclusion), 4 for the cross-host merge (interleaving, unresolved hosts, tie-break stability, cap), 4 for `resolveSidebarServerIds`. Existing sidebar projection, status view-model, and i18n suites pass. `npm run typecheck`, `npm run lint`, `npm run format` all clean.

**Known gaps I'd rather flag than hide:**
- The tail refreshes on archive/unarchive *from the sidebar*. Unarchiving via the workspace route's recovery action or from another client leaves a stale row until the 30s stale time lapses.
- Every seeded workspace sat in `done`, so the tail's visual separation from a populated `needs_input`/`Working` list isn't stress-tested.
- Translations for the 7 non-English locales are machine-produced; the type requires full parity, so they had to be filled in. Worth a native check.

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran
- [ ] UI changes include screenshots or video for every affected platform — **web + compact only; desktop/iOS/Android unverified, see table above**

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -90,6 +90,29 @@ leave both Paseo's `archivedAt` and the provider's native archive state unchange
 the only transition back to an interactive runtime: it runs the provider's native unarchive hook
 (including Codex `thread/unarchive`) before the normal agent resume and timeline hydration flow.
 
+### The recently-archived tail
+
+Status mode renders a greyed-out **Recently archived** group after every real status group, newest
+archive first, so a mis-archive is one click from being undone. It is an undo affordance, not an
+archive browser.
+
+The rows come from `workspace.archived.list.request` (capability
+`server_info.features.archivedWorkspacesList`), **not** from the workspace directory. This split is
+deliberate and load-bearing: every consumer of the session store's workspace map — project mode, the
+workspace switcher, status counts — treats a present descriptor as live. Archived workspaces must
+never enter that map, so they live only in the `archivedWorkspaces` query
+(`packages/app/src/hooks/use-archived-workspaces.ts`) and only status mode reads it. If you ever need
+archived workspaces somewhere else, add another reader of that query; do not widen the directory.
+
+`archived` is not a status bucket. It is deliberately excluded from `STATUS_BUCKET_ORDER` and from the
+1–9 keyboard shortcut index — archived workspaces have no live status and nothing to navigate to. The
+only action on a row is unarchive, which goes through the existing
+`workspace.recovery.restore.request`, so the daemon (not the client) still decides between plain
+unarchive and recreating a Paseo-owned worktree.
+
+The daemon caps its response and drops workspaces whose owning project is archived; unarchiving one
+would surface a workspace under a project the user cannot see.
+
 Provider session connection owns every process it spawns until the session is registered with
 `AgentManager`. If initialization, persisted-session resume, or initial history hydration fails,
 `connect()` must dispose that process before rethrowing; the manager cannot clean up a session it never
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -113,6 +113,11 @@ unarchive and recreating a Paseo-owned worktree.
 The daemon caps its response and drops workspaces whose owning project is archived; unarchiving one
 would surface a workspace under a project the user cannot see.
 
+While status mode is mounted, its archived query listens to the existing `workspace_update`
+broadcast. A live-directory removal can mean a new archive, and an upsert for a row already in the
+archived cache means that row was restored elsewhere, so either transition refreshes the tail across
+clients without polling.
+
 Provider session connection owns every process it spawns until the session is registered with
 `AgentManager`. If initialization, persisted-session resume, or initial history hydration fails,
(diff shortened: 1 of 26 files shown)


colonelpanic8:
Superseded by #2529, which now includes the recently archived tail, the live-to-archived transition animation and pending spinner, and the outstanding review fixes.

PR #2539, opened 2026-07-28
fix(pi): use get_available_commands RPC name for OMP compatibility

### LINKED ISSUE #1692: bug: OMP/Pi provider slash commands & skills never load — daemon calls renamed RPC `get_commands` (omp uses `get_available_commands`), times out after 30s
### What's broken

With the `omp` provider `available` and `Enabled` (`paseo provider ls` is green on 0.1.99), the agent UI never loads slash commands or skills:

- The model selector in the composer is stuck on **"Loading…"**.
- Typing `/` shows **no** command/skill suggestions.
- Invoking a skill command (e.g. `/ctx stats`) makes the agent reply that the skill doesn't exist (the text is sent as a normal message because Paseo's command list is empty).

This is **not** the `unavailable` regression from #1665/#1657 (fixed in #1660 / 0.1.99). The provider is available; only command/skill discovery is broken.

### Root cause

The Pi/OMP RPC command for listing commands was **renamed** in OMP. Paseo still sends the old name.

- `packages/server/src/server/agent/providers/pi/cli-runtime.ts` → `PiCliRuntimeSession.getCommands()` sends:
  ```ts
  const data = (await this.request({ type: "get_commands" })) as { commands?: PiRpcSlashCommand[] };
  return data.commands ?? [];
  ```
- OMP `16.1.16` (`omp --mode rpc`) no longer accepts `get_commands`. It returns:
  ```json
  {"type":"response","command":"get_commands","success":false,"error":"Unknown command: get_commands"}
  ```
  Note this error response has **no `id`**, so `handleResponse()` (which requires `response.id`) drops it. The pending request therefore never settles and hits `DEFAULT_TIMEOUT_MS` (30 000 ms).

Daemon log shows the symptom repeating on every UI `list_commands_request`:
```
{"level":50,"msg":"Failed to list commands","err":"Pi RPC request timed out for get_commands"}
{"level":40,"msg":"ws_slow_request","requestType":"list_commands_request","durationMs":31037}
```

The current accepted request type is **`get_available_commands`**, which:
- returns `success:true` and **echoes the request `id`**, and
- carries the payload at `data.commands` — the exact shape `getCommands()` already expects (`handleResponse` resolves `response.data`).

Probe of OMP 16.1.16 RPC over stdio:

| request type sent | result | `id` echoed? |
|---|---|---|
| `get_commands` (what Paseo sends) | `success:false`, `Unknown command` | ❌ no → Paseo times out |
| `get_available_commands` | `success:true`, `data.commands` (58 entries) | ✅ yes |
| `get_state`, `get_available_models`, `get_session_stats`, `get_messages` | `success:true` | ✅ yes |

(OMP also proactively pushes `{"type":"available_commands_update","commands":[…]}` on boot, but Paseo's pull request is what `listCommands()` depends on.)

### Suggested fix

In `packages/server/src/server/agent/providers/pi/cli-runtime.ts`, `getCommands()`:

```diff
-    const data = (await this.request({ type: "get_commands" })) as {
+    const data = (await this.request({ type: "get_available_commands" })) as {
       commands?: PiRpcSlashCommand[];
     };
     return data.commands ?? [];
```

Update the `PiRpcCommand` union in `packages/server/src/server/agent/providers/pi/rpc-types.ts` accordingly (rename the `get_commands` member to `get_available_commands`).

Response shape already matches, so no other change is needed. If back-compat with older OMP/Pi that only understand `get_commands` is desired, fall back to `get_commands` when `get_available_commands` returns `Unknown command`. (Separately, OMP's `Unknown command` error response omitting the `id` is what turns a fast error into a 30s hang — Paseo could log unmatched error responses to surface this faster.)

### Verified

Patched the installed `@getpaseo/server` `cli-runtime.js` with the one-line rename, restarted the daemon, and called `PiCliRuntime#getCommands()` directly:

```
getCommands() returned: 58 commands
skill commands: 27 (skill:codebase-design, skill:context-mode, skill:ctx-doctor, skill:ctx-stats, …)
```

Before: 30s timeout, empty list. After: full 58-command list incl. all skills and plugin commands. Daemon log: 0 `Failed to list commands`.

### Where did this happen

Desktop app + daemon (CLI `paseo provider ls` shows the provider as available; the failure is in `list_commands`).

### Paseo version

0.1.99 (`@getpaseo/server` 0.1.99)

### OS version

macOS 15 (Darwin 25.5.0), arm64

### Agent provider

omp (also affects `pi` — same `PiCliRuntime` code path)

### Provider configuration

```json
"omp": { "enabled": true, "command": ["/Users/<user>/.bun/bin/omp"] }
```
OMP version: 16.1.16

BrianHotopp:
The Pi CLI runtime used `get_commands` which OMP renamed to `get_available_commands`. This caused slash commands and skills to fail loading with OMP. OMP returns an error without an `id` field for unknown commands, which caused Paseo to time out after 30s instead of failing fast.

Fixes #1692

### DIFF
--- a/packages/server/src/server/agent/providers/pi/agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.test.ts
@@ -1398,7 +1398,7 @@ describe("PiRpcAgentClient", () => {
     ]);
   });
 
-  test("lists Pi compact even when RPC get_commands omits built-in slash commands", async () => {
+  test("lists Pi compact even when RPC get_available_commands omits built-in slash commands", async () => {
     const { pi, session } = await createSession();
     pi.latestSession().commands = [
       { name: "review", description: "Review changes", source: "extension" },
@@ -1418,7 +1418,7 @@ describe("PiRpcAgentClient", () => {
     });
   });
 
-  test("preserves known argument hints when RPC get_commands returns built-in slash commands", async () => {
+  test("preserves known argument hints when RPC get_available_commands returns built-in slash commands", async () => {
     const { pi, session } = await createSession();
     pi.latestSession().commands = [
       { name: "compact", description: "Compact from RPC", source: "extension" },
--- a/packages/server/src/server/agent/providers/pi/agent.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.ts
@@ -1208,7 +1208,7 @@ function createRuntime(logger: Logger, runtimeSettings?: ProviderRuntimeSettings
     logger,
     runtimeSettings,
     command: [PI_BINARY_COMMAND],
-    commandsRpcName: "get_commands",
+    commandsRpcName: "get_available_commands",
   });
 }
 
--- a/packages/server/src/server/agent/providers/pi/cli-runtime.test.ts
+++ b/packages/server/src/server/agent/providers/pi/cli-runtime.test.ts
@@ -242,7 +242,7 @@ describe("PiCliRuntime", () => {
     expect(events).toEqual([{ type: "turn_start" }]);
   });
 
-  test("lists commands through the default Pi get_commands RPC", async () => {
+  test("lists commands through the default Pi get_available_commands RPC", async () => {
     const child = createPiChild();
     const commandTypes: string[] = [];
     replyToCommands(child, (command) => {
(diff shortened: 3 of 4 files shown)


boudra:
The OMP command-discovery bug was fixed in [#1698](https://github.com/getpaseo/paseo/pull/1698), which is merged. Closing this PR.

PR #2556, opened 2026-07-28
Build Paseo Desktop from the Nix flake on macOS

### LINKED ISSUE #2555: Expose the desktop app from the Nix flake on macOS
## Problem

The repository flake exposes `.#desktop` only on Linux even though nixpkgs provides Electron on Darwin. Nix-darwin users therefore cannot install a current Paseo Desktop build from the same pinned source as the daemon, and may fall back to an older published cask.

## Proposed flow

Expose `packages.<darwin-system>.desktop` and produce a native `Applications/Paseo.app` bundle backed by the nixpkgs Electron runtime and the checkout-built Paseo main process, renderer, client, and daemon. Keep the existing Linux package behavior unchanged.

Expected usage:

```console
nix build .#desktop
```

On macOS, the result should contain `Applications/Paseo.app` and `bin/paseo-desktop`, with the Paseo bundle ID, icon, and URL scheme registered in `Info.plist`.

colonelpanic8:
Closes #2555.

## Problem

The flake exposed the Desktop package only on Linux, so nix-darwin configurations could not install a current Desktop build from the same pinned Paseo source as the daemon.

## Change

- expose `packages.<system>.desktop` on Darwin
- use electron-builder to produce the native macOS bundle layout, helper names, bundle identifiers, resources, and URL scheme
- source Electron from nixpkgs through electron-builder's custom distribution support instead of downloading it
- stage a writable copy of the read-only Nix Electron app so electron-builder can update the helper metadata
- skip code signing and hardened runtime for the local Nix output
- retain the existing Linux launcher and desktop entry behavior
- document `nix build .#desktop` on Linux and macOS

## QA

- `nix flake show --all-systems`
- `nix build .#desktop --no-link` on x86_64-linux
- built PR head `b12e3b63b` on aarch64-darwin with `nix build`
- `plutil -lint` passed for the main app and helper app plists
- verified bundle version `0.2.3` and bundle ID `sh.paseo.desktop`
- launched the exact Nix store output on Apple Silicon and verified the main, GPU, network, and renderer processes remained alive
- verified no new macOS crash report was created
- verified Paseo Desktop remained client-only (`manageBuiltInDaemon=false`) and the separately managed launchd daemon retained the same PID through app startup
- `npm run format`
- `npm run lint`
- `npm run build:server`

`npm run typecheck` was also run. Every workspace except `@getpaseo/expo-two-way-audio` passed; that workspace currently fails on the branch baseline because `expo-module-scripts` is missing and React Native/DOM globals conflict. This change does not touch that workspace.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -104,3 +104,27 @@ jobs:
 
       - name: Build Nix desktop package
         run: nix build .#desktop -o result-desktop
+
+  build-desktop-darwin:
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+
+      - name: Build Nix desktop package
+        run: nix build .#desktop -o result-desktop
+
+      - name: Verify macOS application bundle
+        run: |
+          set -euo pipefail
+
+          app="result-desktop/Applications/Paseo.app"
+          test -x "$app/Contents/MacOS/Paseo"
+          test -L result-desktop/bin/paseo-desktop
+          test "$(/usr/libexec/PlistBuddy -c 'Print :CFBundleIdentifier' "$app/Contents/Info.plist")" = "sh.paseo.desktop"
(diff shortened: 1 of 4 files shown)


boudra (approved):
Rebased onto the traced Linux runtime and verified by both Ubuntu and macOS Nix builds. The added macOS job validates the executable, launcher, bundle identifier, and embedded plists.

PR #2558, opened 2026-07-28
feat(workspaces): make worktree setup launch-safe

panta82:
---

Hey, I've forked Paseo to add a few features I need for my own setup and fix some bugs that bothered me.

I did adversarial reviews with different agents, but I'm not deep enough in the codebase to do the final human review or judge whether this is a good fit for the project.

I'm keeping this PR open in case there is interest in the change. It matters in my setup because worktree setup can include dependency installation, database creation, and other commands that take several minutes. Starting an agent before those commands finish creates a mess.

The rest of this description is AI-generated.

---

## Summary

This makes worktree setup a readiness boundary that agent work cannot race.

For a new branch-off workspace with an initial agent, the daemon now generates the title and branch before it creates the worktree. It uses the generated branch for the worktree directory too, and applies collision suffixes to both names together. Setup receives the final `PASEO_BRANCH_NAME` and path from its first command. If naming fails, creation stops before the worktree or setup process exists.

The rest of the change:

- persists setup state and restores interrupted setup as a visible failure after daemon restart;
- coordinates workspaces that share one physical worktree and sends progress to each workspace;
- returns the workspace and idle agent while setup continues, reveals the live Setup tab, keeps the queued prompt visible, and marks setup work as active in the sidebar;
- waits for successful setup before starting session or MCP agent turns and configured terminals;
- removes the client's fixed 60-second creation deadline and deduplicates create requests by client message ID;
- retains final setup state, throttles progress persistence, and cancels the process tree when the last workspace using a worktree is archived;
- aborts and awaits active setup during daemon shutdown;
- cleans up MCP-created worktrees whose setup fails;
- avoids seeding a source-only committed `paseo.json` into a worktree based on a ref that does not contain it.

## Why

Setup can legitimately take longer than a minute. A client timeout does not stop the daemon process, and launching an agent into a half-built checkout turns missing dependencies or databases into misleading agent failures.

The old naming order added another race. Paseo created an animal-named branch, started setup, then asked a model to rename the branch. Long-lived setup shells and environment variables kept the placeholder even when Git changed underneath them. Generating the name first removes that split identity.

## Verification

- Focused server and client tests: 354 passed, 1 platform skip.
- Focused app tests: 69 passed.
- Playwright browser test `reveals setup while the initial agent waits for workspace readiness`: passed.
- `npm run build:server`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

All checks ran on the isolated one-commit contribution at `5ece226ea`, based on `upstream/main` at `0dad1d7a6`.


panta82:
Rebased onto current upstream and addressed the remaining review findings in `600ae9cfe`.

- Daemon shutdown now aborts every distinct active worktree setup and awaits all setup completions before continuing teardown, so setup subprocesses cannot outlive the daemon.
- Added a regression covering the daemon-wide cancellation barrier.
- Corrected the test snapshot to use the complete wire-valid `worktree_setup` detail shape.
- Changed the Git object spec to the portable `HEAD:paseo.json` form; the real POSIX worktree suite passes.

Verification on the isolated one-commit branch: 213 focused tests passed, 1 platform skip, plus full typecheck, lint, and format check.

panta82:
Updated the single commit and rebased it onto current upstream/main (3b721dc1a). This revision also fixes legacy git.createWorktree launches creating an orphan workspace for the source checkout before the intended worktree workspace. Explicit agent titles are now applied during initial worktree creation, and the legacy-created worktree is available to placement and failed-create cleanup.

Additional verification on the rebased branch:
- npm run build:server
- npm run typecheck
- npm run lint
- npm run format:check
- npx vitest run packages/server/src/server/worktree-session.test.ts --bail=1 (28 passed)
- npx vitest run packages/server/src/server/daemon-e2e/git-operations.e2e.test.ts -t "runs paseo.json setup asynchronously and reports workspace status" --bail=1 (1 passed, 10 skipped by filter)

PR #2578, opened 2026-07-29
feat(app): show file-type icons in the Changes flat list

cleiter:
### Linked issue

Discussed in discord. This is the leftover half of #1918, which added the file-type icons
to the Changes list but wired them into tree mode only.

### Type of change

- [ ] Bug fix
- [X] New feature (with prior issue + design alignment)
- [ ] Docs

### What does this PR do

Renders the file-type icon on flat-list rows, the same icon tree mode has
had since #1918.

The icon was behind a `showDir ? null : ...` branch — `showDir` is true in
flat mode and false in tree mode, so the icon only ever appeared in the
tree. Dropping the branch renders it in both. `showDir` now controls only
the muted inline directory suffix, which is what its name says.

Nothing else about the flat list changes: the directory suffix stays inline
on the row, rows stay unindented, and the item model is untouched. Tree mode
is unchanged — it already rendered this icon. Client-only, no protocol or
server change, and the icon comes from the existing `getFileIconSvg`.

Three insertions, five deletions, one file.

### How did you verify it

- `npm run typecheck`, `npm run lint`, `npm run format` clean.
- Live in Desktop App against a real diff: icons
  resolve per file type in the flat list, the directory suffix and `+/-`
  stats are unchanged, and toggling to tree and back leaves tree mode as it
  was.
- No test changes: the change is one JSX branch in the header row, and the
  `src/git/` unit tests cover the pure item-model modules, which this does
  not touch. The `diff-row-alignment` e2e helper asserts the flat row
  contains both the filename and its directory — still true.

Before:

image

After:

image

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/app/src/git/diff-pane.tsx
+++ b/packages/app/src/git/diff-pane.tsx
@@ -1195,11 +1195,9 @@ const DiffFileHeader = memo(function DiffFileHeader({
         style={showDir ? styles.fileHeaderLeft : [styles.fileHeaderLeft, styles.fileHeaderLeftTree]}
       >
         {showDir ? null : <View style={styles.fileChevronSpacer} />}
-        {showDir ? null : (
-          <View style={styles.fileIcon}>
-            <MaterialFileIcon fileName={fileName} size={WORKSPACE_TREE_ICON_SIZE} />
-          </View>
-        )}
+        <View style={styles.fileIcon}>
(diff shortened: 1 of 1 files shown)


cleiter:
Superseded by sidebar layout changes in 0.5/0.6.

PR #2588, opened 2026-07-29
Align inline code chips with the prose line on native

### LINKED ISSUE #2580: Inline code background looks misaligned and uneven on mobile
On native (Android in the screenshot, same mechanism on iOS), inline `code` spans inside assistant markdown render with a background chip that floats above the surrounding text baseline, and wrapped lines containing inline code get uneven vertical spacing.

**Cause**

On native a markdown paragraph is a `flexDirection: "row"` wrapping `<View>` whose children are sibling `<Text>` spans (`packages/app/src/components/markdown-text.android.tsx`, `.ios.tsx`). An inline code span is therefore a *flex item*, not an inline run inside a shared line box. `code_inline` in `packages/app/src/styles/markdown-styles.ts` sets `paddingVertical: 2` and `lineHeight: round(fontSize.code * 1.45)`, both of which make that flex item taller than its plain-text neighbours. With the paragraph's `alignItems: "flex-start"`, the taller item is top-aligned, so the background box sits high relative to the text baseline and the row it lands on grows.

**Fix**

Drop the vertical padding on native and match the inline-code `lineHeight` to the prose line height (`round(fontSize.base * 1.4)`) so code chips are the same height as the text spans they sit beside. Web keeps the padded chip, where inline code really is an inline run.

colonelpanic8:
### Linked issue

Closes #2580

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

On mobile, inline `code` spans render with their background chip floating above the surrounding text baseline, and wrapped lines containing inline code get uneven vertical spacing. See #2580 for the screenshot.

On native a markdown paragraph is a `flexDirection: "row"` wrapping `<View>` whose children are sibling `<Text>` spans (`markdown-text.android.tsx` / `.ios.tsx`), so an inline code span is a *flex item*, not an inline run inside a shared line box. `code_inline` set `paddingVertical: 2` and `lineHeight: round(fontSize.code * 1.45)`, both of which make that flex item taller than its plain-text neighbours. With the paragraph's `alignItems: "flex-start"` the taller item is top-aligned, so the chip background sits high relative to the text baseline and the row it lands on grows.

The fix drops the vertical padding on native and matches the inline-code line height to the prose line height (`round(fontSize.base * 1.4)`), so every span in the row is the same height. Web is unchanged — there inline code really is an inline run, so the padded chip stays.

### How did you verify it

Honest accounting of what was and wasn't checked:

- `npm run typecheck` — passes (all workspaces)
- `npm run lint` — passes, 0 warnings / 0 errors
- `npm run format` — ran (Biome), no changes
- `packages/app/src/styles/markdown-styles.test.ts` — passes
- Added `packages/app/src/styles/markdown-styles-native.test.ts` — the existing test file evaluates with `isWeb` true, so the native branch needs its own file with `vi.mock("@/constants/platform")` in place before the module is imported (the flag is read once at module scope). It asserts inline code gets `paddingVertical: 0` and the prose line height on native.
- **No "after" screenshots yet.** The bug was reported and reproduced on Android (screenshot in #2580); I have not captured post-fix visuals on Android, iOS, or web. The diagnosis is from reading the native paragraph layout, not from watching the pixels change. Happy to hold this until "after" shots are attached if you want the visual proof before review — flagging it rather than implying testing that did not happen.

### Checklist

- [x] `npm run lint` passes
- [ ] UI changes include screenshots or video for every affected platform — **not done, see above**

### DIFF
--- a/packages/app/src/styles/markdown-styles-native.test.ts
+++ b/packages/app/src/styles/markdown-styles-native.test.ts
@@ -24,7 +24,7 @@ describe("createMarkdownStyles on native", () => {
     });
     expect(createCompactMarkdownStyles(darkTheme).code_inline).toMatchObject({
       paddingVertical: 0,
-      lineHeight: proseLineHeight,
+      lineHeight: 20,
     });
   });
 });
--- a/packages/app/src/styles/markdown-styles-native.test.ts
+++ b/packages/app/src/styles/markdown-styles-native.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it, vi } from "vitest";
+import { createCompactMarkdownStyles, createMarkdownStyles } from "./markdown-styles";
+import { darkTheme } from "./theme";
+
+// markdown-styles reads `isWeb` once at module scope, so the native branch needs
+// its own file with the mock in place before the import is evaluated.
+vi.mock("@/constants/platform", () => ({
+  isWeb: false,
+  isNative: true,
+}));
+
+describe("createMarkdownStyles on native", () => {
+  it("keeps inline code the same height as the prose line it sits in", () => {
+    // Native paragraphs are wrapping flex rows of sibling <Text> spans, so a
+    // taller inline-code item is top-aligned rather than baseline-aligned: the
+    // chip background floats above the surrounding text and the wrapped row
+    // grows. Matching the prose line height (and dropping the vertical padding)
+    // keeps every span in the row the same height.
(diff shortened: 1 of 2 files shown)


boudra:
fixed in https://github.com/getpaseo/paseo/commit/6ae6231d1fa745061808bf4873819d057aea5123 and will ship as 0.3.1 reopen if fix not right

Issue #1840, opened 2026-07-01
Feature request: Pin conversations (agents) to keep them at the top of the sidebar


EatSweetJellyfish:
## Problem

In the sidebar, conversations (agents) are ordered by recent activity. Important or frequently-used conversations keep getting pushed down as newer activity bumps other items to the top, and there is no way to keep a specific one reliably in view.

## Proposal

Add a **Pin** action for conversations (and ideally workspaces) so pinned items stay at the top of the list regardless of last-activity time.

## Suggested behavior

- "Pin" / "Unpin" entry in the right-click context menu
- Pinned items shown in a group above the activity-sorted list
- Pin state persists across app restarts

## Current workaround

Using labels + the filter control to resurface a conversation, but it does not actually keep it pinned to the top.

## Environment

- Paseo desktop app (macOS)

leibole:
This would be great, pinning should work the way it does in the codex app

Issue #356, opened 2026-04-14
Feature: Group sessions by project folder in Sessions view


cjh-store:
## Problem

Currently the Sessions screen displays all agents in a **flat list sorted by creation date**. When working across multiple projects, it quickly becomes hard to find the session you need — you end up scrolling through dozens of unrelated sessions mixed together.

The sidebar already groups agents by project (via `deriveProjectKey` / `deriveRemoteProjectKey`), but the Sessions screen and mobile view don't leverage this grouping at all.

## Proposed Solution

Add a **folder/project grouping mode** to the Sessions view, so sessions are organized by their working directory (project). Specifically:

1. **Group by project folder** — Sessions sharing the same `cwd` or `projectKey` should be visually grouped together (collapsible sections, similar to the sidebar's active project groups).
2. **Toggle between views** — Allow users to switch between the current flat chronological list and the folder-grouped view (e.g., a toggle button or dropdown: "By Date" / "By Project").
3. **Mobile support** — The `sessions-screen.tsx` on mobile should also support this grouping, not just the desktop sidebar.

## Context

The codebase already has most of the building blocks:
- `agent-grouping.ts` → `deriveProjectKey(cwd)` and `groupAgents()` for project-based grouping
- `use-sidebar-agent-sections.ts` → sidebar sections already group by project
- Agent data model includes `cwd` and `projectPlacement` with `projectKey`/`projectName`

The main gap is that `sessions-screen.tsx` renders a flat `AgentList` sorted by `createdAt` without using any of this grouping infrastructure.

## Use Case

As a developer working on 5+ projects simultaneously with Paseo, I want to quickly see "all sessions for project X" without scanning through a long mixed list. This is especially painful on mobile where the sidebar grouping isn't available.

Issue #287, opened 2026-04-12
Feature req: mini collapse mode in sidebar


ryanmargono:
Compact view to be able to see progress in different projects wihtout having it fully expanded like in Superset

PR #2597, opened 2026-07-29
Expandable agent/terminal tree in the sidebar

### LINKED ISSUE #730: Request: Ability to minimize tabs like run or other background scripts
For my workflow I run different background scripts in different terminals and I would like the ability to group or minimize them so it's easier to navigate if needed.

bendavid:
### Linked issue

Closes #

No prior issue — flagging that up front given CONTRIBUTING. Adjacent open issues this touches on rather than closes: #730 (grouping/minimizing background terminals), #287 (mini collapse mode), #1840 (pinning agents), #356 (collapsible project grouping). Happy to scope this down or split it if you'd rather see it land in pieces.

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Makes each sidebar workspace row expandable into the agents and terminals it contains, so you can reach a specific agent or terminal without opening the workspace first. Subagents nest under their parent to arbitrary depth.

The rows render through the tabs' own `WorkspaceTabIcon`, so an agent looks the same in the sidebar as in its tab: provider icon, status dot overlaid on the icon corner, or the synced spinner in the icon's place while running. That reuse is what keeps the two surfaces from drifting, and it's also what makes the layout work — because status rides on the icon there is no separate status column, so agent and terminal rows share one icon/label column.

Three alignment rules are enforced rather than incidental, in `tree-layout.ts`:

- every row reserves the chevron column whether or not it can expand, so a row never shifts sideways when it gains its first child;
- a workspace's direct children — top-level agents and terminals alike — all render at `TREE_ROOT_DEPTH`, from one shared constant;
- indentation is a pure function of depth, clamped at 4 levels so deep subagent chains keep a readable label.

The active agent or terminal takes the workspace row's own selected treatment (same `surfaceSidebarHover` token), resolved from the focused tab of the active workspace's layout. A row's tab target is built once and used for *both* navigation and the active-row check, so a row can't open one tab while highlighting another.

Cost control: the terminals list is the only thing here that needs an RPC, and it's fetched only when a workspace is actually expanded — collapsed workspaces issue no `listTerminals` and hold no push subscription. Agents come from the session store, which is already snapshot-then-subscribe. Provider subagents that existed before the app connected are hydrated on expand, once per agent per connection.

### How did you verify it

Driven in a real browser (Playwright against a local dev daemon, Chromium), plus unit tests. Screenshots below are from those runs.

**Active row highlighting** — cursor parked off-sidebar so hover couldn't be mistaken for selection. Clicking the agent highlights it (`aria-selected=true`); clicking a terminal moves the highlight and flips the agent to `false`. The workspace row stays highlighted throughout as the active workspace.

| Agent active | Terminal active |
| --- | --- |
| Agent row highlighted | Terminal row highlighted |

**Running spinner** — a real running agent (the repo's built-in `mock` load-test provider, so no model spend) and a terminal driven to `working` through the real hook path (`paseo hooks claude UserPromptSubmit` from inside the terminal, `HOOK_RC=0`). Sampled the DOM 400ms apart: both rows have zero `<svg>` and six dot views whose opacities cycle `[0,0,1,.34,.78,.56] → [.56,.78,.34,1,0,0]`, i.e. the icon is genuinely replaced by an animating loader. Idle rows keep their glyph.

Running agent and terminal

**Provider subagent hydration** — A/B against real data (an existing session with 5 real Claude subagents on disk). With hydration disabled the parent agent gets no chevron and zero subagent rows; with it, the chevron appears and expanding shows the 5 real rows at depth 1.

| Before (no hydration) | After |
| --- | --- |
| No subagents | Subagents listed |

**Alignment**, measured from the DOM rather than eyeballed. With a pinned workspace present, every workspace chevron sits at left 8 / icon 11; inside the tree an expandable agent, a non-expandable agent, and terminals all land on icon 51 / title 76; depth 1 is 60/63/88 and depth 2 is 72/75/100 — one 12px step per level.

**Request behaviour** — confirmed in the daemon log that expanding a workspace emits one `agent.provider_subagents.list.request` per Paseo agent from that origin, and that collapsed workspaces emit none.

**Unit tests**: 44 across tree construction (nesting, ordering, cross-workspace subagents, parentage cycles), the layout invariants, and the row presentation/status mapping.

**Known limitation, pre-existing and not addressed here:** OMP subagents don't appear. An OMP agent with 26 real subagent transcripts on disk returns an empty `listProviderSubagents`, both on expand and after opening the agent — the daemon's in-memory registry isn't populated from OMP history the way Claude's sidechain tracker is. This looks like the same root cause as #2232. Claude subagents work.

Not verified: native iOS/Android (Linux-only checkout here), and drag-reorder re-measurement when an expanded row changes height inside the draggable project list.

### Checklist

- [x] `npm run lint` passes


bendavid:
Thanks — the cross-workspace navigation issue was real, and fixed in a5bc5143c.

The tree pulls subagents in by parentage regardless of workspace, so a row can be displayed under one workspace while belonging to another; rows were passing the containing workspace to `navigateToWorkspace` unconditionally, which put the tab in the wrong layout. `resolveRowWorkspaceId` now gives the agent's own workspace precedence and falls back to the containing one only when the agent has none — the same precedence #2462 establishes for the subagent navigation resolver. Provider subagents carry no workspace of their own and stay scoped to their parent.

The active-row highlight had the mirror-image bug, so `useActiveTreeTab` now returns the focused tab together with the workspace it lives in and rows match on both. Without that, a cross-workspace subagent would open in its own workspace and then never highlight; it also stops a row matching a colliding tab id from another workspace.

Covered by three new cases in `row-presentation.test.ts` (own-workspace precedence, blank/whitespace fallback, provider-subagent scoping).

Also rebased onto current `main` and fixed the `SessionState` test fixture for `messageSubmissions`, which is what the typecheck job was failing on.

boudra:

Issue #94, opened 2026-03-07
Multiple drafts created when opening a new workspace with no agents


boudra:
## Bug

When creating a new workspace that has no agents, multiple draft agents are created instead of just one.

## Expected behavior

A single draft agent should be created when opening a workspace with no existing agents.

## Steps to reproduce

1. Create/open a new workspace with no agents
2. Observe that multiple drafts appear

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2614, opened 2026-07-29
fix(app): prevent duplicate draft agent creation

lincolnaleixo:
## What changed

- Treat an already-active create attempt in the shared create-flow store as an in-flight submission.
- Reject a concurrent submit even before React has rendered the reducer's `creating` state.
- Add a regression test that invokes the submit handler twice in the same render cycle and verifies that only one create request starts.

## Root cause

`handleCreateFromInput` guarded on `isSubmitting`, which is derived from the reducer state captured by the current render. Multiple submit callbacks in the same event loop all observed `isSubmitting === false`, generated independent message IDs, and started independent `createAgent` requests before React could render the first `SUBMIT` transition.

The create-flow Zustand store is updated synchronously before the request starts, so checking its active attempt closes that render gap while preserving retries after failed attempts are cleared.

## Observed impact

In a real Paseo 0.2.3 desktop session, one new-workspace prompt produced seven `create_agent_request` messages from the same client, session, and connection within one millisecond. All seven became real Codex agents.

This may be related to #94, although that issue describes duplicate draft tabs rather than confirmed duplicate provider agents.

## Validation

- `npm run build:app-deps`
- `npm run test --workspace=@getpaseo/app -- src/composer/draft/create-flow.test.ts`

All passed locally.


boudra:
Superseded by #4442, which takes over the duplicate draft-agent workflow with the shared active-attempt check, stable per-draft creation/message identities, and durable daemon receipts. The canonical PR includes failing-first browser coverage for rapid submissions and replayed requests.

PR #2616, opened 2026-07-29
fix(app): reconnect stale host connections on foreground

### LINKED ISSUE #2615: Add Project directory autocomplete fails silently after hours idle (zombie WebSocket)
## Summary

After the app sits idle for hours (iOS suspends it), the Add Project → Browse directory autocomplete doesn't populate. Sending a chat message to any agent on that host "fixes" it — the next attempt populates.

## Root cause (from code)

The WebSocket becomes a zombie while the app is suspended: dead server-side, but the client never saw a close frame and its liveness heartbeat was frozen, so it still reports `connected`. There is no `AppState → active` hook forcing an immediate liveness check, so after waking there's a ~10–30s window where the client trusts the dead socket (`packages/client/src/daemon-client.ts:959-961` — 10s interval, 2 failures before reconnect).

Opening Add Project in that window fires `getDirectorySuggestions` into the dead socket. The request is `skipQueue: true` (`daemon-client.ts:1743`), so it hangs until the heartbeat tears the socket down or the 60s RPC timeout hits, then fails — and the query is `retry: false` with a 15s staleTime (`packages/app/src/components/add-project-flow.tsx:395-411`), so the failure sticks. The flow shows empty results.

Chatting "fixes" it because the heartbeat (or the chat write itself) tears down the zombie and reconnects in the meantime; the next query goes over the fresh socket.

## Repro (no iOS device needed)

Any platform — the iOS suspension is just the trigger:

1. Run the app (web works) against a daemon.
2. `kill -STOP <daemon-pid>` — the socket stays half-open, writes vanish, the client still thinks it's connected. This simulates the post-suspend zombie exactly.
3. Open Add Project → Browse, type in the directory search → nothing populates.
4. `kill -CONT <daemon-pid>`, wait for reconnect (or send a chat message) → autocomplete works again.

## Suggested fixes (any of)

- Force a liveness check / reconnect on `AppState → active` (and/or when opening Add Project against a stale connection).
- Give the directory suggestions query retries so a zombie-window failure self-heals.
- Gate the query on connection status and show "Reconnecting…" instead of firing into a dead socket.

## Environment

- iOS app, but platform-independent mechanism
- Reported from Linux; no iOS device available for verification

BrianHotopp:
## Problem

After the app sits idle for hours (iOS suspends it), the Add Project directory autocomplete doesn't populate (#2615). The WebSocket dies while the app is suspended, but the client never sees a close frame and its liveness heartbeat is frozen — so it still reports `connected`. Requests fired in that window hang against the dead socket and fail without retry. Unrelated traffic (a chat send) tears the zombie down, which is why the feature "starts working again" afterward.

The mechanism is platform-independent: any OS suspension of the JS runtime produces the same zombie.

## Fix

- **`DaemonClient.reconnectIfStale()`** (client): one bounded liveness probe when `connected` — on failure it reconnects immediately, bypassing the steady-state failure threshold (which exists for flap avoidance, the wrong tool for a one-shot staleness decision). When `disconnected`, it skips any pending reconnect backoff and connects now. Healthy connections cost one ping, no churn.
- **`HostRuntimeStore.reconnectStaleConnections()`** (app): pokes every live client; wired to `AppState → active` in the root layout.
- **Add Project directory query**: `retry: 2` so a first attempt that races connection recovery self-heals.

## Tests

- Client: healthy probe (no reconnect), failed probe → immediate reconnect, backoff skip, post-close no-op.
- Store: `reconnectStaleConnections` reaches every live client.

## Verification

Reproduced the zombie against a dev daemon frozen with `SIGSTOP` (socket half-open, writes vanish — the same state iOS suspension leaves): the directory query stalls on "Loading…", and after `SIGCONT` the connection re-establishes and autocomplete populates. Client + app test suites green (109 + 60), typecheck/lint/format clean.

The AppState → active path is native; verified by unit tests (web's visibility-driven suspension doesn't fully freeze timers the same way).

Fixes #2615

### DIFF
--- a/packages/app/src/app/_layout.tsx
+++ b/packages/app/src/app/_layout.tsx
@@ -92,6 +92,7 @@ import { queryClient } from "@/data/query-client";
 import {
   getHostRuntimeStore,
   hasConfiguredLocalDaemonOverride,
+  reconnectStaleHostConnections,
   useHostRegistryLoaded,
   useHostMutations,
   useHostRuntimeClient,
@@ -977,7 +978,12 @@ export default function RootLayout() {
     const subscription = AppState.addEventListener("change", (nextState) => {
       if (nextState !== "active") {
         void flushDraftPersistStorage();
+        return;
       }
+      // OS suspension freezes client heartbeats and can kill sockets without
+      // a close frame; reconnect anything stale now rather than waiting out
+      // the heartbeat backoff.
+      reconnectStaleHostConnections();
     });
     return () => subscription.remove();
   }, []);
--- a/packages/app/src/components/add-project-flow.tsx
+++ b/packages/app/src/components/add-project-flow.tsx
@@ -412,7 +412,9 @@ export function AddProjectFlow({ request, onClose }: AddProjectFlowProps) {
     },
     enabled: Boolean(client && searchesDirectories),
     dataShape: "value",
-    retry: false,
+    // Retry: the first attempt can race connection recovery (e.g. right
+    // after the app regains foreground) and fail into an empty list.
+    retry: 2,
     staleTimeMs: 15_000,
   });
   const githubQuery = useFetchQuery({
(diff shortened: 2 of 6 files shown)


boudra:
Closing for now: this needs before/after evidence that returning from iOS suspension restores Add Project directory results when the socket still reports connected, without sending a chat message. Please reopen with that evidence.

boudra:

PR #2618, opened 2026-07-29
fix(server): append standard ports (80, 443) to connection hint

shasderias:
### Linked issue

Not sure if a Discord thread qualifies as a issue, but happy to open an issue if required.

https://discord.com/channels/1481169421832814616/1529502102433566770/1529502102433566770

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

The connection hints generated by `web-ui.ts` don't contain a port number if request was made over a standard port (80, 443). This requires the user to manually add a host with the correct connection string. This should ideally be seamless.

This PR fixes the above issue by appending the port number if it is appropriate to do so.

This PR also makes the connection hint resolver respect `X-Forwarded-Host` if trusted proxy = true. This was already being done for `X-Forwarded-Proto` when we use `req.protocol` (c.f. [Express.js docs](https://expressjs.com/en/4x/api/request/#reqprotocol)).

Re. lifting of implementations from Express v5 and Go: the implementation can probably be simplified into one function, but making a mistake here probably breaks connections for a non-trivial number of users. There are probably edge cases handled in the code that I lifted that I would miss if I were to write them myself.

#### IPV4 Testing

- rebuilt, accessed Paseo using the following links
  - https://paseo-throwaway.exe.xyz:8000 - continued to work
  - https://paseo-throwaway.exe.xyz - was failing before this fix

#### IPv6 Testing

- tested over loopback with the following script that spun up a paseo instance, prodded it with a set of headers to see what the connection hint is, then attempted a connection with the connection hint
  - [test-paseo-ipv6-connection-hint.sh](https://github.com/user-attachments/files/30514147/test-paseo-ipv6-connection-hint.sh)
  - [ipv6-test-output.txt](https://github.com/user-attachments/files/30514148/ipv6-test-output.txt)

### Checklist

- [x] `npm run lint` passes

### DIFF
--- a/packages/server/src/server/bootstrap-web-ui.test.ts
+++ b/packages/server/src/server/bootstrap-web-ui.test.ts
@@ -97,6 +97,24 @@ describe("daemon web UI bootstrap", () => {
         headers: { "x-forwarded-proto": "https" },
       }),
     );
+    const noExplicitPortHttpsHint = readInjectedConnectionHint(
+      await fetchDaemonWebUi({
+        port: daemonHandle.port,
+        headers: {
+          host: "daemon.example.test",
+          "x-forwarded-proto": "https",
+        },
+      }),
+    );
+    const forwardedHostHttpsHint = readInjectedConnectionHint(
+      await fetchDaemonWebUi({
+        port: daemonHandle.port,
+        headers: {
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the narrower connection-hint fix in open PR [#3832](https://github.com/getpaseo/paseo/pull/3832).

PR #2635, opened 2026-07-30
DRAFT - feat: Groups for Projects and Workspaces

cleiter:
⚠️ **Please do not review this line by line yet.** It is a draft, and it is here so it can be
run rather than read. The design is aligned (see below) but the implementation still has
open ends, listed at the bottom, and I expect more passes on them before this
is worth a real review. ⚠️

What I would find most useful right now: someone other than me driving it — especially on
mobile and other platforms than macOS.

### Linked issue

N/A - Discussed and aligned in Discord.

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Adds groups ("folders") to the sidebar, at both levels — projects can be grouped, and each
project's workspaces can be grouped — plus persistent ordering of groups and of their
members.

The interesting part is not the UI, it is where the state lives.

**One replicated layout document, not group attributes on records.** The first version of
this modelled a group as `groupId`/`groupName` fields on the project and workspace records,
the way `pinnedAt` works. That cannot represent two things the feature needs:

1. **An empty group cannot exist.** A group is only real while something points at it, so
   dragging the last workspace out silently destroys it, and creating a group before
   filling it is impossible.
2. **Groups cannot be ordered.** There is no group entity to carry an order, and order
   cannot be inferred from members without contradiction.

Promoting groups to their own records does not fix it either, and the reason is specific
to Paseo: **a sidebar project is a client-side merge across N daemons** (`hosts[]` is
plural — the same repo on two machines is one row). So a group is a *user-level* object no
single daemon owns. An empty group has no members, therefore no member hosts, therefore no
answer to "which daemons does this group live on". And `order: number` is unrepresentable:
two hosts holding different membership produce contradictory global orderings with no
merge rule to resolve them.

So instead: `$PASEO_HOME/projects/sidebar-layout.json`, one document, replicated whole.

```ts
{
  revision: number,          // bumped on write
  updatedAt: string,
  projectGroups:   [{ id, name, projectKeys: string[] }],
  workspaceGroups: [{ id, name, projectKey, workspaceKeys: string[] }],
  ungroupedProjectKeys: string[],
  ungroupedWorkspaceKeysByProject: Record<string, string[]>,
}
```

**Array order is the order.** No `order: number` anywhere, so there is nothing to renumber
and nothing to conflict. Writes carry `expectedRevision`; on mismatch the daemon rejects
and returns the authoritative document, and the client re-applies its intent against it.
The daemon stores and serves the document and never interprets it — the reconciler never
sees a group, so `workspace-reconciliation-service.ts` is untouched.

Keys that cannot be resolved (a group referencing a workspace on a host you are not
connected to) are **pruned at render, retained in storage**, which is what lets a
multi-daemon layout survive being viewed from a machine that only sees half of it.

Notable: this is *smaller* on the server than the attribute version it replaces. It rides
`useReplicaQuery` + `RECONNECT_REPAIR_POLICIES`, which already do snapshot + live push +
repair-on-reconnect, rather than adding another bespoke sync channel.

**Capability gate:** `server_info.features.sidebarLayout`, one `COMPAT(sidebarLayout)`
comment marking the cleanup site. No fallback path — an old daemon just doesn't get the
feature.

#### The one behavioural change

Group order and member order are the same document, deliberately. If group order synced
but member order did not, a fresh device would render *groups in the right order with
their contents scrambled*, which reads as a bug and is worse than either extreme.

The cost, stated plainly: **users who never create a group now get their drag order
persisted server-side** instead of per-device in AsyncStorage. Not a visual change, but a
behavioural one, and it is the only part of this that touches shipped functionality. This
is the piece that was explicitly called out and signed off in the Discord thread.

#### Drag and drop

Web only (`react-native-draggable-flatlist` cannot drag between lists at all, so on native
moving a row between groups goes through the row menu). The drag machinery is
level-agnostic — it speaks item/group/header, and both levels share it.

The decisions live in a pure, tested policy module rather than inside dnd-kit event
handlers, because every drag bug in this feature turned out to be a policy bug and none of
them were visible from inside the handlers. Six invariants that each cost a real bug are
written up in `docs/sidebar-drag.md`.

#### Also in here

- Long groups cap at the same 20-row window the rest of the sidebar uses, behind
  "show more". This needed the reorder writes to change first: a capped list hands a drag
  back only the *visible* rows, so writing that subset wholesale would have ejected
  everything below the cap out of the group. The writes now merge the visible reorder into
  the stored order in place.
- The ungrouped remainder is a first-class collapsible section at both levels, so it
  behaves like a group without being one (italic title, no glyph, no menu, no drag handle).

### How did you verify it

**Testing status: incomplete, and that is the main reason this is a draft.**

Automated — run locally, all green:

- `packages/app/src/sidebar/sidebar-layout-edits.test.ts` — the document edits, including
  the direction cases and the capped-group cases (the hidden tail survives a visible-row
  reorder at every level)
- `packages/app/src/components/sidebar/sidebar-group-drag-policy.test.ts` — one case per
  drag bug that shipped, each named for the symptom
- `packages/app/src/sidebar/sidebar-groups.test.ts`,
  `sidebar-projection.test.ts`, `use-sidebar-pins.test.ts`,
  `utils/sidebar-reorder.test.ts`, `stores/sidebar-collapsed-sections-store/state.test.ts`
- `packages/server/src/server/sidebar-layout-store.test.ts`,
  `wire-compat.test.ts`
- `npm run typecheck`, `npm run lint` (full repo), `npm run format`

The north-star assertion, if you only look at one test: **a user with no groups gets a
byte-identical sidebar view-model.** This must not change the sidebar for anyone who
doesn't opt in.

Manual — what I actually did:

- Dogfooded on macOS Desktop and Web against a real daemon: created,
  renamed, deleted, reordered groups at both levels; dragged rows within and between
  groups, including into an empty group and back out to the remainder; collapsed and
  reloaded; confirmed order and collapse state survived a daemon restart.
- Confirmed a sidebar with no groups is unchanged from before.
- Cross-device test: Group on desktop → open the web client → same sidebar.

Manual — what I have **not** done, and where I most want help:

- **No iOS or Android testing.** Drag is web-only so the drag paths are unreachable there,
  but the menu-driven move, the group headers, the collapse state, and the cap are all
  live on mobile and I have not exercised any of them on a device.
- **No two-daemon test.** The revision-conflict and unresolvable-key paths are the whole
  reason for this design and they are currently proven only by unit tests. The real check
  is: group across hosts, kill one daemon, rename + reorder, restart it, confirm it
  self-heals to the higher revision.

Regular view, also showing compat for project without workspace groups:

image

Ungrouped workspace:

image

Two projects in one group:

image

Ungrouped project:

image

Menu:

image

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran
- [ ] **UI changes include screenshots or video for every affected platform** — not yet;
      see above. Blocking on mobile testing.

### Known open ends before this is review-ready

1. Screenshots/video for iOS, Android, web and desktop.
2. Mobile pass — the feature is reachable there via menus and completely untested.
3. Two-daemon verification.
4. A **"New group"** affordance. Groups are currently created from the row menu
   ("Move to group → New group…"), which is discoverable enough to test with but is
   probably not the final entry point.
5. The `COMPAT(sidebarLayout)` marker names a version that has since been overtaken —
   needs correcting to whatever this would actually land in.
6. Whether the ungrouped remainder should show a header at all when a user has exactly one
   group. Right now it does, and it may read as clutter.


cleiter:
Superseded by #3206

PR #2645, opened 2026-07-30
fix(app): recalculate file explorer modified time on mount (#2567)

### LINKED ISSUE #2567: bug: file explorer shows a stale "modified X ago" in the file actions menu
## What's broken

The file actions menu in the file explorer has a header showing the file's size and when it was last modified. That "modified X ago" value is computed when the **row** renders, not when the **menu** opens, and it's frozen in a `useMemo` keyed only on `entry.modifiedAt` / `entry.size`.

So if the explorer pane has been open for a while, opening a file's actions menu shows a timestamp calculated back when the row first rendered. A file modified 40 minutes ago can report "just now" simply because that's what was true when the pane loaded.

## Steps to reproduce

1. Open a workspace and show the file explorer pane.
2. Leave the pane open, untouched, for several minutes.
3. Open the actions menu (⋯) on any file.
4. The "modified" value reflects the time the row first rendered, not the present.

## Where

`packages/app/src/components/file-explorer-pane.tsx` — the `metaHeader` `useMemo` (~line 138) calls `formatTimeAgo(new Date(entry.modifiedAt))`, and the resulting element is passed to `FileActionsMenu` as `header`.

## Notes

Pre-existing; not introduced by #2566. Calling it out because that PR audited every `formatTimeAgo` caller and deliberately left this one alone: the obvious fix — subscribing the row to the new `useMinuteTick` — would re-render every file row in the directory once a minute to keep a value that is invisible more than 99% of the time. Computing the label when the menu opens, rather than freezing it at row render, is the better shape.

## Related follow-up

Two surfaces still run their own hand-rolled tickers that duplicate `useMinuteTick`, each with a `setInterval` plus local state:

- `packages/app/src/git/commits-section/commits-section.tsx` (10s interval, gated on panel active + expanded)
- `packages/app/src/components/provider-diagnostic-sheet.tsx` (10s interval, gated on `visible`)

Both are correct today and were left alone in #2566 because the shared hook ticks while mounted and doesn't express their visibility gating. Worth consolidating if `useMinuteTick` grows a "only while visible" option — both also tick 6× more often than `formatTimeAgo`'s output can now change.

PTK030:
### Linked issue

Closes #2567

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Extracts \FileExplorerMetaHeader\ component in \ile-explorer-pane.tsx\ to recalculate \ormatTimeAgo\ dynamically on mount instead of capturing stale values in memoized closures. Adds unit test suite in \ile-explorer-pane.test.tsx\.

### How did you verify it

- Added unit tests in \packages/app/src/components/file-explorer-pane.test.tsx\ using fake timers to verify dynamic time-ago computation on mount/remount.
- Ran \
pm test --workspace=@getpaseo/app -- src/components/file-explorer-pane.test.tsx\ (passed).
- Ran \
px oxlint\ and \
px oxfmt\ across modified files.

### Checklist

- [x] \
pm run test\ passes
- [x] \
px oxlint\ passes
- [x] \
px oxfmt\ ran


boudra:
The file-actions metadata header was removed in merged [#3826](https://github.com/getpaseo/paseo/pull/3826). Closing this PR.

PR #2656, opened 2026-07-30
Use Ghostty for native mobile terminals

colonelpanic8:
## Summary

- adapt T3 Code's MIT-licensed native terminal module for Paseo
- use custom-I/O `GhosttyKit` on iOS and `libghostty-vt` plus a Canvas/JNI renderer on Android
- keep the existing xterm WebView renderer as an automatic fallback when the native module is unavailable
- preserve snapshot restore, input-mode tracking, hardware/software keyboard input, resize/focus handling, scrolling, Android selection/copy, and terminal panel swipe gestures
- discard renderer-generated terminal query replies because Paseo's daemon already owns those responses
- feed live output directly and retain only the latest snapshot payload, avoiding T3's unbounded replay-buffer growth

## Packaging and compatibility

- vendors pinned Ghostty artifacts, headers, rebuild scripts, licenses, and T3 Code attribution
- verifies all Android Ghostty/JNI libraries use 16 KB segment alignment
- raises the iOS deployment target from 15.1 to 16.1 to match the vendored `GhosttyKit` build
- excludes the binary-backed native module from F-Droid autolinking; F-Droid and Expo Go continue to use the WebView fallback

The checked-in native module adds approximately 55 MB to the source tree, mostly the two iOS static-library slices.

## Known follow-ups

- native rendering does not yet activate Paseo's local file links
- iOS Ghostty selection/clipboard support remains disabled by the upstream custom-I/O integration
- custom terminal font-family selection is not yet bridged to the native renderers

## Verification

- `npm run build:app-deps`
- `npm run build:server`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npx vitest run packages/app/src/terminal/runtime/terminal-snapshot.test.ts --bail=1`
- Expo autolinking discovery for Android and Apple
- Android `:getpaseo-mobile-terminal-native:assembleDebug` across arm64-v8a, armeabi-v7a, x86, and x86_64
- F-Droid Gradle project configuration without the native terminal module
- `readelf -lW` alignment check for vendored `libghostty-vt.so` and generated `libpaseoterminal.so` across all four ABIs

iOS prebuild and autolinking were verified on Linux; an Xcode compile/runtime pass still needs macOS CI or a device build.


boudra:
closing this in favour of https://github.com/getpaseo/paseo/pull/1607

that pr was dormant because it was low priority, i am picking it up now and the terminal is already way better than the webview one

this pr is too much to maintain, separate android / ios implementations, vendored stuff, custom tooling, different fallbacks per platform. it also doesnt implement selection which is something that 1607 already supports

the only benefit i see of ghostty is it's gpu rendering, but it's not a requirement for a mobile terminal, for ocassional commands and TUI

also, i do not want to adapt any third party code to paseo, because i will be inherting decisions and maintenance debt that we do not know what trade offs they made or how this was designed, any prs that copy implementations from other apps wholesale like this will be closed

thanks

PR #2669, opened 2026-07-30
fix(app): seed metadata endpoint fields after save

iamriajul:
## Summary
- Fixes blank Base URL / API key / model fields on the host settings metadata custom endpoint card after save.
- `FormTextInput` → `AdaptiveTextInput` intentionally ignores controlled `value` (RN flicker fix). The card now seeds with `initialValue` and remounts via a persisted-snapshot `resetKey`.
- Model chips also remount the model field so selection appears immediately.
- Unit coverage for reset-key identity and reading saved config.

Closes backlog task: Settings metadata endpoint form doesn't show any values after save.

## Test plan
- [x] `npx vitest run packages/app/src/screens/settings/metadata-custom-endpoint-config.test.ts --bail=1`
- [x] lint + typecheck via pre-commit
- [ ] Manual: Host settings → enable metadata custom endpoint → fill fields → Save → values remain visible
- [ ] Manual: leave and re-enter Host settings → saved values still show
- [ ] Manual: refresh models → pick a chip → model field updates


iamriajul:
Opened against upstream by mistake; this is fork-only work on iamriajul/paseo.

PR #2679, opened 2026-07-31
fix(acp): recover stale foreground turn on replace

### LINKED ISSUE #1936: Hermes ACP follow-up can fail with “A foreground turn is already active” and leave confusing agent state
## Problem

When a Hermes/ACP-backed agent is still processing a foreground turn, sending a follow-up prompt through Paseo can fail with:

```text
A foreground turn is already active
```

In the observed run, the failed follow-up left the agent in an error/closed path while a replacement agent was needed. This is hard to reason about because Paseo metadata can look stale even while daemon logs show provider stream activity.

## Why this matters

The user flow was: create a Hermes planning agent, then try to send a follow-up instruction before accepting the plan. The follow-up was not queued and failed with a system error. A user watching the UI cannot easily tell whether the agent is still doing background/provider work, whether Paseo is stale, or whether it is safe to send another prompt.

Expected behavior could be one of these, but it should be explicit and non-destructive:

- Queue the follow-up until the active foreground turn completes.
- Reject immediately with a clear “agent is busy, retry later” response without moving the agent into an error path.
- Expose a clear busy/foreground-turn-active state in agent metadata/UI so automation does not send a prompt into the active turn.

## Observed evidence

Daemon log stack from the failure:

```text
Error: A foreground turn is already active
    at ACPAgentSession.startTurn (.../server/agent/providers/acp-agent.js:874:19)
    at Object.startTurn (.../server/agent/provider-registry.js:192:47)
    at AgentManager.streamForwarder (.../server/agent/agent-manager.js:1153:52)
    at .../server/agent/agent-prompt.js:32:30
    at startAgentRun (.../server/agent/agent-prompt.js:50:7)
    at sendPromptToAgent (.../server/agent/agent-prompt.js:119:12)
    at .../server/agent/tools/paseo-tools.js:1106:9
```

Related structured log line:

```json
{
  "module": "agent",
  "component": "agent-manager",
  "provider": "hermes",
  "lifecycle": "error",
  "activeForegroundTurnId": null,
  "error": "A foreground turn is already active",
  "msg": "handleStreamEvent: turn_failed"
}
```

A nearby runtime metric also showed active stream activity for the replacement Hermes agent even though normal agent-list metadata did not make that progress obvious:

```json
{
  "outboundAgentStreamTypesTop": [["timeline:tool_call", 48]],
  "outboundAgentStreamAgentsTop": [["<replacement-agent-id>", 48]],
  "agents": { "byLifecycle": { "idle": 11, "error": 1, "running": 1 }, "withActiveForegroundTurn": 1 }
}
```

## Repro shape

1. Create a Hermes/ACP-backed agent in a workspace.
2. Start a long foreground turn.
3. Before that turn finishes, send a follow-up prompt to the same agent using Paseo tooling or UI.
4. Observe that the follow-up can fail with `A foreground turn is already active` instead of being queued or cleanly rejected as “busy.”
5. Observe that agent status/metadata may be confusing: the user may see stale `updatedAt`, error state, or no clear indication that provider-side work is still active.

## Suggested fix direction

The issue appears to be in the foreground-turn gate around `ACPAgentSession.startTurn` and the prompt path through `sendPromptToAgent`/`agent-prompt`. A minimal fix might be to detect an active foreground turn before invoking `startTurn`, then return a typed busy response to callers. A fuller fix would queue the follow-up or expose `foregroundTurnActive` in agent metadata so watchdogs and UI clients can avoid unsafe prompt injection.

## Environment notes

- Paseo desktop user agent observed in daemon logs: `Paseo/0.1.103` on macOS.
- Provider path involved: Hermes / ACP.
- This report is based on local daemon logs and live agent metadata from a real run.

## Existing issue search

Searched open and closed issues in `getpaseo/paseo` for:

- `"A foreground turn is already active"`
- `"foreground turn"`
- `ACPAgentSession startTurn hermes`
- `Hermes ACP foreground turn active activeForegroundTurnId`

No matching issues were found.

## AI disclosure

Prepared with AI assistance from Codex in a local agent harness, using the GitHub and Paseo connectors. The underlying evidence came from local Paseo daemon logs and live agent metadata observed during the run.

rinseaid:
## Problem

If a provider acknowledges cancellation but never emits a terminal turn event, Paseo force-cancels its manager-side run after the timeout. The provider session can still retain its foreground turn. The next replacement prompt then fails with `A foreground turn is already active`.

## Fix

After that forced cancellation path, reload the agent session before starting the replacement prompt. Reloading resumes the persisted provider session while retaining the Paseo agent, timeline, labels, and workspace. Normal cancellations keep their existing result shape and do not reload.

## Regression coverage

The test models a provider that accepts cancellation, omits the terminal event, and rejects another foreground turn. It verifies that Paseo closes the stale session, resumes a replacement session, and completes the queued replacement prompt.

## Validation

- `npx vitest run src/server/agent/agent-manager.test.ts`
- `npm run lint -- packages/server/src/server/agent/agent-manager.ts packages/server/src/server/agent/agent-manager.test.ts`
- `npm run format:check:files -- packages/server/src/server/agent/agent-manager.ts packages/server/src/server/agent/agent-manager.test.ts`

Fixes #1936

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -879,7 +879,7 @@ test("reload leaves a closed durable snapshot when shutdown starts during the sw
   }
 });
 
-test("reload closes both sessions when the closed snapshot cannot be persisted", async () => {
+test("reload retains the existing session when the closed snapshot cannot be persisted", async () => {
   const workdir = mkdtempSync(join(tmpdir(), "agent-manager-reload-persist-failure-test-"));
   const storagePath = join(workdir, "agents");
   const storage = new AgentStorage(storagePath, logger);
@@ -906,8 +906,6 @@ test("reload closes both sessions when the closed snapshot cannot be persisted",
     writeFileSync(storagePath, "blocks the storage directory");
 
     const reload = manager.reloadAgentSession(agentId).catch((error: unknown) => error);
-    await client.waitForCloseToStart();
-    client.finishClosing();
 
     expect(await reload).toBeInstanceOf(Error);
     await manager.flushForShutdown();
@@ -915,9 +913,9 @@ test("reload closes both sessions when the closed snapshot cannot be persisted",
       agents: manager.listAgents(),
       originalSessionClosed: client.originalSessionClosed,
       replacementSessionClosed: client.replacementSessionClosed,
-    }).toEqual({
-      agents: [],
-      originalSessionClosed: true,
+    }).toMatchObject({
+      agents: [{ id: agentId, lifecycle: "idle" }],
+      originalSessionClosed: false,
       replacementSessionClosed: true,
     });
   } finally {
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the recovery implementation in open [#4041](https://github.com/getpaseo/paseo/pull/4041).

PR #2684, opened 2026-07-31
fix(opencode): reconcile turns after event stream EOF

### LINKED ISSUE #2271: OpenCode turns still fail with "event stream ended before the turn reached a terminal state" when the global SSE stream drops mid-turn
### What happens

Foreground OpenCode turns fail with:

```
[System Error] OpenCode event stream ended before the turn reached a terminal state
```

while the underlying OpenCode session is still busy and usually finishes normally (the completed assistant message is visible in `session.messages` / `opencode export` afterwards). Same user-visible symptom as #861, but on current versions and with a different root cause on the Paseo side.

### Environment

- Paseo `0.1.110` (the relevant code path is unchanged at current `main`)
- OpenCode `1.18.3`
- macOS arm64 (Apple Silicon), daemon managing multiple concurrent OpenCode agents

### Why it happens

Since #916 the OpenCode adapter trusts the global event stream (`client.global.event`, subscribed with `sseMaxRetryAttempts: 0`) as the only signal for turn progress, and the old messages-API recovery path was removed. `consumeEventStream` treats any end/error of that stream while a foreground turn is active as fatal: it fails the turn immediately. There is no reconnect attempt and no reconciliation against session state.

That makes a single dropped SSE connection — for any environmental reason: host under load, proxy/timeout in between, transient socket error — kill an otherwise healthy turn. The May regression that prompted #861 was OpenCode-version-specific; this one is structural.

### Repro / evidence

Reliable trigger for me is host load. With 9 concurrent OpenCode agents against one `opencode serve` on a loaded machine (daemon event-loop stalls up to ~4.4 s in `ws_runtime_metrics`, concurrent git operations timing out), 9 turns failed this way within ~5 minutes, all with `provider.opencode.stream.eof` traces and no stream error beyond EOF:

```
18:12:20 turn_failed  agent adfddc7f… ses_08317b288ffe… "OpenCode event stream ended before the turn reached a terminal state"
18:12:27 turn_failed  agent a5660155… ses_0891420bfffe… (same)
18:16:45 … 18:17:45   7 more identical failures
```

In each case OpenCode's session status was still `busy` (or later `idle` with a completed assistant message) — only Paseo's view of the turn died.

A deterministic unit-level repro is: end (or error) the mocked global event stream mid-turn while `session.status` reports `busy` → the turn fails instead of reconnecting.

### Proposed fix (PR follows)

On stream end with an active foreground turn, reconcile before giving up:

- read `session.status`, `session.messages`, and pending question/permission requests;
- if OpenCode is still busy or has pending requests, reconnect the stream (capped exponential backoff) and bridge any assistant parts that streamed while disconnected;
- if the session completed while disconnected, recover the completion (including usage) instead of failing;
- only fail the turn when the session state is genuinely unresolvable.

The stream stays the source of truth — the messages API is used once per disconnect to bridge the gap, not polled, so this is not a return to the pre-#916 recovery path.

I've had this fix cherry-picked onto `v0.1.110` running my production daemon (≈19 real OpenCode agents) since 2026-07-20 with no EOF turn failures since.

timigod:
### Linked issue

Closes #2271

Supersedes #2272 with the same bounded outcome rebuilt and re-reviewed from the latest stable release, `v0.2.5` (`6fc491e6220fba6543bbbe4bf1b1f58cfe59228b`). The old PR remains unchanged so its discussion and evidence stay readable.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### What does this PR do

The OpenCode adapter can lose the terminal portion of a turn when the global event stream ends before the provider session has finished. This change reconciles the active turn from persisted provider state instead of treating stream EOF alone as terminal truth.

The focused replacement:

- keeps turn ownership explicit while reconciling persisted provider messages;
- recovers multi-step turns without accepting ambiguous or unrelated assistant output;
- preserves completed usage and cost, including failed and canceled terminal turns;
- isolates persisted terminal errors from later turns;
- keeps the normal live event stream as the primary path; and
- preserves current upstream autonomous-turn behavior: idle persisted user content does not start a turn, while busy status and permission-first live activity do.

It contains no Plexer policy, host topology, fleet routing, release bump, or machine-local behavior.

### How did you verify it

The current head is `a69c283a4bff686834df67e62cc0c2721c0298ca`. It retains the stable-derived change and merges current upstream `main` at `70ed70d36e2eabd3876a22931954319ec202c902` without force.

- 114 focused OpenCode adapter tests passed on Node 24.
- Full server and CLI dependency build passed.
- Server/CLI/protocol dependency typecheck passed.
- Targeted lint and formatting passed.
- The all-workspace pre-commit lint, format, and typecheck passed.
- Greptile reviewed the exact current head and reported 5/5 confidence with no blocking defect.
- GitHub reports the PR mergeable with no content conflict.

The focused tests cover disconnected completion, missing suffix recovery, pending permission/question states, provider turn ownership, multi-step persisted messages, failure/cancel outcomes, usage and cost, ambiguous-state rejection, idle persisted user content, busy autonomous activity, and permission-first activity.

No UI or desktop bundle is changed. This branch has not been installed or activated on either live daemon.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

This contribution was developed with AI assistance; every verification claim above comes from an actual run against the submitted branch.


timigod:
Fixed in `2238d3b416417124fc59af2a526779c250efd661`. The pre-dispatch history lookup now uses the active turn AbortSignal, and the continuation checks cancellation before dispatch. Added interrupt and close regressions proving the pending lookup, `startTurn`, and close all settle without dispatching the slash command. Verification: 92 focused OpenCode tests, full server build, repository-wide typecheck, and pre-commit lint/format/typecheck all pass.

boudra:
Thanks for the PR. This workflow is superseded by https://github.com/getpaseo/paseo/pull/3395, so I’m closing this one.

PR #2730, opened 2026-08-01
fix(app): make selectors keyboard searchable

### LINKED ISSUE #2726: bug: host pickers are never searchable — search is gated behind an unreachable 10-host threshold
### What's broken

None of the host pickers are searchable, while every picker next to them is — project, starting ref, provider, model, and thinking effort all let you type to filter.

This is not a missing feature so much as an unreachable one: `HostPicker` already implements search, but it is gated behind two conditions that can never both hold in practice.

`packages/app/src/components/hosts/host-picker.tsx`:

```ts
const SEARCHABLE_THRESHOLD = 10;
...
const isSearchable = searchable === true && orderedHosts.length > SEARCHABLE_THRESHOLD;
```

The `searchable` prop has no default, so it is `undefined` unless a call site opts in. Only one of the four call sites does:

| Call site | `searchable` |
| --- | --- |
| `components/left-sidebar.tsx:439` (sidebar footer host picker) | `searchable` (true) |
| `screens/settings-screen.tsx:914` (Settings → switch host) | `false` |
| `screens/new-workspace-screen.tsx:1395` (new-workspace Host badge) | `false` |
| `components/hosts/host-filter.tsx:64` (Filter by host) | `false` |
| `components/schedules/schedule-form-sheet.tsx:718` (schedule Host field) | `false` |

So three of the pickers can never search, and the one that opts in also needs more than 10 hosts before the field appears. With a realistic number of hosts (I have 5) the search field never renders anywhere.

For contrast, the standalone host chooser modal (`src/hosts/host-chooser.tsx`) always shows a search box and matches on both label and serverId — so the app is already inconsistent with itself about whether host lists deserve search.

### Steps to reproduce

1. Add 4 or more hosts (Settings → Hosts).
2. Open any host picker: the sidebar footer host button, Settings → switch host, the Host badge in the new-workspace composer, or the "Filter by host" control on the Sessions/Schedules screens.
3. Try to type to narrow the list.

Expected: a "Search hosts" field, as in the project and starting-ref pickers right next to it.
Actual: no search field; the only way to pick a host is to scan the list visually.

### Where did this happen

Desktop (Electron), and the same code path runs on web and mobile.

### Paseo version

0.2.5

### OS version

NixOS (Linux 6.18)

### Agent provider

Not relevant

### Anything else

I have a fix and will open a PR linking this issue.

colonelpanic8:
Fixes #2726

## Problem

Selector searchability was controlled independently by wrappers, call sites, and option-count thresholds. `Combobox` already defaults to searchable, but:

- `HostPicker` only enabled search when explicitly opted in and more than 10 hosts existed.
- `SelectField` defaulted search off.
- Provider, thinking effort, and schedule mode selectors only enabled search above six options.
- Compact thinking effort, workspace tabs, cadence, isolation, and import-provider selectors explicitly disabled search.

That made keyboard filtering depend on the surface and option count. In particular, normal thinking-effort lists were too short to cross the threshold.

## Change

- Make `SelectField` searchable by default, matching `Combobox`.
- Remove all product-level `searchable={false}` overrides and option-count search thresholds.
- Make host pickers unconditionally searchable and remove their opt-out prop.
- Keep host action rows such as **Add host** and **Enable built-in daemon** visible while filtering, after matched hosts.
- Match host search against both label and server ID through the shared combobox scorer.
- Prevent isolated E2E daemons from inheriting the developer daemon's `PASEO_PASSWORD`; tests can still opt into password auth explicitly.

Every option-list `Combobox` and `SelectField` in the app is now keyboard searchable regardless of list length. The low-level primitives still support `searchable={false}` if a non-product use needs it later.

## Tests

- Extended the existing two-host new-workspace browser case to type into host search, verify the non-match disappears, and select the match.
- Extended the Codex controls browser case to search a normal short thinking-effort list for **Extra high** and verify **Low** is filtered out.
- Added unit coverage proving always-visible host actions remain after filtered matches.

## QA

Local checks:

```text
npm run format
npm run typecheck
npm run lint -- <changed files>
npx vitest run packages/app/src/components/ui/combobox-options.test.ts --bail=1
npm run test:e2e --workspace=@getpaseo/app -- \
  e2e/browser/new-workspace-codex-mode-preferences.spec.ts \
  --grep "keeps Full Access"

Test Files  1 passed (1)
Tests      16 passed (16)
Playwright 1 passed (1)
```

The first local Playwright attempt exposed that Paseo-hosted shells pass `PASEO_PASSWORD` to child processes. The isolated E2E daemon inherited it while its seed client intentionally connected without credentials, so fixture setup timed out. The harness now removes inherited password auth, and the targeted effort-search case passes without unsetting the parent environment. The local run used the Nix-packaged browser and CI-pinned Codex 0.105. Mobile bottom-sheet behavior was not exercised locally.

### DIFF
--- a/packages/app/e2e/browser/new-workspace-codex-mode-preferences.spec.ts
+++ b/packages/app/e2e/browser/new-workspace-codex-mode-preferences.spec.ts
@@ -114,6 +114,12 @@ async function expectThinkingOptionsFit(page: Page): Promise<void> {
   await expect(popup).toBeVisible({ timeout: 10_000 });
   await expectNoTruncation(popup);
 
+  const searchInput = popup.getByPlaceholder("Search...");
+  await expect(searchInput).toBeVisible();
+  await searchInput.fill("Extra high");
+  await expect(popup.getByText("Extra high", { exact: true })).toBeVisible();
+  await expect(popup.getByText("Low", { exact: true })).toHaveCount(0);
+
   await page.keyboard.press("Escape");
   await expect(page.getByTestId("combobox-desktop-container")).toHaveCount(0, { timeout: 5_000 });
 }
--- a/packages/app/e2e/browser/new-workspace.spec.ts
+++ b/packages/app/e2e/browser/new-workspace.spec.ts
@@ -280,6 +280,10 @@ test.describe("New workspace flow", () => {
       await page.keyboard.press("Escape");
 
       await page.getByTestId("host-picker-trigger").click();
+      await page.getByPlaceholder("Search hosts").fill("Empty");
+      await expect(
+        page.getByTestId(`new-workspace-host-picker-option-${primaryServerId}`),
+      ).toHaveCount(0);
       await page.getByTestId(`new-workspace-host-picker-option-${emptyServerId}`).click();
       await expect(projectTrigger).toContainText("Choose project");
       await projectTrigger.click();
--- a/packages/app/e2e/support/helpers/isolated-host-daemon.ts
+++ b/packages/app/e2e/support/helpers/isolated-host-daemon.ts
@@ -161,6 +161,9 @@ export async function startIsolatedHostDaemon(
       cwd: serverDir,
       env: withDisabledE2ESpeechEnv({
         ...process.env,
+        // Test hosts must not inherit auth from the developer's daemon. A test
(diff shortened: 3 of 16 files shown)


boudra:

PR #2732, opened 2026-08-01
fix(server): retain and reap owned provider helper processes

timigod:
## Outcome

Makes managed OpenCode helper ownership durable across startup, shutdown races, daemon crashes, and retryable cleanup failures without overstating the OS-level ownership boundary.

## Changes

- retains exited OpenCode generations until cleanup succeeds or durable retry state is written
- rejects unresolved readiness when startup is terminated and prevents late exec confirmation from reviving it
- serializes managed-process ledger writes and persists captured identity after transient initial-write failures
- scopes bootstrap reconciliation and every retry to the fixed set of ledger records captured before daemon construction; helpers recorded later are never included
- lifecycle-owns the initial ledger read and active stale reap, propagates cancellation through supported file/process operations, and joins reconciliation during daemon shutdown
- prevents stopped daemons from mutating the ledger or racing a replacement daemon after delayed list/reap work completes
- joins uncancellable positive-PID tree-kill callbacks before abort completes and removes abort listeners on every callback, failure, and abort path
- disables cleanup retry timers for every OpenCode generation entering manager shutdown while preserving incomplete cleanup in the durable ledger
- contains reconciliation scheduler and canceller failures, consumes tracked promise rejections, and continues later daemon teardown
- makes the real POSIX fixture persist the escaped child PID before IPC and clean both owner and escaped processes even when setup fails early
- uses the original direct child handle for live Windows cleanup
- documents that POSIX `setsid` descendants and Windows descendants after leader exit require stronger OS supervision primitives
- integrates current `main` while preserving upstream `baseEnv`, runtime and launch environment overlays, relay runtime lifecycle, and the final generated managed identity-token overlay
- removes the local-only patch manifest from the upstream diff

## Verification

- latest race-focused run: 48 passed, 5 platform-skipped across 3 test files
- prior integrated focused run: 65 passed, 5 platform-skipped across 5 test files
- server/client/protocol/relay/CLI build stack passed on Node 24
- full workspace typecheck passed
- repository lint passed with zero findings
- repository format check passed
- branch-delta diff check passed
- final commit hooks passed

Exact integrated head: 35fe90bce7f0cafb09dc864230943e1c4424b2fa
Shutdown-race repair commit: 35fe90bce7f0cafb09dc864230943e1c4424b2fa
Lifecycle repair commit: 8233f2bd30dc73446b9fe485d5f9d12d69d2c98c
Repair-first commit: dd256d075789f02bc40ebeaaf6fc3081f549b20c
Integrated current main: 70ed70d36e2eabd3876a22931954319ec202c902
Official stable base: v0.2.5 / 6fc491e6220fba6543bbbe4bf1b1f58cfe59228b

This PR remains draft pending independent review and Windows CI coverage.


timigod:
Superseded by #3102, which reimplements the same generic helper-ownership failure class directly on the v0.3.0 stable base with exact process-identity cleanup and focused regressions.

PR #2738, opened 2026-08-01
fix(opencode): recover stuck stop boundaries

timigod:
## Summary

- classify OpenCode stop-boundary start failures and retry the unsent prompt exactly once
- isolate the replacement OpenCode runtime generation so closing the old session cannot cancel replacement work
- atomically register the replacement before closing the old session so Stop never loses agent ownership during recovery
- revalidate the original registered agent immediately before replacement install so a concurrent close cannot be resurrected
- close a stale replacement and preserve the durable closed record when reload loses ownership
- make old-session close and recovery cancellation abortable and bounded for Stop, close, and daemon shutdown
- fence asynchronous OpenCode turn startup with a generation canceled by Stop or shutdown before provider submission
- reject pending local startup when autonomous OpenCode activity wins slash-command discovery
- classify derived OpenCode providers through typed base-provider provenance rather than their visible provider ID
- preserve current event-stream reconnect and provider-status reconciliation behavior

## Testing

- `npx vitest run packages/server/src/server/agent/providers/opencode-agent.test.ts --bail=1` (92 passed)
- `npx vitest run packages/server/src/server/agent/agent-manager.test.ts --bail=1` (165 passed)
- `npx vitest run packages/server/src/server/agent/provider-registry-wrap.test.ts --bail=1` (2 passed)
- held-storage regression for close before reload replacement commit
- successful follow-up prompt after Stop during recovery handoff
- named shutdown coverage during asynchronous slash-command discovery
- `npm run build:server`
- `npm run typecheck`
- `npm run lint`
- `npm run format`
- `npm run format:check`
- `git diff --check`


timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. Its stop-boundary recovery overlaps focused PR #2684 and the persistent service-owner work in #3081. The discussion and commits remain available. Any uncovered current stop defect should return as a small PR from current upstream main.

PR #2741, opened 2026-08-01
feat(server): add durable coordinator resume outbox

timigod:
Adds a durable coordinator-resume outbox for agent-scoped background create and follow-up turns. The event is armed before child execution, bound to the exact child turn, persisted through process restarts, retried after failed or canceled delivery, and acknowledged only after the exact coordinator continuation completes.

This branch replays original verified commit 6cc341f3e88081f449ae358a2f1c2da7fb74abcf onto current upstream main 70ed70d36e2eabd3876a22931954319ec202c902 with an identical stable patch id and zero merge conflicts.

Original focused evidence:
- coordinator resume store and worker: 10 passed
- turn binding/start failure regressions: 3 passed
- prompt and create integration: 13 passed
- MCP notification regression: 1 passed
- full workspace typecheck: passed
- full lint: passed

This draft exists to obtain exact replay-tree CI and review. It must not be merged until those checks are green and the active Paseo release owner coordinates deployment.


timigod:
Closing this draft after exact-head correctness review and architecture decomposition. The submitted coordinator/outbox design has five release-blocking failure classes: non-composite child turn identity, terminal settlement blocked on outbox I/O, lost permission-blocked wakes, notification promised before eligibility, and startup recovery that can change execution truth or duplicate delivery. Those behaviors are not safe to repair as one broad Paseo patch. Cross-host/cross-turn orchestration and durable execution finality are being kept in the external Plexer control plane; any genuinely generic Paseo scheduler correction will return as a small current-main PR with its own regression and no dependency on this draft. The history remains here for reference; this head is excluded from the stable candidate and has no live deployment.

PR #2742, opened 2026-08-01
test: require opt-in for checkout ship GitHub e2e

timigod:
Summary
- require the checkout-ship-specific PASEO_CHECKOUT_SHIP_LIVE_GITHUB_E2E=1 opt-in before any live repository mutation
- keep the operation token out of Git remote URLs, subprocess arguments, checkout protocol/status facts, errors, and audit logs by using a bounded run-owned Git credential-cache socket
- inject a narrowly test-scoped GitHub service into the real daemon checkoutPrCreate path, pin every command to github.com, and override hostile ambient GH_HOST/GH_TOKEN values with the exact verified operation login/token
- arm remote cleanup only after create returns immutable repository ID, node ID, full name, and a run marker; refuse deletion if any ownership fact changes, delete by immutable node ID, and require marker confirmation plus exact-slug 404 readback
- bound every test-owned git/gh subprocess and aggregate primary, agent, remote, credential-cache, and filesystem cleanup failures

Regression coverage
- logged-in gh alone, unauthenticated gh, missing delete_repo scope, and an inactive scoped account cannot enable mutation
- hostile ambient GitHub host/token values cannot redirect or rebind the daemon PR path
- operation credentials stay out of argv, errors, Git config/remotes, protocol facts, and audit records
- cleanup is unarmed before successful create and refuses name collisions/replacements with different immutable ownership
- failed/timeout deletion, lost access behind 404, and incomplete confirmation cannot certify cleanup
- all primary and cleanup failures remain visible

Verification
- focused checkout-ship daemon e2e with live opt-in unset: 14 passed, 1 live test skipped
- oxfmt check and oxlint on the three changed files
- @getpaseo/server typecheck and build
- repository pre-commit gate: format, lint, and all-workspace typechecks
- git diff --check
- no live GitHub mutation test was run


timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. This test-only checkout opt-in is unrelated to the current production-backed fleet invariant set. Its history remains available; re-propose only with a fresh current-main need.

PR #2746, opened 2026-08-01
fix(server): strip GUI credential helpers from agent environments

BrianHotopp:
## Problem

On a desktop host, the daemon inherits the login session's environment (`inheritLoginShellEnv`) — including GUI credential helpers like `SUDO_ASKPASS`. Agent tool calls run without a TTY, so an ordinary `sudo apt install` falls back to the askpass helper: an invisible desktop dialog the remote user can never see or answer. The agent's turn parks until the tool timeout (5 minutes of silence under omp's default), with nothing in the chat explaining why.

Notable: the failure only exists when running the agent *through paseo* — the same agent in a real terminal prompts visibly. It's an environment-leak bug, not an agent bug.

The codebase already treats this class as a bug elsewhere: `GIT_TERMINAL_PROMPT=0` in all forge/git services, `NO_BROWSER` on ACP probes, `SSH_ASKPASS=/usr/bin/false` + BatchMode in the test harness, and the `PARENT_SESSION_ENV_VARS` strip for `CLAUDECODE`. Agent spawn env was the gap.

## Fix

- `createProviderEnvSpec` now strips `SUDO_ASKPASS`, `SSH_ASKPASS`, and `GIT_ASKPASS` (in addition to the existing parent-session vars), so credential prompts **fail fast with a legible error** the agent can react to. Values set explicitly via `runtimeSettings.env` are kept — deliberate configuration, not a leak.
- The pi and omp providers built launch env by hand and **bypassed the spec entirely**; they now route through it, gaining both strips. This was the exposed path for omp agents (which additionally harden `SSH_ASKPASS`/`GIT_TERMINAL_PROMPT` at their tool layer but not `SUDO_ASKPASS`).

Deliberately kept:
- `DISPLAY` — GUI-aware tools (browser automation) legitimately need it.
- `DBUS_SESSION_BUS_ADDRESS` — secret-service reads work silently when the keyring is unlocked; stripping would break working flows to fix locked-keyring edge cases.
- Paseo terminals — they have a real PTY, so credential prompts are visible to the remote user. That path is a feature.

Residual, documented non-goals: gpg GUI pinentry and locked-keyring D-Bus prompts (conditional, unfixable without breaking working cases).

## Tests

- Spec: strips inherited askpass vars (incl. `GIT_ASKPASS`), keeps `DISPLAY`, keeps explicit overrides.
- pi/omp launch builders: askpass + parent-session markers marked for deletion; explicit override preserved.
- Full pi + omp provider suites: 198 green. Typecheck/lint/format clean.

### DIFF
--- a/packages/server/src/server/agent/provider-launch-config.test.ts
+++ b/packages/server/src/server/agent/provider-launch-config.test.ts
@@ -265,6 +265,39 @@ describe("createProviderEnv", () => {
     expect(env.CLAUDE_AGENT_SDK_VERSION).toBeUndefined();
     expect(env.CLAUDE_CODE_ENABLE_SDK_FILE_CHECKPOINTING).toBe("true");
   });
+
+  test("strips inherited askpass helpers so credential prompts fail fast", () => {
+    const base = {
+      PATH: "/usr/bin",
+      SUDO_ASKPASS: "/usr/bin/zenity-askpass",
+      SSH_ASKPASS: "/usr/bin/ssh-askpass",
+      GIT_ASKPASS: "/usr/bin/git-askpass",
+      DISPLAY: ":1",
+    };
+
+    const env = createProviderEnv({ baseEnv: base });
+
+    expect(env.SUDO_ASKPASS).toBeUndefined();
+    expect(env.SSH_ASKPASS).toBeUndefined();
+    expect(env.GIT_ASKPASS).toBeUndefined();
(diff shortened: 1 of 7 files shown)


boudra:
Closing for now: this needs before/after output from a real Paseo agent showing the inherited askpass hang becomes a prompt error, with an explicit helper override still working. Please reopen with that evidence.

boudra:

PR #2747, opened 2026-08-01
Cache repository-common Git facts

timigod:
## Summary
- cache repository-common Git facts by normalized physical `gitCommonDir` while keeping worktree-specific config and status local
- bound ready facts, CWD mappings, and metric cardinality while preserving in-flight single-flight and invalidation safety
- invalidate and refresh facts after mutations, fetches, repository ref changes, external-state notifications, and periodic self-heal
- invalidate cached checkout diffs at the generic Git mutation boundary so an immediate post-commit non-forced read is fresh

## Context
Ten sibling worktrees could independently repeat the same repository-common Git reads, producing an active Git count of 8, a pending peak of 108, queue p95 of 15-34 seconds, and event-loop stalls up to 8.9 seconds. Exact-CWD coalescing could not share those reads.

## Safety
- effective Git config remains per worktree for `extensions.worktreeConfig`, `config.worktree`, and `includeIf` correctness
- ready caches are capped at 128 repositories, 64 facts per repository, and 512 CWD mappings
- in-flight loads are separate from ready LRU state and invalidated loads cannot populate or satisfy newer reads
- forced refresh coordination is generation-aware and distinguishes fresh recomputation from forced notification
- unsupported recursive ref watching falls back to a 60-second self-heal

## Verification
- `workspace-git-service.test.ts`: 35 passed
- `workspace-git-service.primitive.test.ts`: 54 passed
- `checkout-git.test.ts`: 131 passed, 1 skipped
- `git-mutation-service.test.ts`: 13 passed
- `npm run typecheck`: passed
- `npm run lint`: passed with 0 warnings and 0 errors
- `npm run format:check`: passed
- `git diff --check`: passed
- independent final review: no findings

## Base
Built directly from stable v0.2.5 commit `6fc491e6220fba6543bbbe4bf1b1f58cfe59228b` without importing current-main-only implementation history.


timigod:
Superseded by #2833, rebuilt as one focused commit on current upstream without force-rewriting this review branch.

PR #2751, opened 2026-08-01
feat(app): unify Add Project search across local and GitHub sources

BrianHotopp:
## Problem

The Add Project flow opened on a method page that forced an eager choice: pick a search domain (directory / GitHub / new directory) before seeing any data. The method page's search box only filtered methods — typing a directory name there showed "No matching options," which reads as a broken autocomplete.

## Change

The method page becomes a unified search:

- **One input, concurrent fan-out.** Local directory suggestions and GitHub repository search run in parallel. Results group under fixed-order Local/GitHub headers with per-group empty states; groups never reorder on arrival.
- **Source-filter pills** under the input (All / Local / GitHub, horizontally scrollable, gated on the host's clone capability). Default is All; explicit selections are sticky **per host** — only non-default choices are persisted, so the default never stomps an untouched host, and a stored filter that outlives its capability falls back to All. Arrow keys cycle the filter with the same keyboard parity as row navigation.
- **Continuations ride on result selection.** Directories open the project; repositories continue to the clone-destination page; New directory and Browse remain as action rows in the Local group. Manual GitHub URL / owner-repo entry works without the account-search capability, as before.
- **Lawful fan-out.** Results are generation-tagged by query string, so stale completions are never rendered regardless of finish order; each domain keeps its own error state; the rate-limited GitHub search debounces at 600ms behind the 250ms local scan.

## Tests

- Model: unified page replaces method in the state machine; back-stack behavior preserved.
- Options: source availability, capability-fallback of stored filters.
- Filter store: default-All, per-host persistence, All removes the stored pref.
- Smoke-tested in the web app: grouped fan-out with live data, pill filtering, stickiness across close/reopen, ArrowLeft/Right cycling.

Draft while I decide whether the group headers want a visual pass.

### DIFF
--- a/packages/app/src/add-project-flow/filter-store.test.ts
+++ b/packages/app/src/add-project-flow/filter-store.test.ts
@@ -1,0 +1,41 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { readAddProjectSourceFilter, useAddProjectFilterStore } from "./filter-store";
+
+vi.mock("@react-native-async-storage/async-storage", () => ({
+  default: {
+    getItem: vi.fn().mockResolvedValue(null),
+    setItem: vi.fn().mockResolvedValue(undefined),
+    removeItem: vi.fn().mockResolvedValue(undefined),
+  },
+}));
+
+describe("add-project source filter store", () => {
+  beforeEach(() => {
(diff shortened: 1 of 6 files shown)


boudra:

PR #2744, opened 2026-08-01
fix(providers): decouple health checks from daemon status

timigod:
## Summary
- make provider health checks nonblocking, generation-aware, and requested-provider scoped
- propagate cancellation through provider catalog, executable, session, transport, and OpenCode helper-server cleanup
- drain all owned work before replacement or shutdown and preserve initiating errors across cleanup failures
- keep protocol compatibility while exposing explicit checking, stale, available, and unavailable health states

## Verification
- affected server tests: 615 passed, 1 platform skip
- `npm run build:server`
- `npm run typecheck`
- `npm run lint` (0 warnings, 0 errors)
- `npm run format:check`
- `git diff --check`
- independent final review: no findings

## Related dependency
- https://github.com/timigod/codex-capacity/commit/b85c6a488a5530d7022eb6c0fed455246efded50

PR #2734, opened 2026-08-01
fix(server): own provider generation lifecycle through shutdown

timigod:
## Outcome

Makes provider generations, snapshots, managed loads, session startup, and provider-owned processes retain truthful ownership through refresh, timeout, replacement, abort, and daemon shutdown.

## Changes

- drain refresh and load operations before unregistering ownership
- bound provider-load draining before terminal client shutdown
- make cleanup run on both fulfillment and rejection
- prevent registration after shutdown begins
- cancel acquisition and replacement work at the owning boundary
- close superseded clients and preserve managed-process truth
- add protocol-compatible provider lifecycle receipts and regressions

## Startup cancellation amendment

- propagate shutdown cancellation through direct and wrapped create, resume, and reload startup
- make ACP initialization abortable and bound courtesy close while terminating child processes concurrently
- dispose Codex children that resolve after startup cancellation and own late cleanup rejection
- retain OpenCode acquisition after session creation begins, delete late-created sessions, own cleanup failures, and drain cleanup before shutdown
- pass OpenCode acquisition cancellation to the server manager so canceled callers do not retire reusable shared generations

## OpenCode teardown amendment

- forward startup cancellation into the native `session.create` request
- terminally bound failed-create cleanup so shutdown releases the helper runtime when the SDK ignores cancellation
- start native close abort without waiting behind event-stream teardown and bound both non-cooperative operations
- fence native abort and ephemeral deletion by server-scoped session generation so a superseded reload cannot abort its replacement

## Final lifecycle amendment

- register create and resume startup ownership before asynchronous acquisition begins
- cancel active startup during client shutdown and drain its terminally bounded cleanup before server-manager shutdown
- make startup `provider.list` signal-aware and bounded while retaining the raw SDK operation until actual settlement
- bound ephemeral `session.delete` before releasing the server while retaining the raw delete operation
- use completion order for same-session generations and concurrent manager reload swaps
- serialize reload swaps with explicit close so each displaced replacement is closed and cannot be orphaned

## Current-main integration

- merged official `main` at `70ed70d36e2eabd3876a22931954319ec202c902` with an ordinary two-parent merge
- resolved the sole textual conflict by retaining both the provider ownership regressions and current-main base-environment coverage
- preserved current-main OpenCode busy-status and turn-cancellation behavior

## PR #2744 integration guidance

Do not combine this branch with draft PR #2744. During later integration, normalize all OpenCode acquisition methods to one options-object shape:

[...]

PR #2758, opened 2026-08-01
fix(providers): decouple health from daemon status

timigod:
## Outcome

Replaces draft #2744 with the provider-health/status repair, now combined with the accepted provider lifecycle ownership change from #2734. This PR remains draft and does not close or modify #2744.

## Changes

- return cached `checking` / `stale` / `available` / `unavailable` provider truth from daemon status without awaiting live discovery, while preserving daemon identity and version
- keep the legacy `available` boolean conservative: missing or unverified providers report `false` to older clients
- run one bounded fresh check for the selected provider before agent creation or resume mutates state; background status probes yield so status can return first
- invalidate and detach superseded probes immediately, so a replacement check starts even when an old provider ignores abort forever
- prevent invalidated or late probe work from publishing, observe late rejection, and remove settled/current probe bookkeeping without retaining never-settling work
- preserve optional `AbortSignal` forwarding across wrapped providers and integrate #2734's provider/session/process shutdown ownership
- render explicit daemon health in the CLI and preserve mixed-version protocol parsing

## Integrated inputs

- #2734 exact accepted head: `dc5107d589d4186fba6366365a8c4a7bab851a58`
- latest `origin/main` integrated and verified: `76abc50f822c4c3370d15614c88e7254a25411c6`

## Regression coverage

- missing/unverified providers remain unavailable to legacy boolean-only clients
- provider overrides and wrapped clients preserve availability/bootstrap compatibility
- an invalidated never-settling probe does not block its replacement, publish late state, leave current probe bookkeeping, or create an unhandled rejection
- a late old success cannot overwrite the replacement result
- startup, resume, shutdown, ACP, OpenCode, OMP, Pi, Codex, daemon-session, protocol, and CLI status paths are exercised together

## Verification

- combined material matrix: 23 files passed; 880 tests passed, 1 skipped
- server integration matrix: 3 files passed; 12 tests passed
- full server unit matrix: 303 files passed, 3 skipped; 4,477 tests passed, 40 skipped; two unrelated failures reproduced:
  - `workspace-service-port-allocator.test.ts`: macOS `/var` versus `/private/var` path identity
  - `opencode-agent.slash-command-timeout.test.ts`: event stream ended before terminal state
- `npm run build:server`
- `npm run typecheck`
- `npm run lint` (0 warnings, 0 errors across 3,033 files)
- `npm run format:check` (3,233 files)
- `git diff --check origin/main...HEAD`

No package publication, deployment, service restart, or production change is included.


timigod:
Independent exact-head rereview: READY at `2adf9c7587a1e78606e0c2e4b680794c26da0bc0` (tree `97af3a94df179b1b3315a515c250f37521224f6c`, base `048b82f2a7b3ecc9b62bc5b817729f4ca2741863`). No P0–P2 findings. One monotonic 8s graceful deadline stays below the 10s worker force-exit; ingress/provider loads abort promptly; agent/provider/hub/schedule/relay drains begin concurrently; later teardown still runs with remaining budget; late provider children remain owned; old/new provider-health payloads remain compatible. Focused shutdown/startup/generation verification, protocol/server/CLI typechecks, and `git diff --check` passed.

timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. This provider-health branch is broad and stale. Plexer-specific provider failure classification belongs in the external Plexer companion. Any upstream-generic health primitive should return as a small PR from current upstream main with a fresh reproduction.

PR #2764, opened 2026-08-01
feat: Background Tasks track, steer, interaction lock, ui_state

iamriajul:
## Summary
- Claude shell **Background Tasks** track under Subagents (live list, icon-only Stop, focus-gated log tab) with protocol/server/client plumbing and post-review hardening for live EOF, terminal cleanup, and Bash command correlation.
- Native **Steer** for Claude/Codex/OMP plus cleaner send defaults.
- Android **whole-app interaction lock** with biometric/device unlock, plus BottomSheet input-scope crash fix.
- Host **ui_state** draft sync and related protocol/server/client support already on this branch.

## Test plan
- [x] Targeted unit tests for background-task store/mappers/EOF/correlation and app focus gate
- [x] `npm run typecheck` / lint / format on changed stacks via pre-commit
- [ ] Manual: Claude `run_in_background` (`npm run dev` / `sleep`) shows track under Subagents
- [ ] Manual: Stop icon stops task and row disappears after membership/terminal cleanup
- [ ] Manual: open task tab; logs poll only while focused; missing `outputFile` does not permanently kill subscription
- [ ] Manual: Android lock/unlock with biometrics/device credential
- [ ] Manual: Steer mid-turn on Claude/Codex/OMP

## Notes
Multi-feature branch intended for merge to `main` then stable release prep. Background-task blockers from pre-PR review are fixed in `57c229824`.

PR #2766, opened 2026-08-01
feat: Claude Background Tasks track under Subagents

iamriajul:
## Summary
- Add a collapsible **Background Tasks** track under Subagents for Claude Code long-running shell processes (e.g. `npm run dev`).
- Protocol + daemon live set from Claude SDK membership signals, with list/stop/output RPCs and focus-gated log polling.
- App store, track UI with icon-only **Stop**, and a workspace detail tab for logs.
- Hardening from pre-PR review:
  - do not permanently EOF/unsubscribe while a task is still live (even without `outputFile` yet)
  - remove terminal tasks from the live set on completion/stop/fail notifications
  - unwrap nested Bash tool-result payloads so command correlation works

## Test plan
- [x] Unit tests: store, mappers, EOF policy, bash correlation, app focus gate (23 targeted tests)
- [x] Protocol/client/server typecheck green
- [ ] Manual: Claude `run_in_background` shows track under Subagents
- [ ] Manual: Stop removes the row after membership/terminal cleanup
- [ ] Manual: open task tab; logs poll only while focused; missing `outputFile` keeps subscription while live
- [ ] Manual: command label appears when Bash returns `backgroundTaskId`

## Notes
Focused branch rebased onto current `upstream/main` (cherry-picked Background Tasks commits only). Separate multi-feature branch PR #2764 can be closed in favor of this.


iamriajul:
Closing: PR should target the fork (iamriajul/paseo), not upstream. Re-opened there.

PR #2771, opened 2026-08-01
fix(cli): bound WebSocket close handshake

timigod:
## Summary

- keep the Node CLI WebSocket graceful close path, then terminate it after a one-second bound if the peer does not complete the handshake
- cancel the fallback on normal close and unref its timer so it cannot retain the CLI process
- apply the behavior at the CLI-only WebSocket factory, covering direct, IPC, and relay command cleanup without changing shared desktop or mobile transports

## Live reproduction

Before this patch, a live `paseo agent finish <id> --json` run against official main `048b82f2a7b3ecc9b62bc5b817729f4ca2741863` committed agent archival and workspace removal, emitted no terminal receipt, and retained only the CLI process for more than 90 seconds until it was interrupted.

## Tests

- `npx vitest run packages/cli/src/utils/client.test.ts --bail=1`
- `npx tsx packages/cli/tests/28-client-ipc-targets.test.ts`
- `npm run build:server`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

### DIFF
--- a/packages/cli/src/utils/client.test.ts
+++ b/packages/cli/src/utils/client.test.ts
@@ -1,0 +1,98 @@
+import { describe, expect, it } from "vitest";
+import { installBoundedWebSocketClose } from "./client";
+
+class FakeCliWebSocket {
+  readyState = 1;
+  closeCalls: Array<{ code?: number; reason?: string }> = [];
+  terminateCalls = 0;
+  private closeListeners: Array<() => void> = [];
+
+  send(): void {}
+
+  close(code?: number, reason?: string): void {
+    this.closeCalls.push({ code, reason });
+  }
+
+  terminate(): void {
+    this.terminateCalls++;
+  }
+
+  once(event: "close", listener: () => void): void {
+    if (event === "close") {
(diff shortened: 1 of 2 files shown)


timigod:
Independent exact-head review is READY at 90fcae4472fccd0956980116c5240562e4507ff0 (tree 42a40241b3385958f7e2c3a4dd4ce94aa0f0ad27; base 048b82f2a7b3ecc9b62bc5b817729f4ca2741863). A real CLI workspace-archive reproduction returned the success receipt, the server deliberately withheld the close acknowledgement, and the CLI exited 0 in 1066ms with no lost output. Focused CLI/archive tests, build, typecheck, lint, format and diff checks passed; no P0-P2 findings. Ready-for-review mutation is pending only because the authenticated GitHub GraphQL quota is currently exhausted.

timigod:
Replaced by #2818, which preserves the bounded-close behavior as one clean v0.2.5-derived commit, removes the export-only test seam, and adds a real `ws` regression. Closing this superseded branch so review can continue on the smaller unit.

PR #2740, opened 2026-08-01
fix: add GitHub rate-limit cooldown

timigod:
## Summary
- Share GitHub API rate-limit cooldowns per resolved host.
- Parse retry metadata without retaining API response bodies in errors.
- Block repeat API calls until the advertised or bounded retry time.

## Validation
- `npx vitest run packages/server/src/services/github-service.test.ts --bail=1`
- `npm run typecheck`
- Focused sidebar and drag tests (15 passing)

PR #2773, opened 2026-08-01
fix(server): bound workspace git refresh pressure

timigod:
## Outcome

Keep workspace inventory responsive during large Git refresh/watch waves by bounding the higher-level workspace pipelines that feed Git and forge subprocess work.

This change is intentionally independent of #2740. That PR handles per-host/provider cooldown and serial admission after a rate-limit response; this PR handles workspace-level admission pressure and truthful cached inventory. The two changes have no semantic dependency, although both touch `workspace-git-service.ts` and may need a mechanical conflict resolution if they land together.

## Incident evidence

The resident 0.2.8 daemon remained alive and listening, but became temporarily unable to answer control/status requests during a workspace refresh burst:

- 120 workspace records, 116 Git subscriptions/watch directories, and 86 active targets
- internal queues observed at 121 and 145 entries
- GitHub GraphQL rate-limit responses during self-heal polling
- `fetch_workspaces` responses at 21.3s and 16.7s
- maximum event-loop delay at 21.3s
- websocket `bufferedAmount` near 13 MB
- daemon status timed out while the listener, process, and port remained alive, then recovered after the burst drained

`fetch_workspaces` was already a passive `peekSnapshot` projection on current `main`. The remaining fan-out was registration and aligned self-heal: every observed cwd could independently enter a full refresh and observation-setup pipeline at once. The lower-level Git command limiter capped individual commands, but did not bound the number of workspace pipelines waiting on Git/forge work or their completion fan-out.

## Change

- Admit at most four whole workspace refresh pipelines concurrently.
- Admit at most two observation-setup pipelines concurrently.
- Defer normal observation setup until an admitted refresh has produced the first snapshot; self-heal setup remains bounded and can still recover a failed initial setup.
- Preserve per-cwd single-flight/coalescing while giving each refresh generation its own completion: a queued force/forge upgrade replays even if the current generation rejects, and its callers receive that later generation's result.
- Coalesce ordinary watcher/self-heal signals received in flight into one later generation, keeping cached facts stale until that generation succeeds.
- Reconcile once through bounded refresh admission after watchers attach, closing the facts-before-watch race without duplicate unchanged emissions.
- Reject pending refresh and observation admissions on disposal, settle queued/active generation callers with `AbortError`, and ignore late dependency results without starting follow-on work.
- Keep workspace inventory passive and cached. Cached Git/GitHub facts now expose additive optional `isStale: true` markers while refresh work is queued/running or after it fails; successful refresh clears the marker.
- Expose admission active/pending counts in service metrics.

The refresh cap is four rather than two because the existing lower Git command pool admits eight commands. Four refresh pipelines plus two observation pipelines bound the higher-level demand at six, preserving two command slots and event-loop/subprocess headroom for health/control work without doubling the cold-start tail that a cap of two would impose. Regression assertions reference the exported configured caps, not elapsed timing.

## Verification

- Protocol workspace schema tests: 37 passed.
- Focused workspace Git/session/observer tests: 214 passed, 4 skipped.
- Existing workspace response projection tests: 2 passed.
- Added deterministic 120-workspace coverage for configured admission caps, coalesced repeated scheduling, aligned 60-second self-heal pressure, forced refresh completion behind the last queued cwd, cached inventory while refreshes are blocked, and stale/fresh/failed snapshot transitions.
- Added review regressions for upgrade replay after a rejected generation, ordinary watcher replay with continuous staleness, a ref change during watcher attachment, clearing both admission queues on disposal, immediate caller settlement, and late-result suppression.
- Protocol and server typechecks passed; the commit hook also typechecked every package workspace.
- Protocol and server builds passed.
- Changed-file lint and format checks passed with zero warnings/errors.
- `git diff --check` passed.

The broad server suite was also attempted. Three environment-sensitive/pre-existing failures reproduced outside this patch's paths: installed Codex availability versus an unavailable expectation, `/private/var` canonicalization versus `/var`, and two zsh integration cases under full-suite load (both zsh cases pass when isolated).


timigod:
Exact-head independent rereview: READY.

Accepted commit: `2c40f9fb5f6be6cd9f918fd62b811dc1092f9f2e`
Tree: `b2a4efb493ea1e4c4112592d35025d8750d32f87`
Base: `048b82f2a7b3ecc9b62bc5b817729f4ca2741863`

The prior starvation reproduction no longer holds: each refresh generation re-enters the global FIFO, so a queued fifth workspace is admitted before a hot workspace successor. The prior late-watcher reproduction also no longer holds: disposal during blocked or re-entrant setup leaves no published target and closes every created watcher exactly once. Failed post-install reconciliation still reuses one watcher generation.

Independent gates: workspace Git 33/33, primitive 54/54, session workspaces 102 passed/4 skipped, protocol 37/37, server/protocol typechecks, protocol codegen with no diff, targeted lint, repository format, and diff check. Worktree remained clean; no live mutation.

timigod:
Superseded by #2831, rebuilt as one focused commit on current main. The replacement preserves the original backpressure goal while avoiding the stale branch conflicts and unrelated historical changes.

PR #2784, opened 2026-08-01
Fix worktree setup with production npm environments

colonelpanic8:
## Summary

- Force dev dependencies for Paseo worktree and Docker installs.
- Resolve `patch-package` directly instead of relying on the lifecycle `PATH`.
- Handle intentionally omitted dev dependencies without an `ENOENT` failure.
- Document the `NODE_ENV=production` interaction.

## Root cause

The daemon environment can include `NODE_ENV=production`. npm interprets that as `--omit=dev`, so `patch-package` is not installed while the postinstall script still attempts to spawn it.

## Validation

- `node scripts/postinstall-patches.mjs` — all patch groups applied successfully.
- `npm run lint`
- `npm run typecheck`
- `npm run format:check`


boudra:
Closing in favor of the setup fix in open [#2650](https://github.com/getpaseo/paseo/pull/2650).

PR #2743, opened 2026-08-01
Fix Kimi usage refresh, 5 hour session limit, and reasoning levels

UnbrokenHunter:
### Linked issue

Closes #2683 and #2705

### Type of change

- [x] Bug fix
- [ ] New feature (with prior issue + design alignment)
- [ ] Refactor / code improvement
- [ ] Docs

### What does this PR do
Fixes Kimi usage showing as unavailable due to stale tokens not requesting a refresh.

Later PR rescope also fixes Kimi 5 hour limit not being displayed in usage panel, and fixes incorrect Kimi reasoning levels. 

### How did you verify it
Forced a refresh by interacting with Kimi, then waited 15 minutes for the token auth to go stale. After it went stale, it requested another correctly.

Also, used the patch for the past 24 hours and has worked fine. 

Later rescope: I have continued using this and have had no problems. I have verified in app that the reasoning levels and 5 hour usage UI displays correctly on Windows 11, but have not tested any other platforms. The changes reuses existing infrastructure though, so they should appear identically to how other providers appear on all platforms. 

image

image

### Checklist

- [x] One focused change. Unrelated cleanups split out.
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` ran (Biome)
- [ ] UI changes include screenshots or video for every affected platform
- [x] Tests added or updated where it made sense

PR #2825, opened 2026-08-03
fix(acp): probe per-model thinking options in the provider catalog

### LINKED ISSUE #2823: bug: ACP provider thinking options don't follow the selected model
### What's broken

I added Kimi Code CLI as an ACP custom provider with several models, each declaring different `efforts` levels in `config.toml`. When I switch the model in the Paseo desktop app (e.g. from `deepseek-v4-flash` to `gpt-5.6-terra`), the effort options in the composer stay the same (`off / low / high / max`) options in `deepseek`, not the `gpt-5.6` options (`off / medium / high / xhigh / max`).

### Steps to reproduce

1. Configure Kimi Code CLI (0.31.1) with two models declaring different
   `support_efforts`, e.g. `deepseek-v4-flash` → `["low","high","max"]`
   and `gpt-5.6-terra` → `["medium","high","xhigh","max"]`, both with
   `capabilities = ["thinking"]`.
2. Add "Kimi Code CLI" from the ACP provider catalog in Host settings
   (default model = `deepseek-v4-flash`), let the snapshot probe finish.
3. Open the composer model selector, switch to `gpt-5.6-terra`.
4. Observe the thinking-depth options: still `off / low / high / max`
   (expected `off / medium / high / xhigh / max`).

### Where did this happen

Desktop (Electron)

### Paseo version

0.2.5

### OS version

Linux

### Agent provider

Custom provider

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

_No response_

MikoyChinese:
Closes #2823

## What

ACP provider catalog probing only read the thinking options of the session's default model, so switching models in the composer kept showing the old model's effort levels. `fetchCatalog` now probes each selectable model and attaches per-model thinking options (and per-model defaults) to the catalog definitions.

## Testing

- 3 new unit tests: per-model option derivation, per-model probing during `fetchCatalog`, and fallback to shared options when a single model probe fails
- Full `acp-agent.test.ts` suite passes: 99/99
- lint + format clean on the changed files

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -2502,6 +2502,85 @@ describe("ACPAgentClient fetchCatalog", () => {
       vi.useRealTimers();
     }
   });
+
+  test("stalled probes run concurrently, so many stalled models stay within one per-model budget", async () => {
+    // Both model probes never answer; the whole catalog must still finish
+    // after a single PER_MODEL_THINKING_PROBE_TIMEOUT_MS, not N × that.
+    const setSessionConfigOption = vi.fn(() => new Promise(() => {}));
+
+    class TestACPAgentClient extends ACPAgentClient {
+      protected override async spawnProcess(): Promise<SpawnedACPProcess> {
+        return {
+          child: { kill: vi.fn(), exitCode: 0, signalCode: null, once: vi.fn() },
+          connection: {
+            newSession: vi.fn().mockResolvedValue({
+              sessionId: "sess-1",
+              modes: null,
+              models: null,
+              configOptions: [
+                {
+                  id: "model",
+                  name: "Model",
+                  category: "model",
+                  type: "select",
+                  currentValue: "model-a",
+                  options: [
+                    { value: "model-a", name: "Model A" },
+                    { value: "model-b", name: "Model B" },
+                  ],
+                },
+                {
+                  id: "thinking",
+                  name: "Thinking",
+                  category: "thought_level",
+                  type: "select",
+                  currentValue: "high",
+                  options: [
+                    { value: "off", name: "Off" },
+                    { value: "low", name: "Low" },
+                    { value: "high", name: "High" },
+                    { value: "max", name: "Max" },
+                  ],
+                },
+              ],
+            }),
+            setSessionConfigOption,
+          },
+          initialize: { agentCapabilities: {} },
+        } as SpawnedACPProcess;
+      }
+
+      protected override async closeProbe(): Promise<void> {}
+    }
+
+    const client = new TestACPAgentClient({
+      provider: "pi",
+      logger: createTestLogger(),
+      defaultCommand: ["test-acp"],
+      defaultModes: [],
+    });
+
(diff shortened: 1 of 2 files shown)


MikoyChinese:
Thanks for the review — the stalled-probe concern is valid and is now fixed in `ccbfc07a0`. @greptile-apps

Confirmed root cause: the ACP SDK's `sendRequest` (`@agentclientprotocol/sdk/dist/acp.js:894`) has no request timeout, so a model-switch request that never settles held the whole `fetchCatalog` hostage inside its single 60s budget (`ACP_CATALOG_TIMEOUT_MS`) and failed the entire provider snapshot.

Fix: each per-model probe is now wrapped in `withTimeout` with `PER_MODEL_THINKING_PROBE_TIMEOUT_MS = 5_000` (exported). A stalled probe rejects after 5s, lands in the existing catch, logs a warning, and the model falls back to the shared options — same path as a rejected probe. Several stalled probes stay well within the 60s catalog budget.

Added test: `a stalled per-model probe times out and falls back without failing the catalog` — mocks a never-settling probe for one model and asserts the catalog still completes with the stalled model on shared options. Suite is now 100/100 passing.

MikoyChinese:
Agreed — serial probing kept a `models × 5s` worst case, so a 15-model agent with several stalled probes could still exhaust the 60s catalog budget. Probes are now concurrent (`f1df694b1`). @greptile-apps

The loop is now a `Promise.all` over the model list; each probe keeps its own `withTimeout` so the total worst case is **one** per-model timeout regardless of model count. This is safe on the wire because ACP is JSON-RPC — responses are matched per request id (`#pendingResponses` in the SDK) and local agents process the sequential writes one at a time anyway. A failed or stalled probe logs a warning and returns `null`, and the map build skips it — same shared-options fallback as before.

Added test: `stalled probes run concurrently, so many stalled models stay within one per-model budget` — both probes never settle and the catalog completes after a single `PER_MODEL_THINKING_PROBE_TIMEOUT_MS` advance. Suite is 101/101 passing.

MikoyChinese:
You're right on both counts — the serial version kept a `models × 5s` worst case, and the concurrent version could associate a session-wide `configOptions` response with the wrong model when an agent processes overlapping `setSessionConfigOption` requests. Correctness wins, so probes are serial again with a 2s per-probe timeout (`f359c5fbe`). @greptile-apps

An ACP session has one shared model state; serial order guarantees each response is built after exactly the requested model was applied, so there is no cross-model contamination. The serial worst case is `models × 2s` — 30 stalled models reach the 60s catalog budget, and realistic model counts (single digits to low teens) with normal millisecond-round-trip probes stay far below it. A stalled probe still times out into the existing warn-and-skip path and falls back to shared options.

Tests updated: the concurrency test is now `multiple stalled probes each time out serially without failing the catalog` (two never-settling probes, each advancing one `PER_MODEL_THINKING_PROBE_TIMEOUT_MS`, catalog completes). Suite is 101/101 passing.

boudra:
Closing in favor of the Kimi per-model catalog implementation merged in [#2743](https://github.com/getpaseo/paseo/pull/2743).

PR #2831, opened 2026-08-03
fix(server): bound workspace git pipelines

timigod:
## Summary
- limit whole workspace Git refresh generations to four concurrent pipelines
- limit workspace observation setup to two concurrent pipelines
- bound watcher subscription setup with a 10-second deadline so stalled native subscriptions release global admission slots
- make service disposal terminal, reject queued limiter work, and clean subscriptions that resolve after timeout or disposal
- expose active and waiting admission counts in runtime diagnostics
- preserve per-workspace coalescing while re-entering the global FIFO limit for queued successors

## Context
Supersedes #2773 with a focused reconstruction on current `main`. The daemon-wide Git process scheduler limits individual subprocesses, but workspace registration could still start an unbounded number of higher-level refresh and watcher-setup pipelines. During the reported 120-workspace incident, that fan-out coincided with 21-second event-loop delay and multi-megabyte websocket buffering.

The original admission limits still allowed the two observation slots to remain occupied forever when `@parcel/watcher.subscribe()` never settled. Disposal also cleared service state without settling queued limiter callers. This repair gives watcher setup a finite deadline, rejects queued work on disposal, prevents post-disposal fallback or successor work, and cleans up subscriptions that resolve after they are no longer usable.

## Test plan
- `npx vitest run src/server/workspace-git-service.test.ts --bail=1` (26 passed)
- `npx vitest run src/server/workspace-git-service.observation.test.ts --bail=1` (16 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git diff --check`


boudra:
Thanks for the PR. Closing, superseded by https://github.com/getpaseo/paseo/pull/3033. Your original commits and authorship were preserved in the delivered change.

PR #2832, opened 2026-08-03
fix(cli): list worktrees from registered projects

timigod:
## Summary
- resolve global worktree inventory from registered Git projects only when the request explicitly sends `allRegisteredProjects: true`
- preserve the legacy error for unscoped requests that omit the global flag
- report only observed registered-project failures through the optional positive-only `repositoryErrors` field; absence does not claim complete inventory
- fail `paseo worktree ls` with `WORKTREE_LIST_PARTIAL` instead of returning known-partial output
- label text, JSON, and YAML output as current registered, non-archived Git projects only; archived or removed projects remain excluded, so the result does not claim to contain every managed worktree
- require `--cwd` or `--repo-root` before `paseo worktree archive <name>` connects to the daemon or resolves a destructive target
- keep the CLI's scoped preflight for useful not-found and ambiguity errors, but do not send its cached path as destructive authority
- make the daemon force a fresh scoped Git read, require exactly one identity match, verify that match still has the expected path, and retain Paseo-owned-path validation before mutation

## Regression
Repository A can expose a matching worktree while repository B cannot be inspected. The daemon returns the observed worktrees with `repositoryErrors`; `paseo worktree ls` rejects that known-partial result, and archive cannot use global inventory because it requires an explicit repository selector before any RPC.

The archive regression primes the real 15-second worktree cache, renames a branch in place, reuses the old branch name at another path, and confirms an ordinary read remains stale. Archive then forces a fresh read and rejects the rebound identity because its path no longer matches. Neither worktree, workspace records, nor agents are archived.

## Protocol compatibility
`allRegisteredProjects` is an optional literal request field, `repositoryErrors` is an optional response field, and the expected archive identity/path fields are optional.

- old CLI + old daemon: the unchanged unscoped list request retains the old `cwd or repoRoot is required` error
- old CLI + new daemon: omission of `allRegisteredProjects` retains that explicit error
- new CLI + old daemon: the old request schema strips the unknown global-list flag, then the old daemon returns the same explicit error
- new CLI + new daemon: `allRegisteredProjects: true` enables registered-project inventory
- new archive CLI + old daemon: the old schema strips the expected identity/path fields, receives no `worktreePath`, and fails instead of mutating
- new archive CLI + new daemon: the daemon resolves the identity from a forced fresh scoped list and verifies the expected path
- older scoped clients that send `worktreePath` retain their existing behavior

Older clients ignore additive response fields, and the daemon omits `repositoryErrors` unless it observed at least one repository failure.

## Verification
- `npm run build:client`
- `npx vitest run packages/protocol/src/messages.test.ts packages/client/src/daemon-client.test.ts packages/cli/src/commands/worktree/archive.test.ts packages/cli/src/commands/worktree/ls.test.ts packages/server/src/server/worktree-session.test.ts --bail=1` - 5 files, 175 passed
- `npm run typecheck`
- `npm run lint` - 0 warnings, 0 errors across 3148 files
- `npm run format`
- `npm run format:check` - 3349 files matched
- `git diff --check`
- pre-commit format, lint, and typecheck hooks

Reviewed repair commit: `6ce5d1286cfaffbb59411843cfb63d0b647d4760`.

### DIFF
--- a/packages/cli/src/commands/worktree/archive.test.ts
+++ b/packages/cli/src/commands/worktree/archive.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, it } from "vitest";
 import type { DaemonClient } from "@getpaseo/client/internal/daemon-client";
 import { runArchiveCommandWithDeps } from "./archive.js";
+import { createWorktreeCommand } from "./index.js";
+
+const testOptions = { host: "localhost:6767" };
 
 function createFakeDaemonClient(
   overrides: Partial<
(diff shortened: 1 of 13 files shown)


timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. This exact-worktree inventory branch is broader than the current upstream contribution set, and current orchestration uses the canonical workspace APIs plus the external Plexer companion. Its history remains available; any missing generic inventory primitive should return as a small PR from current upstream main.

PR #2833, opened 2026-08-03
Cache repository-common Git facts

timigod:
## Summary
- cache repository-common default-branch, ref-existence, and main-repository-root reads by normalized physical `gitCommonDir`
- keep effective Git config, current branch, status, dirty state, and other worktree-specific reads local to each checkout
- bound ready values to 128 repositories, 64 facts per repository, and a 60-second freshness window while keeping in-flight single-flight state separate
- invalidate repository facts at existing mutation/refresh boundaries and when the final repository observer closes
- fail closed when an external-state event arrives before its cwd can be mapped: invalidate all repository facts, fence older loads, and detach that cwd's base and uncommitted diff reads
- preserve repository-scoped invalidation for mapped cwds
- pass repository identity and cache context through base-mode checkout-diff default-branch and ref-existence reads so sibling worktrees share one load

## Safety
- concurrent sibling reads share one repository-common load
- invalidated in-flight loads cannot repopulate the ready cache or replace a newer value
- a newly observed sibling cannot join an older repository-fact or checkout-diff load after its external-state event
- direct checkout reads expire even when no workspace observation target exists
- removing the final listener prevents repository facts from surviving observer shutdown
- cache keys use physical repository identity and fall back to resolved absolute paths when realpath lookup is unavailable
- mapped invalidation retains the repository-local fast path; only unknown ownership uses the global fail-closed fallback

## Verification
- `workspace-git-service.primitive.test.ts`: 65 passed
- `workspace-git-service.observation.test.ts`: 16 passed
- `workspace-git-service.test.ts`: 20 passed
- `checkout-diff-manager.test.ts`: 12 passed
- `checkout-git.test.ts`: passed
- exact unmapped-sibling regression: failed before the repair and passed after it
- `npm run build:server`: passed
- server and root workspace typechecks: passed
- root `npm run lint`: 0 warnings and 0 errors
- root `npm run format:check`: passed
- `git diff --check`: passed
- pre-commit hook: formatting, lint, and full workspace typecheck passed

## Base
Merged upstream `main` commit `777fe968f5bab50f63f25a61188482f6a15789fc` (tree `4c06280b94ecab9e2d98c4c5326ce8bdc3095928`) without conflicts.

Replaces #2747 with a smaller implementation based on the current upstream observation and refresh architecture.

### DIFF
--- a/packages/server/src/services/github-service.test.ts
+++ b/packages/server/src/services/github-service.test.ts
@@ -105,6 +105,35 @@ function createDeferredRunner(): TestRunner {
   };
 }
 
+function createQueuedDeferredRunner(): TestRunner & {
+  resolveCall: (index: number, stdout: string) => void;
+} {
+  const calls: RunnerCall[] = [];
+  const pending = new Map<number, (stdout: string) => void>();
+
+  return {
+    calls,
+    runner: (args: string[], options: GitHubCommandRunnerOptions) => {
+      const index = calls.length;
+      calls.push({ args, cwd: options.cwd, envOverlay: options.envOverlay });
+      return new Promise((resolve) => {
+        pending.set(index, (stdout: string) => {
+          pending.delete(index);
+          resolve({ stdout, stderr: "" });
+        });
+      });
+    },
+    resolveNext: () => {},
+    resolveCall: (index: number, stdout: string) => {
+      const resolve = pending.get(index);
+      if (!resolve) {
+        throw new Error(`Runner call ${index} is not waiting for resolution.`);
+      }
+      resolve(stdout);
+    },
+  };
+}
+
 interface FakeGitHubCliFixture {
   cwd: string;
   logPath: string;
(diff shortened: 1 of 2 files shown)


timigod:
Addressed the P1 stale-load finding in `a7d47fbeb`.

The root cause was the unknown-cwd branch in `invalidateRepositoryFacts`: before a direct sibling read had established either a workspace target or a cwd-to-repository mapping, external-state invalidation did nothing. A forced sibling diff could also join the older outer diff promise.

The repair now:

- invalidates base and uncommitted diff entries for the event cwd, detaching older in-flight outer reads;
- uses repository-scoped invalidation when the cwd is mapped;
- uses a fail-closed global repository-fact invalidation when ownership is still unknown, marking every older in-flight fact load invalid so it cannot repopulate the cache.

The exact regression starts an old `main` repository-fact load and an unmapped sibling diff load, emits the external-state event for that sibling, then performs a forced direct sibling diff. Before the repair it failed with only two diff loads because the forced read joined stale work. It now starts a third diff load and a second repository-fact load, returns `base:develop`, and proves that later completion of the old `main` and outer-diff promises cannot replace the fresh value.

Focused verification:

- workspace Git primitive: 65 passed
- observation: 16 passed
- workspace service: 20 passed
- checkout diff manager: 12 passed
- checkout-git file: passed
- server build and server typecheck: passed
- root workspace typecheck, lint, and format check: passed
- diff check and pre-commit gates: passed

PR remains open and unmerged.

timigod:
Addressed the remaining P1 invalidation race in 69b170c60.

The repair now fences every cache layer that can outlive repository-fact invalidation:

- checkout-diff cache entries carry a generation token, so invalidation detaches an old in-flight read and its eventual completion cannot repopulate the entry;
- unknown-cwd invalidation fences every base-diff entry, while mapped invalidation remains repository-scoped across sibling checkout consumers;
- checkout-facts loads capture the target generation and retry against the current generation instead of publishing an invalidated result to latestFacts;
- known external changes mark affected public snapshots stale, and snapshot refreshes retry if their generation changes before publication;
- scheduled self-heal/fetch audits preserve the prior low-latency snapshot while their already-scheduled replacement loads.

Regression coverage now proves:

- after the stale outer diff and old main fact complete, rereading both the triggering sibling and the original checkout returns develop;
- an in-flight public snapshot cannot publish the invalidated main result;
- an already-cached public snapshot becomes unreadable through peekSnapshot and refreshes to develop after unknown-cwd invalidation.

Focused verification:

- workspace-git-service primitive: 67 passed
- observation, workspace service, and checkout diff manager: 48 passed
- checkout-git: 138 passed, 1 skipped
- server dependency/server/CLI build: passed
- all-workspace typecheck: passed
- root lint and format check: passed
- diff check and the repository pre-commit hook: passed

Exact pushed head: 69b170c607262059e65bf1cb308fe24532baad97
Tree: c0857f7d50fb9212a9d388c4cd28d066eb614fd0

The worktree is clean. The PR remains open and unmerged. No daemon, Desktop app, live agent, service, deployment, or restart was touched. Per repository policy, no full local test-suite claim is made.

timigod:
Closed the remaining forge-poll generation-fencing P1 in `752cfb348edce42b264e1e34915e8621adcaeb2f` (tree `052a3ce369823ad48388d876d87d986ba5a322c5`).

- Forge/PR poll subscription identity now includes repository `factsGeneration`, so an unchanged textual key no longer preserves an old-generation poll.
- A generation or key change stops the prior subscription and revalidates immediately.
- Retained-adapter status/error callbacks and generic async completions, errors, and rescheduling all reject stale captured key/generation identities.
- Added a deterministic deferred-poll regression: start an unresolved generic forge poll, globally invalidate through an unknown sibling cwd, run `getSnapshot(..., { includeForge: false })`, require an immediate fresh same-key poll, and prove the late old result cannot overwrite the fresh PR state. The regression failed on the prior head with one poll instead of two, then passed after the fix.

Verification on the exact pushed tree:
- `packages/server/src/server/workspace-git-service.primitive.test.ts`: 68 passed
- `workspace-git-service.test.ts` + `workspace-git-service.observation.test.ts`: 36 passed
- `npm run build:server-deps`
- `npm run typecheck:server`
- `npm run lint`
- `npm run format:check`
- repository pre-commit hook: staged format/lint plus full all-workspace typecheck

The branch was pushed by normal fast-forward. No daemon, provider, or Desktop runtime was launched, restarted, or deployed.

timigod:
Closed the retained-GitHub lifecycle gap in `6194bec6b394fe1816bd779f992d3b347bee2f81` (tree `0c28ca6666d799313040521d987c20ce415c178a`).

The prior workspace regression correctly covered generic forge polls but disabled `retainCurrentPullRequestStatusPoll`, so it did not cover the production GitHub adapter cache. In that lane, a newly retained same-key target could join the old lifecycle’s unresolved cached request, and the old completion then looked up and notified the replacement target by textual key.

Provider-level correction:
- releasing the last retained subscriber expires only that target’s cached/in-flight PR-status read, forcing a replacement lifecycle to issue its own request;
- every PR-status completion captures the exact `GitHubPollTarget` object present when the request starts;
- completion delivery and rescheduling require that exact object to still be the active map entry, so an old same-key completion cannot notify its replacement;
- the generic-forge key/generation fencing and all repository fact/diff/snapshot fixes remain intact.

New deterministic production-path regression uses the actual GitHub retained poll implementation: hold the old GitHub request unresolved, unsubscribe, retain the identical target again, prove a second request starts, resolve the fresh request first, then resolve the stale request and prove only the fresh lifecycle receives one fresh status.

Verification on the exact pushed tree:
- `github-service.test.ts`: 93 passed
- three workspace-git suites: 104 passed
- 197 focused tests total
- `npm run build:server-deps`
- `npm run typecheck:server`
- `npm run lint`
- `npm run format:check`
- repository pre-commit hook: staged format/lint and full all-workspace typecheck

Pushed by normal fast-forward. No daemon, provider, Desktop runtime, restart, or deployment was used.

timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. GitHub fact caching is unrelated to the current fleet reliability invariants. The discussion and commits remain available. Any current rate-limit defect should return as a focused PR from current upstream main.

PR #2834, opened 2026-08-03
fix(server): fence workspace release against agent creation

timigod:
## Summary
- reserve workspace ownership while agent creation, persistence resume, or session import is registering
- release an owned workspace only after the exact finished agent is archived and no live, stored, or pending rival owns it
- route Hub finish cleanup through the atomic `AgentManager` fence
- add deterministic coverage for both race orderings and the same-workspace Hub regression

## Verification
- `npx vitest run src/server/agent/agent-manager.test.ts --bail=1` (152 passed)
- `npx vitest run src/server/hub/daemon-executions.test.ts --bail=1` (11 passed)
- changed Hub regression passed again in isolation after the final assertion update
- pre-existing Hub cleanup test passed in isolation after one later full-file run hit its 30-second timeout
- `npm run build` in `packages/server`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -6433,6 +6433,96 @@ test("archiveAgent persists archivedAt and updatedAt before emitting closed stat
   expect(lifecycles.slice(-2)).toEqual(["idle", "closed"]);
 });
 
+test("does not release a workspace while another agent is registering", async () => {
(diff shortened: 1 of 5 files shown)


timigod:
Superseded by #2840, which contains the leak-free failed-registration repair and the Hub harness wakeup fix, mechanically replayed onto current upstream main with exact verification receipts. The original branch is intentionally preserved.

PR #2818, opened 2026-08-03
fix(cli): bound the WebSocket close handshake

timigod:
Replaces #2771 with a clean, independently reviewable change based on the latest stable v0.2.5 line and verified to merge cleanly with current upstream main.\n\nThis keeps the production behavior but removes the export-only test seam: the bounded close logic now lives in `websocket-close.ts` and is covered with a real `ws` server regression.\n\nScope: 3 files, one commit, no unrelated fleet/runtime changes.\n\nVerification previously run on the exact commit: focused WebSocket-close tests, CLI typecheck, lint, formatting, and diff check.\n\nSupersedes: #2771.

PR #2835, opened 2026-08-03
Fix CLI attach exit on agent completion

timigod:
## Summary
- keep `paseo attach` connected when stale, autonomous, canceled-for-replacement, or other non-current terminal events arrive while authoritative exact-agent state is still running
- verify target terminal events with an exact-agent readback and exit only on authoritative non-attachable state or exact-target removal
- race subscription bootstrap, exact readback, and timeline catch-up against completion; close the daemon client once and always remove stream/update/signal listeners
- suppress expected timeline warnings when Ctrl-C closes an in-flight catch-up request
- use a shared bounded 60-second session-RPC-scale history timeout for both `paseo logs` and `paseo attach`, while retaining caller-specific timeout overrides

## Integration
- merged upstream `origin/main` `777fe968f5bab50f63f25a61188482f6a15789fc` normally in `b940898a`
- repaired head: `5902a93820dd79f76a9154c31875ba9f0eb2131b`
- repaired tree: `9d277a8abfb7873af495e0e17f549e7321aa414a`

## Tests
- `npx vitest run packages/cli/src/commands/agent/attach.test.ts packages/cli/src/commands/agent/logs.test.ts packages/cli/src/utils/timeline.test.ts --bail=1` - 3 files, 19 tests passed
- `npx vitest run packages/server/src/server/agent/agent-manager.test.ts --bail=1 -t "ignores stale autonomous terminals|autonomous events arriving during foreground run"` - 2 passed, 148 skipped
- `npm run build --workspace=@getpaseo/cli` - passed
- `npm run typecheck --workspace=@getpaseo/cli` - passed
- targeted `npm run lint -- ...` - 0 warnings, 0 errors
- `npm run format` and targeted format check - passed
- `git diff --check` - passed

### DIFF
--- a/packages/cli/src/commands/agent/attach.test.ts
+++ b/packages/cli/src/commands/agent/attach.test.ts
@@ -1,0 +1,301 @@
+import type { AgentTimelineItem } from "@getpaseo/protocol/agent-types";
+import type { AgentStreamEventPayload } from "@getpaseo/protocol/messages";
+import { describe, expect, it, vi } from "vitest";
+import {
+  runAttachSession,
+  type AttachAgentState,
+  type AttachAgentUpdate,
+  type AttachSessionClient,
+  type AttachSignal,
+  type AttachSignalSource,
+} from "./attach.js";
+
+const TARGET_ID = "11111111-1111-4111-8111-111111111111";
+const OTHER_ID = "22222222-2222-4222-8222-222222222222";
+
+class FakeAttachClient implements AttachSessionClient {
+  readonly streamListeners = new Set<(agentId: string, event: AgentStreamEventPayload) => void>();
+  readonly updateListeners = new Set<(update: AttachAgentUpdate) => void>();
+  private readback: AttachAgentState | null | Promise<AttachAgentState | null>;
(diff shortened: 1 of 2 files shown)


timigod:
Independent exact-head review at `5902a93820dd79f76a9154c31875ba9f0eb2131b` found the attach/timeline behavior itself sound (19/19 focused CLI, 2/2 server lifecycle, 192/192 isolated CLI tests), but identified a delivery dependency: natural completion and signal detach now await `DaemonClient.close()`, while this base can retain an unacknowledged `ws` close for about 30 seconds. The generic bounded-close repair is already isolated in #2818 (`1c4442ec1c5fd4de1db3d0cc4f78438aae4700c0`). Therefore #2835 should be treated as depending on #2818 and reverified on the combined tree after #2818 lands; duplicating the close fix here would make the PRs less independent. No additional #2835 code finding was identified.

timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. Automatic CLI attach exit is a lower-priority convenience and should not remain as unreviewed queue debt. Preserve the history; submit a small current-main PR if a fresh reproduction still justifies it.

PR #2837, opened 2026-08-03
fix(cli): scope daemon sessions to each process

timigod:
## Summary

- generate one reconnect-stable client ID per CLI OS process instead of persisting one identity across every invocation
- leave legacy `~/.paseo/cli-client-id` files inert and remove the accidentally tracked identity artifact
- cover same-process reuse, concurrent process separation, stale-file immunity, isolated stream fanout, and independent session cleanup

## Verification

- CLI client-ID regressions: 2 passed
- CLI unit suite with isolated `PASEO_HOME`: 175 passed
- client daemon/reconnect suite: 106 passed
- server WebSocket session suite: 22 passed
- `npm run build:server`
- `npm run typecheck:server`
- pre-commit format, lint, and all-workspace typecheck


timigod:
Closing as part of the v0.3.1 contribution-stack consolidation. Process-scoped CLI sessions are useful ergonomics, but they are not required by the current production-backed invariant set. The history remains available; submit a small current-main PR if a fresh reproduction still justifies it.

PR #2841, opened 2026-08-03
Fence client shutdown of service-managed daemons

timigod:
## Why

An externally managed Paseo daemon can currently be stopped by an ordinary client shutdown request. That is unsafe for systemd, launchd, Docker, or another service supervisor: a routine CLI call can take the shared control plane down, and the existing signal fallback can bypass a daemon-side refusal when the socket is unavailable.

## What changes

- Add an explicit generic `PASEO_SERVICE_MANAGED=1` ownership declaration, separate from Desktop-spawn provenance.
- Reject an unqualified shutdown at the daemon enforcement point with structured `shutdown_rejected` / `service_managed` output and emit no lifecycle stop intent.
- Refuse the CLI signal fallback from service-managed pid metadata.
- Preserve unmanaged and Desktop-managed behavior.
- Retain an explicit maintenance route: `paseo daemon stop --service-maintenance`.
- Mark the shipped NixOS and Docker service definitions as service-managed.

## Verification

- service-managed server shutdown: 3/3
- service-managed local daemon CLI: 5/5
- daemon client: 108/108
- Desktop daemon-manager and quit lifecycle: 21/21
- protocol/wire/AOT: 34/34
- server config/environment/pid/lifecycle boundary: 23/23
- CLI supervision/status: 13/13
- server build, workspace typecheck, lint, and formatting: passed

The branch is one focused commit on current upstream `main` (`777fe968f5bab50f63f25a61188482f6a15789fc`). It contains no machine names, local paths, launchd labels, credentials, or private provider configuration.


timigod:
Mixed-version and Desktop-inheritance follow-up is now pushed at `1cf8466491c4d3b79c7252c2a870f37ef94fad19` (tree `102ab66819734755c1efa592e14f505d057c8fdc`).

- A service-managed supervisor treats worker signals without an authorized worker-relayed shutdown as stray and respawns the worker.
- An unauthorized supervisor stop exits `75`, so `Restart=on-failure` service managers recover instead of accepting a clean shutdown from a legacy CLI fallback.
- Authorization remains the session-fenced worker `paseo:shutdown` route used by explicit maintenance.
- Every final Desktop-owned daemon environment deletes `PASEO_SERVICE_MANAGED`, including inherited, login-shell, and shell-resolution-failure paths, while preserving Desktop ownership.

Verification: 259 tests across 24 files passed; all-workspace typecheck, server build, lint, formatting, diff check, and pre-commit hooks passed. Both new regression suites were proven non-vacuous by temporarily neutralizing their production fixes.

The compatibility boundary is truthful: a pre-fence CLI can still signal the supervisor, but it can no longer make a service-managed host remain down when its manager has the documented restart policy. A bundled upgrade must still stop the old daemon before replacing the daemon/CLI bundle; new CLI to an old daemon cannot retroactively create this contract.

timigod:
Superseded by #3081.

#3081 now carries the complete macOS service-ownership state machine as one focused change: persistent LaunchAgent ownership, explicit stop, forced stop, and the related desktop, CLI, and supervisor regressions. It has been rebased onto current upstream `main` at `02c4bc76a6cdd0f0a2665cc85ff4702e6b28f169`; its current head is `062857814fa40068514459de0785743d241eb608` and GitHub reports it mergeable.

Keeping #2841 open would maintain two competing implementations of the same service-managed shutdown boundary. Closing this older partial design in favor of #3081.

PR #2849, opened 2026-08-03
feat(server): support layered daemon config

colonelpanic8:
## Motivation

Today `$PASEO_HOME/config.json` is a single file that the daemon read-modify-writes wholesale whenever settings change. That makes the config impossible to manage externally: anyone keeping their machines in a dotfiles repo (Nix/home-manager, chezmoi, plain git) has to choose between a source-controlled config and a working settings UI. Symlinking the file doesn't work either, since saves are atomic replaces.

This adds a general layering mechanism so most settings can live in a shared, source-controlled file, machine-specific settings can layer on top, and the settings UI still works.

## What changed

Two new optional root-level keys, both fully backward compatible — with neither present, behavior is unchanged:

- **`imports: string[]`** (allowed in any file) — flattened depth-first, later imports win, and each file overrides its own imports. Paths resolve relative to the referencing file; `~/` expands to the home directory. Missing imports and cycles are hard errors naming the importer / full cycle chain.
- **`writeTo: string`** (root file only) — selects which layer receives settings changes. Must name the root or a file in its import graph. Defaults to the root.

A typical layered setup:

```json
{
  "imports": ["~/dotfiles/paseo/shared.json", "./machine.json"],
  "writeTo": "~/dotfiles/paseo/shared.json"
}
```

### Saves write a diff, not the merged config

The key design point: persisting writes only the selected layer's *difference* from the merge of all other layers, recomputed on every save. Writing the merged whole would freeze the other layers' values into the target on the first UI edit, so later upstream/dotfiles changes would silently stop applying. Diffing also means a settings change shows up in `git diff` as exactly the field that changed, and keys that come to match another layer drop back out on the next save.

Before writing, the save re-merges all layers with the candidate content and compares against the desired config. If a later layer shadows a patched field (or a removal targets a value owned by a read-only layer), it throws without writing and names the owning file:

```
[Config] Cannot save config:
  - daemon.mcp.injectIntoAgents is defined in /home/u/.paseo/machine.json, which overrides the writable config file. Change it there instead.
```

This generalizes the existing `relayEnabledMutable` idea ("Relay is controlled by a daemon launch override…") so machine-pinned settings are visibly pinned instead of silently masked.

### Other notes

- Imported files are read without permission enforcement or chmod, so shared layers can live in a Nix store path or a git checkout. The root config keeps the existing `ensurePrivateFile` treatment, and writes stay atomic + private.
- `DaemonConfigStore` moves to the stack-aware API; its rollback path restores the previous write-target bytes.
- `onboard` and `daemon set-password` route through the same write path — `set-password` now reports the file it actually wrote.
- Structural keys (`imports`, `writeTo`, `version`, `$schema`) of the written file are preserved verbatim.
- Documented in `docs/data-model.md`.

## Testing

`packages/server` typecheck, lint, and format are clean; 115 tests pass across the config suites. New coverage includes: unchanged single-file behavior, nested/multiple import precedence, relative and `~` resolution, missing-import and cycle errors, `writeTo` outside the graph and in an imported file, delta-only writes, stale-key dropout on re-save, shadowed-patch rejection with nothing written, structural-key preservation, provider removal from the target vs. a read-only layer, and write-target restoration on a failed field transition.

---

## Part two: app and desktop settings

The daemon config was only one of the places settings live. App-level preferences sit in the renderer's AsyncStorage (localStorage in a LevelDB on Electron desktop — not a file you can manage), and desktop-shell preferences in `desktop-settings.json`, which the app atomically rewrites on every load, so dotfiles managers have to render-and-merge it with `jq` at login. This PR now extends the same layering model to both.

### `settings-seed.json`

A single optional file in the Electron `userData` directory (`~/.config/Paseo/settings-seed.json` on Linux). It is a read-only layer merged **under** the locally persisted values — dotfiles supply defaults, the settings UI keeps writing machine-local deltas on top. The app never writes it, so it can be a plain symlink into a dotfiles repo or a Nix store path; no render-and-merge hoops.

```json
{
  "app": {
    "appSettings": { "theme": "dark" },
    "keyboardShortcutOverrides": { "<binding-id>": "<combo>" },
    "preferredEditor": "zed",
    "changesPreferences": { "layout": "split" },
    "createAgentPreferences": { "provider": "claude" }
  },
  "desktop": {
    "releaseChannel": "stable",
    "daemon": { "manageBuiltInDaemon": true, "keepRunningAfterQuit": false }
  }
}
```

### Layering is opt-in per key

Most of the renderer's storage keys are machine-local state that must not be shared across machines — caches, drafts, window/panel layout, the client id, the daemon registry. Only the five keys above are layerable, declared in `packages/app/src/storage/settings-seed/registry.ts`; every other key passes through the layered storage byte-for-byte untouched. A key has to opt in, and the bar is: a human would author it in a dotfile, it is meaningful across machines, and merged-under semantics cannot corrupt runtime state.

### Same rules as the daemon config

- **Reads** merge the seed under the local value (local wins per field).
- **Saves persist only the diff against the seed**, recomputed on every save — same reasoning as above: writing the merged whole would freeze seed values into local storage on the first UI edit, and later dotfiles changes would silently stop applying. A field changed back to the seed's value drops out of local storage. The stores that used to persist their full defaults on first run no longer do, for the same reason.
- **Removing a seed-provided value fails loudly naming the file**: resetting a seeded keyboard shortcut or clearing a seeded editor throws `[Settings] Cannot remove "<entry>": it is defined in <seed path>. Change it there instead.` The local layer sits on top, so overriding always works; only deletion is unrepresentable.

The `desktop` section does the same for `desktop-settings.json` in the main process, which makes the login-time `jq` merge workaround unnecessary: with a seed present the store persists only the delta against defaults ⊕ seed; with no seed it writes byte-identical files to today.

### Platform behavior

The app is cross-platform and most targets have no filesystem to read. The seed source is a Metro platform split: the Electron build fetches the seed once per page load over a new `get_settings_seed` desktop command; the base module (iOS, Android, plain browser web) returns null unconditionally. With no seed — wrong platform or file absent — every path through the layered storage is a byte-for-byte passthrough, verified by tests.

### Testing (part two)

- `packages/app`: 26 new layered-storage tests (passthrough with no seed on registered and unregistered keys, the layerable/non-layerable boundary, struct/record/scalar merge + delta + dropout, shadow errors naming the seed file, malformed local values and wrong-shaped seeds degrading to passthrough); updated store tests for the removed first-run default writes. Package typecheck and lint clean.
- `packages/desktop`: 20 new tests across the seed reader (absent/invalid/symlinked files), the seed-aware settings store (no-seed byte-compatibility, merge-under, delta persistence and dropout, unreadable seed degrading to defaults), and the `get_settings_seed` command. Package typecheck and lint clean.
- Server config suites re-run after the change: 74 tests in `persisted-config` + `daemon-config-store` plus the CLI `set-password` script test, all passing — the daemon half is untouched.


boudra:

Issue #1931, opened 2026-07-07
feat: run `worktree.setup` (or a restore hook) when a worktree is recreated on unarchive


cleiter:
### Prior search

- [x] I searched existing issues and discussions, and this isn't already proposed.

### What's the problem

Unarchiving a worktree-backed workspace recreates the worktree directory from the kept branch, but it does **not** run the `worktree.setup` commands from `paseo.json`. The restored worktree only contains what was committed to the branch — anything `worktree.setup` produces (installed dependencies, generated env files, symlinks, local config) is missing, so the workspace comes back non-functional until the user re-runs setup by hand.

This is the mirror image of create: `worktree.setup` runs on create and `worktree.teardown` runs on archive, but restore — which is effectively a second create of the same worktree — silently skips setup.

Concretely, archive deletes the worktree working tree (`git worktree remove --force` + `rm -rf`), keeping only the branch. On unarchive, `recreateOwningWorktreeForRestore` re-checks-out that branch into a fresh worktree, hardcoding `runSetup: false`:

- `packages/server/src/server/session.ts:3823-3830` — `createWorktree({ source: { kind: "checkout-branch", branchName: branch }, runSetup: false, ... })`
- Setup commands are `worktree.setup` in `paseo.json`, run by `runWorktreeSetupCommands` only when `runSetup` is true (`packages/server/src/server/utils/worktree.ts`).

The asymmetry means a round-trip (archive → unarchive) is lossy for any project whose worktrees depend on setup-generated, gitignored artifacts — which is the common case for anything with a dependency install or a local `.env` step.

### What would solve it

Run `worktree.setup` when a worktree is recreated on restore — i.e. pass `runSetup: true` in `recreateOwningWorktreeForRestore` (`session.ts:3827`), so a restored worktree reaches the same usable state a freshly-created one does.

Open questions worth a decision before any code:

- **Reuse `worktree.setup` vs. a dedicated restore hook.** Reusing `setup` keeps the config surface small and matches user intent for most projects. A separate `worktree.restore` hook would let setup and restore diverge (e.g. skip a heavy first-time bootstrap on restore), at the cost of another config key. Reusing `setup` seems like the right default.
- **Surfacing progress/failure.** Setup on restore should report the same way create does (progress + a toast on failure), so a slow or failing setup during unarchive is visible rather than a silent partial restore.

### Alternatives you considered

[...]

PR #2852, opened 2026-08-04
Fix checkout version and archived worktree base metadata

yooztech:
## Summary

- resolve daemon versions from an exact checkout release tag before falling back to `package.json`
- restore persisted worktree base metadata when recreating an archived workspace
- cover release-tag version resolution and archive-to-restore checkout status

## Why

Archived worktrees were recreated as a plain branch checkout, which rewrote their base metadata to the feature branch. Checkout status then could not detect changes from the original base branch, hiding the Create PR action. This is related to the incomplete archived-worktree recreation path described in #1931.

Development daemons built from tagged fork or beta checkouts also reported the package version instead of the active release tag.

## Verification

- `npx vitest run packages/server/src/server/session/workspace-recovery/workspace-recovery-service.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`


yooztech:
Opened by mistake; please disregard.

PR #2855, opened 2026-08-04
Scope Paseo tools by provider

hoangnb24:
## Workflow

I run different Codex provider profiles through Paseo for different orchestration roles:

- lead/supervisor profiles coordinate work and need Paseo tools;
- peer profiles perform bounded engineering work and should not be able to create or manage other Paseo agents.

Today, enabling Paseo tools injects them into every launched provider. Profiles such as `codex-lead` and `codex-peer` therefore receive the same orchestration capability even when both extend `codex` for otherwise shared runtime behavior.

The desired flow is to enable Paseo tools globally, then allow them only for exact configured provider IDs.

## What changed

- Add optional `daemon.mcp.injectIntoProviders` exact-provider-ID allowlisting.
- Preserve existing behavior when the field is omitted.
- Treat `[]` as deny-all.
- Support `null` in the mutable patch API to reset the allowlist back to omission.
- Apply one policy predicate to MCP fallback and native Paseo tools across create, resume, import, and reload paths.
- Strip stale daemon-owned `/mcp/agents` entries before policy-controlled reinjection.
- Advertise `providerScopedPaseoTools` through `server_info.features`.
- Prevent new SDK clients from sending provider-policy patches to older daemons that cannot enforce them.
- Update the protocol/config schemas and documentation.

Example:

```json
{
  "daemon": {
    "mcp": {
      "injectIntoAgents": true,
      "injectIntoProviders": ["codex-supervisor", "codex-lead"]
    }
  }
}
```

With this configuration, `codex-peer` does not receive Paseo tools even when it extends the same base `codex` provider.

## Root cause

Paseo tool delivery had only a global `injectIntoAgents` gate. The launch path did not retain a provider-scoped policy decision when constructing either the internal MCP server config or the native tool launch context.

Adding the config field also required two lifecycle protections:

1. mutable configuration needed an explicit reset that removes persisted state and notifies runtime listeners;
2. separately released clients and daemons needed feature negotiation so an old daemon could not accept and echo a policy it does not enforce.

## Compatibility

- Old clients accept the new optional server capability.
- New clients block provider-policy patches before sending them to old daemons.
- Unrelated daemon config patches remain compatible.
- Direct persisted configuration remains backward compatible.
- No Codex runtime fork or provider-specific prompt restriction is used.

## QA evidence

Rebased candidate reviewed at `d082ff3992f8e2df77dfdf984dcba00b9480059c`, based on `df32e6ddcd8dc6785b249aee4d7ba107a2e1655e`.

Focused candidate tests:

```text
messages.wire-compat.test.ts                 7 passed
client/src/index.test.ts                     8 passed
agent-manager.test.ts                      153 passed
config.test.ts                               5 passed
daemon-config-store.test.ts                 19 passed
persisted-config.test.ts                    43 passed
websocket-server.relay-reconnect.test.ts    21 passed

Total: 256 passed
```

Upstream-overlap checks after rebase:

```text
server/session.test.ts                     153 passed
utils/checkout-git.test.ts                 143 passed, 1 skipped
protocol ws-outbound.test.ts                 8 passed
```

Static checks:

```text
npm run build:client                       passed
npm run typecheck:server                   passed
targeted npm run lint                      0 warnings, 0 errors
targeted format check                      passed on 20 files
git diff --check origin/main...HEAD        passed
```

A fresh independent read-only review accepted the exact rebased SHA. `git range-diff` mapped all three pre/post-rebase commits unchanged, and a synthetic merge tree matched the rebased HEAD.

## Platforms tested

- Tested: macOS, deterministic offline/unit and integration-style test clients.
- Not tested: Windows, Linux, real-provider sessions, and the full daemon E2E suite.
- UI screenshots/video: not applicable; this PR has no UI changes.

## Development note

AI assistance was used for implementation and independent review. The scope, architecture decisions, candidate acceptance, and publication were explicitly approved by the repository owner, and the reproducible QA evidence is included above.

## Related context

There is no linked upstream issue or discussion yet. This is opened as a draft for maintainer feedback on the workflow and direction before requesting final review.


boudra:
Superseded by #4277, which lands exact-provider controls across native and MCP catalogs and adds individual tool exclusions. Thank you @hoangnb24 for the workflow and implementation work.

PR #2856, opened 2026-08-04
fix(desktop): declare NSLocalNetworkUsageDescription for macOS direct LAN connections

lvntbkdmr:
## Summary

- Adds `mac.extendInfo.NSLocalNetworkUsageDescription` to `packages/desktop/electron-builder.yml` so the packaged macOS app's `Info.plist` declares local-network usage.

## Problem

Adding a **direct connection** to a daemon's plain LAN IP (e.g. `tcp://192.168.1.155:6767?password=...`) from Paseo Desktop on macOS fails silently:

- No error surfaces in the app.
- Nothing is logged in `~/Library/Logs/Paseo/main.log` or the daemon's own `~/.paseo/daemon.log` — the connection attempt never reaches the target host at all.
- The app never appears under **System Settings → Privacy & Security → Local Network**, and macOS never shows a consent prompt for it.
- This is specific to the macOS desktop app: the same LAN IP is reachable from `curl`/`nc` on the same machine, and both the mobile app and a Tailscale-IP direct connection (`100.64.0.0/10`) work fine.

## Root cause

Captured via the macOS unified log while reproducing:

```
kernel: tcp drop outgoing [<client>:50262<->192.168.1.155:6767] interface: en0
  process: Paseo Helper:55899 t_state: SYN_SENT so_error: 0 reason: NECP
```

The outbound `SYN` is dropped by the kernel (`reason: NECP`) before it ever leaves the machine.

macOS's Local Network privacy boundary only offers a user consent prompt (and only lists the app under Privacy & Security → Local Network) if the app's `Info.plist` declares `NSLocalNetworkUsageDescription`. Checked `packages/desktop/electron-builder.yml` and `packages/desktop/build/entitlements.mac.plist` — neither the key nor the equivalent `mac.extendInfo` entry exists anywhere in the repo. Without it, macOS doesn't fall back to prompting; it silently denies local-network connections at the kernel level and never surfaces the app in Privacy Settings, since the app was never eligible for the consent flow to begin with.

This also explains why non-RFC1918 direct connections (Tailscale IPs) are unaffected — macOS's Local Network address-space check doesn't apply to CGNAT (`100.64.0.0/10`) addresses, so those connections never hit this gate regardless of the missing declaration.

## Fix

Declare the usage description via `electron-builder`'s `mac.extendInfo`, matching how the entitlements/Info.plist are already configured for this target:

```yaml
mac:
  extendInfo:
    NSLocalNetworkUsageDescription: >-
      Paseo connects directly to the Paseo daemon on your local network when you add a
      direct connection by LAN IP address.
```

This should let macOS present the normal one-time consent prompt on first direct-LAN-connection attempt (or at minimum let the app appear under Local Network privacy settings for a manual grant), instead of an unexplained, unlogged connection failure.

## Test plan

- [ ] Build the macOS desktop app with this change and confirm `Info.plist` in the packaged `.app` contains `NSLocalNetworkUsageDescription`.
- [ ] On a clean macOS profile (or after resetting the app's TCC entry via `tccutil reset All sh.paseo.desktop`), add a direct connection to a LAN IP (e.g. `tcp://192.168.x.x:6767`) and confirm macOS shows the Local Network consent prompt.
- [ ] Confirm the app now appears under System Settings → Privacy & Security → Local Network, and that granting it allows the direct LAN connection to succeed.


boudra:
Closing for now: this needs a packaged macOS build showing the Local Network permission flow and a successful direct LAN connection after granting access. Please reopen with that evidence.

boudra:

PR #2862, opened 2026-08-04
fix: show Z.ai quota usage

### LINKED ISSUE #2440: bug: Z.ai provider usage always reports "unavailable"
### What happens

With a Z.ai custom provider configured and working for agents, the provider usage screen always shows **Z.ai — unavailable**. No plan, no usage bars.

### Root cause

`packages/server/src/services/quota-fetcher/providers/zai.ts` resolves the token only from process env:

```ts
const token = process.env["ZAI_API_KEY"] || process.env["GLM_API_KEY"];
if (!token) return unavailableUsage(this);
```

But Z.ai is documented as a **custom provider** ([docs/custom-providers.md](../blob/main/docs/custom-providers.md)), so the token lives in `~/.paseo/config.json`:

```
agents.providers.<id>.env.ANTHROPIC_AUTH_TOKEN   (+ ANTHROPIC_BASE_URL = https://api.z.ai/api/anthropic)
```

The daemon process has neither `ZAI_API_KEY` nor `GLM_API_KEY`, so `fetchUsage()` bails out before making any request. Anyone who followed the custom-provider docs gets `unavailable` forever.

### Steps to reproduce

1. Configure Z.ai as a custom provider following `docs/custom-providers.md` (token in `config.json`, no env vars).
2. Confirm agents run fine on that provider.
3. Open the provider usage screen → **Z.ai: unavailable**.
4. Confirm the daemon environment has neither variable (`ps eww <daemon-pid> | tr ' ' '\n' | grep -E '^(ZAI|GLM)_API_KEY='` returns nothing).

### Second problem behind the first

Even with a token present, the provider only calls `https://api.z.ai/api/biz/subscription/list`, which returns subscription metadata (`productName`, `status`, `purchaseTime`, `valid`) and no usage. The code sets `windows: []`, so the card would render with a plan label and no usage bar.

The usage numbers come from a different endpoint, which does work with the same token:

```
GET https://api.z.ai/api/monitor/usage/quota/limit
-> data.level ("lite") and data.limits[] with { type, unit, number, percentage, nextResetTime }
   TOKENS_LIMIT entries carry the percentage used per window; TIME_LIMIT tracks the billing cycle.
```

### Environment

- Paseo daemon 0.2.1, macOS
- Plan: GLM Coding Lite, provider id `zai`, `extends: claude`

### Note

I have a working fix for the token resolution (env vars keep priority, falling back to the configured custom provider whose `ANTHROPIC_BASE_URL` host is `api.z.ai`) plus the usage windows from the endpoint above, with tests. Happy to open a PR for the token-resolution bug on its own if you prefer to keep it focused — the window rendering and the label/detail choices are behavior changes I'd rather align with you first.

ele-yufo:
Closes #2440.

## Summary

- Resolve Z.ai / GLM credentials from documented custom-provider configuration when daemon-level `ZAI_API_KEY` or `GLM_API_KEY` is absent.
- Read 5-hour, weekly, and monthly MCP quota windows from `/api/monitor/usage/quota/limit` for both `api.z.ai` and `open.bigmodel.cn`.
- Keep the existing Z.ai subscription metadata when available, while treating it as optional so quota remains visible if that endpoint fails.

## Safety and compatibility

- Custom-provider credentials are only used when `ANTHROPIC_BASE_URL` has an exact supported hostname; unrelated provider tokens are ignored.
- Monitor requests use the API's raw-token authorization format, tokens are never logged, and no protocol schema changes are required.
- API-level error envelopes and nullable quota fields are handled explicitly.

## Validation

- `npx vitest run --bail=1 packages/server/src/services/quota-fetcher/providers/zai.test.ts packages/server/src/services/quota-fetcher/service.test.ts` — 65 tests passed.
- `npm run build --workspace=@getpaseo/server` — passed.
- `npm run typecheck` — all workspaces passed.
- Pre-commit lint and format checks — passed.
- Live check against a GLM China Coding Plan returned the expected 5-hour, weekly, and monthly MCP windows. Credential and raw response data were not printed.

Tested on Linux. The app UI and wire protocol are unchanged, so there is no platform-specific client behavior in this patch.

### DIFF
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -111,6 +111,8 @@ See [Codex with a custom OpenAI-compatible endpoint](#codex-with-a-custom-openai
 
 - `ANTHROPIC_AUTH_TOKEN` is used instead of `ANTHROPIC_API_KEY` — this is the z.ai API key
 - The `API_TIMEOUT_MS` env var extends the request timeout (z.ai can be slower than direct Anthropic)
+- Plan usage automatically reads this provider credential and supports both `api.z.ai` and the
+  China-region `open.bigmodel.cn` endpoint
 - If you get auth errors, run `/logout` inside Claude Code before switching to the z.ai provider
 - Web search (`WebSearch` tool) is an Anthropic-only server-side feature — third-party endpoints don't support it. Add `"disallowedTools": ["WebSearch"]` to avoid errors.
 - Automated setup is also available: `npx @z_ai/coding-helper`
--- a/packages/server/src/server/websocket-server.ts
+++ b/packages/server/src/server/websocket-server.ts
@@ -76,6 +76,8 @@ import {
   type WebSocketRuntimeDiagnosticSnapshot,
 } from "./websocket/runtime-metrics.js";
 import { ProviderUsageService } from "../services/quota-fetcher/service.js";
+import type { ProviderUsageAgentProviderConfigs } from "../services/quota-fetcher/provider.js";
+import { loadPersistedConfig } from "./persisted-config.js";
 import { getProcessMemoryDiagnostics, getProcessUptimeSeconds } from "./process-diagnostics.js";
 import {
   CLIENT_SHUTDOWN_RPC_REASON,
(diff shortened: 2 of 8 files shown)


ele-yufo:
Addressed both P2 points from the Greptile review (`4dfa19a55`).

**1. Credential selection ambiguity (P2)**

`resolveZaiCredential` no longer does first-match across persisted profiles. Profile selection is now bound to **provider identity + enabled state**:

- A profile with `enabled: false` is skipped, so quota is never reported for an account the user has turned off.
- Only profiles that resolve to a Z.ai / GLM China identity (`ANTHROPIC_BASE_URL` on `api.z.ai` / `open.bigmodel.cn`) and carry a token are candidates.
- When several enabled profiles target Z.ai, the **lowest configured `order`** wins (stable by profile id as tiebreak) — so the user's primary account is surfaced rather than whichever entry iterates first.

To thread `enabled`/`order` through, `ProviderUsageAgentProviderConfig` gained those fields and the websocket-server call site reads them from the persisted provider overrides (the parsed entries retain `enabled`/`order` even though the narrowed `agents.providers` type hid them).

Added tests: lowest-order wins over first-match across two enabled Z.ai profiles; a disabled profile is skipped even with the lowest order.

**2. Optional subscription metadata delaying quota (P2 note)**

The monitor call is authoritative for availability; subscription only enriches `planLabel`/details. The two now run **concurrently**, and the optional subscription fetch is bounded by a tighter timeout (`4s`, vs the `15s` monitor timeout). A subscription failure or hang degrades to monitor-only data instead of blocking/delaying the result — fast failures add ~0 latency (they overlap the monitor round-trip), and a hang is bounded rather than propagating the full monitor timeout. Early-return paths no longer await the optional metadata.

Added tests: quota stays available with monitor data when subscription throws; plan label falls back to the monitor `level` when subscription has no product name; a hanging subscription endpoint does not block the result (resolves via the bounded timeout).

`pnpm test:unit` for `quota-fetcher/` + the websocket-server / persisted-config suites: 166 passed.

ele-yufo:
Closing — the related provider-usage visibility PR (#2863) was closed pending a Discussions-first proposal, so this one is not worth keeping open on its own. Happy to revisit if there's interest.

PR #2863, opened 2026-08-04
feat: hide unused usage providers

ele-yufo:
Links https://github.com/getpaseo/paseo/discussions/2861

## Summary
- add host-scoped Shown providers switches to Settings > Usage
- persist hidden provider IDs in daemon.providerUsage.hiddenProviders
- capability-gate the UI for mixed client/daemon versions
- keep the active-provider context tooltip unchanged

Missing or empty configuration shows every provider, and unknown hidden IDs are preserved when settings change.

## Compatibility
- new protocol fields are optional
- older daemons retain the existing Usage UI
- no provider credential, quota-fetching, or relay behavior changes

## Validation
- 82 focused Vitest tests passed
- full workspace typecheck passed
- server build, targeted lint, and formatting passed
- Playwright provider-usage suite: 4/4 passed
- browser UI inspected at desktop and responsive widths; no platform-specific UI code added

## Screenshot
Provider usage visibility settings

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -199,6 +199,7 @@ snapshot so a mixed edit can apply its live subset and still name the paths that
     hostnames: true | string[],   // legacy alias `allowedHosts` is migrated on load
     trustedProxies: true | string[], // defaults to ["loopback"]; Express proxy names/CIDRs
     mcp: { enabled: boolean, injectIntoAgents: boolean },
+    providerUsage: { hiddenProviders: string[] }, // host-scoped provider cards hidden from Usage
     git: { maxProcessesPerSecond: number, maxProcessConcurrency: number },
     appendSystemPrompt: string,    // appended to supported provider system/developer prompts
     terminalProfiles: TerminalProfile[],  // named shell commands; omitted means DEFAULT_TERMINAL_PROFILES
@@ -313,6 +314,10 @@ session is created, resumed, imported, or reloaded, so configuration changes aff
 session rather than an already-running one.
 
 `agents.metadataGeneration.providers` controls the preferred structured-generation fallback order for daemon-side metadata tasks such as commit messages, PR text, branch names, and generated agent titles. Entries are tried first in the configured order, then Paseo falls through to dynamically discovered defaults and finally the current selection when available.
+
+`daemon.providerUsage.hiddenProviders` stores provider IDs hidden from that host's Usage page. An
+empty or missing list shows every provider. This preference does not hide the active provider from
+the context-window tooltip.
 
 ### Git process limits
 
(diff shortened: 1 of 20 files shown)


ele-yufo:
Pushed `aa7f80250` (rebased onto `main`, branch was 12 commits behind). Addresses both P1 items from the Greptile review.

## P1 #1 — persistence breaks daemon rollback
`DaemonConfigStore` now writes `daemon.providerUsage` to `config.json` **only when at least one provider is hidden**, and strips any stale block when the list becomes empty (`packages/server/src/server/daemon-config-store.ts`, `mergeMutableConfigIntoPersistedConfig`). The daemon object schema stays strict; an older daemon that does not yet know `providerUsage` can still load a config written by this daemon after a no-op upgrade/rollback — the default state no longer pollutes the file. (The persisted shape when non-empty is unchanged, so the v1 JSON schema and `data-model.md` stay accurate.)

## P1 #2 — concurrent clients overwrite each other
Root cause: clients computed a **full-array** `hiddenProviders` patch from their own stale snapshot, and the server's `deepMerge` replaces arrays wholesale → last writer wins. Fix has two halves:
- **Client** (`packages/app/src/provider-usage/visibility.ts`) now emits a single-provider delta — `{ showProviders: [id] }` / `{ hideProviders: [id] }` — and no longer reads the local snapshot to build the patch.
- **Server** (`DaemonConfigStore.patch`) resolves each delta against its authoritative current set inside the synchronous read-modify-write (`resolveProviderUsagePatch`), so sequential patches from concurrent clients each observe the prior result. Toggling different providers can no longer silently undo each other.

The wire patch schema (`packages/protocol/src/messages.ts`) keeps `hiddenProviders` (full replace) for back-compat and stays `.passthrough()`/permissive, and adds `hideProviders` / `showProviders`.

## Tests
- `daemon-config-store.test.ts`: no `providerUsage` key on disk at default / when all unhidden (rollback contract); concurrent `hide`/`hide`/`show` deltas merge without lost updates.
- `messages.browser-automation.test.ts`: delta fields parse permissively.
- `visibility.test.ts`: client emits single-provider deltas.

Verified locally: protocol (494), targeted server `daemon-config-store`/`persisted-config`/`config-provider-usage` (67, +3 new), app `provider-usage` (4), client `daemon-client` (118), plus full-workspace `typecheck` / `oxlint` / `oxfmt` via the pre-commit hook. The Playwright `provider-usage-settings` e2e was not run locally (needs full app+daemon+browser harness); its assertions are end-state/RPC-based and remain compatible with the delta + conditional-persist changes.

cc @greptileai

boudra:

PR #2875, opened 2026-08-04
fix(app): detect Paseo worktrees from daemon data, not path convention

colonelpanic8:
## Problem

Three utilities in `packages/app` decided "is this cwd a Paseo-managed worktree?" by string-matching the literal path segment `.paseo/worktrees`:

- `agent-grouping.ts` — derives the parent repo from a worktree cwd for sidebar grouping
- `new-agent-routing.ts`
- `agent-working-directory-suggestions.ts`

This is wrong under a custom `worktrees.root`, which is a supported setting today. **It is also wrong under the default.**

The daemon creates worktrees at `<worktreesRoot>/<project-hash>/<slug>`, where the default root is `$PASEO_HOME/worktrees` — i.e. `~/.paseo/worktrees/...`, not repo-local. So `deriveProjectKey("~/.paseo/worktrees/<hash>/<slug>")` walked back to the segment before `.paseo/worktrees` and returned `/home/<user>`, collapsing **every** worktree agent across **every** repo into one bogus project group. The marker only ever made sense for a repo-local layout the daemon no longer uses.

And no path convention can fix it: the project hash is one-way, so the parent repo is not recoverable from the path. That fact drove the design.

## Change

Clients now get the answer from the daemon instead of guessing.

- **Protocol:** new optional `server_info.worktreesRoot`, the daemon's own resolved worktrees base root. Additive `z.string().optional()`, no transform, nothing narrowed or removed. Old daemons omit it; old clients ignore it via `.passthrough()`. Presence of the value *is* the capability signal, so no paired `features` flag.
- **App:** one shared helper `packages/app/src/utils/paseo-worktree-path.ts` matches `<worktreesRoot>/<hash>/<slug>`. It reports `mainRepoRoot: null` under the current layout rather than fabricating a repo path — only the daemon can answer that.
- Call sites prefer daemon-computed data where it exists: `agent.projectPlacement.projectKey` for grouping, `checkout.mainRepoRoot` for "which repo was this cut from". `resolveNewAgentWorkingDir` without a checkout now stays in `cwd` instead of guessing.

The legacy `.paseo/worktrees` branch survives as the single fallback for pre-`worktreesRoot` daemons, tagged `// COMPAT(worktreesRoot): added in v0.3.0, remove after 2027-08-04` at the constant, the function, the schema, and each test block covering it.

Separators are normalized 1:1 and slices come off the original string, so Windows paths stay Windows paths.

## Evidence

```
$ NODE_ENV=test npx vitest run packages/app/src/utils/paseo-worktree-path.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  13 passed (13)

$ NODE_ENV=test npx vitest run packages/app/src/utils/agent-grouping.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  12 passed (12)

$ NODE_ENV=test npx vitest run packages/app/src/utils/new-agent-routing.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  13 passed (13)

$ NODE_ENV=test npx vitest run packages/app/src/utils/agent-working-directory-suggestions.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  4 passed (4)

$ NODE_ENV=test npx vitest run packages/protocol/src/messages.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  21 passed (21)
```

The 13 new `paseo-worktree-path` cases cover: default root, custom root, nested cwd, trailing separator, prefix-sibling rejection (`/root-other` must not match `/root`), base-root and per-project-container rejection, Windows paths, and the legacy fallback. `websocket-server.relay-reconnect.test.ts` gained assertions that the emitted `server_info` carries the right root under both default and custom config.

`npm run build:client`, `npm run build:server`, `npm run typecheck`, `npm run lint`, `npm run format` all clean.

## Platforms

Pure logic in shared (non-platform-gated) modules plus one optional protocol field.

| Platform | Tested | Notes |
| --- | --- | --- |
| Desktop Linux | ✅ | Suites above |
| Web / iOS / Android / macOS / Windows | — | No platform-gated code paths touched; Windows path handling is unit-covered explicitly |

Old-daemon/new-app and new-daemon/old-app are both covered by the field being optional plus the COMPAT fallback.

## Two things for you to decide

1. **`groupAgents`, `resolveNewAgentWorkingDir`, and `collectAgentWorkingDirectorySuggestions` have no production callers** — only tests. The one live path is `deriveProjectKey` → `deriveProjectPlacementFromCwd` → `resolveProjectPlacement` in `agent-directory-sync.ts`, the fallback for an agent that arrives without a `projectPlacement`. I fixed all three as asked rather than deleting them, but if they are dead they should probably go; happy to strip them in a follow-up.
2. **`DaemonServerInfo.worktreesRoot` is required (`string | null`)** in the app-side type, so every producer must decide — that cost 17 one-line fixture edits. Optional would have been quieter but lets a caller silently skip it. Say the word if you'd rather have the quiet version.

Also noticed, pre-existing and untouched: `ServerInfoStatusPayloadSchema` already carries a `.transform()`, which the wire-purity rule in `docs/protocol-compatibility.md` forbids.

Context: part of a small series toward a dotfile-friendly config/state layout. Independent of the others — merges on its own.


boudra:
Closing for now: this needs a before/after repro through the current app showing incorrect worktree grouping and its correction with a custom worktrees.root. Please reopen with that evidence.

boudra:

PR #2878, opened 2026-08-05
Preview CSV and TSV files as a sortable table

nickmaglowsch:
### Linked issue

None. There's no issue or discussion for this one — I hit it in my own workflow and it's a small, contained addition to the file pane. Closing it as "not the direction" is a fine outcome; I'd rather you see the shape first.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Agents produce CSVs constantly — benchmark runs, exported query results, generated fixtures, `usage.csv` from a script in the workspace. Opening one in the file pane today gives highlighted plain text, so past three or four columns you're counting commas to work out which value belongs to which column, and you can't answer "which row has the worst pass rate" without leaving Paseo for a spreadsheet.

This renders `.csv` and `.tsv` as a table in Preview, with `Source` still one toggle away, using the same Preview/Source chrome Markdown and HTML previews already use. Each column header sorts and carries a filter box, so the two things you actually do with a data file — order it and narrow it — happen in the pane.

### Goals

- Render `.csv` and `.tsv` through the existing `Preview`/`Source` toggle; `Source` keeps the raw text and the editor.
- Parse per RFC 4180: quoted cells keep their delimiters and newlines, `""` unescapes, CRLF and a trailing newline are handled, ragged rows pad to the widest row.
- Detect the delimiter from the header row, so semicolon- and pipe-separated exports work without another extension.
- Sort per column: ascending, descending, off. Numeric when both cells are numbers, case-insensitive for text, blanks last in both directions, ties keep file order.
- Filter per column, ANDed across columns, with the surviving row count visible.
- Keep a large export scrollable — virtualized rows, no layout pass over a file's worth of cells.
- Follow live file updates, the same way the Markdown and HTML previews do.

### Non-goals

- No editing through the table. `Source` is the editor; the table is a viewer.
- No `.xlsx`/`.docx`/spreadsheet-binary support, and no new dependency — the parser is ~60 lines.
- No column resize, column hide, multi-column sort, or aggregate row.
- No change to Markdown, HTML, image, or code previews beyond moving the three renderable branches of `FilePreviewBody` into one `RenderedTextPreview` component (the `complexity` lint rule caps that function, and a fourth branch went over it).

### QA

Ran on macOS 15 (Darwin 25.4), against a real daemon and a real Chromium through the app's Playwright harness.

**New Playwright cases — real browser, real daemon, real file on disk**

```
$ npm run test:e2e --workspace=@getpaseo/app -- file-editing.spec.ts \
    -g "previews a CSV as a table|header-only CSV|resizes table columns|loses the column"

Running 4 tests using 1 worker
  ✓  1 [browser] › file-editing.spec.ts:516:7 › previews a CSV as a table that sorts, filters, and follows the file (4.1s)
  ✓  2 [browser] › file-editing.spec.ts:590:7 › survives a CSV that loses the column it was sorted and filtered by (3.4s)
  ✓  3 [browser] › file-editing.spec.ts:613:7 › resizes table columns when the code font size changes (3.4s)
  ✓  4 [browser] › file-editing.spec.ts:641:7 › names a header-only CSV empty rather than filtered (3.0s)
  4 passed (21.9s)
```

The first writes `metrics.csv` in the workspace, opens it from the explorer, and asserts: the parsed row order, that `runs` sorts `9` before `10` (numeric, not lexicographic) with the blank cell last in both directions, that a third click clears the sort and returns file order, that the accessible header label reads `runs, sorted ascending` / `runs, sorted descending`, that a column filter narrows the rows and the footer reads `1 of 3 rows`, that a filter matching nothing shows `No matching rows`, that a row appended on disk appears without reopening the file, and that `Source` shows the raw CSV text in the editor and Preview brings the table back.

The second sorts and filters a column, then rewrites the file without it. That used to crash the pane — `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` — because sort and filter state outlives the parse; the test reproduces that on the pre-fix commit.

The third raises Code font size in Appearance settings and asserts the header cell gets wider, because column widths and row height are derived from a token that gets patched at runtime rather than through a React render. It fails on the commit before the fix — `Expected: > 97, Received: 97` — and passes after it, 97px to 140px.

The fourth opens a header-only file and asserts it reads `No rows` with a `0 rows` count — an empty file and a filtered-to-nothing file say different things.

**Unit tests**

```
$ npx vitest run packages/app/src/file-pane/table/model.test.ts \
    packages/app/src/components/file-pane-render-mode.test.ts \
    packages/app/src/i18n/resources.test.ts

 Test Files  3 passed (3)
      Tests  69 passed (69)
```

`model.test.ts` covers the parser (quoted delimiters and newlines, doubled quotes, tab/semicolon/pipe detection, a quoted header cell containing the delimiter, `.tsv` forcing tabs over a comma in the header, rows of only empty cells surviving while blank lines are dropped, ragged rows, CRLF, blank content, header-only) and the view model (case-insensitive substring filters, ANDed columns, whitespace-only filter ignored, numeric ordering, case-insensitive text ordering, blanks last both ways, stable ties, no mutation of the source rows, and the sort cycle).

**Checks**

```
$ npm run typecheck        # 0 errors
$ npm run lint             # Found 0 warnings and 0 errors (3213 files)
$ npm run format:check     # all matched files use the correct format
```

**Regression pass over the rest of the file pane**

`npm run test:e2e --workspace=@getpaseo/app -- file-editing.spec.ts` → 10 passed, 5 failed. The same 5 fail on a clean `origin/main` checkout on this machine (`git stash` + rerun, same 5), and they all fail on CodeMirror keybindings that differ on macOS — `Control+A` and `Control+Home` don't select-all / jump-to-start here, so the helper types instead of replacing. Not touched by this PR; the HTML preview, image preview, Vim-mode, and unsaved-draft cases all pass.

**Screenshots** — desktop web at 1280px, then compact at 430px, light and dark:

| Default | Sorted `runs` descending |
| --- | --- |
|  |  |

| Two column filters | Filter matches nothing |
| --- | --- |
|  |  |

| Compact, light | Compact, dark |
| --- | --- |
|  |  |

Descending `runs` reads `1042, 512, 256, 128, 77, 64, 15, 9, 3` with the empty cell last, which is the point of sorting numerically. Note the blank `pass rate` and `runs` cells stay at the bottom either way.

**Video of the whole interaction** (open → sort → filter → empty → clear → compact):

CSV table preview interaction

Source `.webm` if the GIF is too lossy: [csv-table-preview.webm](https://raw.githubusercontent.com/nickmaglowsch/paseo/qa/csv-table-preview/csv-table-preview.webm). Both live on a `qa/csv-table-preview` branch in my fork, not in this diff.

**Platform matrix**

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | ❌ | No simulator on this machine. See the native note below. |
| Android | ❌ | No emulator on this machine. See the native note below. |
| Web | ✅ | Chromium via the app Playwright harness, real daemon, light and dark, 1280px and 430px. |
| Desktop macOS | ⚠️ | Not launched separately. Same renderer and same code path as browser web; this change has no Electron-specific branch. |
| Desktop Windows | ❌ | No machine. |
| Desktop Linux | ❌ | No machine. |

**What I did about native without a device**

Two things in this feature could only break on native, so I kept them off the risky path rather than guess:

- The filter box does **not** use `FormTextInput`. On a compact native layout that primitive renders `@gorhom/bottom-sheet`'s `BottomSheetTextInput`, which is only valid inside a sheet — this row lives in a pane. The cell takes its chrome from `createControlGeometry` directly, so it stays on the same field geometry tokens without the sheet coupling. Comment at `packages/app/src/file-pane/table/view.tsx`.
- Vertical rows go through a `FlatList` nested in a **horizontal** `ScrollView`, so the orientations differ and `nestedScrollEnabled` is set. Row height is fixed and fed to `getItemLayout`.

If you'd rather someone with a device confirm the compact filter row and nested scroll before this lands, that's a reasonable gate and I'd expect it.

### Review notes

Two independent reviewers went over this before I opened it, and six findings turned into the three commits after the first:

- A `.tsv` whose header holds `last, first` has as many commas as tabs, and counting alone split the name in half. An extension that names its delimiter now wins whenever that delimiter appears in the header.
- Rows of only empty cells (`,` and `""`) were dropped along with blank lines, because both parse to empty strings. The reader now tracks whether a line said anything, so those rows survive and a blank line still does not.
- Column widths and the row height came from a `UnistylesRuntime` read, which does not re-render when appearance settings patch the code font size. The grid sits behind `AppearanceStyleBoundary` now; sort and filter state stays outside it, and the filter box seeds from that state so a remount under an active filter keeps its term. There's a Playwright case for this that fails on the pre-fix commit.
- A header-only file said `No matching rows` when nothing had been filtered.

- Sort and filter state outlived the parse, so a live file that lost the column being sorted or filtered on crashed the pane on the next render. `tableViewWithinColumns` resolves the requested view against the columns the file currently has, which also drops a filter whose box is no longer on screen to clear.
- A `.tsv` whose header held no tab still fell through to comma sniffing, so a single-column TSV of prose split on its commas. An extension that names its delimiter is now taken at its word.

One finding I argued instead of fixing: the parse walks the whole file on the UI thread. The daemon read already decodes the entire file into one string before any preview runs, and the code view this replaces then calls `highlightCode()` over all of it and renders one non-virtualized `View` per line. For the same large CSV the table does strictly less work than what ships today. Chunked or worker parsing belongs to the file-read path for every preview kind, not to this feature.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### Notes for review

- Column widths are estimated from the header plus the first 200 rows rather than measured, so a wider row later truncates instead of widening its column. That bounds the cost on a large file; the ceiling is a comment at the constant.
- Sorting deliberately avoids `localeCompare`: Hermes, JSC, and V8 disagree, and the same file has to sort the same on every platform.
- The row-count footer is right-aligned because the compact layout floats a control over the pane's bottom-left corner.
- Blank records are dropped while parsing (a trailing newline, a CRLF CRLF ending, a stray gap). Row indexes are view positions, not file line numbers, and nothing claims otherwise.
- Maintainer edits are enabled.

### DIFF
--- a/packages/app/e2e/browser/file-editing.spec.ts
+++ b/packages/app/e2e/browser/file-editing.spec.ts
@@ -1,11 +1,14 @@
 import { mkdir, readFile, writeFile } from "node:fs/promises";
 import path from "node:path";
+import type { Locator } from "@playwright/test";
 import { expect, test, type Page } from "../support/fixtures";
 import {
   openFileExplorer,
   openFileFromExplorer,
   expectFileTabOpen,
 } from "../support/helpers/file-explorer";
+import { openSettings } from "../support/helpers/app";
+import { openSettingsSection, clickSettingsBackToWorkspace } from "../support/helpers/settings";
 import { installDaemonWebSocketGate } from "../support/helpers/daemon-websocket-gate";
 import { openAgentRoute, seedMockAgentWorkspace } from "../support/helpers/mock-agent";
 
@@ -49,6 +52,11 @@ async function openWorkspaceFile(page: Page, filename: string): Promise<void> {
   if (!(await tree.isVisible())) await openFileExplorer(page);
   await openFileFromExplorer(page, filename);
   await expectFileTabOpen(page, filename);
+}
+
+async function columnWidth(header: Locator): Promise<number> {
+  const box = await header.boundingBox();
+  return box ? box.width : 0;
 }
 
 function tableRows(page: Page): Promise<string[][]> {
@@ -577,6 +585,34 @@ test.describe("CodeMirror workspace file editing", () => {
     await expect(editor(page)).toContainText("katherine,42");
     await selectFileView(page, "Preview");
     await expect(page.getByTestId("file-table-preview")).toBeVisible();
+  });
+
+  test("resizes table columns when the code font size changes", async ({ page, withWorkspace }) => {
+    test.setTimeout(120_000);
+    const workspace = await withWorkspace({ prefix: "file-editing-table-appearance-" });
+    await writeFile(
+      path.join(workspace.repoPath, "sizes.csv"),
+      "region,owner\nus-east,ada\n",
+      "utf8",
+    );
+    await workspace.navigateTo();
+    await openWorkspaceFile(page, "sizes.csv");
+
+    const header = page.getByTestId("file-table-sort-0");
+    await expect(header).toBeVisible();
+    const initialWidth = await columnWidth(header);
+
+    await openSettings(page);
+    await openSettingsSection(page, "appearance");
+    const codeSize = page.getByLabel("Code font size");
+    await codeSize.fill("22");
+    await codeSize.press("Enter");
+    await clickSettingsBackToWorkspace(page);
+
+    // Column widths and row height are derived from the code font size, and that
+    // token is patched at runtime rather than through a React render.
+    await expect(header).toBeVisible();
+    await expect.poll(() => columnWidth(header)).toBeGreaterThan(initialWidth);
   });
 
   test("names a header-only CSV empty rather than filtered", async ({ page, withWorkspace }) => {
--- a/packages/app/e2e/browser/file-editing.spec.ts
+++ b/packages/app/e2e/browser/file-editing.spec.ts
@@ -573,8 +573,23 @@ test.describe("CodeMirror workspace file editing", () => {
     await selectFileView(page, "Source");
     await expect(page.getByTestId("file-source-editor")).toBeVisible();
     await expect(page.getByTestId("file-table-preview")).toHaveCount(0);
+    await expect(editor(page)).toContainText("name,runs");
+    await expect(editor(page)).toContainText("katherine,42");
     await selectFileView(page, "Preview");
     await expect(page.getByTestId("file-table-preview")).toBeVisible();
+  });
+
+  test("names a header-only CSV empty rather than filtered", async ({ page, withWorkspace }) => {
+    test.setTimeout(90_000);
+    const workspace = await withWorkspace({ prefix: "file-editing-table-empty-" });
+    await writeFile(path.join(workspace.repoPath, "headers.csv"), "region,owner\n", "utf8");
+    await workspace.navigateTo();
+    await openWorkspaceFile(page, "headers.csv");
+
+    await expect(page.getByTestId("file-table-preview")).toBeVisible();
+    await expect(page.getByTestId("file-table-sort-0")).toHaveText("region");
+    await expect(page.getByText("No rows", { exact: true })).toBeVisible();
+    await expect(page.getByTestId("file-table-row-count")).toHaveText("0 rows");
   });
 
   test("runs inline scripts without allowing fetch egress", async ({ page, withWorkspace }) => {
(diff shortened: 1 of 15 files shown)


boudra:

PR #2879, opened 2026-08-05
fix(server): finish-watcher wakeups — closed children, response cap, repeat finishes

thomasvan:
### Linked issue

No linked issue — found while running long-lived agent-to-agent delegation (a caller agent watching children it spawned via MCP). Happy to open one if you'd like it tracked separately.

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [x] Refactor
- [ ] Docs

### Reasoning

Three problems in `setupFinishNotification`, all hit by the same workload: an agent that delegates to children and waits on their finish notifications.

**A closed child never wakes its caller.** When an agent creates a child with `create_agent` and waits, killing or archiving that child from the UI (or `paseo kill`) sends nothing: the `closed` lifecycle is treated as terminal bookkeeping (`fired = true` + unsubscribe). The calling agent waits forever on a dead delegation and the user sees a silently stalled orchestration.

**A verbose child can blow out the caller's context.** The notification embeds the child's entire last assistant message, so a long research dump lands whole in the caller's next prompt.

**The watcher goes silent after the first finish.** It disarms on notify, so a child that takes several turns — because the caller prompts it again, or because it orchestrates children of its own — reports its first turn and nothing after. The caller has no way to hear about the rest short of polling `get_agent_activity`.

### Goals

- Fire a `"was closed"` finish notification when a watched child is killed/closed before finishing, through the same `notifySafely` path as `finished`/`errored`/`needs permission`.
- Cap the embedded child response at 4000 chars, with a pointer to `get_agent_activity` for the full text.
- Keep the watcher armed across finishes so a caller hears about every turn of a long-lived child.
- Cover all three with unit tests.

### Behavior change to call out

The watcher now re-notifies on every finish instead of once. For the common case — spawn a child, it finishes once — nothing changes. It differs only for children that run multiple turns, which previously went unreported after the first.

Two conditions disarm it, so it can't leak or spam indefinitely:

- the child closes (terminal — the final `"was closed"` notification still fires);
- the caller is archived (it can never hear us again).

Permission requests are deliberately excluded from the re-arm: they fire mid-turn while the child is still running, so re-arming there would swallow the turn's own finish.

An earlier revision of this PR gated the behavior behind a `notifyMode: "once" | "each"` parameter threaded through `create_agent`, `send_prompt`, and the MCP tool schemas. The last commit removes it. Watchers only exist for agent callers, so there was no caller that wanted the old silence — the parameter was a knob with one sensible setting, and dropping it returns `create.ts`, `paseo-tools.ts`, and `agent-prompt.test.ts` to byte-identical with `main`. Net diff is now one source file plus one test file.

### Non-goals

- No change to notification transport, permission flow, or the other lifecycle reasons.
- No config knob for the cap — fixed limit, callers can always fetch the full response via `get_agent_activity`.
- No new MCP tool surface. The tool schemas are untouched by the final state of this PR.

### QA

- Repro before: caller agent watching a spawned child; `paseo kill <child>` mid-run → caller never receives a wakeup and stays blocked. After: caller is prompted with `Agent <id> (<title>) was closed.` Verified live on a macOS daemon with CLI monitors watching the caller's prompt stream, across kill-mid-run and archive cases.
- Truncation: child finishing with a >4000-char final message wakes the caller with the capped message plus `[truncated N chars — use get_agent_activity for the full response]`; full text still retrievable via `get_agent_activity`.
- Re-arm: a child driven through two finish cycles produces two caller notifications and the subscription stays live; a child whose caller is archived produces none and unsubscribes.
- `npx vitest run packages/server/src/server/agent/agent-prompt.test.ts packages/server/src/server/agent/agent-prompt.finish-notification.test.ts` → 11 passed (7 existing + 4 new). Run without `--bail=1` on purpose: bailing early masked a harness bug where the caller resolved as neither a live nor a stored agent, which `sendPromptToAgent` → `ensureAgentLoaded` rejects.
- New scenarios live in `agent-prompt.finish-notification.test.ts` rather than `agent-prompt.test.ts`, because they need a caller that resolves as a live agent. The existing test file is unchanged.
- This patch has been running for several days on a fork driving a multi-agent supervisor/lead/peer setup with delegation wakeups as the core loop.

### Checklist

- [x] One focused change — three commits, one subsystem (the finish watcher); happy to split if you'd rather review them separately
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/agent-prompt.finish-notification.test.ts
+++ b/packages/server/src/server/agent/agent-prompt.finish-notification.test.ts
@@ -1,0 +1,177 @@
+// Finish-watcher coverage: a closed child wakes its caller, oversized child
+// responses are capped, and the watcher re-arms across finishes. Separate
+// from agent-prompt.test.ts because these scenarios need a caller that
+// resolves as a live agent (notify() delivers through sendPromptToAgent).
+import { expect, test } from "vitest";
+
+import { createTestLogger } from "../../test-utils/test-logger.js";
+import { AgentManager } from "./agent-manager.js";
+import { AgentStorage } from "./agent-storage.js";
+import { formatSystemNotificationPrompt, setupFinishNotification } from "./agent-prompt.js";
+import type { AgentManagerEvent, ManagedAgent } from "./agent-manager.js";
+
+interface FinishWatcherScenarioOptions {
+  childLastAssistantMessage?: string | null;
+  callerArchivedAt?: string | null;
+}
+
+interface FinishWatcherScenario {
+  startWatchingChild(): void;
+  finishChild(): void;
+  closeChild(): void;
+  flush(): Promise<void>;
+  parentPrompts: string[];
+  isSubscribed(): boolean;
+}
+
+function createFinishWatcherScenario(
+  options?: FinishWatcherScenarioOptions,
+): FinishWatcherScenario {
+  let subscriber: ((event: AgentManagerEvent) => void) | null = null;
+  const parentPrompts: string[] = [];
+
+  const childAgent: ManagedAgent = Object.create(null);
+  Reflect.set(childAgent, "id", "child-agent");
+  Reflect.set(childAgent, "lifecycle", "idle");
+  Reflect.set(childAgent, "config", { title: "Child Agent" });
+
+  // The caller must resolve as a live agent: notify() delivers through
+  // sendPromptToAgent → ensureAgentLoaded, which throws if the caller is
+  // neither live nor in storage.
+  const callerAgent: ManagedAgent = Object.create(null);
+  Reflect.set(callerAgent, "id", "caller-agent");
+  Reflect.set(callerAgent, "lifecycle", "idle");
+  Reflect.set(callerAgent, "config", { title: "Caller Agent" });
+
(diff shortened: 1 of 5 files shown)


wilgon456:
I reproduced the closed-child wakeup failure on Paseo 0.3.1 on macOS:

- an agent-scoped delegated child completed work in an isolated worktree;
- PR-merge cleanup archived the worktree and the child transitioned to `closed`;
- the parent received no completion callback and stayed asleep until the user manually asked for the result;
- no `Failed to notify caller agent` record appeared because current `main` handles `closed` by calling `stop()` without attempting delivery.

This validates problem A from this PR against a real orchestration run. Since #3177 changed the same watcher and this branch now conflicts with `main`, I opened a focused successor containing only the `running -> closed` wakeup and its regression test: #3192.

I preserved Thong Van's authorship with a `Co-authored-by` trailer and linked this PR in the new PR body. Response truncation and repeated-finish behavior remain scoped to this PR.

boudra:
Thanks for this. The finish-watcher work is superseded by https://github.com/getpaseo/paseo/pull/3192, which carries the closed-child wakeup and response truncation onto current main with your authorship preserved, so I am closing this one.

PR #2885, opened 2026-08-05
Stop shipping a hardened runtime on unsigned mac builds

colonelpanic8:
Reported by a user: the arm64 DMG is ad-hoc signed but carries the hardened runtime flag, so dyld library validation refuses to load Electron Framework with a "different Team IDs" error and the app dies instantly on launch. The only recovery is `codesign --force --deep --sign - /Applications/Paseo.app` after install, which no downloader should have to run.

## Cause

An ad-hoc signature has no Team ID. The hardened runtime turns on dyld library validation, which requires every loaded library to share the main executable's Team ID. Together they cannot work.

`electron-builder.yml` set `hardenedRuntime: true` unconditionally. Official CI passes a real certificate via `secrets.APPLE_CERTIFICATE`, so it is correct there — but any build without credentials still gets an ad-hoc signature (arm64 macOS refuses to run a fully unsigned binary), and ships the broken combination. That covers fork CI, where the Apple secrets are empty, and a local `npm run build:desktop` on a Mac without certificates.

This is already known in the repo, incidentally: `nix/desktop-package.nix` works around it by hand with `--config.mac.hardenedRuntime=false --config.mac.notarize=false` and a comment explaining that an unsigned build cannot satisfy a hardened runtime. It was fixed there instead of at the source.

## The change

YAML cannot express the condition, so `electron-builder.yml` stays the source of truth and a thin `electron-builder.config.cjs` reads it and applies one override when no signing identity is available.

| Environment | `hardenedRuntime` |
| --- | --- |
| Official CI (`CSC_LINK` set) | `true` — unchanged |
| Fork CI (`CI=1`, no certs) | `false` |
| Local Mac, keychain identity | `true` — unchanged |
| `CSC_IDENTITY_AUTO_DISCOVERY=false` | `false` |

Notarization goes off with it, since it requires the hardened runtime and a Developer ID and could only fail.

Fixing this in the build config rather than in the workflow means every invocation path benefits — fork CI, the Nix build, and local builds — instead of only the one workflow.

### Why not the entitlement

The other fix is a `com.apple.security.cs.disable-library-validation` entitlement. It clears the same crash, but by weakening the *signed* release for every user: it permits loading third-party libraries into the app. Dropping the hardened runtime only changes builds that are already unsigned and un-notarizable.

## Notes

- `js-yaml@4.1.1` was already in the tree via `electron-builder`; this declares it in `packages/desktop` instead of relying on a transitive hoist. `npm install --package-lock-only` wanted to churn 419 unrelated lines of peer/optional drift, so the lockfile edit is the single added line.
- `nix/desktop-package.nix` points at the shared config and keeps its explicit flags, which now restate the shared rule rather than owning it.

## Test plan

- Two tests cover the switch, including that it leaves the rest of the mac config (entitlements, appId) untouched for a signed build.
- `desktop-packaging.test.ts`: 9 tests pass. `npm run typecheck` and `npm run lint` clean.

**Needs verification on a Mac** — I could not test this from Linux. Specifically: that the resulting DMG launches on Apple Silicon without the `codesign` workaround, and that a signed release build still notarizes. The table above confirms the config values are unchanged with a certificate present, but that is not the same as a green notarization.

### DIFF
--- a/nix/desktop-package.nix
+++ b/nix/desktop-package.nix
@@ -118,11 +118,13 @@ buildNpmPackage {
       chmod -R u+w "$electron_dist/Electron.app"
       (
         cd packages/desktop
-        # The Nix output is not a distributable DMG, so leave it unsigned and
-        # disable the hardened runtime that requires a matching signature.
+        # The Nix output is not a distributable DMG, so leave it unsigned.
+        # CSC_IDENTITY_AUTO_DISCOVERY=false also tells the shared config to drop
+        # the hardened runtime, which an unsigned build cannot satisfy; the
+        # flags below restate that so this stays readable on its own.
         CSC_IDENTITY_AUTO_DISCOVERY=false \
           ../../node_modules/.bin/electron-builder \
-            --config electron-builder.yml \
+            --config electron-builder.config.cjs \
             --dir \
             --mac \
             --publish never \
--- a/package-lock.json
+++ b/package-lock.json
@@ -37473,6 +37473,7 @@
         "@types/ws": "^8.5.14",
         "electron": "41.2.0",
         "electron-builder": "26.8.1",
+        "js-yaml": "^4.1.0",
         "typescript": "5.9.3",
         "unzip-crx-3": "^0.2.0",
         "vitest": "^4.1.6",
--- a/packages/desktop/electron-builder-signing.cjs
+++ b/packages/desktop/electron-builder-signing.cjs
@@ -1,0 +1,30 @@
+const { execFileSync } = require("node:child_process");
+
+function hasConfiguredSigningIdentity(env) {
+  return ["CSC_LINK", "CSC_NAME", "CSC_IDENTITY"].some((name) => (env[name] ?? "").trim() !== "");
+}
+
+function hasKeychainIdentity(env, options = {}) {
+  if (env.CSC_IDENTITY_AUTO_DISCOVERY === "false") return false;
+  if ((options.platform ?? process.platform) !== "darwin") return false;
+
+  const run = options.run ?? execFileSync;
+  try {
+    const output = run("security", ["find-identity", "-v", "-p", "codesigning"], {
+      encoding: "utf8",
+      stdio: ["ignore", "pipe", "ignore"],
+    });
+    return /\b[1-9]\d* valid identities found\b/.test(output);
+  } catch {
+    return false;
+  }
+}
+
+function isMacSigningAvailable(env, options) {
+  return hasConfiguredSigningIdentity(env) || hasKeychainIdentity(env, options);
+}
+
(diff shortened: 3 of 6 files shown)


boudra:
Closing for now: this needs a macOS Apple Silicon build-and-launch check showing the unsigned app starts without re-signing, plus a successful signed/notarized build. Please reopen with that evidence.

colonelpanic8:
Partial macOS evidence is now available; the signed/notarized requirement remains outstanding, so I am leaving this PR closed.

Tested source branch: `b5f9c4913bb74490109365ce0f37d34ccbe699bd` (the original PR plus a focused workflow fix that unsets `CSC_LINK`/`CSC_KEY_PASSWORD` when the certificate value is empty).

The original run at `c523a90a87` failed before signing because Electron Builder treated the empty certificate path as `packages/desktop`, reporting `not a file`. The follow-up changes only that empty-variable handling; configured certificate values remain unchanged.

The Apple Silicon job in [run 34278532216](https://github.com/colonelpanic8/paseo/actions/runs/34278532216) succeeded on `macos-14`, arm64, with publishing disabled. Raw log excerpts:

```text
falling back to ad-hoc signature for macOS application code signing
signing file=release/mac-arm64/Paseo.app platform=darwin type=distribution identityName=- identityHash=none provisioningProfile=none
skipped macOS notarization reason=`notarize` options were set explicitly `false`
Packaged desktop smoke passed: real renderer and preload loaded; renderer-started desktop daemon pid 41014, listen 127.0.0.1:49228; CLI shim daemon status and terminal smoke succeeded
```

The smoke launches the packaged `.app` during the build, without manual re-signing. The produced DMG is `Paseo-0.5.0-beta.2-arm64.dmg`, SHA-256 `ebdbcb020e4273d1d6c70b129f33846882b4e8ab226e8a75ca5addc2118df871`. Read-only signature inspection of its extracted arm64 app and Electron Framework reports `CodeSignatureFlags(ADHOC)` with no runtime flag. That inspection was performed on Linux, not through macOS `codesign`.

Limits: no signed/notarized run was possible with this account's pull-only upstream access and no signing identity on the fork. This does not claim Gatekeeper acceptance or a separate launch from a mounted DMG. A maintainer with access to the upstream Apple signing credentials still needs to perform the certificate-backed build and notarization check.

The workflow change passed formatting, desktop typecheck, and Bash syntax checking. The repository lint script does not accept YAML files (`No files found to lint`). The original config suite passed 11 tests; the actual Apple Silicon build/smoke above validates the workflow change.

[Complete raw logs and signature inspection](https://github.com/colonelpanic8/paseo/tree/32f060fa4001643a66e23ee79240ee23fae660e7/pr-2885).

boudra:

Issue #2522, opened 2026-07-28
feat: follow XDG base directory spec instead of a single flat ~/.paseo


colonelpanic8:
## Problem

Everything Paseo persists lives in a single flat `PASEO_HOME` (`~/.paseo` by default), which mixes four categories of data with very different lifecycles:

```
~/.paseo/
  config.json               # hand-authored config  -> want in source control
  daemon-keypair.json       # secret
  push-tokens.json          # secret
  server-id                 # machine identity
  cli-client-id             # machine identity
  projects/{projects,workspaces}.json  # state (absolute paths, per-machine)
  agents/ loops/ schedules/ runtime/   # state
  worktrees/                # data (checkouts)
  models/ opencode-home/    # cache (models/ can be large)
  daemon.log                # log (currently 5.6 MB here)
  paseo.pid                 # runtime
```

This makes a few ordinary things awkward:

1. **Dotfiles / source control.** `config.json` is the only file a user authors and would want to commit, but it sits in the same directory as private keys, push tokens, machine-specific absolute paths, and a multi-megabyte log. You can't check in or symlink the directory; you have to cherry-pick one file out of it and ignore the other 13 entries.
2. **Declarative config management (NixOS / home-manager / Ansible / chezmoi).** These tools want to render a read-only config file into a config directory. Today that means dropping a read-only symlink into a directory the daemon also writes state, secrets, and logs into — which works, but only by accident of the daemon not rewriting `config.json`.
3. **Backup and sync.** No clean boundary between "back this up" (config, maybe projects/schedules) and "never sync this" (worktrees, models, logs, runtime, machine identity). Users end up syncing gigabytes of worktrees and model files, or nothing.
4. **Cache cleanup.** `models/` and `opencode-home/` have no distinct location, so a "clear caches" story requires knowing the internal layout.

The XDG Base Directory spec exists precisely to separate these: `XDG_CONFIG_HOME` / `XDG_DATA_HOME` / `XDG_STATE_HOME` / `XDG_CACHE_HOME`.

## Proposal

Support XDG paths, with a compatibility path so nothing breaks:

| Category | Contents | Location |
|---|---|---|
| Config | `config.json` | `$XDG_CONFIG_HOME/paseo/` (`~/.config/paseo`) |
[...]

PR #2900, opened 2026-08-05
Separate configuration from mutable Paseo state

colonelpanic8:
Everything Paseo persists currently mixes hand-authored configuration with secrets, machine-specific state, logs, caches, or Electron's mutable browser profile. That makes daemon and desktop preferences awkward to manage declaratively.

This implements the first practical slice of #2522 without automatically migrating an existing installation.

## Daemon paths

`resolvePaseoPaths(env)` selects `config`, `data`, `state`, and `cache` roots. Only config diverges in this PR; data, state, and cache remain together.

Resolution order:

1. `PASEO_HOME` or `--home` is set: use the historical flat layout at that exact root.
2. macOS or Windows: keep the historical flat layout.
3. A Linux XDG layout marker exists: keep using XDG, even if `~/.paseo` later appears.
4. `~/.paseo` exists: keep the historical flat layout.
5. Otherwise, on a fresh Linux install: put config under `$XDG_CONFIG_HOME/paseo` and everything else under `$XDG_DATA_HOME/paseo`.

The selected XDG layout is recorded in the data root and cached for the process lifetime. Files cannot move because an environment variable changes during runtime, and a later legacy directory cannot switch an XDG installation back after restart.

Daemon config supports layered imports and a writable target. CLI onboarding and password updates use the same resolved config stack, so XDG routing, imports, and writes cannot disagree.

## Desktop settings

Desktop preferences move from browser localStorage into two flat JSON files:

- mutable `settings.json` remains in Electron `userData` and stores only local differences;
- read-only `settings-seed.json` supplies declarative defaults underneath it.

On Linux, a new seed is read from `$XDG_CONFIG_HOME/paseo/settings-seed.json` (default `~/.config/paseo/settings-seed.json`). An existing seed in Electron `userData` keeps its historical path. macOS and Windows remain unchanged, and `PASEO_SETTINGS_SEED_FILE` provides an explicit override on every platform.

Only the five registered preference fields participate in layering. Local values win per field, saves retain only the difference from the seed, and removing a seed-owned value reports the seed path instead of silently shadowing it.

## Compatibility

- `resolvePaseoHome()` remains the historical accessor and returns the data root under XDG.
- Empty and relative `PASEO_HOME` values retain their previous behavior.
- Relative `XDG_*_HOME` values are invalid under the XDG spec and fall back to defaults; `~/...` is expanded before validation.
- Explicit `--home` remains flat even when its path equals the automatic XDG data root.
- Existing daemon installations and desktop seed files retain their current layouts.
- Browser, iOS, and Android settings storage is unchanged.
- There is no automatic migration.

## Verification

Focused local checks on the combined branch:

- 68 server config/path tests passed;
- 43 desktop seed/settings/IPC tests passed;
- desktop package typecheck passed;
- lint and formatting passed for all integration-resolution and XDG seed files.

The real-Electron settings regression was extended to assert the XDG seed path through the actual IPC bridge. Its local run was blocked before Electron startup by the checkout's missing `@anthropic-ai/claude-agent-sdk`; the unit/integration paths above and CI use the same production resolver and wiring.


boudra:

PR #2903, opened 2026-08-05
Fix context usage for Anthropic-compatible providers

ele-yufo:
## Summary

- Treat zero-token `message_start` usage as incomplete so output deltas cannot replace the context meter with an output-only value.
- Reconcile each parent assistant frame’s per-request input, cache-read, cache-creation, and output usage into live and completed context usage.
- Keep subagent usage excluded through the existing parent-frame routing.

## Root cause

Some Anthropic-compatible gateways emit a zeroed `message_start` usage object, then publish accurate usage on the completed assistant frame. Paseo accepted the zero input as authoritative, added the later output delta, and displayed only that output count (for example, `799 / 1m`). It did not read the accurate assistant-frame usage, even though that frame contained the full cached context.

## Validation

- `npx vitest run packages/server/src/server/agent/providers/claude/agent.test.ts --bail=1` (62 passed)
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -2182,6 +2182,84 @@ describe("ClaudeAgentSession context window usage", () => {
     }
   });
 
+  test("assistant usage corrects incomplete Anthropic-compatible stream usage", async () => {
+    const session = await createSessionForTurns(
+      [
+        [
+          createInitMessage(),
+          createMessageStartEvent({
+            input_tokens: 0,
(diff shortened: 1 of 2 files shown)


boudra:
Closing for now: this needs a redacted trace from an affected gateway and before/after context-meter readings for the zero-input usage case. Please reopen with that evidence.

ele-yufo:
Author-side reopen isn't permitted after the close, so the requested evidence (redacted gateway traces + before/after context-meter readings) and the follow-up commit it motivated are now in #4466.

boudra:

PR #2920, opened 2026-08-06
fix(server): shut down provider-owned resources when a provider is disabled

### LINKED ISSUE #2107: Disabled OpenCode provider still starts opencode serve on macOS
### What's broken

On macOS, Paseo starts an `opencode serve` child process even when the OpenCode provider is disabled in **Settings -> Providers**.

In my case, this led to a large accumulation of temporary `.dylib` files under `/private/var/folders/.../T`, eventually consuming about 25 GB of disk space.

### Environment

- Paseo Desktop: `0.1.107`
- macOS: `26.5.2` (`25F84`)
- Architecture: `arm64`
- OpenCode CLI: `1.15.3`
- OpenCode provider setting: disabled in Paseo UI

### Observed behavior

After restarting Paseo and disabling OpenCode in the Providers settings, I still saw a Paseo-managed OpenCode sidecar:

```sh
/Users/ellic/.opencode/bin/opencode serve --port 59453
```

Process tree showed it was launched under Paseo:

```sh
Paseo Supervisor
└─ Paseo Daemon
   └─ /Users/ellic/.opencode/bin/opencode serve --port 59453
```

The OpenCode row in the UI showed disabled, but the sidecar process was still running.

Earlier, a long-running Paseo-managed `opencode serve` generated thousands of temp files matching:

```sh
/private/var/folders/.../T/.fbea*-00000000*
```

At peak:

```sh
12790 files, ~25.0 GB physical
```

Killing `opencode serve` and deleting only those temp files restored disk space from about 11 GiB available to about 36 GiB available.

### Expected behavior

If OpenCode is disabled in Providers:

1. Paseo should not spawn `opencode serve`.
2. Disabling OpenCode should stop any existing Paseo-managed `opencode serve`.
3. Provider snapshot/model discovery should skip disabled providers.
4. Any managed OpenCode sidecar should be cleaned up on disable / daemon restart / app quit.

### Actual behavior

Paseo started `opencode serve` even though OpenCode was disabled in the UI.

### Related issues / PRs

Possibly related, but this seems distinct:

- #1683 macOS OpenCode helper CPU issue
- #1697 Stop OpenCode helper servers from leaking
- #2016 acquireDedicated spawns new OpenCode server per agent
- #1178 external OpenCode server lifecycle discussion
- #673 orphaned opencode serve cleanup

hooni0918:
## Summary

Disabling a provider in Settings → Providers did not stop the provider's running resources. `ProviderSnapshotManager.applyMutableProviderConfig` rebuilds the registry and replaces the whole client map without disposing the displaced clients, so a live `opencode serve` sidecar kept running until the daemon exited (#2107). `ProviderSnapshotManager.shutdown()` also only materialized clients for *enabled* providers, so a provider disabled at runtime never got its `shutdown()` called on daemon stop either.

Fixes #2107

## Changes

- `applyMutableProviderConfig` now shuts down displaced clients whose provider no longer has an enabled consumer. A base provider still counts as consumed while any provider derived from it remains enabled, because derived providers share the base client's resources (the OpenCode sidecar is a process-wide singleton).
- `ProviderSnapshotManager.shutdown()` now materializes clients for **all registered providers, including disabled ones**, so a provider disabled at runtime still releases its resources on daemon stop.

Single file changed (`provider-snapshot-manager.ts`) plus its test file.

### Regression tests fail on the old code

Three new tests reproduce the reported behavior on unpatched `main` (the fourth is a negative guard that passes on both):

```
$ git stash push -- packages/server/src/server/agent/provider-snapshot-manager.ts
$ npx vitest run src/server/agent/provider-snapshot-manager.test.ts
 ❯ src/server/agent/provider-snapshot-manager.test.ts (47 tests | 3 failed)
     × shuts down the displaced client when its provider is disabled
     × keeps the base client alive while a derived provider remains enabled
     × shutdown releases provider resources even when the provider is disabled
 Tests  3 failed | 44 passed (47)
```

With the fix applied:

```
$ npx vitest run src/server/agent/provider-snapshot-manager.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  47 passed (47)
```

### Live repro against a real opencode binary (macOS)

Using the ad-hoc in-process daemon harness from `docs/ad-hoc-daemon-testing.md` with a real `opencode` 1.18.14 binary: create an opencode agent (holds a sidecar acquisition), verify `opencode serve` is running via `pgrep`, disable the provider through the same `set_daemon_config_request` RPC the UI uses, then poll for the process.

**Before (unpatched `main`)** — sidecar survives the disable:

```
[3] opencode serve processes after discovery:
29348 .../node_modules/.bin/opencode serve --port 56877
[4] disabling the opencode provider via set_daemon_config_request...
[5] opencode serve processes after disable (15s grace):
29348 .../node_modules/.bin/opencode serve --port 56877
RESULT: FAIL — sidecar survived the provider disable (#2107 behavior)
```

**After (this branch)** — sidecar stops:

```
[3] opencode serve processes after discovery:
28564 .../node_modules/.bin/opencode serve --port 56842
[4] disabling the opencode provider via set_daemon_config_request...
[5] opencode serve processes after disable (15s grace):
  (none)
RESULT: PASS — disabling the provider stopped the sidecar
```

<details>
<summary>Repro script (run with <code>npx tsx</code>, not committed)</summary>

```typescript
// Starts an isolated in-process daemon with a real opencode binary on PATH,
// triggers the opencode sidecar via agent creation, disables the provider
// through the same RPC the UI uses, and checks whether the sidecar dies.
import os from "node:os";
import path from "node:path";
import { execSync } from "node:child_process";
import { mkdir, mkdtemp, rm } from "node:fs/promises";
import pino from "pino";
import { createPaseoDaemon } from "./bootstrap.js";
import { DaemonClient } from "./test-utils/daemon-client.js";

const OPENCODE_BIN_DIR = process.env.QA_OPENCODE_BIN_DIR;
if (!OPENCODE_BIN_DIR) {
  throw new Error("Set QA_OPENCODE_BIN_DIR to the directory containing the opencode binary");
}
process.env.PATH = `${OPENCODE_BIN_DIR}:${process.env.PATH}`;

function opencodeServeProcesses(): string {
  try {
    return execSync("pgrep -fl 'opencode.*serve --port' || true", {
      encoding: "utf8",
      shell: "/bin/bash",
    }).trim();
  } catch {
    return "";
  }
}

async function sleep(ms: number): Promise<void> {
  await new Promise((resolve) => setTimeout(resolve, ms));
}

const logger = pino({ level: "warn" });
const paseoHomeRoot = await mkdtemp(path.join(os.tmpdir(), "paseo-qa-2107-"));
const paseoHome = path.join(paseoHomeRoot, ".paseo");
await mkdir(paseoHome, { recursive: true });
const staticDir = await mkdtemp(path.join(os.tmpdir(), "paseo-qa-2107-static-"));

const daemon = await createPaseoDaemon(
  {
    listen: "127.0.0.1:0",
    paseoHome,
    corsAllowedOrigins: [],
    hostnames: true,
    mcpEnabled: false,
    staticDir,
    mcpDebug: false,
    agentClients: {},
    agentStoragePath: path.join(paseoHome, "agents"),
    relayEnabled: false,
    relayEndpoint: "relay.paseo.sh:443",
    appBaseUrl: "https://app.paseo.sh",
  },
  logger,
);

await daemon.start();
const target = daemon.getListenTarget();
const port = target!.type === "tcp" ? target!.port : null;

const client = new DaemonClient({
  url: `ws://127.0.0.1:${port}/ws`,
  appVersion: "0.1.70",
});

let exitCode = 1;
try {
  await client.connect();
  await client.fetchAgents({ subscribe: { subscriptionId: "qa-2107" } });

  console.log("[1] opencode serve processes before discovery:");
  console.log(opencodeServeProcesses() || "  (none)");

  console.log("[2] creating an opencode agent — holds a sidecar acquisition...");
  const cwd = await mkdtemp(path.join(os.tmpdir(), "paseo-qa-2107-ws-"));
  const agent = await client.createAgent({ provider: "opencode", cwd });
  console.log(`  agent created: ${agent.id}`);

  await sleep(1_000);
  const afterSpawn = opencodeServeProcesses();
  console.log("[3] opencode serve processes after discovery:");
  console.log(afterSpawn || "  (none)");
  if (!afterSpawn) {
    console.log("FAIL: sidecar never spawned — cannot exercise the repro");
    process.exit(1);
  }

  console.log("[4] disabling the opencode provider via set_daemon_config_request...");
  await client.patchDaemonConfig({ providers: { opencode: { enabled: false } } });

  let remaining = afterSpawn;
  for (let i = 0; i < 15; i++) {
    await sleep(1_000);
    remaining = opencodeServeProcesses();
    if (!remaining) break;
  }
  console.log("[5] opencode serve processes after disable (15s grace):");
  console.log(remaining || "  (none)");

  if (remaining) {
    console.log("RESULT: FAIL — sidecar survived the provider disable (#2107 behavior)");
    exitCode = 1;
  } else {
    console.log("RESULT: PASS — disabling the provider stopped the sidecar");
    exitCode = 0;
  }
} finally {
  await client.close().catch(() => undefined);
  await daemon.stop().catch(() => undefined);
  await rm(paseoHomeRoot, { recursive: true, force: true });
  await rm(staticDir, { recursive: true, force: true });
}
process.exit(exitCode);
```

</details>

### Checks

- `npm run typecheck` — passes (also ran in pre-commit)
- `npm run lint` on the changed files — 0 warnings, 0 errors
- `npm run format:check:files` — clean

### Platform matrix

Daemon-only change, no UI.

| iOS             | –      | not affected (daemon change) |
| Android         | –      | not affected (daemon change) |
| Web             | –      | not affected (daemon change) |
| Desktop macOS   | ✅     | live repro above (macOS 15, arm64, opencode 1.18.14) |
| Desktop Windows | –      | code path is platform-independent; covered by unit tests |
| Desktop Linux   | –      | code path is platform-independent; covered by unit tests |

## Out of scope (observed while investigating, happy to file/follow up)

- `wrapClientProvider` does not forward the optional `shutdown()` method, so an opencode provider with model overrides (or a derived provider) yields a client whose sidecar can never be released via this path.
- An in-flight `refreshProvider`/`fetchCatalog` that already passed the enabled guard can still spawn a sidecar right after a disable. The fixed `shutdown()` now reaps it at daemon stop, and the managed-process ledger reaps it at next boot, but the window itself still exists.

## Notes

Maintainer edits are welcome. Developed with AI assistance; every command and repro above was run and verified on a real machine.

### DIFF
--- a/packages/server/src/server/agent/provider-snapshot-manager.test.ts
+++ b/packages/server/src/server/agent/provider-snapshot-manager.test.ts
@@ -1072,6 +1072,80 @@ describe("ProviderSnapshotManager applyMutableProviderConfig", () => {
       manager.destroy();
     }
   });
+
+  test("shuts down the displaced client when its provider is disabled", () => {
+    const shutdown = vi.fn(async () => {});
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      providerOverrides: {
+        claude: { enabled: false },
+        codex: { enabled: false },
+        copilot: { enabled: false },
+        pi: { enabled: false },
+      },
+      extraClients: {
+        opencode: createExtraClient("opencode", { shutdown }),
+      },
+    });
+    try {
+      manager.applyMutableProviderConfig({ opencode: { enabled: false } });
+      expect(shutdown).toHaveBeenCalledTimes(1);
+    } finally {
+      manager.destroy();
+    }
+  });
+
+  test("does not shut down clients whose provider remains enabled", () => {
+    const shutdown = vi.fn(async () => {});
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      providerOverrides: {
+        claude: { enabled: false },
+        copilot: { enabled: false },
+        pi: { enabled: false },
+      },
+      extraClients: {
+        opencode: createExtraClient("opencode", { shutdown }),
+      },
+    });
+    try {
+      manager.applyMutableProviderConfig({ codex: { enabled: false } });
+      expect(shutdown).not.toHaveBeenCalled();
+    } finally {
+      manager.destroy();
+    }
+  });
+
+  test("keeps the base client alive while a derived provider remains enabled", () => {
+    const shutdown = vi.fn(async () => {});
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      providerOverrides: {
+        claude: { enabled: false },
+        codex: { enabled: false },
+        copilot: { enabled: false },
+        pi: { enabled: false },
+      },
+      extraClients: {
+        opencode: createExtraClient("opencode", { shutdown }),
+      },
+    });
+    try {
+      manager.applyMutableProviderConfig({
+        opencode: { enabled: false },
+        "my-opencode": { extends: "opencode", label: "Mine", enabled: true },
+      });
+      expect(shutdown).not.toHaveBeenCalled();
+
+      manager.applyMutableProviderConfig({
+        opencode: { enabled: false },
+        "my-opencode": { extends: "opencode", label: "Mine", enabled: false },
+      });
+      expect(shutdown).toHaveBeenCalledTimes(1);
+    } finally {
+      manager.destroy();
+    }
+  });
 });
 
 describe("ProviderSnapshotManager lifecycle", () => {
@@ -1121,6 +1195,29 @@ describe("ProviderSnapshotManager lifecycle", () => {
     listener.mockClear();
     manager.applyMutableProviderConfig({});
     expect(listener).not.toHaveBeenCalled();
+  });
+
+  test("shutdown releases provider resources even when the provider is disabled", async () => {
+    const shutdownSpy = vi
+      .spyOn(OpenCodeAgentClient.prototype, "shutdown")
+      .mockResolvedValue(undefined);
+    const manager = new ProviderSnapshotManager({
+      logger: createTestLogger(),
+      providerOverrides: {
+        claude: { enabled: false },
+        codex: { enabled: false },
+        copilot: { enabled: false },
+        opencode: { enabled: false },
+        pi: { enabled: false },
+      },
+    });
+    try {
+      await manager.shutdown();
+      expect(shutdownSpy).toHaveBeenCalled();
+    } finally {
+      shutdownSpy.mockRestore();
+      manager.destroy();
+    }
   });
 });
 
--- a/packages/server/src/server/agent/provider-snapshot-manager.ts
+++ b/packages/server/src/server/agent/provider-snapshot-manager.ts
@@ -81,6 +81,22 @@ function omitProviderOverrides(
   return Object.keys(nextOverrides).length > 0 ? nextOverrides : undefined;
 }
 
+// Derived providers reuse the base provider's client resources (the OpenCode
+// sidecar server is process-wide), so a base client still counts as consumed
+// while any provider derived from it remains enabled.
+function clientHasEnabledConsumer(
+  registry: Record<AgentProvider, ProviderDefinition>,
+  provider: AgentProvider,
+): boolean {
+  for (const [id, definition] of Object.entries(registry)) {
+    if (!definition.enabled) continue;
+    if (id === provider || definition.derivedFromProviderId === provider) {
+      return true;
+    }
+  }
+  return false;
+}
+
 type ProviderSnapshotChangeListener = (entries: ProviderSnapshotEntry[], cwd: string) => void;
 
 export interface ProviderSnapshotManagerOptions {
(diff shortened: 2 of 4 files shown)


hooni0918:
The disable/re-enable race flagged in the review is real and is now fixed in 65a1ebe.

`OpenCodeServerManager.shutdown()` cleared `currentServer` only **after** awaiting the kills, so during the SIGTERM grace window a racing acquire could receive the dying server — and a fresh server installed by that acquire was then detached by the pending shutdown's cleanup, leaving it running untracked.

The fix detaches the captured servers synchronously before awaiting the kills, so a racing acquire always starts a fresh server that the pending shutdown no longer touches.

Regression test (`an acquire racing shutdown gets a fresh server that shutdown does not clear`) uses a termination gate in the fake runtime to hold the kill in flight deterministically. On the previous commit it fails exactly as described:

```
AssertionError: expected 'http://127.0.0.1:4493' to be 'http://127.0.0.1:4492'
```

(the re-enabled acquire's server 4492 was wiped by the pending shutdown, forcing a third spawn). With the fix, both server-manager and snapshot-manager suites pass (65 passed, 1 pre-existing skip), plus typecheck/lint/format in pre-commit.

PR #2932, opened 2026-08-06
fix(app): copy code without Markdown fences or a trailing newline

### LINKED ISSUE #2667: bug: copying a selection from a message loses markdown formatting (Copy turn preserves it)
### What's broken

Copying a *selection* from an assistant message loses the markdown structure. Lists are the most visible case: each bullet arrives as its own line with blank lines inserted between items, so a compact list pastes out as loose, unformatted prose that has to be repaired by hand.

The turn-level **Copy turn** / **Copy message** actions are fine — #1156 put both `text/plain` markdown and `text/html` on the clipboard, and formatting survives. But those actions are all-or-nothing, and whenever I only want part of an answer the only option is to select with the mouse, which takes an entirely different path.

Roughly what a selected three-item list looks like on paste (schematic, not a verbatim capture):

```
Some intro line

First item

Second item

Third item
```

instead of:

```
Some intro line
- First item
- Second item
- Third item
```

**Expected:** a selection pastes with the same fidelity the Copy actions already provide, at least as plain markdown.

**Why it happens (as far as I can tell):** there is no `copy` event handler anywhere in the app, so a manual selection is serialised by the browser straight from the rendered DOM, and the `rich-clipboard` path never runs. Since the renderer builds markdown into React Native Web primitives via `react-native-markdown-display`, every paragraph and list item becomes its own block-level element, and the browser inserts a line break at each boundary — which is exactly the "one bullet per line, blank line between" shape above. The existing `markdown-text-selection.tsx` only governs *what* is selectable (prose vs table cells), not how a selection is serialised.

A `copy` listener on the chat surface that rewrites `clipboardData` from the source markdown for the selected range would bring this in line with the buttons. `react-native-markdown-display` is markdown-it based and its block tokens carry `map` line ranges, so a selection can be mapped back to source lines; the markdown → plain/HTML conversion already exists in `rich-clipboard.ts` and could be reused. Even a coarse version that snapped the selection to whole blocks and emitted their source markdown would fix the common case.

### Steps to reproduce

1. Ask an agent something that answers with a bulleted or numbered list.
2. Select part of the rendered answer with the mouse (not the Copy turn button) — e.g. an intro line plus a few list items.
3. Copy with Ctrl/Cmd+C.
4. Paste into any plain-text or markdown editor.
5. The list markers are gone and every item sits on its own line separated by blank lines.

Reproduces every time. Using **Copy turn** on the same message instead pastes correctly, which is the contrast that makes the selection path stand out.

### Where did this happen

Desktop app (Linux).

### Paseo version

0.2.3

### OS version

Arch Linux / CachyOS, X11

### Agent provider

Claude (Claude Code 2.1.220), though this looks provider-independent since it's purely a rendering/clipboard path.

cleiter:
### Linked issue

No issue — reported on Discord by ramblurr, confirmed by nikus and reproduced by me.

Follow-up to #2808 (which closed #2667).

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Copying a snippet out of an assistant message dragged Markdown syntax along with it, so
it had to go through a scratch editor before it could be pasted into a shell or a file.
Selecting a single word inside a code block came back wrapped in a fence. The only way to
get clean code out of a block was its Copy button, which takes every line at once.

#2808 re-serializes the selected DOM back to Markdown, which is right for prose and lists
— a partial list selection keeps its bullets. It reaches code too.
`wrapSelectionWithAncestors` rebuilds the `pre`/`code` ancestor chain around the cloned
fragment, so Turndown emits a whole fence around whatever was highlighted. Inline
`` `code` `` picks up backticks the same way.

This adds one guard before the Markdown path: a selection that sits entirely inside
rendered code copies as the code itself. `text/plain` is what was highlighted, indentation
intact; `text/html` keeps it marked up so a rich target still sees code. That matches what
the block's own Copy button puts on the clipboard.

A Turndown rule could not have done this. By the time Turndown runs, a one-word selection
and a whole-block selection are byte-identical DOM, because the `pre`/`code` wrapper is
manufactured *after* `cloneContents()`. Only a check against the live `Range` has the
information.

Two newline fixes ride along, because both make a paste run code you did not ask to run:

- Newlines between rendered code lines are their own text nodes, so a selection that runs
  off the end of a line carries the break with it. Pasting that into a terminal executes
  the line. One trailing break, plus any indentation after it, is now stripped.
- The block's Copy button was handed the raw fence body, which ends in a newline. It now
  gets the same text that is rendered.

### Goals

- Selecting part of a code block and copying yields the code, with no fence or backticks.
- Multi-line selections keep their newlines and leading indentation.
- Neither selection copy nor the Copy button leaves a trailing newline on the clipboard.
- The block's hover Copy button never leaks its label into the copied text.
- A rich paste target still receives a code block (or inline code), with the language class.
- Prose, list, link, table, and whole-message selections behave exactly as before — the
  existing byte-exact whole-message assertion passes unchanged.

### Non-goals

Three deliberate calls, listed so they are not read as oversights:

- **Selecting a *whole* fence, or a whole inline code span, now also yields bare code**
  rather than the Markdown construct. That is the behavior the reports asked for and it
  matches the Copy button. **Copy turn** still produces fenced Markdown for anyone who
  wants the source form.
- **Mixed code-and-prose selections still fence the code fragment.** A selection that
  starts inside a block and ends in the sentence after it is unchanged. Deferred on
  purpose: a correct fix needs a "is this fence fully selected?" test, and the block's
  hover Copy button lives *inside* the `pre`, so a node-based containment check reports
  *partial* for a fence the user completely highlighted — it would strip the fence off
  exactly the case that must keep it. Containment would have to be defined on text content
  excluding `[data-paseo-markdown-ignore]` nodes. Worth doing, not here.
- **@ramblurr's original ask is broader than this PR.** It covers prose too — "always
  includes the markdown formatting" — which points at a copy-as-plain-text preference, or
  flipping `text/plain` to rendered text globally. That reverses what #2667 explicitly
  asked for, so it is a product call and belongs in its own issue. This PR fixes the code
  case, which is the part that regressed.

Also out of scope: native selection copy (`assistant-selection-copy/surface.tsx` is a
passthrough on native — there is no handler to change), and tool-call / file-preview code
blocks, which carry the same attributes but render outside `[data-testid="assistant-message"]`
where the copy handler never sees them.

### QA

**Manual** — verified on Electron desktop. Before/after videos below. Native is
untouched: `surface.tsx` has no copy handler at all, so this is not a claimed native pass.

**Before**

https://github.com/user-attachments/assets/a65216f4-454d-42ba-803d-6d668dcf3cc5

**After**

https://github.com/user-attachments/assets/9d8bbd6f-bdde-47da-a40d-f53e80b25789

Steps run:

1. Double-click a word inside a fence, Cmd+C, paste into a terminal → the word alone, no
   trailing newline, nothing executes.
2. Drag-select two indented lines, paste into a plain editor → indentation intact, no fences.
3. Paste the same two lines into a rich target → still a code block.
4. Select part of `` `inline` `` → no backticks.
5. Select the whole fence plus the sentence after it → fence still there.
6. Select the whole message, paste → fenced Markdown, unchanged.
7. Click the block's Copy button, paste into a terminal → no trailing newline.

**Automated**

- `packages/app/e2e/browser/assistant-selection-copy.spec.ts` — five new cases against the
  real renderer, each doing a real Cmd+C (or a real button click) and reading the system
  clipboard back: a word inside the ```` ```typescript ```` fence, a two-line range inside
  the ````` ````text ````` fence, `apply` inside `` `apply_patch` ``, a whole fence line
  selected through its trailing line break, and the Copy button. The last two were each
  verified to **fail with their fix reverted**, so they reproduce the bug rather than only
  asserting the outcome. The `typescript` fence in the fixture is two lines now; at one
  line it had no line break node to select through.
- The pre-existing whole-message assertion in that spec is byte-identical and still passes
  — that is the regression guard for #2808.
- `packages/app/src/assistant-selection-copy/content.browser.test.ts` — new, 16 cases in
  the real-Chromium vitest project. First direct test of `content.web.ts`. Covers the
  element-vs-text common ancestor, indentation and newline preservation, trailing-break
  stripping, the ignored Copy button, whitespace-only and ignored-only selections, a fence
  with no info string, a fence info string containing `"`, inline code partial and whole,
  and the two paths that must stay on Markdown.
- `npm run typecheck`, `npm run lint`, `npm run format` pass at the repo root, after
  rebasing onto current `main`.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/assistant-selection-copy.spec.ts
+++ b/packages/app/e2e/browser/assistant-selection-copy.spec.ts
@@ -309,6 +309,28 @@ test("copying an assistant selection preserves Markdown structure and links", as
     expect(columnSliceClipboard.html).not.toContain("<table>");
     expect(columnSliceClipboard.html).toContain("ready");
     expect(columnSliceClipboard.html).toContain("<s>obsolete</s>");
+
+    await selectAssistantText(page, "answer");
+    await copySelection(page);
+
+    const wordInFenceClipboard = await readRichClipboard(page);
+    expect(wordInFenceClipboard.plainText).toBe("answer");
+    expect(wordInFenceClipboard.html).toContain("<code>answer</code>");
+    expect(wordInFenceClipboard.html).not.toContain("<pre>");
+
+    await selectAssistantTextRange(page, "before", "after");
+    await copySelection(page);
+
+    const fenceLinesClipboard = await readRichClipboard(page);
+    expect(fenceLinesClipboard.plainText).toBe("before\n```\nafter");
+    expect(fenceLinesClipboard.html).toContain('<pre><code class="language-text">');
+
+    await selectAssistantText(page, "apply");
+    await copySelection(page);
+
+    const inlineCodeClipboard = await readRichClipboard(page);
+    expect(inlineCodeClipboard.plainText).toBe("apply");
+    expect(inlineCodeClipboard.html).toContain("<code>apply</code>");
   } finally {
     await agent.cleanup();
   }
--- a/packages/app/e2e/browser/assistant-selection-copy.spec.ts
+++ b/packages/app/e2e/browser/assistant-selection-copy.spec.ts
@@ -36,6 +36,7 @@ const ASSISTANT_MARKDOWN = [
   "",
   "```typescript",
   'const answer = "yes";',
+  "  return answer;",
   "```",
   "",
   "````text",
@@ -132,6 +133,44 @@ async function selectAssistantTextRange(
     },
     { startText, endText },
   );
+}
+
+/**
+ * Select from `startText` through the line break that ends its rendered code line.
+ *
+ * Highlighted code splits a line across one text node per syntax token and puts the
+ * newline in a node of its own, so this is what a drag or double click that runs off
+ * the end of a line produces.
+ */
+async function selectAssistantThroughLineBreak(page: Page, startText: string): Promise<void> {
+  const assistantMessage = page.getByTestId("assistant-message").filter({
+    hasText: "Direct matches:",
+  });
+  await assistantMessage.evaluate((element, text) => {
+    const walker = document.createTreeWalker(element, NodeFilter.SHOW_TEXT);
+    let startNode: Node | null = null;
+    let startOffset = -1;
+    let textNode = walker.nextNode();
+    while (textNode) {
+      if (!startNode) {
+        const offset = textNode.textContent?.indexOf(text) ?? -1;
+        if (offset >= 0) {
+          startNode = textNode;
+          startOffset = offset;
+        }
+      } else if (textNode.textContent?.startsWith("\n")) {
+        const range = document.createRange();
+        range.setStart(startNode, startOffset);
+        range.setEnd(textNode, 1);
+        const selection = window.getSelection();
+        selection?.removeAllRanges();
+        selection?.addRange(range);
+        return;
+      }
+      textNode = walker.nextNode();
+    }
+    throw new Error(`Could not find a line break after: ${text}`);
+  }, startText);
 }
 
 async function copySelection(page: Page): Promise<void> {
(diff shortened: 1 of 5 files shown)


cleiter:
Partly obsoleted by #2930 - will create follow-up PR to keep it clean.

PR #2964, opened 2026-08-07
Plugin system and marketplace

nickmaglowsch:
## What this is

A plugin system for Paseo, plus the marketplace plumbing to install from one. VS Code is the model: a manifest declares contribution points, the host renders the plugin in a sandbox, and the plugin talks back over a narrow bridge.

Two things that would otherwise have been core features are plugins instead:

- **File previews** — a plugin can claim any extension core does not already render, CSV being the worked example, instead of adding a branch to `pane.tsx`. `use-file-preview-renderer.ts` picks the renderer. HTML was to be one of these too; upstream landed it in core mid-branch, and the merge below leaves it there.
- **A sidebar pet** — the "misc plugin" case, shipped as `examples/plugins/hello-paseo` so the contribution points have a worked example rather than only prose.

50 commits. `docs/plugins.md` is the contract; `CONTRIBUTING.md` and `SECURITY.md` point at it.

## Shape

| Layer | What landed |
|---|---|
| `packages/protocol` | Manifest and RPC schemas, dotted-namespace names, all new fields optional |
| `packages/server` | `plugin/{store,service,registry-client,version-range}` — install/enable/uninstall, crash recovery, registry trust, atomic writes |
| `packages/app` | `plugins/` — host bridge, sandbox (web + native twins), settings section, query layer |
| `packages/cli` | `paseo plugin browse/install/ls/uninstall/set-enabled` |
| `packages/desktop` | Window-manager wiring for the plugin surface |

## Sandbox model

The plugin runs in an `<iframe sandbox="allow-scripts">` — an opaque origin, deliberately without `allow-same-origin`. Native loads that same iframe inside a relay host document in `react-native-webview` rather than making the plugin the WebView's top document.

Two invariants that cost the most to get right, both documented at the code:

**The parser door is reachable but starved.** `Document.prototype.open` is not denied; `write`/`writeln`, `parseHTMLUnsafe`, `setHTMLUnsafe`, `setHTML`, `parseHTML`, and both `XSLTProcessor` transforms are. That list was found short five times during review — treat it as incomplete, not finished.

**On both native platforms the guest can post to the host directly, bypassing the relay that does the stamping.** Android registers `addWebMessageListener(view, "ReactNativeWebView", Set.of("*"), …)` and the listener discards `isMainFrame`. iOS *looks* safe — the `window.ReactNativeWebView` shim is `forMainFrameOnly:YES` — but the handler behind it goes on the whole `WKUserContentController` via `addScriptMessageHandler:name:`, which has no frame scoping to give, and `didReceiveScriptMessage` never reads `frameInfo.isMainFrame`. Neither injected object can be removed: both live in the guest's cross-origin realm. So the host stamps each document with an unguessable id (~52 bits, `crypto.getRandomValues`) and ignores anything that does not carry it. That stamp is the entire mitigation, which is why it gets its own tests.

Every denial is proven from inside the plugin realm, not by reading `child.contentWindow.X` from the host — that throws on an opaque origin either way and proves nothing.

## Merged with upstream `main`

The branch was cut from a base 112 commits behind, so `8abf23f33` merges `upstream/main` in. Five files conflicted, and one was a design collision rather than a textual one: **upstream landed a native HTML preview in core** while this branch was making previews plugin-driven.

Resolved with the built-ins winning. An extension core already renders is closed to plugins, and `pluginFilePreviewsForPath` returns nothing for one, so the pane and the Settings conflict report cannot disagree about who renders a file. Upstream's `mode` prop and line-target fallback are gone because the renderer model subsumes them.

That reroutes `.html` away from a plugin and into `FileHtmlPreview`, so it is worth stating what that renderer is: a sandboxed `allow-scripts` iframe on web and a navigation-locked `WebView` on native, both fed a document with a prologue-supplied CSP of `default-src 'none'`. It has **no host bridge at all** — no `onMessage`, no `postMessage` channel. The plugin sandbox's denial list and document stamp exist to protect a bridge core's preview does not have, so this narrows the surface rather than widening it.

## Review

26 rounds against two standing reviewers (one security, one correctness), reusing the same two agents so each kept its memory of what it had already found rather than rediscovering it. Round 22 was the first double APPROVE; the upstream merge then took three more rounds of its own.

The recurring failure worth naming: **a fix ships behind a fully green gate with no test that can fail, and the missing test hides a second defect.** That happened enough times that every fix in the later rounds had to be proven red first — copy the file aside, `git checkout HEAD -- <file>`, watch it fail, restore.

## Disclosures

Read these before approving; they are the known gaps, not a summary of the good parts.

- **Cross-family review was unavailable** (Codex, HTTP 401). Every round was Claude-on-Claude, so the blind spots are correlated.
- **Zero iOS/Android device runtime evidence.** The native findings above are source-level, read out of vendored `RNCWebView.java` / `RNCWebViewImpl.m`. Nobody ran this on a phone.
- **`plugins.paseo.sh` is NXDOMAIN.** The registry the client points at does not exist yet; `https://plugins.paseo.sh/index.json` 404s at the DNS layer.
- **Residual, not closed:** a plugin can append `<link rel="preconnect">` and get one DNS lookup plus a TCP/TLS connection to a host of its choosing. Measured to produce no connection in Chromium; no web platform primitive denies it. Disclosed open.
- **Core's HTML preview can navigate itself**, on web and (narrowly) on Android. No CSP directive prevents it — `navigate-to` is dead and a meta refresh needs no script — so a hostile page can reach a server by navigating, carrying what is in the preview. The opaque origin bounds it. Upstream's residual, carried in SECURITY.md; this branch routes `.html` to that renderer but did not introduce it.
- **Frame removal is empirical**, observed in Chromium, not spec-guaranteed.
- `attachShadow` always returns an open root.
- `will-attach-webview` was deliberately not tightened — Electron 41 gives it no `WebFrameMain`, and the IPC alternative is async while attach is sync.

Plugins cannot embed `iframe`/`object`/`embed`/`link`/`webview`/`fencedframe`.

### Platform matrix

| Platform | Tested | How |
|---|---|---|
| Web (Chromium) | **Yes** | Playwright `Desktop Chrome` against a real daemon, plus the Vitest browser project (real Chromium) for every sandbox denial |
| iOS | **No** | No simulator run, no device run |
| Android | **No** | No emulator run, no device run |
| Desktop (Electron) | **No** | Not exercised. The `frame-src 'none'` meta the `paseo://` handler serves is asserted by a unit test reading `public/index.html`, not by a running Electron window |

Read that as the headline, not a footnote. `packages/app/src/plugins/sandbox.tsx` is a **separate implementation** from the web one — `react-native-webview` with `originWhitelist: ["*"]` and a deny-by-default `onShouldStartLoadWithRequest` — and that navigation gate is one of the four controls `docs/plugins.md` says must hold to keep previewed file content on the machine. It is the control with zero runtime evidence here. Both exfiltration holes found in review round 1 passed typecheck, lint and the whole suite, and one of them was `originWhitelist` in exactly this file. Before this reaches a phone, someone has to open a plugin preview on real iOS and real Android and confirm a plugin that navigates, submits a form, or opens a URL goes nowhere.

### End-to-end, against a real daemon

`packages/app` · `npx playwright test --project=browser e2e/browser/plugin-system.spec.ts`

```
Running 6 tests using 1 worker
[e2e] Worker 0 daemon started on port 36219, home: /tmp/paseo-e2e-worker-0-eVNJPd
  ✓  1 … Settings lists a hand-installed plugin and a broken one as unavailable (3.2s)
  ✓  2 … the enable toggle round-trips through the daemon and survives a reload (7.7s)
  ✓  3 … a claimed extension renders the plugin preview, and the toolbar switches back to source (4.7s)
  ✓  4 … the plugin's sidebar panel gets a tab and renders with the workspace context (4.4s)
  ✓  5 … a user-selected theme change reaches the plugin, not just the OS colour scheme (10.4s)
  ✓  6 … disabling the plugin removes its preview and its sidebar tab; re-enabling restores them (11.8s)

  6 passed (53.6s)
```

The daemon is the one `e2e/global-setup.ts` starts in a fresh `mkdtemp` `PASEO_HOME`. The plugin is installed the way `docs/plugins.md` says to install one by hand — a directory dropped into `$PASEO_HOME/plugins/` — so this exercises the real scan, the real RPCs, and the real sandbox.

### Screenshots

**Settings → Plugins.** A hand-installed plugin and a deliberately broken one. The broken row is the point: a manifest whose `entry` does not exist stays listed, disabled, with the reason on it, rather than disappearing or crashing the section.

Settings plugins list

**File preview from a plugin.** `notes.hello` is an extension core does not render, so the plugin's sandboxed preview takes the pane and the toolbar offers Lines / Source. `.md` and `.html` never reach here — built-ins win, see the merge section above.

Plugin file preview

**Sidebar panel.** The "misc plugin" contribution point — the pet gets its own tab beside Changes and Files, and receives the workspace path as context.

Plugin sidebar panel

**Theme reaches the plugin.** Same preview after the user picks a dark theme in Settings — not the OS colour scheme, the in-app choice. The plugin repaints because the host pushes `theme` over the bridge; it has no access to the host's CSS.

Plugin preview under a dark theme

### Automated tests

| Suite | Result |
|---|---|
| `packages/app` e2e (`plugin-system.spec.ts`) | 6 passed |
| `packages/app` browser project (`sandbox*.browser.test.ts`, `host-document.browser.test.ts`) | 39 passed — every sandbox denial, proven from inside the plugin realm |
| `packages/app` unit (`plugins/`, `file-pane-render-mode`) | 143 passed |
| `packages/server` (`plugin/`, websocket) | 110 passed |
| `npm run typecheck` / `npm run lint` / `npm run format:check` | exit 0, 0 warnings, 0 errors |

## Verification

CI has now run, and found four things the local targeted runs could not. All four are fixed in `09cc4ccf2`:

- **`server-tests` (both platforms).** The plugin runtime was a fifth *required* service, but it is roughly the thirtieth positional constructor argument and every pre-existing test stops before it. A daemon with no plugin runtime is a state the feature already models — it never advertises `features.plugins` — so it is optional now and `dispatchPluginMessage` falls through to the unknown-RPC answer rather than throwing at construction.
- **`playwright (shard 3/4)`.** `git mv`-ing the e2e helper into `e2e/support/` added a directory level and the two `../`-counted paths to the repo root went with it. Anchored on one `REPO_ROOT`.
- **`app-tests`.** The sibling-impersonation test read the victim's *first* report, but the victim reports after every message and the attacker's own report can reach the host first. It waits for both now, counted across both buckets, so a victim that wrongly accepts the forgery still fails the assertion rather than timing out.
- **`package-lock.json`.** An earlier `npm install` rewrote 190 lines of unrelated optional-peer bookkeeping. Cut to the nine lines this branch needs.

`nix/npm-deps.hash` is stale against that lockfile, and `prefetch-npm-deps` is the only thing that can compute the replacement.

Locally: typecheck clean across all workspaces, lint 0/0, formatted. 143 app unit, 39 app browser, 110 server plugin + websocket tests green.


nickmaglowsch:
@greptile-apps

boudra:
Thanks for the work on this proposal. Paseo has since shipped its native plugin system, and we are building extensions on that runtime. Closing this alternative iframe/WebView plugin system and marketplace proposal.

PR #2973, opened 2026-08-07
Keep wide characters two columns wide in the native terminal

lightofjeju:
Draft until I confirm the rebuilt Android APK on the device that produced the evidence below.

## Problem

Every CJK character in the native terminal renders as the character plus a literal space, and everything after it on the row drifts one column to the left per character. Long lines then wrap early and mixed Korean/Latin rows visibly break apart.

A double-width cell is stored as the character followed by a spacer cell. `terminalCellCount` treated `cell.width` as authoritative even when it reported `1` for a character the terminal drew in two columns, so the spacer became a rendered cell of its own.

`renderTerminalSnapshotToAnsi` has the same loss on the restore path: the daemon serialises the spacer as `" "`, so replaying the snapshot writes a real space after every wide character and the terminal it replays into allocates a third column.

## Evidence

Captured from a Galaxy Z Fold (SM-F976N, Android 17) with `uiautomator`, reading the terminal row accessibility labels. `python3` printed `한한한한한|AAAAA`:

```text
terminal-row-9  |0000000000000000000000000000000000000000|
terminal-row-11 |한 한 한 한 한 |AAAAA|
```

The ruler row is intact, so the row model — not the font — is inserting the columns. A temporary build that appended each run's `cellCount` to the label showed the whole row as a single 112-cell run whose text already contained the spacers.

Both regression tests fail on `main` with exactly that string:

```text
AssertionError: expected [ { text: '한 글 |', cellCount: 5 } ] to deeply equal [ { text: '한글|', cellCount: 5 } ]
AssertionError: expected '한 글  A…' to contain '한글 A'
```

## Fix

- Treat a reported width of `1` on an intrinsically double-width character as lost information and fall back to the spacer cell, which is the surviving evidence of the second column. A width of `2` or more stays authoritative, so the existing "trust xterm over the code point" behaviour is unchanged.
- Skip the spacer cell when replaying a snapshot to ANSI.
- Move the wide-character range table into `@getpaseo/protocol/terminal-char-width` so the renderer and the snapshot replay share one definition.

## Verification

```text
$ npx vitest run packages/app/src/terminal/native-renderer/terminal-row-model.test.ts packages/protocol/src/terminal-snapshot.test.ts packages/app/src/terminal/runtime/terminal-snapshot.test.ts
Test Files  3 passed (3)

$ npm run typecheck
✓ typecheck (12.10 seconds)

$ npm run lint -- packages/app/src/terminal/native-renderer/terminal-row-model.ts packages/protocol/src/terminal-snapshot.ts

$ npm run format:check

$ ./gradlew assembleRelease --no-daemon --max-workers=2 -Dorg.gradle.parallel=false
BUILD SUCCESSFUL
```

`packages/app/src/terminal/runtime/terminal-snapshot.test.ts` round-trips a snapshot through a real terminal and still passes, so the ANSI replay change does not alter ordinary rows.

## Platform coverage

- Galaxy Z Fold, Android 17: bug reproduced and captured above; the rebuilt APK is not yet re-checked on the device.
- iOS/iPadOS, browser, Electron: not manually tested. The snapshot replay change is shared, and its round-trip test covers it.

### DIFF
--- a/packages/app/src/terminal/native-renderer/terminal-grid-metrics.test.ts
+++ b/packages/app/src/terminal/native-renderer/terminal-grid-metrics.test.ts
@@ -5,6 +5,7 @@ import {
   resolveTerminalGridMetricsMeasurement,
   resolveMeasuredTerminalCellMetrics,
   resolveTerminalCursorOffset,
+  resolveTerminalWideRunLetterSpacing,
 } from "./terminal-grid-metrics";
 
 describe("native terminal grid metrics", () => {
@@ -74,5 +75,23 @@ describe("native terminal grid metrics", () => {
         cellHeight: 16,
       }),
     ).toEqual({ cellWidth: 7.3, cellHeight: 16 });
+  });
+});
+
+describe("wide run letter spacing", () => {
+  it("stretches a narrow wide-glyph advance up to two terminal columns", () => {
+    expect(resolveTerminalWideRunLetterSpacing({ cellWidth: 20, wideAdvance: 27.5 })).toBe(12.5);
+  });
+
+  it("adds nothing when the glyph already advances two columns", () => {
+    expect(resolveTerminalWideRunLetterSpacing({ cellWidth: 20, wideAdvance: 40 })).toBe(0);
+  });
+
+  it("never compresses a glyph that advances past two columns", () => {
+    expect(resolveTerminalWideRunLetterSpacing({ cellWidth: 20, wideAdvance: 48 })).toBe(0);
+  });
+
+  it("ignores an unmeasured advance", () => {
+    expect(resolveTerminalWideRunLetterSpacing({ cellWidth: 20, wideAdvance: 0 })).toBe(0);
   });
 });
--- a/packages/app/src/terminal/native-renderer/terminal-grid-metrics.ts
+++ b/packages/app/src/terminal/native-renderer/terminal-grid-metrics.ts
@@ -27,6 +27,11 @@ export interface TerminalCustomGlyphCellTransformInput {
   cellHeight: number;
 }
 
+export interface TerminalWideRunLetterSpacingInput {
+  cellWidth: number;
+  wideAdvance: number;
+}
+
 export function resolveTerminalCustomGlyphCellTransform(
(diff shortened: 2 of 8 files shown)


lightofjeju:
Second commit added: `ad8aeea39` pads wide glyphs to the two columns the grid allocates them.

Fixing the cell model alone was not enough. A run is drawn as one `Text`, and the font that supplies CJK glyphs advances by its own em rather than by two monospace cells, so the glyphs still walked off the grid inside a run and only snapped back where a style change started the next run. Measured against the ruler row on the device, the pipe in `한한한한한|AAAAA` landed 2.65 columns left of column 10, and the run filled 69% of its allocated width.

The renderer now measures the wide-glyph advance alongside the existing `mmmmmmmmmm` measurement and pads each wide glyph back to two columns. Wide cells no longer share a run with single-column text, since the padding only applies to them.

I also added an end-to-end check of the restore path — daemon-shaped snapshot cells, `renderTerminalSnapshotToAnsi`, a real headless terminal, then the row model — which now reproduces the original row exactly:

```text
label: '메모리 사용량 점검 (2h ago)'
runs : [['메모리', 6], [' ', 1], ['사용량', 6], [' ', 1], ['점검', 4], [' (2h ago)…', 22]]
```

```text
$ npx vitest run packages/app/src/terminal/native-renderer/terminal-row-model.test.ts packages/app/src/terminal/native-renderer/terminal-grid-metrics.test.ts packages/protocol/src/terminal-snapshot.test.ts packages/app/src/terminal/runtime/terminal-snapshot.test.ts packages/app/src/terminal/native-renderer/headless-terminal-output-drain.test.ts packages/app/src/terminal/native-renderer/terminal-custom-glyph.test.ts
Test Files  6 passed (6)
Tests  33 passed (33)

$ npm run typecheck
✓ typecheck (7.97 seconds)

$ npm run format:check
```

Still draft: the device this was captured on disconnected before I could re-check the rebuilt APK.

boudra:
Closing for now: this needs before/after evidence from the patched Android app showing CJK columns stay aligned during normal output and after snapshot restore. Please reopen with that evidence.

boudra:

PR #2984, opened 2026-08-07
fix(pi): adapt message_update to delta-only events from Pi >=0.84

NOknow95:
## Problem

Pi 0.84.0 removed the cumulative `message` field from RPC `message_update` events ([pi #7290](https://github.com/earendil-works/pi/issues/7290)) — they now carry only `assistantMessageEvent` deltas with a `contentIndex`. Paseo's `handleMessageUpdate` read `event.message.role` on every streamed delta, so the first `message_update` threw `TypeError: Cannot read properties of undefined (reading 'role')` inside the JSONL subscriber loop, which has no error boundary. The daemon died mid-turn (`Uncaught exception — daemon crashing` in `daemon.log`), making Pi agents unusable from the desktop app after upgrading pi.

## Fix

- `agent.ts`: `handleMessageUpdate` consumes only `assistantMessageEvent` deltas. The legacy `message` field is used as an optional fallback when the binary predates 0.84 (role filter + `responseId`), tagged `// COMPAT(piMessageUpdateDeltas)`.
- `rpc-types.ts`: `PiAssistantMessageEvent` matches the 0.84 protocol (`contentIndex` plus `text_*` / `thinking_*` / `toolcall_*` deltas); `message_update.message` is now optional.
- `agent.test.ts`: stream tests use the 0.84 delta-only shape; added a regression test that replays a real 0.84 event sequence (thinking + text blocks across `contentIndex`) and a test asserting toolcall deltas/content markers produce no timeline items.

## QA

**Regression test fails on the pre-fix code with the exact production error** (`event.message` is `undefined`):

```
❯ packages/server/src/server/agent/providers/pi/agent.test.ts (56 tests | 1 failed)
    × streams assistant text, reasoning, and tool calls from Pi events
TypeError: Cannot read properties of undefined (reading 'role')
```

**Unit tests** (pi provider, 91 tests):

```
 Test Files  5 passed (5)
      Tests  91 passed (91)
```

**Live probe against pi 0.84.1** (`pi --mode rpc`, raw JSONL): `message_update` events contain only `assistantMessageEvent`; `message_start` / `message_end` / `agent_end` still carry the full message; `tool_execution_start` still fires for bash tool calls.

```
EVENT: {"type":"message_update","assistantMessageEvent":{"type":"thinking_delta","contentIndex":0,"delta":"The"}}
  -> has message field: false | keys: type,assistantMessageEvent
```

**Static checks**: `oxlint` 0 warnings/errors, `oxfmt` clean, `tsgo --noEmit` passes.

## Platform coverage

Server-only change (Pi provider event handling). No UI/platform surface touched. Tested on daemon headless path; not tested on iOS/Android/Web (unaffected — server code).

### DIFF
--- a/packages/server/src/server/agent/providers/pi/agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.test.ts
@@ -500,18 +500,15 @@ describe("PiRpcAgentSession", () => {
     });
     fakeSession.emit({
       type: "message_update",
-      message: { role: "assistant", content: [], responseId: "response-1" },
-      assistantMessageEvent: { type: "text_delta", delta: "hel" },
+      assistantMessageEvent: { type: "text_delta", contentIndex: 1, delta: "hel" },
     });
     fakeSession.emit({
       type: "message_update",
-      message: { role: "assistant", content: [], responseId: "response-1" },
-      assistantMessageEvent: { type: "text_delta", delta: "lo" },
+      assistantMessageEvent: { type: "text_delta", contentIndex: 1, delta: "lo" },
     });
     fakeSession.emit({
       type: "message_update",
-      message: { role: "assistant", content: [] },
-      assistantMessageEvent: { type: "thinking_delta", delta: "thinking" },
+      assistantMessageEvent: { type: "thinking_delta", contentIndex: 0, delta: "thinking" },
     });
     fakeSession.emit({
       type: "tool_execution_start",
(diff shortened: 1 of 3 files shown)


boudra:
Thanks for the fix and the real Pi 0.84 protocol evidence. We are consolidating this in #2978, which covers the same root cause and is verified with before/after real-provider QA. Closing this duplicate in favor of #2978.

PR #2989, opened 2026-08-07
fix(pi): handle pi >= 0.84 message_update events without message field

WhiteWorld:
pi 0.84.0 changed RPC message_update events to emit only assistantMessageEvent deltas, dropping the cumulative message field (pi issue #7290). Paseo's PiRpcAgentSession.handleMessageUpdate reads event.message.role unconditionally, crashing the daemon with 'Cannot read properties of undefined (reading role)' on the first assistant delta, which surfaces as stuck agents / reconnect loops.

Make message optional on the message_update event type and gate the role check on its presence; message_start/message_end remain the authoritative source of the full message.

### Linked issue

Closes #

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Checklist

- [ ] One focused change
- [ ] `npm run typecheck` passes
- [ ] `npm run lint` passes
- [ ] `npm run format` passes
- [ ] QA evidence

### DIFF
--- a/packages/server/src/server/agent/providers/pi/agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.test.ts
@@ -553,6 +553,31 @@ describe("PiRpcAgentSession", () => {
     ]);
   });
 
+  test("streams assistant deltas from pi >= 0.84 message_update events without a message field", async () => {
+    const { pi, session, events } = await createSession();
+    const fakeSession = pi.latestSession();
+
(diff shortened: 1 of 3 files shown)


boudra:
Thanks for the focused fix and regression test. We are consolidating this in #2978, which covers the same root cause and is verified with before/after real-provider QA. Closing this duplicate in favor of #2978.

PR #2992, opened 2026-08-07
feat: render ```mermaid fences as diagrams on web

fredriccliver:
## Problem

Markdown fence blocks (\`\`\`\`lang ... \`\`\`\`) always route through `HighlightedCodeBlock` regardless of language — the render rule in `packages/app/src/components/markdown/renderer.tsx` never branches on the fence's info string. A ```` ```mermaid ```` block (e.g. a gantt chart pasted from a repo's markdown docs) always shows as plain syntax-highlighted text instead of a diagram, because there's no diagram renderer wired up at all yet.

I hit this concretely trying to view a Gantt-chart status dashboard (`docs/projects/mordor/DASHBOARD.md` in an internal repo) in Paseo's file preview — the fence just showed as text.

## What this does

Adds `MermaidDiagram`, a web-only component (`.web.tsx` / native-stub `.tsx` pair, following the existing `markdown-text.web.tsx`/`.ios.tsx` pattern in this codebase) that:

- Dynamic-imports `mermaid` (kept out of the initial web bundle — it pulls in d3, cytoscape, khroma)
- Renders with `securityLevel: "strict"` (the mermaid default — DOMPurify-sanitizes labels, refuses raw HTML/script). This matters because fence content is repo- or agent-authored text, not something the user typed and trusted themselves, matching the trust model this codebase already applies to `FileHtmlPreview`'s sandboxed iframe.
- Falls back to the existing `HighlightedCodeBlock` (i.e. today's behavior) on:
  - parse errors (invalid mermaid source)
  - native (iOS/Android) — there's no SVG renderer there yet; that's a real gap, not attempted in this PR, and explicitly called out below

`renderer.tsx`'s `fence` rule now checks the fence's info string (first whitespace-delimited token, case-insensitive — `isMermaidFence()` in `mermaid-fence.ts`) and routes to `MermaidDiagram` only for `mermaid` fences; every other language is untouched.

## What I deliberately did NOT do

- **Native rendering.** No react-native-svg/WebView work here — native falls back to the pre-existing code-block rendering, so nothing regresses, but mermaid diagrams won't render on iOS/Android after this PR. Flagging this explicitly per the QA guide since it's a real platform gap, not an oversight.
- **Fixing the `expo-clipboard` / Vite JSX-loader gap.** While building this I found that anything statically or dynamically importing `@/components/highlighted-code-block` from a Vitest test hits `The JSX syntax extension is not currently enabled` (`node_modules/expo-clipboard/build/ClipboardPasteButton.js` has untranspiled JSX in a `.js` file, and neither the "unit" nor "browser" vitest project has an esbuild loader override for that). Metro/Babel handles it fine in the real app; Vite's transform doesn't. This is pre-existing and unrelated to fence rendering, so I designed around it (`MermaidDiagram` takes a `fallback: ReactNode` prop built by the caller) rather than fixing it — happy to open a separate issue/PR for that if useful.

## QA

- `npm run typecheck` (packages/app): clean
- `oxlint`/`oxfmt` on changed files: 0 errors, 0 warnings — one justified `// oxlint-disable-next-line react-perf/jsx-no-jsx-as-prop` (the `fence` handler is a `react-native-markdown-display` render rule invoked once per AST node, not a component with a re-render cycle for that rule to protect against)
- `npx vitest run` (packages/app): **475/475 test files, 3990/3990 tests** pass (full suite, including this PR's new tests)
- New tests:
  - `mermaid-fence.test.ts` — unit tests for the fence-language detection helper
  - `mermaid-diagram.browser.test.tsx` — **real Chromium via Playwright** (vitest browser mode), asserting an actual `<svg aria-roledescription="gantt">` renders for valid mermaid source, and that invalid source falls back to the highlighted code block
- Ran the workspace-wide pre-commit hook (`lefthook`) locally — lint/format/typecheck all pass across every package

**Platforms tested:** web only (Chromium, via the browser-mode test above). Not tested: iOS, Android, Electron — native explicitly falls back to unchanged behavior so I don't expect regressions there, but haven't run it on-device.

**Screenshot:** I'll attach the real Chromium screenshot from the test run (a gantt chart rendering as an actual SVG diagram, not raw text) in a follow-up comment.

MermaidDiagram--web--renders-a-mermaid-gantt-fence-as-an-inline-SVG-diagram--not-the-raw-source-1

Opening as a draft since this adds a feature without a prior Discussion thread per CONTRIBUTING.md; happy to open one and link it back here, or take feedback directly on the approach/scope in this PR.


boudra:
I prefer diagram presentation as a plugin, using custom timeline renderers and daemon-side rendering. Those APIs are on main for the upcoming 0.8 release: [plugin documentation](https://paseo.sh/docs/plugins/v0.8/reference#timeline-items). Closing the core PR in favor of that approach.

PR #3008, opened 2026-08-08
feat(app): find workspaces by PR/MR number in the command center

cleiter:
### Type of change

- [x] Enhancement

### Reasoning

When you work across a lot of workspaces you remember the work by its
change-request number - a review comment lands on `#2986` and you want the
workspace for it - not by its branch name. The command center matches
workspaces on title, host, project, and branch, so the number you have at hand
finds nothing and you have to translate it back into a branch name first.

Workspace results now carry their pull/merge request number, so typing it
reaches the workspace directly.

### Goals

- `42`, `#42`, `!42`, `pr 42`, and `mr 42` all resolve to the workspace whose
  PR/MR is 42.
- The number is compared exactly, so `42` does not match PR 142, PR 420, or a
  branch containing those digits.
- An exact number hit sorts above ordinary text matches for the same query.
- Workspaces without a PR/MR behave exactly as before.

### Non-goals

- **No visual change.** The number is a search key only. Result rows render
  exactly as before — no PR/MR badge, no subtitle change. Showing the reference
  is a separate question and I did not want to bundle it here.
- No new Playwright coverage. `seedWorkspace` builds a local repo with no
  remote and the seed client has no seam to attach a `githubRuntime` PR status,
  so a seeded workspace cannot have a change request today. Doing it properly
  needs a fake-forge fixture, which is its own change. Coverage here is
  unit-level.
- No fuzzy or multi-token matching. The number path is exact, and everything
  else keeps the existing substring behavior.

### QA

Opened the command center against a workspace with an open PR and typed the
number. Before, it returned nothing; after, the workspace is the top result.

**Before**

image

image

**After**

image

**Tests:**

- `packages/app/src/command-center/workspace-search.test.ts` (new) — parsing of
  every change-request spelling (`42`, `#42`, `!42`, `pr 42`, `MR!42`), rejection
  of strings that merely contain digits, exact-number matching and the ranking
  flag it sets, and that title/branch text matching still works.
- `packages/app/src/utils/projects.test.ts` (extended) — resolving the number
  onto the workspace summary: the daemon's numeric field wins over parsing the
  URL, the URL is the fallback for GitHub and GitLab, an absent `forge` still
  resolves, and no PR or an unparseable URL gives null.
- `packages/app/src/command-center/results.test.ts` and
  `packages/app/src/screens/projects-screen.test.tsx` (extended) — fixtures only,
  for the new field.

```
Tested:     manual pass above (macOS)
Not tested: iOS, Android
```

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/command-center/command-center.tsx
+++ b/packages/app/src/command-center/command-center.tsx
@@ -56,6 +56,7 @@ import { navigateToAgent } from "@/utils/navigate-to-agent";
 import { formatTimeAgo } from "@/utils/time";
 import { shortenPath } from "@/utils/shorten-path";
 import { useCommandCenterContributions } from "./provider";
+import { filterAndRankWorkspaces } from "./workspace-search";
 import {
   buildContributionSections,
   filterAndRankBuiltInResults,
@@ -121,10 +122,6 @@ function agentSearchFields(result: CommandCenterAgentResult): CommandCenterSearc
   return { visible: [result.title, result.subtitle], hidden: [result.agent.cwd] };
 }
 
-function workspaceSearchFields(result: CommandCenterWorkspaceResult): CommandCenterSearchFields {
-  return { visible: [result.title, result.subtitle], hidden: [] };
-}
-
 /**
  * Build every pinned row, in its default order. Deliberately not keyed on `query`: none of this
  * work depends on what was typed, and rebuilding it per keystroke means re-running an Intl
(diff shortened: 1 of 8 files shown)


boudra:
@paseo-bot deliver

PR #3016, opened 2026-08-08
Fix dictation STT stream finalization

### LINKED ISSUE #3015: Dictation can drop or duplicate transcript text
Dictation can produce incomplete or duplicated text for a single recording.

Observed behavior:
- Longer dictation may miss a large middle section.
- A single recording may insert the same final text twice.

Expected behavior:
- Dictation should preserve spoken audio in order.
- Each recording should insert one final transcript.

yuruiz:
Fixes #3015

## Summary
- Snapshot STT audio at commit time so in-flight transcription cannot clear later recorded audio.
- Guard dictation stream finalization so each stream emits only one final transcript.
- Add regression coverage for OpenAI STT, local Parakeet realtime STT, and re-entrant dictation finalization.

## Tests
- npx vitest run packages/server/src/server/speech/providers/openai/stt.test.ts packages/server/src/server/speech/providers/local/sherpa/sherpa-parakeet-realtime-session.test.ts packages/server/src/server/dictation/dictation-stream-manager.test.ts --bail=1
- npm run typecheck --workspace=@getpaseo/server
- npm run lint
- npm run format:check

## Notes
- Full npm run typecheck still fails in @getpaseo/app because the local react-native-draggable-flatlist patch is not applied; server, client, protocol, and other workspaces pass before that app error.

### DIFF
--- a/packages/server/src/server/dictation/dictation-stream-manager.test.ts
+++ b/packages/server/src/server/dictation/dictation-stream-manager.test.ts
@@ -63,6 +63,15 @@ class FakeSttProvider implements SpeechToTextProvider {
   }
 }
 
+class SyncFinalSession extends FakeRealtimeSession {
+  override commit(): void {
+    this.commitCalls += 1;
+    const segmentId = `seg-${this.commitCalls}`;
+    this.emitCommitted(segmentId);
+    this.emitTranscript(segmentId, `transcript ${this.commitCalls}`, true);
+  }
+}
+
 const buildPcmBase64 = (sampleValue: number, sampleCount: number): string => {
   const samples = new Int16Array(sampleCount);
   samples.fill(sampleValue);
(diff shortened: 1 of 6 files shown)


boudra:
hi @yuruiz can you confirm if you tested this and saw the problem go away?

andkulikov:
Tested this PR successfully on macOS with OpenAI gpt-4o-transcribe.

Before the patch, Paseo 0.8.0 could drop audio recorded while the previous 15-second segment was being transcribed. With this branch, a ~27-second real microphone dictation crossed the auto-commit boundary and remained complete.

I also ran the three targeted suites from the PR description: 16/16 tests passed, including the OpenAI in-flight audio regression and duplicate-finalization coverage.

The fix is confirmed on my setup, and I observed no dictation regressions. This looks safe to merge.

PR #3035, opened 2026-08-08
feat(app): choose when a new workspace opens its Setup tab

cleiter:
### Type of change

- [x] New feature

### Reasoning

Every new workspace in a repo with `worktree.setup` commands gets a Setup tab opened for it, whether or not setup did anything worth looking at. When setup reliably succeeds that tab is noise on every single workspace, and the only way to avoid it is to close it by hand every time. The information you actually want from setup is that it broke. This adds a per-device preference in Settings > General with three modes: Always, the current behaviour and still the default; Until setup succeeds, which opens the tab while setup runs and closes it again on a clean finish; and Only when setup fails. Every mode opens the tab when setup fails, because the Setup tab is the only place a failed setup is visible anywhere in the app, so no setting can turn that signal off.

### Goals

- Add a `setupTabAutoOpen` app setting with three modes: `always` (default, current behaviour), `untilSuccess`, `onFailure`, surfaced as a dropdown in Settings > General.
- `untilSuccess` opens the Setup tab while setup runs and closes it again once the run completes cleanly.
- Every mode opens the tab when setup fails. No mode can hide a failed setup.
- The default is unchanged, so existing behaviour is identical until the setting is changed.
- Auto-close only ever closes a tab the app opened itself in this session. A tab opened from Show Setup, or one restored after a reload, is left alone.
- Auto-close is skipped once the tab has been switched to, so a run that finishes while you are reading the log does not close it under you.

### Non-goals

- No "never" mode. Total suppression was considered and dropped: the Setup tab is the only place a failed setup shows up, so an off switch would silently hide broken workspaces.
- No per-project override. `worktree.setup` is per project, so wanting the tab for one project and not another is coherent, but it needs a second settings tier.
- No protocol or daemon change. The behaviour is entirely client side.
- Changing the preference does not close an already-open Setup tab.

### QA

**Tests:**

- `packages/app/e2e/browser/workspace-setup-auto-open.spec.ts` (new) - one spec per mode against a real daemon: `untilSuccess` opening the tab mid-run and closing it on a clean finish, `untilSuccess` keeping it for a failed run, `onFailure` opening for a failed run and staying away from a clean one.
- `packages/app/src/stores/workspace-setup-store.test.ts` (extended) - the open rule across all three modes by status and snapshot age, the close rule including adopted tabs and failed commands inside a completed run, and that a failed run opens the tab in every mode.
- `packages/app/src/stores/workspace-layout-store.test.ts` (extended) - the tab-adoption predicate, including that a tab alone in its pane is not treated as chosen.
- `packages/app/src/hooks/use-settings/storage.test.ts` (extended) - the new setting's default, a stored value round-tripping, and unknown or non-string values falling back to `always`.
- `packages/app/e2e/browser/workspace-setup-streaming.spec.ts` (extended) - the existing auto-open spec now asserts the Setup tab appears before opening the panel, so it can tell auto-open apart from a manual open.

Two notes on how the e2e specs are built, since both cost a debugging pass:

- They assert on the Setup entry in the tab strip, not through `expectSetupPanel`. That helper falls back to opening the panel from the header menu, and auto-open opens the tab in the background, so any test routed through it passes whether auto-open works or not.
- The auto-close spec creates its workspace through the UI, because `workspace_setup_progress` is only emitted to the session that started the run. It also waits for the app to be told the run completed before requiring the tab to be gone, so "setup never finished" cannot pass as "the tab closed".

**Manual QA (desktop, Electron on macOS):**

- Opened Settings > General, picked each of Always, Until setup succeeds and Only when setup fails from the Setup tab dropdown, and confirmed the choice stuck.
- With Until setup succeeds, created a workspace in a repo that defines `worktree.setup`: the Setup tab appeared while setup ran and disappeared once it finished cleanly.
- Created a workspace whose setup command exits non-zero and confirmed the Setup tab opened and stayed, with Show Setup still available in the header menu.

image

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


cleiter:
Superseded by #3634

PR #3040, opened 2026-08-08
feat(desktop): confirm before quitting when it stops the daemon

cleiter:
### Type of change

- [x] New feature

### Reasoning

Quitting Paseo Desktop is instant and irreversible. Cmd+Q on macOS, or closing the last window on Windows and Linux, tears the app down, and because `daemon.keepRunningAfterQuit` defaults to `false` it also stops the desktop-managed daemon and every agent it was running.

That is not only what a fresh install gets. The `daemonStopOnQuitDefaultApplied` migration reset installs that still had the old `keepRunningAfterQuit: true` value on disk, so stop-on-quit applies to everyone who has not explicitly turned it back on.

I hit Cmd+Q by accident and lost in-progress agent work. Q sits one key from W and one key from Tab, so Cmd+Q is a slip away from both Cmd+W and Cmd+Tab, which are the two things you press constantly while working.

In this app that shortcut destroys running processes doing real work, not an unsaved buffer you can recover. It deserves the same guard any other destructive action gets. This adds a confirmation, but only on the quits that actually stop something.

### Goals

- Confirm before a quit that would stop a running desktop-managed daemon, and only then. If the daemon is already stopped, was started outside the app, or "Keep daemon running after quit" is on, the quit stays immediate and silent.
- Cover every quit entry point rather than just the shortcut: Cmd+Q, Dock Quit, the app menu, and Alt+F4 on the last window.
- Default the dialog to Cancel so pressing Enter does not quit.
- Offer "Leave the daemon running" in the dialog, so quitting the desktop app without interrupting agents is one click at the moment you need it. It applies to that quit only and never writes settings, because "just this once" and "from now on" are different intents and the durable one already exists as "Keep daemon running after quit".

### Non-goals

- No new setting. "Keep daemon running after quit" already silences the prompt, and it does so by removing the risk rather than hiding it. A separate "confirm before quitting" toggle would be a second way to answer the same question, and a no-op whenever the first one is on. The trade-off: there is no way to keep stop-on-quit and never be asked. Happy to add one if you want it.
- No warning for a daemon Paseo did not start. The prompt is scoped to desktop-managed daemons, since Paseo cannot claim ownership of a daemon it did not launch. Quitting with agents on a manually started daemon stays silent.
- The Windows and Linux overlay gap is left as is. Quitting by closing the last window destroys that window before the daemon-stop overlay can render, so the app disappears and the process lingers briefly. This is pre-existing behaviour, documented rather than fixed. The confirmation itself runs before the window is destroyed, so it is unaffected.

### QA

**Tests:**

- `packages/desktop/src/features/quit-confirmation.test.ts` (new) - the whole decision table for whether a quit needs a dialog, the one-shot checkbox and its refusal to leak past Cancel, dialog deduplication when two quit paths race, fail-open on every throwing dependency, and parent-window selection excluding hidden, minimized and destroyed windows.
- `packages/desktop/src/window/window-close-policy.test.ts` (new) - when closing a window should veto on Windows and Linux, including the update-in-flight and session-end exemptions and the destroyed-window counting rule.
- `packages/desktop/src/settings/quit-dialog-copy.test.ts` (new) - round-trip of the localized copy store, plus per-key English fallback for a missing, corrupt or partial file.
- `packages/desktop/src/features/dialogs.test.ts` (new) - that the extracted message-box builder preserves `cancelId` and `defaultId` for the existing confirm paths, and that the quit dialog defaults to Cancel.
- `packages/desktop/src/daemon/quit-lifecycle.test.ts` (extended) - an unconfirmed quit runs none of the side effects below the gate, the one-shot keep-running override outranks the stored setting, and a spurious second quit is not misread as an update handoff.
- `packages/desktop/src/window/window-manager.test.ts` (extended) - a normal Cmd+W close still flushes geometry, and a cancelled quit does not trip the window-state store's one-way finalized latch.
- `packages/desktop/src/settings/desktop-settings.test.ts` (extended) - the synchronous settings accessor agrees with the async one, which is what lets the gate decide before `preventDefault()`.

**Behaviour verified by hand** on macOS, desktop build, against a checkout-local `PASEO_HOME` with a managed daemon running:

- Cmd+Q shows the confirmation. Cancel leaves the app fully usable with agents still streaming.
- Ticking "Leave the daemon running" and choosing Quit exits the app, and the daemon and its agents are still alive afterwards.

Not verified by hand: that Enter defaults to Cancel, the auto-update and Windows session-end exemptions, and anything on Windows or Linux.

image

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/app/_layout.tsx
+++ b/packages/app/src/app/_layout.tsx
@@ -29,6 +29,7 @@ import { AddProjectFlowHost } from "@/components/add-project-flow-host";
 import { WorktreeSetupCalloutSource } from "@/components/worktree-setup-callout-source";
 import { DownloadToast } from "@/components/download-toast";
 import { QuittingOverlay } from "@/components/quitting-overlay";
+import { QuitDialogSync } from "@/desktop/components/quit-dialog-sync";
 import { KeyboardShortcutsDialog } from "@/components/keyboard-shortcuts-dialog";
 import { AppDiagnosticHost } from "@/components/app-diagnostic-host";
 import { LeftSidebar } from "@/components/left-sidebar";
@@ -574,6 +575,7 @@ function AppContainer({ children, chromeEnabled: chromeEnabledOverride }: AppCon
       <KeyboardShortcutsDialog />
       <AppDiagnosticHost />
       <QuittingOverlay />
+      <QuitDialogSync />
     </View>
   );
 
--- a/packages/app/src/desktop/components/quit-dialog-sync.tsx
+++ b/packages/app/src/desktop/components/quit-dialog-sync.tsx
@@ -29,9 +29,12 @@ export function QuitDialogSync(): null {
       quitLabel: i18n.t("desktop.quitConfirm.quit"),
       cancelLabel: i18n.t("desktop.quitConfirm.cancel"),
       keepDaemonRunningLabel: i18n.t("desktop.quitConfirm.keepDaemonRunning"),
-    }).catch(() => {
+    }).catch((error: unknown) => {
       // Non-fatal: main keeps whatever copy it already has, falling back to
-      // its English constants. A failed push must never break the app.
+      // its English constants. A failed push must never break the app, but it
+      // must not be invisible either — an English dialog on a translated app is
+      // otherwise impossible to explain.
+      console.warn("[quit-dialog-sync] Failed to push quit dialog copy", error);
     });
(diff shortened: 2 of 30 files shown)


boudra:

PR #3048, opened 2026-08-08
feat: add file manager content search

nikuscs:
### Linked issue

No matching open issue found.

### Type of change

- [ ] Bug fix
- [ ] Docs

### What does this PR do

Adds workspace content search to the Files pane so you can find text without leaving Paseo.

- The Search action in the Files header switches between the file tree and content search. The file sort control is hidden while search is active because it does not affect search results.
- Results are grouped by file with the match count kept visible. Each group can be collapsed on its own, and the options menu can collapse or expand every group.
- Matching text is highlighted in each line preview. Selecting a match opens the file at the matching line.
- Include and exclude patterns stay visible beside the search field. Match case, Match whole word, and Use regular expression live in the options menu.

The app waits 300 ms before searching so typing does not send a request for every keypress. Newer searches replace older searches, and stale responses cannot replace the current result set.

The daemon prefers ripgrep and streams its JSON output. It falls back to Git grep when ripgrep is not installed. Searches stay inside the workspace root and do not follow results outside it. Runtime, file size, output line size, line preview length, per-file matches, and total matches are all bounded.

The protocol adds correlated `fs.search.request` and `fs.search.response` messages. Support is advertised through the optional `fileContentSearch` feature, and the client checks that feature before sending a request.

### How did you verify it

- Searched a real workspace in an isolated web preview and confirmed grouped results, exact highlights, line navigation, per-file collapse, collapse all, expand all, and every search option.
- Confirmed the live daemon advertises `fileContentSearch` and returns bounded correlated results.
- Protocol, client, daemon, search model, and file navigation coverage: 138 tests passed.
- Re-ran the file search service tests after the final parser cleanup: 6 tests passed.
- `npm run typecheck`
- `npm run lint`
- `npm run format`

### Risk surface

Protocol, daemon, client, and app. The protocol additions are optional and capability-gated. Search commands run only against a validated workspace root and are bounded by cancellation, timeout, file size, output size, and result limits. Hosts without the capability reject the client request before any search message is sent.

### Screenshots

CleanShot 2026-08-11 at 10 06 00@2x
CleanShot 2026-08-11 at 10 07 32@2x

### Checklist

- [x] `npm run lint` passes


boudra:

PR #2504, opened 2026-07-28
feat(app): cycle sidebar grouping with Cmd/Ctrl+Shift+G

cleiter:
### Type of change

- [x] Enhancement

### Reasoning

The sidebar's Group by preference is buried in the Display preferences dropdown, so switching it costs three pointer actions: open the gear menu, move to Group by, click the mode. I switch constantly - Project when I'm looking for something in a specific repo, Status when I want to see what needs attention - and that round trip adds up over a day. This binds Cmd+Shift+G (Ctrl+Shift+G off Mac) to cycle Group by, so a view I change dozens of times a day costs one keystroke.

### Goals

- Cmd+Shift+G on macOS, Ctrl+Shift+G elsewhere, cycles the sidebar Group by preference.
- The shortcut appears once under Settings > Keyboard Shortcuts > Panels and is rebindable there like any other.
- The existing Display preferences dropdown keeps working and its checkmark follows the shortcut.
- The shortcut is inert while the terminal or the command center has focus, matching the sibling panel shortcuts, and still fires with the composer focused.
- Adding a third grouping mode later requires only a new entry in `GROUP_MODE_ORDER`.
- Automated coverage for both platform combos, the suppression contexts, and the help row.

### Non-goals

- No change to what the grouping modes are or how the sidebar renders them; this only changes how you reach the existing preference.
- No mobile surface: keyboard shortcuts are desktop and web only, and the Settings section already says so on native.
- No unbind affordance: a shortcut cannot currently be removed, only rebound or reset to its default, and changing that is a separate change to the shortcut system.
- No Command Center entry for grouping. That is a second discovery surface and belongs in its own change.

### QA

Pressed Cmd+Shift+G with the sidebar visible and watched the grouping switch between Project and Status, with the Display preferences checkmark following it. Opened Settings > Keyboard Shortcuts > Panels and confirmed the Cycle sidebar grouping row shows the combo.

Tested on desktop (Electron, macOS). Not tested: web, and the Ctrl+Shift+G side of the binding on Windows or Linux.

image

https://github.com/user-attachments/assets/97b9fdab-18ff-4379-b0d1-9381676e3039

**Tests:**

- `packages/app/src/keyboard/keyboard-shortcuts.test.ts` (extended) - that each platform's combo resolves to the cycle action and the other platform's modifier does not, that the terminal and the command center suppress it, and that the help row lands in Panels with the right binding id and default keys.
[...]

PR #3063, opened 2026-08-09
feat(app): switch sidebar Group by from the command center

cleiter:
### Type of change

- [x] New feature

### Reasoning

Switching the sidebar between Project and Status grouping takes three clicks today: open the display preferences menu, open Grouping, pick the mode. It is a view preference people flip several times a day depending on whether they are thinking in projects or in what is running right now, and it is the only sidebar preference with no fast path. The command center already covers the rest of the shell, so grouping belongs there too. Typing "group" in Cmd-K now offers a single entry that switches you to the mode you are not in.

### Goals

- One command center entry that switches the sidebar grouping mode, matching "group", "grouping", "sort", "sidebar", "project", and "status".
- The entry always names the mode you are not in, so it never reads as a no-op and you can see where you will land before running it.
- The default empty-query Cmd-K list is unchanged: the entry only appears once you type.
- Focus returns to whatever you were typing in, and the choice persists across a reload, same as the display preferences menu.

### Non-goals

- No keyboard shortcut. #2504 already proposes Cmd/Ctrl+Shift+G for the same thing, so this deliberately does not add a competing binding.
- The other sidebar display preferences (title source, row items, checks, trailing, host filters) stay in the menu. Title source would flip as cleanly, but this keeps the PR to one idea.
- The other root command center entries are left as they are. A pure builder is introduced for this entry only, rather than rewriting the file, so this does not collide with #3013.

### QA

**Tests:**

- `packages/app/src/command-center/root-contributions.test.ts` (new) - the contribution builder in both grouping modes: which mode it offers, which icon it carries, which mode it applies when run, that its id stays stable across modes so the registry tiebreak does not move, and that it stays query-only at the expected rank and group.
- `packages/app/e2e/browser/command-center-grouping.spec.ts` (new) - the entry is absent from the empty-query list, appears on "group" without its opposite alongside it, flips the sidebar from project to status grouping, survives a page reload, and flips back.

Manual pass on the desktop app (Electron, macOS): opened Cmd-K, typed "group", ran the entry and watched the sidebar regroup. Ran it again from the composer and confirmed the cursor returned to the composer instead of being lost. Reopened the palette afterwards, confirmed the entry then offered the way back, and ran it to return to project grouping.

image

image

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/command-center-grouping.spec.ts
+++ b/packages/app/e2e/browser/command-center-grouping.spec.ts
@@ -1,0 +1,58 @@
+import { expect, test, type Page } from "../support/fixtures";
+import { gotoAppShell } from "../support/helpers/app";
+import { openCommandCenter } from "../support/helpers/command-center";
+import { seedWorkspace } from "../support/helpers/seed-client";
+
+const GROUP_BY_STATUS = "Group by status";
+const GROUP_BY_PROJECT = "Group by project";
+
+// Result rows carry no testID of their own, so the entries are addressed by their visible label.
+async function runGroupingEntry(page: Page, label: string, absent: string): Promise<void> {
+  const panel = await openCommandCenter(page);
+  await panel.getByTestId("command-center-input").fill("group");
+
+  const entry = panel.getByText(label, { exact: true });
+  await expect(entry).toBeVisible({ timeout: 30_000 });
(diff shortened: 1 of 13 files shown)


boudra:
@paseo-bot /deliver this PR

PR #3064, opened 2026-08-09
feat(app): show the project icon in the workspace header

cleiter:
### Type of change

- [x] Enhancement

### Reasoning

The sidebar row shows a workspace's project icon; the workspace header does not. Switching between workspaces from different projects means reading the header text to work out where you are, and on wide layouts the project name is dropped whenever it repeats the workspace name, so there is nothing to read. This adds the icon to the header, leading the title, always visible.

### Goals

- Show the project icon at the start of the workspace header title, at the same 16pt size the sidebar uses.
- Keep the icon visible even when the header drops the redundant text project name on wide layouts.
- Use the same icon the sidebar row shows, including the generated fallback's colour, so a project reads the same in both places.
- Align the icon to the title's line box on compact, where the title and project name stack, instead of centring it between the two lines.

### Non-goals

- No status badge on the header icon. The sidebar's corner badge stays sidebar-only; the header carries agent state elsewhere.
- The icon is not interactive. No press target, no menu, no navigation.
- No placeholder while the project resolves. The generated icon's colour comes from the project's view key, which is not derivable from the workspace descriptor, so a guessed icon would show one colour and then flip to another. The header renders nothing until the real key is known.
- No project icon on other surfaces such as tabs, terminal headers, or browser panes.

### QA

**Tests:**

- `packages/app/src/screens/workspace/workspace-header-project-icon.test.tsx` (new) - the component's five outcomes: no workspace and unresolved project both render nothing and issue no icon request, the daemon-provided image renders, the generated initial is the fallback, the icon survives the header dropping the text project name, and the icon target is taken from this workspace's host rather than the first placement.
- `packages/app/src/utils/sidebar-project-row-model.test.ts` (extended) - the new `preferredServerId` argument: the requested host wins, and the walk falls back to the first usable placement when that host is absent or has no working directory.

Manual, in the desktop app (Electron, macOS): opened a workspace and saw the daemon-provided project icon at the start of the header title. Checked a project with no uploaded icon and confirmed the header's generated square is the same colour and carries the same initial as its sidebar row. Narrowed the window to the compact layout, where the title and project name stack, and confirmed the icon sits on the title's row rather than centred between the two lines.

Project with icon:

image

Project with fallback:

image

Compact:

image

`resolveSidebarProjectIconTarget` is shared with the sidebar, so its existing callers were re-run: `packages/app/src/components/sidebar-workspace-list.test.tsx` and `packages/app/src/projects/host-projects.test.ts` pass unchanged. The new argument is optional and every existing call omits it.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:
Thanks for taking the time on this. I don’t want project icons in the workspace header, so I’m closing it.

Issue #3061, opened 2026-08-09
Daemon restarts silently kill active provider turns, drop notifyOnFinish, and leave no shutdown trace


sjpritchard:
Environment: Paseo 0.3.0 daemon, Windows 11, Codex and Claude providers, orchestration over the MCP agents endpoint (`http://127.0.0.1:6767/mcp/agents`).

We drive multi-agent change workflows through Paseo (one orchestrator, several delegate agents on long 10–30+ minute turns). A daemon restart mid-turn produced a cluster of related supervision gaps:

1. **Active turns do not survive a daemon restart.** A restart mid-turn aborts the provider turn (the Codex rollout records `turn_aborted / reason: "interrupted"`); after the restart the agent snapshot reloads as `idle` with no final message. From an orchestrator's view the agent looks finished-but-silent; the only evidence is in the provider's own session log. Expected: either turn resume, or an explicit `aborted` state on the agent snapshot.

2. **`notifyOnFinish` subscriptions die with the daemon process.** A pending finish-notification is never delivered after a restart, and nothing records that it was dropped. Orchestrators must poll as a fallback. A persisted-and-redelivered notification, or an on-reconnect "these turns aborted while you were away" digest, would fix it.

3. **The daemon logs nothing on shutdown.** Two abrupt restarts show metrics logging simply stopping mid-stream with no shutdown/signal/crash line, so restart forensics start from negative evidence (PID-range comparison in `daemon.log`). A shutdown reason (signal, updater, crash handler) in the log would help.

4. **Prompts to a busy agent hard-fail.** `send_agent_prompt` to an agent with an active run raises `already has an active run` (logged at level 50) instead of queueing or returning a typed busy status. A queue-or-clean-reject contract would remove a race every orchestrator currently has to code around.

5. Minor, same theme of provider supervision: `Claude query operation did not settle cleanly — "ProcessTransport is not ready for writing"`, and a relay client reconnect loop churning connect/disconnect every few seconds.

Log excerpts are available for all five (daemon.log PID ranges 69280 → 50332, Codex rollout session `019fe3f5…`); happy to attach any of them.

sjpritchard:
App version: **0.3.0** (installed desktop app; worker logs report `daemonVersion: "0.3.0"`), Windows 11.

[...]

Issue #2495, opened 2026-07-27
Child completion callbacks are dropped when the parent agent has an active run


kdegeek:
## Summary

When a child agent completes, Paseo attempts to notify its parent. If the parent currently has an active run, the notification path rejects the delivery with `Agent <id> already has an active run`. The failure is logged but not queued or retried, so the parent never receives the child completion.

## Why this matters

This breaks the core orchestration pattern where one long-lived coordinator dispatches child agents and continues working while waiting for their results. Completion reports can disappear permanently, particularly for ACP providers such as Grok.

## Observed behavior

- Child agents completed.
- Paseo logged `Failed to notify caller agent`.
- The underlying error was `Agent <id> already has an active run`.
- Multiple delivery attempts for the same parent were rejected.
- No later callback, inbox item, or retry was delivered after the parent became idle.

## Expected behavior

Child completion should be durably queued against the parent when immediate delivery is unsafe, then delivered when its active run settles. Delivery should be idempotent and survive reconnects/restarts. A visible pending-completion state would also make failures diagnosable.

## Related areas

This appears related to active-turn follow-up handling and background-task notification behavior, but is specifically about Paseo-managed child completion callbacks being lost rather than deferred.

PR #3065, opened 2026-08-09
fix(server): don't report a still-tracked run as canceled

sjpritchard:
### Linked issue

Refs #3061 (item 4), related to #2495

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Sending a prompt to an agent can fail with `Agent <id> already has an active run` even when the caller did nothing wrong. It happens while orchestrating agents over MCP: a coordinator sends a prompt, the provider is slow to accept the turn, and the next prompt is rejected with an error that reads like the caller raced itself. It didn't — the daemon's own bookkeeping is stale.

`streamAgent` creates the pending foreground run synchronously, but its `turnId` is only assigned once `session.startTurn()` returns. That provider round-trip is a window where `run.turnId === null`. `settleTerminalRun` early-returns for a foreground run it cannot match by turn id, so during that window *nothing* can clear the run except its own stream generator — no force-cancel reaches it.

`cancelAgentRun` doesn't account for that. When settlement times out in this window it matches neither force-cancel branch (one needs `run.turnId`, the other needs `kind === "autonomous"`) and falls through to `return { status: "settled" }` while the run is still tracked. `cancelAgentRunBefore` only throws on `refused`, so `replaceAgentRun` continues into `streamAgent`, which sees the tracked run and throws.

So a prompt that should have produced a clear "this agent is busy" signal instead produces an error blaming the caller, and orchestrators have to pattern-match on the message to work around it.

This reports `refused` when the run is still tracked after cancellation, so callers get the typed `AgentRunCancellationError` the codebase already defines for exactly this case.

### Goals

- A cancellation that leaves the run tracked is reported as `refused`, never `settled`.
- `replaceAgentRun` fails with `AgentRunCancellationError` instead of `already has an active run` when the underlying run could not be cleared.
- A warn-level log records the agent, run kind, and turn id when this happens, so it is diagnosable in `daemon.log`.

### Non-goals

- No prompt queueing. #3061 asks for "queue or clean reject"; this is the clean-reject half. Queueing is a larger behavioral change and belongs in its own PR.
- No change to the cases that already work: an acknowledged interrupt that settles still returns `settled`, and an unacknowledged one still returns `refused`.
- Does not change how a pending foreground run gets cleared. Making the null-turn-id window itself cancellable is a deeper change to run tracking; this PR only stops the manager from reporting success it didn't achieve.

### QA

Added `agent-manager-cancel-pending-run.test.ts`, which drives the real failure through `replaceAgentRun` using a session whose `startTurn` never resolves and whose `interrupt` resolves (so cancellation is "acknowledged" but nothing settles).

Reproduced the bug before the fix — the test fails with exactly the error from the field report:

```
AssertionError: expected [Function] to throw error including 'Cannot replace agent 00000000-0000-40…'
  but got 'Agent 00000000-0000-4000-8000-0000000…'
Received: "Agent 00000000-0000-4000-8000-000000003061 already has an active run"
```

Confirmed fixed after — 1 passed.

Regression runs on `packages/server`:

- `agent-manager.test.ts`, `rewind/`, `agent-prompt.test.ts`, `agent-manager-stream-coalescing.test.ts`, plus the new test — **190 passed**
- `npm run test:unit -- src/server/agent` — **1697 passed**, 17 skipped. The 2 failures in that run (`providers/claude/models.test.ts`, `mcp-server.test.ts`) reproduce identically on an unmodified checkout of `01a1d3b`, so they are pre-existing and unrelated.
- `npm run typecheck` — clean
- `npm run lint` — 0 warnings, 0 errors
- `oxfmt --check` on both changed files — correctly formatted

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/agent-manager-cancel-pending-run.test.ts
+++ b/packages/server/src/server/agent/agent-manager-cancel-pending-run.test.ts
@@ -1,0 +1,147 @@
+import { expect, test } from "vitest";
+
+import { createTestLogger } from "../../test-utils/test-logger.js";
+import { AgentManager } from "./agent-manager.js";
+import type {
+  AgentCapabilityFlags,
+  AgentClient,
+  AgentPermissionRequest,
+  AgentPersistenceHandle,
+  AgentProvider,
+  AgentRunResult,
+  AgentSession,
+  AgentSessionConfig,
+  AgentStreamEvent,
+} from "./agent-sdk-types.js";
+
+const CAPABILITIES: AgentCapabilityFlags = {
+  supportsStreaming: true,
+  supportsSessionPersistence: true,
+  supportsDynamicModes: false,
+  supportsMcpServers: false,
+  supportsReasoningStream: false,
+  supportsToolInvocations: false,
+  supportsRewindConversation: false,
+  supportsRewindFiles: false,
+  supportsRewindBoth: false,
+};
+
+/**
+ * A session whose `startTurn` never resolves, modelling a provider that is slow to
+ * accept a turn. The manager has a pending foreground run for it, but that run has
+ * no turn id yet. `interrupt` resolves, so the manager treats the cancellation as
+ * acknowledged even though nothing can settle the pending run.
+ */
+class StalledStartTurnSession implements AgentSession {
+  readonly provider: AgentProvider = "claude";
+  readonly id = "stalled-start-turn-session";
+  readonly capabilities = CAPABILITIES;
+  interruptCount = 0;
+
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the pending-run cancellation implementation merged in [#3742](https://github.com/getpaseo/paseo/pull/3742).

PR #3081, opened 2026-08-09
fix(desktop): add a persistent macOS daemon LaunchAgent

timigod:
## Problem

The desktop-managed daemon is coupled to a detached child process. A GUI or `launchd` start can also have a reduced `PATH`, which makes provider command discovery unreliable.

## Change

- Add a packaged, non-interactive daemon launcher.
- Prepend stable macOS command paths and preserve inherited `PATH` entries.
- Add an owned LaunchAgent definition with atomic reconciliation.
- Refuse to replace an unmanaged service file.
- Do not load, restart, or reload the service during reconciliation.
- Reconcile before graphical startup without creating another window.

The service label and paths are generic to upstream Paseo.

## Verification

- Focused desktop tests: 19 passed.
- Desktop main build: passed.
- Desktop and repository typechecks: passed.
- Lint and format checks: passed.
- Launcher shell syntax check: passed.

Based directly on Paseo `v0.3.0`.

### DIFF
--- a/packages/cli/src/commands/daemon/local-daemon.supervision.test.ts
+++ b/packages/cli/src/commands/daemon/local-daemon.supervision.test.ts
@@ -1,4 +1,5 @@
 import { EventEmitter } from "node:events";
+import { existsSync } from "node:fs";
 import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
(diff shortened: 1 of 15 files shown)


timigod:
Addressed the explicit-stop review finding in `3386d13e21796828a08dd1306d149041d6d5433d`.

- Launchd now uses `KeepAlive.SuccessfulExit = false` instead of unconditional `KeepAlive`.
- The existing supervisor exits `0` after a supported lifecycle shutdown, so an explicit stop stays stopped.
- The supervisor exits nonzero after an unexpected failure, so launchd restarts a crashed daemon.
- Only an unmanaged LaunchAgent ownership conflict remains warning-only. Unexpected reconciliation failures now stop Desktop startup instead of being hidden.

Verification passed: 4 focused LaunchAgent tests, 9 combined Desktop tests, full workspace typecheck, repository lint and formatting, and the Desktop main build.

boudra:

PR #2890, opened 2026-08-05
fix: show imported Codex session titles in workspaces

qmm0523:
## Summary
- sync an imported provider session title onto a newly created workspace
- preserve an existing workspace title when it was set manually
- add a regression test for imported workspace titles

## Root cause
Codex import persisted the session title on the agent record, but the newly provisioned workspace kept `title: null`. Workspace display then correctly fell back to the branch name, making the imported session appear untitled in the workspace list.

## Validation
- `npx vitest run packages/server/src/server/session.workspaces.test.ts --bail=1 -t "import_agent_request registers a workspace for a never-seen cwd"`
- `npm run typecheck:server`
- `npm run lint`
- `npm run format`
- pre-commit hook: full workspace typecheck, lint, and format check

The full workspace test file also ran; the changed test passed, while an unrelated existing worktree archive test exceeded its 5-second timeout.

Issue #2636, opened 2026-07-30
feat: make older sessions reachable in Import Session (search or pagination beyond the recent window)


gentoosys:
### Prior search

- [x] I searched existing issues and discussions, and this isn't already proposed.

Nothing found about reaching sessions beyond the recent window. The Import Session issues I could find are about other aspects of it: #2513 (titles), #2512 (spinner), #2574 (sessions vanishing), #1279 (import timeout), #2065 (listing performance).

### What's the problem

Import Session only ever offers the most recent sessions, and there is no way to reach anything older. The sheet requests a fixed page:

```ts
// packages/app/src/components/import-session-sheet-view-model.ts
export const PER_PROVIDER_LIMIT = 15;
```

and that number is the whole window. There is no search field in the sheet, no "load more", and no way to raise the limit — the only controls are a provider filter and a refresh. On the server the same limit bounds how far the provider even looks:

```ts
// providers/claude/agent.ts, listImportableSessions()
const limit = options?.limit ?? 20;
const candidates = await collectRecentClaudeSessions(sessionsRoot, limit * 3, ...);
...
return parsed.filter(...).slice(0, limit);
```

Sessions are ordered by mtime, so anything that hasn't been touched recently falls off the end and becomes unreachable through the UI. One of my project directories has 116 Claude Code sessions in it; 15 of them are offerable at any time.

The part that makes this more than a papercut is that the natural workaround is circular. To pull an older session up the list you have to bump its mtime by opening it in Claude Code — but to do that you must already know which session it is, and finding it is exactly what the import sheet is for. I went through this yesterday: opened a session in the CLI and sent a message purely to make it recent enough to appear.

It's also worth noting the code already recognises that the window can be silently eaten, and compensates for one cause of it:

```ts
// agent/import-sessions.ts
const sessions = await agentManager.listImportableSessions({
  limit: limit + importedSessions.count,
  ...
```

Already-imported sessions are added back to the budget so they don't shrink the visible list. That's the right instinct — this request is the same concern applied to sessions that are merely *old* rather than already imported.

### What would solve it

Any one of these would resolve it; the first is the most useful:

1. **Search in the Import Session sheet.** A text field matching on title and prompt preview. It has to be server-side, since the client only ever holds 15 descriptors — a client-side filter over the current page would look like search while still being unable to find anything older.
[...]

Issue #2513, opened 2026-07-28
bug: Claude Code /rename titles are ignored in Import Session


isCopyman:
### What's broken

Claude Code sessions renamed with the native `/rename` command still appear in Paseo's **Import Session** sheet under their first user prompt instead of the saved custom title.

Claude Code successfully persists the rename as a `custom-title` JSONL record and shows `Session renamed to: ...`. Reopening/resuming the native Claude Code session retains that title. Paseo's import listing does not display it.

Expected behavior: when a Claude session has a valid `custom-title` record, Paseo should use that title in the Import Session list. A reasonable precedence would be the latest `custom-title`, then `ai-title`, then the first user prompt as a fallback.

### Steps to reproduce

1. Start a Claude Code session and send an ordinary prompt.
2. In that same Claude Code session, run `/rename My research session`.
3. Confirm Claude prints `Session renamed to: My research session`.
4. Close or leave the session available for resume.
5. Open Paseo desktop and select **New Agent → Import Session → Claude**.
6. Find the session.
7. Observe that Paseo displays the original first prompt rather than `My research session`.

I reproduced this with Claude Code 2.1.220 and Paseo 0.2.3.

### Where did this happen

Desktop (Electron)

### Paseo version

0.2.3

### OS version

Windows 11 Pro for Workstations, 10.0.26200 build 26200

### Agent provider

Claude Code

### Provider configuration

Claude Code 2.1.220. The behavior reproduces with a normal native Claude Code JSONL session. The API backend does not affect the stored title record.

### Logs

No daemon error is emitted. A reduced synthetic example of the relevant Claude session records is:

```jsonl
{"type":"user","message":{"role":"user","content":"Review this project"},"cwd":"C:\\workspace","sessionId":"example-session"}
{"type":"custom-title","customTitle":"My research session","sessionId":"example-session"}
{"type":"system","subtype":"local_command","content":"<local-command-stdout>Session renamed to: My research session</local-command-stdout>","cwd":"C:\\workspace","sessionId":"example-session"}
```

Paseo lists this session as `Review this project`, not `My research session`.

### Additional reproducible observation

In the current Claude import descriptor parser, the first usable user entry supplies `title`, `sessionId`, and `cwd`. Once those three values exist, the scan stops. `custom-title` and `ai-title` records are not applied to the descriptor, so native Claude renames later in the JSONL cannot affect the import row.

### Screenshots or video

[...]

PR #3092, opened 2026-08-09
fix(server): name an imported Claude agent after its /rename

platanus-kr:
### Linked issue

Refs #2513. Closed by #4216, which fixed the Import Session listing.

### Type of change

- [x] Bug fix

### Reasoning

Renaming a Claude Code session with `/rename` is how you keep parallel sessions
apart. Paseo dropped that name on the way in: Import Session listed every Claude
session under its first prompt, and the imported agent was named after that same
prompt. When the first prompt is `/model` the row carries no usable name at all,
so the mapping had to be rebuilt by hand for every session.

Claude stores the name as its own transcript record: `custom-title` for
`/rename`, `ai-title` for the ones it generates. #4216 landed the listing side,
so the sheet now shows the renamed session. Importing it still produces an agent
named after the first prompt, which is the half its non-goals left open. This
reads the record at import time so the name comes across.

### Goals

- The imported agent keeps an explicit `/rename` as its title.
- The last rename wins when a session was renamed more than once.

### Non-goals

- The listing. #4216 owns it; this PR does not touch the descriptor parser,
  the accumulator or `normalizeClaudeSessionTitle`.
- Naming imported agents from `ai-title`. Agents are named after their first
  prompt line since #1563; taking Claude's generated title would put that LLM
  summary back. Only an explicit rename is carried over.
- Other providers. Codex is being handled separately in #2890.
- Keeping the title in sync after import. This reads it once, at import time.
- Reaching sessions outside the recent window (#2636).

### QA

All evidence uses a generated fixture, not a real `~/.claude`.

The new `agent.import-titles.test.ts` drives the real `importSession` against a
temporary `CLAUDE_CONFIG_DIR`, with `resumeSession` stubbed to an empty history
so the resulting config is observable without spawning Claude Code.

| session | expected `config.title` |
| --- | --- |
| renamed with `/rename` | the rename |
| renamed more than once | the last one |
| later record carries no usable name | the earlier rename survives |
| rename recorded on a sidechain transcript | unset |
| never renamed | unset |
| only a generated `ai-title` | unset |
| no transcript on disk | unset |
| transcript unreadable (EISDIR) | throws |

```
$ npx vitest run src/server/agent/providers/claude/agent.import-titles.test.ts
 Test Files  1 passed (1)
      Tests  8 passed (8)

$ npx vitest run src/server/agent/providers/claude/agent.test.ts
 Test Files  1 passed (1)
      Tests  75 passed (75)

$ npm run typecheck   # all workspaces, exit 0
$ npm run lint        # 0 warnings, 0 errors
$ npm run format
```

`agent.test.ts` is the existing suite for this file, including the
`listImportableSessions` cases #4216 added. Unchanged and still passing.

Daemon-side only; no client code is touched. Tested on macOS.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


fighterhit:
need this feature! cc @boudra

platanus-kr:
Hi @boudra!

I'm one of the people happily using Paseo.
I'd love it if you could review this PR so that people using Claude Code can move over to Paseo more easily!
Please take a look!

platanus-kr:
Reduced to the part #4216 left open: the sheet shows the renamed session, but
importing it still names the agent after its first prompt.

Rebased onto current main as a single commit. The listing is untouched.

PR #3094, opened 2026-08-09
fix(server): notify the caller when a detached child finishes

thomasvan:
## Problem

`create_agent` with `relationship: { kind: "detached" }` and `notifyOnFinish` never notifies. The two options are silently incompatible today — no error, nothing in the logs, the caller just waits forever.

`resolveCreateAgentIntent` strips `paseo.parent-agent-id` for a detached child [on purpose](https://github.com/getpaseo/paseo/blob/main/packages/server/src/server/agent/create-agent/intent.ts):

```ts
if (input.legacyDetached) {
  delete labels[PARENT_AGENT_ID_LABEL];
}
```

But `createAgentCommand` then arms the finish watcher with `requireParentOwnership: true` unconditionally. When the child finishes, the watcher compares its (absent) parent label against the caller:

```ts
if (requireParentOwnership && getParentAgentIdFromLabels(record?.labels) !== callerAgentId) {
  return;
}
```

`null !== callerAgentId` is always true, so the notification is dropped every time. An agent that deliberately creates a detached child *and* asks to be told when it finishes is guaranteed never to hear back.

Measured in an orchestration setup that spawns detached children: the child finished at `13:43:39Z` and its creator did not wake until `14:07:25Z` — 23m46s later, and only because an unrelated scheduled task happened to run. Without that schedule it would have hung indefinitely.

## Fix

```ts
requireParentOwnership: !(input.detached ?? false),
```

Children that carry a parent label are unaffected — the guard still stops notifying a caller once the child is re-parented away, which is what `agent-prompt.test.ts` → *"detaching a child ends its parent-owned finish notification"* covers. That test still passes untouched.

## Why not fix it in `setupFinishNotification`

That was my first attempt (snapshot ownership when the watcher arms, enforce it only for a child that was parented then). It cannot work: at that layer, a child **created** detached and a child **detached after the fact** are the same input — neither has a parent label. The existing test above constructs exactly the former and asserts silence, so narrowing the guard there breaks it.

Only the create call site knows the caller explicitly asked for a detached child, so the decision belongs there.

## Tests

Two tests appended to `create.test.ts`. They drive the **real** finish watcher through `createAgentCommand` and assert the caller is actually prompted — no mocking of `setupFinishNotification`, so the assertion is "did the caller get woken", not "which flag was passed". Both the detached and parented cases are covered; the detached one fails without the change.

- `npx vitest run src/server/agent/create-agent/create.test.ts` → 12 passed
- `npx vitest run src/server/agent/agent-prompt.test.ts` → 7 passed (the guard's existing coverage)
- `npm run typecheck`, `npm run lint`, `npm run format:check` → clean

Verified end-to-end on a live daemon as well: a caller spawned a detached child (`labels: {}` confirmed on the stored record), ended its turn, and received the finish notification 3.9s later.

## Note

Independent of #2879 — that one is entirely in `agent-prompt.ts`, this is one argument in `create.ts`. They can land in either order.

### DIFF
--- a/packages/server/src/server/agent/create-agent/create.test.ts
+++ b/packages/server/src/server/agent/create-agent/create.test.ts
@@ -10,7 +10,7 @@ import { AgentManager } from "../agent-manager.js";
 import { AgentStorage } from "../agent-storage.js";
 import type { CreatePaseoWorktreeWorkflowResult } from "../../worktree-session.js";
 import { createAgentCommand } from "./create.js";
-import type { ManagedAgent } from "../agent-manager.js";
+import type { AgentManagerEvent, ManagedAgent } from "../agent-manager.js";
 
 const logger = createTestLogger();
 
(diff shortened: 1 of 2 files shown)


thomasvan:
Status update: I still carry this patch on `slp/patches` (fork tip `04563655b`), source-reverified through the 2026-09-07 sync to upstream `c424f8292` (0.7.2). Merge-tree against current `main` (`8c89e72a0`) is clean; the detached-child finish-notification fix remains needed — upstream still requires parent ownership for a child that was deliberately created detached.

PR #3097, opened 2026-08-09
feat: bridge Claude prompt suggestions

mikeyzhe:
### Linked issue\n\nNone — this bridges an existing Claude Agent SDK event for a downstream self-hosted client workflow.\n\n### Type of change\n\n- [ ] Bug fix\n- [x] New feature\n- [ ] Enhancement\n- [ ] Refactor\n- [ ] Docs\n\n### Reasoning\n\nClaude Agent SDK can emit a prompt_suggestion after a successful foreground turn, but Paseo currently drops that event. Downstream clients therefore cannot offer the provider-authored next prompt without owning a second Claude integration. This change carries the suggestion as an optional, capability-gated stream event while keeping older clients and non-Claude providers unchanged.\n\n### Goals\n\n- Preserve the originating foreground turn ID after its successful result so a later suggestion stays attached to the correct turn.\n- Emit only nonblank suggestions up to 500 Unicode code points.\n- Advertise and gate the event with the prompt_suggestions client capability, including per-source fanout.\n- Enable Claude suggestions by default while honoring an explicit provider option of false.\n- Clear retained targets on new turns, failure, close, and session mismatch.\n\n### Non-goals\n\n- Add or change client UI.\n- Generate suggestions inside Paseo.\n- Add suggestion support to providers that do not emit this event.\n- Treat suggestions as permissions or required responses.\n\n### QA\n\n- Targeted Vitest run across the new protocol/provider tests plus messages, session, and relay reconnect coverage: 193 tests passed.\n- npm run format:check: passed.\n- npm run typecheck: passed.\n- npm run lint: passed.\n- Server/protocol-only change; no mobile, web, or desktop UI surface changed.\n\n### Checklist\n\n- [x] One focused change\n- [x] npm run typecheck passes\n- [x] npm run lint passes\n- [x] npm run format passes\n- [x] QA evidence\n- [x] Tests added or updated where it made sense

### DIFF
--- a/packages/protocol/src/agent-types.ts
+++ b/packages/protocol/src/agent-types.ts
@@ -375,6 +375,13 @@ export type AgentStreamEvent =
     }
   | { type: "turn_canceled"; provider: AgentProvider; reason: string; turnId?: string }
   | {
+      type: "prompt_suggestion";
+      provider: AgentProvider;
+      turnId: string;
+      suggestion: string;
+      messageId: string;
+    }
+  | {
       type: "timeline";
       item: AgentTimelineItem;
       provider: AgentProvider;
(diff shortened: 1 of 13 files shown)


mikeyzhe:
Closing as stale — conflicted with main and superseded. Branch `feat/prompt-suggestion-quick-replies` (bb74d983) is kept if this is revived.

PR #3119, opened 2026-08-10
fix(workspace): report archive cleanup outcome

timigod:
## Summary

- expose the existing workspace archive `removedDirectory` result in the daemon response
- report `true`, `false`, or `null` from `paseo workspace archive`, with `null` preserving compatibility with older daemons
- keep exact-ID archive retryable so a retained busy managed worktree can be removed once it becomes safe

The archive service already returned the cleanup result, but the session response discarded it. The CLI therefore reported `archived` without saying whether the physical managed directory remained. This change carries the service result through the optional wire field and makes cleanup state explicit without weakening any archive safety check.

## Verification

- protocol and CLI focused tests: 10 passed
- archive session tests: 2 passed, 107 skipped by focus filter
- exact-ID retained-directory retry: 1 passed, 7 skipped by focus filter
- `npm run build:server`
- `npm run typecheck`
- touched-file format, lint, and `git diff --check`

The retry regression proves `removedDirectory: false` while a writer keeps the directory busy, followed by `removedDirectory: true` when the same archived workspace ID is retried after the directory becomes safe.

### DIFF
--- a/packages/cli/src/commands/workspace/archive.test.ts
+++ b/packages/cli/src/commands/workspace/archive.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from "vitest";
+import { buildWorkspaceArchiveResult } from "./archive.js";
+
+describe("workspace archive receipt", () => {
+  it.each([
+    { removedDirectory: true, expected: true },
+    { removedDirectory: false, expected: false },
+    { removedDirectory: undefined, expected: null },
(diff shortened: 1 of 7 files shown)


boudra:
Closing in favor of the workspace archive implementation in [#2987](https://github.com/getpaseo/paseo/pull/2987).

PR #3137, opened 2026-08-10
Persist hidden provider subagents across restarts

etinpres:
## Summary

- persist client-side hidden state for finished provider subagents
- keep hidden provider rows dismissed across app restarts and provider history replay
- add a parent-agent tab action to restore hidden provider subagents on desktop and mobile
- update lifecycle documentation and translations

## Problem

The existing **Hide finished** action only updated in-memory state. After a client restart or a fresh provider history replay, completed provider subagents appeared above the composer again. On mobile, even a collapsed track consumes a line of limited screen space.

## Implementation

Only the hidden provider-subagent keys are persisted through Zustand and AsyncStorage. Runtime descriptors and timelines remain transient. If a hidden subagent becomes active again, the provider update reveals it automatically. The parent agent tab menu exposes **Show hidden subagents** when restorable history exists.

This is intentionally client-local and does not add a daemon or protocol mutation.

## Verification

- focused tests: 4 files, 38 tests passed
- app TypeScript check passed
- formatting check passed
- `git diff --check` passed
- focused lint: 0 errors (3 pre-existing warnings in `workspace-screen.tsx`)

The regression coverage includes two completed provider subagents named `oneshot_supervisor` and `taxi_supervisor`, followed by store recreation and provider-list replay.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -150,7 +150,12 @@ Claude Code announces subagent lifecycle on the SDK stream (`task_started` / `ta
 
 Archived Paseo subagents disappear from the track, by design. To remove one from the track without closing its tab, use the **archive button** on the row — it opens a confirm dialog and archives the subagent on confirm. Provider-owned rows have no individual Paseo lifecycle controls.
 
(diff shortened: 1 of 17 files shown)


etinpres:
Closing because the motivating user-visible issue is resolved in the current Paseo release. The two stale Codex provider subagents that previously remained active and could not be dismissed no longer appear after the upstream Codex lifecycle fix in #3188, so this persistence change is no longer needed for the original report. Thanks!

Issue #964, opened 2026-05-12
bug: The latest version of Mac often can't reconnect after being interrupted.


magicliuxd-dev:
### What's broken

I'm using version Mac0.1.75, and if I manually terminate a task and then try to restart it, the following error occurs. Ah, this error happens all the time.[System Error] Error running remote compact task: unexpected status 503 Service Unavailable: upstream connect error or disconnect/reset before headers. reset reason: connection termination, url: https://chatgpt.com/backend-api/codex/responses/compact, cf-ray: 9faaed8dabf3f8ce-SIN

### Steps to reproduce

Step one: Manually interrupt a task.
Step two: Just let it continue, or let it continue after changing the requirements.

### Where did this happen

Desktop (Electron)

### Paseo version

0.1.75

### Agent provider

None

### Logs

```text

```

boudra:
The captured error is a 503 from Codex’s upstream compaction service. Closing pending the requested current reproduction. Please reopen with diagnostics if interrupting and continuing still leaves the Paseo session unusable.

PR #3146, opened 2026-08-10
perf(client): reconnect immediately on app resume, skipping backoff

italianpastry8:
## Related issue

Related to #964 (reconnect after interruption is slow/unreliable).

## Problem

When the app returns from the background, the WebSocket to the daemon is almost always dead, but recovery waits for the first-attempt backoff (~1500ms) before even trying to reconnect, surfacing a ~3s "Reconnecting" toast on every resume.

## Solution

Add `DaemonClient.resumeConnection()` that clears the pending backoff timer and calls `attemptConnect()` immediately when the client is disconnected; it is a no-op for connecting/connected/disposed states. It is called from `useClientActivity` on `AppState` → active (native) and `visibilitychange` → visible (web).

Also raise the reconnect-toast threshold from 1000ms to 1500ms so that a sub-second resume reconnect never flashes the indicator.

Cuts resume-to-session-ready from ~3s to ~300-500ms on Android; web and Electron get the same fast path on tab/window refocus. Passive backoff for transient drops while the app is running is unchanged.

## QA / testing evidence

Automated checks:

- `npm run typecheck` — all workspaces pass.
- Repository lint and formatting checks pass (lefthook pre-commit).
- Focused Vitest: `packages/client/src/daemon-client.test.ts` — 112 tests pass, including new coverage for `resumeConnection()` (immediate reconnect when disconnected, no-op for connecting/connected/disposed, backoff state reset).

UI QA:

- Android: verified by the author on a physical device during branch development; resume-to-session-ready dropped from ~3s to ~300-500ms and the reconnect toast no longer appears on quick resumes.
- iOS: not tested; no iOS simulator was available. The change only touches reconnect timing via the cross-platform `AppState` API, with no platform-specific code.
- Web / Desktop: not tested interactively; they share the same `visibilitychange` fast path, and the behavior is covered by the client unit tests.

### DIFF
--- a/packages/app/src/hooks/use-client-activity.ts
+++ b/packages/app/src/hooks/use-client-activity.ts
@@ -50,11 +50,15 @@ export function useClientActivity({
   // Track app visibility via AppState (native).
   useEffect(() => {
     const subscription = AppState.addEventListener("change", (nextState) => {
-      tracker.notifyAppVisibility(nextState === "active");
+      const visible = nextState === "active";
+      tracker.notifyAppVisibility(visible);
+      if (visible) {
+        client.resumeConnection();
+      }
       tracker.sendHeartbeat();
     });
     return () => subscription.remove();
-  }, [tracker]);
+  }, [client, tracker]);
 
   // Track user activity and visibility on web.
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


boudra:
Closing in favor of the implementation merged in [#4160](https://github.com/getpaseo/paseo/pull/4160).

PR #3155, opened 2026-08-10
fix(server): resolve Claude history from the provider profile's CLAUDE_CONFIG_DIR

### LINKED ISSUE #2005: bug: issues with loading session history from ccs
### What's broken

I'm using [ccs](https://github.com/kaitranntt/ccs/) to switch sessions. While I can add a "claude" provider in Paseo config as `ccs work` - but then session history doesn't work.

## Issue
When Paseo tries to load session history for a `claude` session started from a non-standard config (e.g. `ccs work`), it fails.

## Potential cause

Some debugging (using `claude`...) revealed that:
```py
private resolveHistoryPath(sessionId: string): string | null {
  const configDir = process.env.CLAUDE_CONFIG_DIR ?? path.join(os.homedir(), ".claude");
  ...
}
```

`resolveHistoryPath` is only checking `CLAUDE_CONFIG_DIR` (otherwise default to `.claude`). I've verified that if I set `CLAUDE_CONFIG_DIR=~/.ccs/instances/work`, things work fine.

But then it'd only work for one claude session at a time - should be at-least support reading this env from claude-work specific config (instead of just `process.env`?)

### Steps to reproduce

1. Configure alternate claude provider in Paseo, which uses a non-standard config location (e.g. `ccs work`)
2. Start Paseo, start a chat.
3. Restart Paseo
4. The chat history from step 2. won't load

### Where did this happen

Web (browser)

### Paseo version

0.1.104

### OS version

Manjaro

### Agent provider

Claude Code

### Provider configuration

2.1.197 (Claude Code) with ccs

### Logs

```text

```

### Screenshots or video

_No response_

xMase:
Fixes #2005.

## What's broken

A custom provider that extends `claude` with its own `env.CLAUDE_CONFIG_DIR` — the documented way to run several accounts ([docs/custom-providers.md](https://github.com/getpaseo/paseo/blob/main/docs/custom-providers.md#multiple-profiles-for-the-same-provider)) — launches Claude Code against that directory, but Paseo reads the transcript back from a different one. Every client shows an empty timeline ("Start chatting with this agent") for an agent whose conversation is safely on disk.

The provider `env` is honored when spawning the CLI and ignored when resolving history. #2005 reported this in July with the same root cause, and the maintainer comment confirmed it and pointed at #2514 as the fix. #2514 was later closed unmerged with "the support already exists in the configuration system and it just needs ui", and #2785 was opened for the UI. The configuration support is real for *writing*, so the history half was never landed — it only existed inside the closed PR.

## Root cause

Both history lookups read the daemon-wide environment, in `packages/server/src/server/agent/providers/claude/agent.ts`:

- `ClaudeAgentSession.resolveHistoryPath()`
- `ClaudeAgentClient.listImportableSessions()`

```ts
const configDir = process.env.CLAUDE_CONFIG_DIR ?? path.join(os.homedir(), ".claude");
```

Both classes already hold `runtimeSettings`, and `ProviderRuntimeSettings.env` is exactly where a profile's `CLAUDE_CONFIG_DIR` lives — the value was available, just never consulted. `project-dir.ts` already accepted an explicit `configDir` override, so the plumbing existed too.

## The fix

`project-dir.ts` exports `resolveClaudeConfigDir()`, adding one precedence level between the explicit override and the process environment:

```ts
options?.configDir ??
options?.runtimeSettings?.env?.CLAUDE_CONFIG_DIR ??
process.env.CLAUDE_CONFIG_DIR ??
join(homedir(), ".claude")
```

Both call sites use it. Behavior is unchanged for anyone without a profile-level `CLAUDE_CONFIG_DIR`.

## Tests

Five cases added to `project-dir.test.ts` pinning the precedence: profile over daemon env, fallback to daemon env, explicit `configDir` winning over both, the `~/.claude` default, and a transcript path resolving under the profile directory.

`packages/server`: 27 files / 348 tests pass for the Claude provider. The repo's pre-commit hooks (oxlint, oxfmt, workspace-wide typecheck) pass clean.

## Manual verification

Same agent, same command, two daemons — an agent created with a custom `claude-sf` provider whose `env.CLAUDE_CONFIG_DIR` points at `~/.claude-sf`:

| daemon | build | `paseo logs <id>` |
| --- | --- | --- |
| published 0.3.1 | unpatched | `No activity to display.` |
| this branch | patched | full transcript restored (306 lines) |

## Out of scope

`models.ts` resolves the model catalog directory the same way, and `provider-registry.ts` never passes `configDir` when constructing `ClaudeAgentClient`, so a profile's catalog is read from the wrong directory too. Left out to keep this focused on the reported bug — happy to fold it in if you'd prefer one PR.


boudra:
Closing in favor of the open fix [#3378](https://github.com/getpaseo/paseo/pull/3378).

PR #3192, opened 2026-08-11
Notify callers when watched children close and cap responses

wilgon456:
## What changed

- notify the caller when a watched child transitions from `running` to `closed`
- cap child responses embedded in finish notifications at 4,000 characters, with a pointer to `get_agent_activity` for the full response
- document the close-notification contract
- add regression coverage for the lifecycle transition and response cap

## Why

`setupFinishNotification()` currently treats `closed` as silent watcher teardown. A delegated child can therefore disappear without waking its caller when PR-merge cleanup, workspace teardown, or an explicit archive closes the runtime before the normal `running -> idle` finish edge is observed.

The same notification path embeds the child's entire last assistant message. A verbose child can therefore consume an outsized portion of its caller's next context window.

I reproduced the closed-child failure on Paseo 0.3.1 on macOS: a delegated child completed its task, its isolated worktree was archived after the PR merged, the persisted child became `closed`, and the parent stayed asleep until the user manually asked for the result. There was no notification failure log because the `closed` branch intentionally called `stop()` without attempting delivery.

This is a focused successor to #2879, rebased onto current `main` after #3177. It carries the closed-child wakeup and response truncation forward. The original #2879 author is credited on both commits. It does not include the repeated-finish watcher behavior.

## Behavior

- A watcher that observes `running -> closed` sends `Agent <id> (<title>) was closed.`
- A child that is already closed when the watcher is installed remains silent, preserving the existing startup guard.
- Archived callers and detached/re-parented children retain their existing guards.
- An assistant response longer than 4,000 characters is truncated in the notification; the full response remains available through `get_agent_activity`.

## Verification

- Regression test before the close implementation: timed out after 5 seconds because the parent was never prompted.
- `npx vitest run packages/server/src/server/agent/agent-prompt.test.ts --bail=1` — 13 passed.
- `npm run lint` — 0 warnings, 0 errors.
- `npm run format` and the pre-commit formatting check — passed.
- Pre-commit hook reran targeted lint, formatting, and workspace typecheck successfully.

Platform coverage: daemon behavior verified on macOS. No app, protocol, or platform-specific code changed; CI provides the repository's routed cross-platform coverage.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -51,6 +51,7 @@ Users can also detach an existing subagent from the subagents track. Detach is d
 `notifyOnFinish` defaults to `true` for agent-scoped creation and background prompt follow-ups because most delegated work needs to report back to the creating agent. Set it to `false` only for truly fire-and-forget agents or prompts.
 Permission requests are notification checkpoints, not the end of that subscription. The caller is notified again after a permission response when the child finishes, errors, or requests another permission.
 The permission notification includes the normalized request plus the child and request IDs, so the caller can inspect it and respond without fetching agent status.
+A watched child that closes before its finish event also notifies the caller so delegated work cannot disappear silently during archive or workspace teardown.
 
 ## Provider-managed child agents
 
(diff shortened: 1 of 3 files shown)


boudra:
I added the 4,000-character child-response cap from https://github.com/getpaseo/paseo/pull/2879 with Thong Van credited as co-author, plus focused regression coverage. I am taking this through CI and merge now.

PR #3206, opened 2026-08-11
feat: workspace labels, filtering, and saved views

cleiter:
### Type of change

- [x] New feature

### Reasoning

Workspaces accumulate faster than they close. Workspaces might sit for days or weeks waiting on someone else, and the sidebar fills with work you cannot act on right now. Once the list is longer than the panel, the sidebar stops being the thing that tells you an agent needs you and becomes a thing you scroll.

The obvious answer is folders, and folders need drag and drop to be worth anything. Cross-list drag does not work in React Native, so folders would have shipped on web and been second-class on iOS and Android. Labels need no drag at all: you assign them from a workspace's menu and filter by tapping a chip, one implementation that behaves the same on all four platforms. You lose manual ordering and gain a sidebar that narrows to what you are working on, identically everywhere.

Folders were tried in #2635 - this is the follow-up.

Labels live on the daemon so they follow you to your phone. The filter is per device, because which slice you want depends on where you are sitting.

### Goals

- Assign labels to a workspace from its menu, with a name and a palette colour. Labels and their colours live on the daemon and sync to every client.
- Filter the sidebar by label from a chip track under the Workspaces header, tri-state per label: neutral, include, exclude, with Cmd- or Shift-click walking the ring the other way. Included labels read as either-of by default, or as all-of from a Match all switch; exclude beats include in both.
- The chip track is put away from a Tag button in the Workspaces header, and away means gone rather than collapsed under a heading that still costs a row. Hidden while filtering, the button carries a dot, so putting the chips away cannot hide that the list is narrowed.
- The label picker follows the project and host filters: a label no workspace in scope carries is not offered, and a label already filtering is never dropped from the picker. Assignment surfaces keep the full catalog.
- Filter by project as a second facet, ANDed against labels. Grouping composes with both filters.
- Group the sidebar by label as a third grouping mode: a label heads a section instead of narrowing the list, and the workspaces carrying none keep a section of their own. Reachable from the display menu, the saved-view editor, and the command palette.
- The workspace you are in is never hidden by the filter, so creating a worktree while filtered does not drop you onto a workspace missing from its own sidebar. A row with a menu open is exempt on the same grounds: labels are assigned from that menu, so the press that stops a workspace matching arrives from inside the row being matched, and without the exemption the row goes and takes the open menu with it.
- The label chips on a workspace row and in its hover card are marks, not controls. A chip there is 14pt inside a row whose job is opening the workspace, so a press aimed at the row that lands a few pixels high rearranged the list instead of opening anything. Only the track filters.
- A rail at the bottom of the list is permanent while filtering, reads "N hidden", and clears every facet when pressed.
- Switch the match mode from a pill after the chips in the track, shown only from two lit labels because below that the two modes produce the same list, and from a permanent row in the label filter menu for when the track is hidden. Saved views capture it.
- Save the current filters and grouping as a named view through the same editor that edits one, apply it from the display menu or the command palette, update it in place once the sidebar has moved off it, and start a new view from inside one. All workspaces heads the switcher, checked when nothing is applied, so leaving a view is a choice among the views rather than a verb wedged after them.
- A Labels section in Settings lists every label across all connected hosts as one merged set, and renames, recolours, or deletes them. Renaming refuses a name another label already holds rather than merging the two, and is all or nothing across hosts because the name is the identity.
- Agents label workspaces over MCP, so a run that finishes can mark the workspace it finished in.

### Non-goals

- No manual ordering. A filter narrows a list, it never arranges one, so there is still no way to say "these four sit at the top, in this order, always". Pinning covers part of it. This is the strongest remaining argument for folders and it is worth raising now rather than after the fact.
- Saved views and the active filter do not sync. Label assignments and colours do, because those are on the daemon. Making views sync needs a replicated client document that does not exist today, so they live in AsyncStorage beside the host filter.
- No per-label match. Intersection is one switch over the whole include facet — Match any or Match all — not a fourth chip state, so there is no way to ask for backend AND (frontend OR docs). A chip that could be included, required, or excluded is a control you decode rather than read, with every chip's meaning conditional on every other chip. The real shape of mixed matching is faceted labels, AND across groups and OR within, which needs label grouping this does not add.
- No project labels. A project filter already inherits to everything under it, so labelling a project would solve a problem projects already solve.
- Renaming a label is a sweep. A workspace stores plain label names, no ids, because two daemons have never seen each other's ids and the name is the only join key for merging their catalogs. Renaming therefore rewrites every workspace carrying the old name, which buys cross-host labels with no id-mapping story.
- No label management over MCP. Agents apply and remove labels on a workspace; creating, recolouring, renaming, and deleting a label stay in the app, because those are catalog-wide and a wrong one is a sweep across every host.

### Labels over MCP

Three tools, alongside `rename_workspace`:

| Tool                      | Function                                                                                                                |
| ------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| `list_workspace_labels`   | The labels this host knows - catalog entries and labels only workspaces carry - each with how many workspaces carry it. |
| `add_workspace_labels`    | Add labels to a workspace, keeping the ones it already has.                                                             |
| `remove_workspace_labels` | Remove labels from a workspace, leaving the rest.                                                                       |

Both writes default to the caller's own workspace, the way `rename_workspace` does. They are one helper with a different transform: read the record, transform the set, put it back through `normalizeWorkspaceLabels`, so an agent cannot write a set the app would refuse to draw. Matching is case-insensitive both ways, so re-adding a label the workspace spells differently keeps the stored spelling. A write that changes nothing is skipped, because `updatedAt` orders the sidebar and a no-op relabel that bumped it would reorder the list to say nothing. Anything over the ten-label cap comes back in `dropped` rather than vanishing. `list_workspaces` and `get_workspace` now report labels, since an agent otherwise cannot see what it is about to change.

`list_workspace_labels` exists because the name is the identity: a label an agent coins itself cannot be merged into an existing one later, so the tool description sends it to look first.

### QA

Tested on desktop (Electron, macOS) and in the browser against a real daemon. Not tested: iOS, Android.

Assigned labels from a workspace's menu, then used the chip track to include and exclude them and watched the sidebar narrow. Lit two labels, flipped the pill to Match all and watched the union become the intersection, and confirmed the hidden rail counted the difference and cleared it. Put the track away with a filter active and confirmed the Tag button keeps its dot, and that the state survives a reload. Set a project filter and confirmed the label picker drops labels nothing in scope carries while keeping the one currently filtering. Saved the current filters as a named view from the full editor, moved the filters, used Update to write them back, and started a second view from inside the first. Renamed, recoloured, and deleted labels from Settings and confirmed the chips on rows and in the track followed.

The MCP tools were driven by an agent against a live daemon, from inside a workspace of my own. `list_workspace_labels` returned the full catalog with usage counts (`Idea`, red, on 8 workspaces). `add_workspace_labels` with `["Idea"]` and no `workspaceId` answered `{"workspaceId":"wks_bcd31d779d8bd36e","labels":["In Progress","Idea"],"changed":true,"dropped":[]}` - the caller's own workspace, its existing `In Progress` kept, and the chip appeared on the row without a reload. Note the daemon builds its tool catalog at startup, so the tools need a daemon restart to show up, and `npm run start` runs `dist/`, so a build has to come first.

Verified in the browser, against a daemon holding real workspaces, with the label track filtered to one label and 60 workspaces hidden: deselecting that label from a row's own menu leaves the row and the menu in place, the row marks itself Outside the filter, a second label can be set from the same open menu, and closing the menu drops the row and takes the hidden count from 60 to 61. Cycled a chip forward and backward under both Cmd and Shift, one press at a time. Confirmed the row and hover-card chips are inert: no button role, and a press on one opens the workspace.

The filter module, the store, the MCP tools, and the daemon side are covered by unit tests; the browser specs drive the whole path against a real daemon.

Rebased onto #3063, which landed the sidebar Group by entries in the command palette. Those were built as a two-mode toggle and could not name a destination once Label became a third mode, so they are now one entry per mode you are not in, driven by the same mode list the display menu and the saved-view editor read.

**Tests:**

- `packages/app/e2e/browser/sidebar-workspace-labels.spec.ts` (new) - labels a workspace against a real daemon, filters to it, asserts the hidden count and that the active workspace is still rendered, unlabels from the row's own menu while that label is the filter and asserts the row and the menu both survive and the label goes back on from the same menu, and collapses and re-expands the Labels section.
- `packages/app/e2e/browser/command-center-grouping.spec.ts` (extended) - the palette reaches every grouping mode, including status to label without passing through project, and never offers the mode the sidebar is already in.
- `packages/app/src/command-center/root-contributions.test.ts` (rewritten) - one entry per mode you are not in, from each of the three starting modes, plus the id and rank rules that keep the palette in the same order as the display menu.
- `packages/app/src/components/sidebar/sidebar-filter.test.ts` (new) - the filter rules, the open-menu exemption and its end when the menu closes, and both safety rails: exclude beating include, an unlabelled workspace failing an include filter, the active workspace surviving an excluding filter, the hidden count, the scoped label picker, the active-filter count behind the Tag button, and the match mode - the intersection under Match all, exclude still winning, a repeated label not counting twice, and the mode reading as Match any below two labels.
- `packages/app/src/hooks/sidebar-workspaces-view-model.test.ts` (extended) - row identity survives a redelivered label list and is dropped when a label changes, so a workspace update does not rebuild every row.
- `packages/app/src/stores/sidebar-view-store.test.ts` (extended) - the v2 to v3 migration, the tri-state cycle, project-filter reconciliation against stale keys, and saved-view CRUD including creating a view from the editor.
- `packages/server/src/server/agent/mcp-server.test.ts` (extended) - the label tools: adding without dropping existing labels, the stored spelling winning a case-insensitive re-add, the no-op write skipping persistence, the cap reported in `dropped`, removal ignoring labels the workspace never had, an archived workspace refused, and the catalog and in-use union with counts.
- `packages/server/src/server/session.workspaces.test.ts` (extended) - the label set and catalog RPCs: persistence, name normalisation, rename refusal on a taken name, and the workspace update they emit.
- `packages/server/src/server/workspace-label-catalog.test.ts` (new) - the file-backed catalog: atomic writes, derived versus chosen colours, and delete.
- `packages/protocol/src/workspace-labels.test.ts` (new) - label normalisation and the rename precondition shared by client and daemon.
- `packages/protocol/src/messages.workspaces.test.ts` (extended) - an old client parsing a payload carrying `labels`, and a new daemon parsing one without it.
- `packages/app/src/components/sidebar/workspace-labels/catalog-writes.test.ts` (new) - fan-out of a catalog write across hosts, including the partial result when one host rejects it.
- `packages/app/src/components/sidebar/workspace-labels/catalog.test.ts`, `known-labels.test.ts`, `edit-plan.test.ts` (new) - cross-host catalog merging, the known-label set derived from workspaces, the list the Labels menu goes on offering while it is open, and the create and edit plans behind the label dialog.
- `packages/app/src/i18n/resources.test.ts` (extended) - both halves of the view switcher label filed under the views the switcher shows. A key put under the wrong parent in all nine locales at once passes both typecheck and the key-sync test, and only the call site notices, by rendering the key.
- `packages/app/src/components/sidebar/saved-views/view-state.test.ts`, `view-editor.test.ts` (new) - view summaries, whether the sidebar still matches the applied view, display ordering, and refusing a name a view already holds.
- `packages/app/src/command-center/saved-view-contributions.test.ts` (new) - the palette entries for applying, saving, and updating a view, including leaving Update out while the sidebar still matches.
- `packages/app/src/components/sidebar/sidebar-groups.test.ts`, `sidebar-projection.test.ts` (new and extended) - grouping by label, and the projection handing the same filtered set to pinned rows, status buckets, and project sections.
- `packages/app/src/components/rename-modal-state.test.ts` (new) - which message the rename dialog shows and when its submit button is dead, extracted from the component so the rule is testable without mounting it.
- `packages/app/src/utils/sidebar-shortcuts.test.ts`, `project-display-name.test.ts`, `workspace-meta-row/meta-items.test.ts`, `command-center/workspace-contributions.test.ts`, `sidebar-collapsed-sections-store/state.test.ts` (extended) - the label group in shortcuts, project display names, the label chips on a row, label search in the palette, and the persisted collapse of the Labels section.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:
Thanks for the PR. This workflow is superseded by [PR #3510](https://github.com/getpaseo/paseo/pull/3510), which takes the core labels workflow forward first and leaves deeper organization decisions for later. I’m closing this one.

PR #3274, opened 2026-08-12
fix(codex): keep archived history resumes inert

wilgon456:
## Summary

- keep Codex sessions opened for archived history in a read-only state
- hydrate persisted timelines with `thread/read` without calling `thread/loaded/list` or `thread/resume`
- add a regression test for the case where Paseo archived the agent record but native Codex archival failed

## Why

Paseo's native provider archive hook is best-effort. If Codex native archival fails, the thread remains unarchived even though the Paseo record is archived. Opening that agent from History currently calls `thread/resume`; because the native thread is still unarchived, the call succeeds and creates a live app-server writer for an agent that should be history-only. A second Codex client can then fail to resume the same thread with an active-writer conflict.

`thread/read` can hydrate the timeline without loading the thread. Skipping `ensureThreadLoaded()` for the existing `purpose: "history"` path preserves the lifecycle contract that only an explicit Unarchive may return an archived agent to an interactive runtime. Interactive resumes still use the existing load/resume path.

## Verification

- `npm exec -- vitest run packages/server/src/server/agent/providers/codex-app-server-agent.test.ts --bail=1` (128 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run format`

### DIFF
--- a/packages/server/src/server/agent/providers/codex-app-server-agent.test.ts
+++ b/packages/server/src/server/agent/providers/codex-app-server-agent.test.ts
@@ -1115,7 +1115,7 @@ describe("Codex app-server provider", () => {
     await session.close();
   });
 
-  test("loads archived Codex history without resuming the native thread", async () => {
+  test("loads archived Codex history without resuming an unarchived native thread", async () => {
     const threadRequests: string[] = [];
     const appServer = createFakeCodexAppServer({
       "thread/loaded/list": () => {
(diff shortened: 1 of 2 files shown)


boudra:
Closing for now: this needs a real Codex check showing that opening Paseo-archived history leaves the native thread unloaded and available to another client when native archival failed. Please reopen with that evidence.

boudra:

PR #3278, opened 2026-08-12
fix(server): avoid false user interrupt on ACP session close

frontend-london:
## Summary

Fixes false "user interrupt" cancellations for ACP agents during daemon/session shutdown.

`ACPAgentSession.close()` previously sent `session/cancel` whenever a foreground turn was active. During daemon shutdown/restart this caused ACP runtimes such as Devin to interpret infrastructure shutdown as an explicit user cancellation, marking active tool calls as failed with messages such as:

`Canceled due to user interrupt`

## Root cause

`close()` used the same cancellation path as an explicit user `interrupt()`.

This conflated:

- session/runtime shutdown
- explicit user interruption

## Fix

- `close()` no longer sends `session/cancel`
- explicit `interrupt()` and deny+interrupt still send `session/cancel`
- client-initiated cancellation state is tracked separately from runtime-originated `stopReason="cancelled"`

## Tests

Added regression coverage for:

- close with active turn does not send `session/cancel`
- explicit interrupt still sends cancel
- explicit interrupt still produces user-cancel semantics
- runtime-originated cancellation is not misclassified as user cancellation
- cancellation state is reset between turns

Validated with:

- ACP agent tests: 104/104
- agent-manager tests: 151/151
- session tests: 144/144
- server typecheck
- lint
- format check
- server build
- git diff --check

Also reproduced and validated end-to-end with Devin/SWE-1.7.

### DIFF
--- a/packages/cli/src/commands/daemon/local-daemon.stop-ownership.test.ts
+++ b/packages/cli/src/commands/daemon/local-daemon.stop-ownership.test.ts
@@ -1,0 +1,56 @@
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, test, vi } from "vitest";
+
+const { shutdownServer, close, tryConnectToDaemon } = vi.hoisted(() => {
+  const shutdownServerMock = vi.fn(async () => {});
+  const closeMock = vi.fn(async () => {});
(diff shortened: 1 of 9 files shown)


boudra:
Closing for now: this needs before/after ACP traces for shutdown during an active turn, with an explicit user interrupt as a control. Please reopen with that evidence.

boudra:

PR #3285, opened 2026-08-12
Give the terminal its own typography settings

eddiemachaalani:
### Linked issue

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

The terminal pane sat on a different surface token from the agent panel and had no
content insets, so opening a terminal next to an agent looked like two unrelated
apps. Fixing that surfaced a worse problem: the terminal borrowed the **code** font
settings, but the terminal is the one surface that needs Nerd Font glyphs for shell
prompts and TUIs, and most code fonts have none. Choosing a nicer font for diffs
silently broke terminal prompts, with no way to opt out.

So the terminal gets its own typography: font, size, line height, and bold text.
Two real bugs surfaced while building it and are fixed here.

### Goals

- Terminal pane matches the agent stream's content insets and surface token.
- Terminal typography is independent of the code font settings.
- `Settings → Appearance → Terminal`: font (web only), size, line height, bold text.
- Nothing changes for existing users on upgrade except line height, which is a
  deliberate default change (1.0 → 1.2). Size defaults to **inherit**, so the
  terminal keeps following Code font size until the user sets a terminal size.
- Works on all three terminal renderers: web xterm, native webview, native grid.

### Non-goals

- No change to the ANSI palette or terminal theme colors.
- No max-content-width cap on the terminal; unlike the agent stream it stays
  full-bleed, so on very wide panes the two still differ horizontally. Deliberate.
- Terminal **font family** row stays web-only, matching the existing UI/code font
  rows (`showFontFamilyRows = !isNative`).

### Two bugs fixed along the way

**1. Font changes never invalidated the WebGL glyph atlas.** `setFont` updated
`terminal.options` and refit, but the WebGL renderer caches rasterized glyphs in a
texture atlas keyed by character and style. The old font's glyphs — bold faces most
visibly — kept being drawn, so the rendered result depended on the *order* settings
were changed rather than their final values. Reproduced repeatedly by hand: the same
font setting produced bold or non-bold text depending on the path taken to reach it.
The runtime now holds the addon and calls `clearTextureAtlas()` on font change,
guarded so a lost GL context can't block the font from applying. Pre-existing —
changing **code font size** hit it too.

**2. An unresolvable font family broke the grid.** A user-entered family with no
fallback (a typo, or a font that isn't installed — e.g. `SFMono-Regular`, which does
not resolve as a CSS family) fell through to the browser's default **proportional**
font. xterm gives every cell the same advance, so proportional glyphs spread out and
the cursor stops lining up. `resolveTerminalFontFamily` now appends a generic
`monospace` backstop to any explicit value.

### QA

**Desktop macOS (Electron)** — driven interactively throughout development against a
dev daemon on an isolated `PASEO_HOME`:

- Padding/surface compared side by side against the agent panel.
- Font changes applied in every order (blank → explicit → blank, and across
  size/line-height changes) — deterministic after the atlas fix; non-deterministic
  before it.
- `printf '\033[1mBOLD-SAMPLE\033[0m plain-sample\n'` used as the bold control:
  heavier with Bold text ON, identical weight with it OFF, colors unchanged in both.
- `system-ui` and `SFMono-Regular` confirmed to break the grid before the monospace
  backstop, and render as normal monospace after.
- Line-height change mid-session leaves a reflow scar in scrollback (a PTY resize,
  same as any terminal); a new terminal is clean.

**iOS 26.3 simulator (iPhone 16e), native grid renderer** — the default renderer on
native and a different code path from xterm, with its own metrics and bold handling:

| | |
|---|---|
| Default (line height 1.2, bold on) | default |
| Bold text off | bold off |
| Settings → Appearance → Terminal | settings |
| Line height 1.4, persisted across relaunch | lh 1.4 |

`BOLD-SAMPLE` is visibly heavier than `plain-sample` with bold on and identical with
it off, applied live without remounting the terminal. Line height 1.4 applies without
clipping or column drift, and survives a relaunch. The Terminal font row is correctly
absent on native.

**Android 17 emulator (Pixel 10), native grid renderer** — same renderer as iOS but a
different default mono stack (`monospace` rather than `ui-monospace`) and different RN
text layout:

| | |
|---|---|
| Bold text on | android bold on |
| Bold text off | android bold off |
| Settings → Appearance → Terminal | android settings |

Padding applies, bold on/off behaves as on iOS, and ANSI colors are preserved with bold
off (the green `ls`/`printf` and orange strings are unchanged between the two shots).

**Browser web (Chromium)** — same xterm renderer as Electron, verified separately:

| | |
|---|---|
| Terminal, bold on | web terminal |
| Settings → Appearance → Terminal | web settings |
| Bold text off | web bold off |

Zero console errors or warnings. The settings shot also shows both copy fixes in place:
the Terminal font hint leads with "Must be a monospace font", and the Code font hint no
longer claims it drives terminal output.

Each client has its own settings store, so iOS, Android and web each started from
defaults independently — which incidentally confirms the shipped defaults on a fresh
install.

| iOS             | Yes    | Simulator, iPhone 16e, iOS 26.3, native grid renderer |
| Android         | Yes    | Emulator, Pixel 10, Android 17, native grid renderer |
| Web             | Yes    | Chromium, zero console errors |
| Desktop macOS   | Yes    | Electron dev build, extensive interactive use |
| Desktop Windows | No     | No machine available |
| Desktop Linux   | No     | No machine available |

**Automated coverage** — 26 new tests:

- `terminal-font.test.ts` — family fallback, the monospace backstop and its
  no-double-append cases, line-height clamping, bold weight mapping.
- `terminal-typography.test.ts` — the inherit sentinel (0 survives rather than being
  clamped up), clamping, fractional line height, defaults, normalization.
- `terminal-grid-metrics.test.ts` — line height scales cell **height** only; width is
  untouched (widening would desync every column); invalid values fall back to 1.0.
- `colors.bold.test.ts` — native bold weight on/off, color preserved when off,
  `themeKey` changes so cached styles invalidate.
- `terminal-emulator-runtime.browser.test.ts` — atlas cleared on font change; font
  still applies when the atlas clear throws.

Coverage is mutation-checked, not just green:

- Reverting the cell-height multiplier turns 2 metrics tests red.
- Reverting the native `boldText` guard turns 1 test red.
- Removing the atlas clear turns 1 browser test red.
- An existing runtime test caught the monospace-backstop change on its own
  (`setFont("Menlo")` now yields `"Menlo, monospace"`); expectation updated and
  extended to assert line height and bold weight.
- One native test initially **failed** because it used a made-up cell shape rather
  than passing vacuously; fixed to the real `TerminalCell`.

`npm run typecheck`, `npm run lint`, `npm run format` pass across all workspaces.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:

PR #3288, opened 2026-08-12
fix(claude): suppress stale compact cancellation

mikeyzhe:
### Linked issue

None — live regression reproduced from provider events.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

When an active Claude query is interrupted and a replacement turn starts, the SDK can report the canceled built-in `/compact` as a zero-token successful result whose text is exactly `Compaction canceled.`. The existing interrupt filter treated every successful result as potentially belonging to the replacement, so that stale sentinel could be translated into visible output and a terminal event for the new turn.

This change recognizes that exact provider sentinel while an interrupt result is pending and drops it before translation. Other successful terminal results still reach the existing replacement-turn routing decision.

### Goals

- Suppress the exact zero-token Claude compact-cancellation sentinel after an interrupt.
- Preserve real successful replacement-turn results.
- Lock the reported sequence down with an automated regression test.

### Non-goals

- Deduplicating `/compact` requests in clients or API gateways.
- Filtering arbitrary assistant text that happens to mention compaction.
- Restarting or changing the production daemon as part of this PR.

### QA

RED: the new regression emitted assistant timeline output plus `turn_completed` for the replacement turn before the fix.

GREEN on current `origin/main`:

- `npm run build:server` — passed.
- `npx vitest run packages/server/src/server/agent/providers/claude/agent.redesign.test.ts packages/server/src/server/agent/providers/claude/agent.interrupt-restart-regression.test.ts --bail=1` — 2 files, 32 tests passed.
- `npm run lint` — 0 warnings, 0 errors.

Not live-tested against the main daemon because restarting port 6767 would interrupt active user agents; the regression routes the exact SDK result shape through `ClaudeAgentSession`.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/providers/claude/agent.redesign.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.redesign.test.ts
@@ -456,6 +456,35 @@ test("interruptActiveTurn only interrupts the active query without info logs", a
   }
 });
 
+test("suppresses a stale successful local-command cancellation after a replacement starts", async () => {
+  const session = await createSession();
+  const events: AgentStreamEvent[] = [];
+  session.subscribe((event) => events.push(event));
+  const internal: {
+    pendingInterruptAbort: boolean;
+    activeForegroundTurnId: string | null;
+    foregroundHasVisibleActivity: boolean;
+    routeSdkMessageFromPump: (message: Record<string, unknown>) => void;
+  } = asInternals(session);
+
+  internal.pendingInterruptAbort = true;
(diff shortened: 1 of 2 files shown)


mikeyzhe:
Closing as stale. Branch `fix/compact-action-lifecycle` (7980a8c3) is kept if this is revived.

PR #3291, opened 2026-08-12
feat(server): expand {repoRoot} in worktrees.root

simonepri:
### Linked issue

None. No prior issue or discussion — happy to open a Discussion and settle the direction there first.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

`worktrees.root` is a single global path resolved at daemon startup, with no repo in scope. If you want each project's worktrees to sit inside that project, next to the checkout they belong to, you have to point the global root at one specific repo — and then every other project's worktrees land inside that repo too, under their own hash directory. The value also has to spell out a home directory, so it can't be shared across machines or with teammates.

A leading `{repoRoot}` expands to the checkout each worktree is cut from, so `"{repoRoot}/.worktrees"` is one setting that keeps every project's worktrees inside that project and holds no user or machine specifics.

### Goals

- `"root": "{repoRoot}/.worktrees"` creates worktrees under the source checkout they belong to
- One value that works for every project and on any machine
- Unchanged layout beneath the root: `<root>/<hash>/<slug>`
- Archive still removes worktrees, including when the git admin dir is already gone
- Absolute and `PASEO_HOME`-relative roots behave exactly as before

### Non-goals

- No per-project override in `paseo.json`; this stays a machine-level setting
- Not dropping the project-hash segment, redundant as it looks under a repo-scoped root. `isPaseoOwnedWorktreeCwd` treats `<hash>/<slug>` as proof Paseo owns a directory, so a one-segment layout would make a hand-made `.worktrees/foo` look Paseo-owned
- `{repoRoot}` is supported only as the first segment, has to be followed by a separator, and cannot climb out of the repo with `..`. Anything else is rejected at config load rather than silently resolving somewhere unexpected
- Paseo does not add the `.gitignore` entry for you; both docs pages tell you to
- No Windows verification

### QA

macOS 15 (arm64, Node 22); not run on Linux or Windows.

Created a workspace through the real daemon and CLI with `"root": "{repoRoot}/.worktrees"`. It landed at `/private/tmp/paseo-inrepo-demo/.worktrees/2yx76c0b/in-repo`, `git worktree list` registered it, `git status` showed `?? .worktrees/` until ignored, and archiving removed and deregistered it.

```
$ npx vitest run src/utils/worktree.posix.test.ts --bail=1   → 48 passed | 1 skipped
$ npx vitest run src/server/config-relay.test.ts --bail=1    → 24 passed
```

Tests added: create/own/list/delete under a repo-relative root against a real repo; ownership with the git admin dir unavailable, plus a wrong-hash look-alike that must not be owned; `loadConfig` keeps the template unresolved and rejects malformed placeholders.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/config-relay.test.ts
+++ b/packages/server/src/server/config-relay.test.ts
@@ -262,4 +262,40 @@ describe("daemon worktree root config", () => {
       path.join(os.tmpdir(), "paseo-custom-worktrees"),
     );
   });
+
+  test("keeps a repo-relative worktrees.root unresolved for per-project expansion", async () => {
+    const home = await createPaseoHome({
+      version: 1,
+      worktrees: { root: "{repoRoot}/.worktrees" },
+    });
+
+    expect(loadConfig(home, { env: {} }).worktreesRoot).toBe("{repoRoot}/.worktrees");
+  });
+
+  test("rejects {repoRoot} outside the first path segment", async () => {
+    const home = await createPaseoHome({
+      version: 1,
+      worktrees: { root: path.join(os.tmpdir(), "shared", "{repoRoot}") },
+    });
+
+    expect(() => loadConfig(home, { env: {} })).toThrow(/first path segment/);
+  });
+
+  test("rejects a repo-relative root that is not separated from the placeholder", async () => {
+    const home = await createPaseoHome({
+      version: 1,
+      worktrees: { root: "{repoRoot}worktrees" },
+    });
+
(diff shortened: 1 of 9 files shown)


simonepri:
@boudra any chance to get this merged? It blocks us from allowing us to define worktrees within the repo root like claude/codex do.

boudra:

PR #3295, opened 2026-08-13
fix(desktop): keep browser annotation card visible and interactive

### LINKED ISSUE #3277: bug: element annotation card renders below the browser plane and is invisible
### What's broken

In the Electron browser, the annotation card opened by **Annotate element** is never visible. The card does mount and work — it takes focus, accepts typing, `Cmd+Enter` attaches the element and `Esc` cancels — but it paints underneath the `<webview>`, so nothing appears on screen.

Expected: after picking an element, the annotation card appears at the bottom of the browser pane, with the element identity, the comment field and the Cancel / Attach buttons.

Actual: no card is visible. The feature only works blind — pick an element, type, then press `Cmd+Enter`; the `browser_element` attachment then shows up in the composer, which proves the card was mounted and focused the whole time.

### Steps to reproduce

1. Open any page in a Paseo desktop browser tab.
2. Click **Annotate element**, then click any element on the page.
3. Observe that no annotation card appears anywhere in the pane.
4. Without clicking anything else, type some text and press `Cmd+Enter`.
5. Observe that the attachment lands in the composer, complete with the selector, outer HTML, computed styles, element screenshot and the typed comment — i.e. the invisible card was fully functional.

### Analysis

`BrowserElementAnnotationCard` renders inline inside the pane tree:

https://github.com/getpaseo/paseo/blob/main/packages/app/src/desktop/browser/pane/index.electron.tsx#L1795

Its `annotationOverlay` style uses a local `zIndex: 1`. The resident webview host is `position: fixed` on the body at `WEB_SURFACE_PLANE.browser` (0):

https://github.com/getpaseo/paseo/blob/main/packages/app/src/desktop/browser/resident-webviews.ts#L112

That local `zIndex: 1` only outranks the browser plane when no ancestor inside `#root` establishes a stacking context, which does not hold in practice. The card therefore paints below the guest page.

Rendering the card in the shared overlay plane — `#overlay-root` at `WEB_SURFACE_PLANE.overlay` (1), the same plane modals and toasts already use — is the natural fix.

**One caveat for whoever fixes it**, verified experimentally: the card must stay inside the React root container. Reparenting the node to `#overlay-root` (a sibling of `#root`) does fix the painting, but React 17+ delegates events at the root container, so clicks inside the card never reach React — Attach, Cancel and the close button silently stop working while the `window`-registered keyboard shortcuts keep working. Portaling into a host that lives inside `#root`, or lifting the card's stacking context in place, avoids that.

### Where did this happen

Desktop (Electron)

### Paseo version

0.3.1

### OS version

macOS 26.6

### Agent provider

Not relevant

### Logs

No errors are emitted; the failure is purely visual. The renderer logs nothing for the annotate flow.

dgk-dev:
## Summary

- render the element annotation card through a React portal into the active resident browser surface
- keep the card in the same paint plane as the `<webview>` so its local z-index places it above the guest
- preserve React ownership so the comment input, Attach, Cancel, and keyboard handlers remain functional
- extend the real Electron browser E2E to verify the input is visible, wins hit testing over the guest, accepts text, and submits through the Attach button

Fixes #3277.

## Why this target

The resident `<webview>` is hosted in a fixed body-level surface outside `#root`, while the card was previously rendered inline under `#root`. Portaling the card into that resident surface puts both in one stacking context. React still owns the portal, so event delegation continues to work; the E2E exercises a real pointer click on **Attach** rather than relying only on keyboard shortcuts.

## Validation

Validated from a clean `origin/main` baseline:

- `npm run test:e2e:browser-tabs --workspace=@getpaseo/desktop`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check:files -- packages/app/src/desktop/browser/pane/index.electron.tsx packages/desktop/e2e/browser-tabs.e2e.mjs`

### DIFF
--- a/packages/app/src/attachments/workspace-attachments-store.test.ts
+++ b/packages/app/src/attachments/workspace-attachments-store.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from "vitest";
+import { collectRetainedAttachmentIds } from "./gc-retention";
 import type { WorkspaceComposerAttachment } from "./types";
 import {
   appendWorkspaceAttachment,
@@ -45,6 +46,32 @@ function reviewAttachment(body: string): WorkspaceComposerAttachment {
           },
         },
       ],
+    },
+  };
+}
+
+function browserElementAttachment(screenshotId: string): WorkspaceComposerAttachment {
+  return {
+    kind: "browser_element",
+    attachment: {
+      url: "https://example.com",
+      selector: "#target",
+      tag: "button",
+      text: "Target",
+      outerHTML: '<button id="target">Target</button>',
+      computedStyles: {},
+      boundingRect: { x: 0, y: 0, width: 100, height: 40 },
+      reactSource: null,
+      parentChain: [],
+      children: [],
+      screenshot: {
+        id: screenshotId,
+        mimeType: "image/png",
+        storageType: "desktop-file",
+        storageKey: `/tmp/${screenshotId}.png`,
+        createdAt: 1,
+      },
+      formatted: "selector: #target",
     },
   };
 }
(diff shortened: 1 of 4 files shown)


boudra:
Closing in favor of the browser annotation editor fix on main: [7d3212e](https://github.com/getpaseo/paseo/commit/7d3212e36588ec6ea98a819f171f9ca87f15b7cf).

PR #3310, opened 2026-08-13
fix(release): refresh nix/npm-deps.hash before the version commit so release tags build

### LINKED ISSUE #3308: Release tags ship a stale nix/npm-deps.hash, so pinning a release breaks the Nix build
Follow-up to #3307. I switched my flake input from tracking `main` to pinning release tags, and the build of `v0.3.1` fails with:

```
error: hash mismatch in fixed-output derivation '/nix/store/...-paseo-0.3.1-npm-deps.drv':
         specified: sha256-RCp5Ogd8AETLmJCZaUebcgSPRk+is26nkUY7+blDb/g=
            got:    sha256-oXz8hMk+5DlTYK8OndUAjB+RJMDbPqobVGXLFeoH++o=
```

The cause seems to be release process timing. The version bump changes `package-lock.json`, the tag gets cut, and only then does CI recompute `nix/npm-deps.hash` on `main`. For v0.3.1 the fix commit (b0a7d49, "fix: update lockfile signatures and Nix hash") landed 7 minutes after the release was published, and it contains exactly the hash my build computed. So `main` is always consistent, but the tags never are.

It looks like this affects every release, not just one:

| Release | Published | Hash fix on main |
|---|---|---|
| v0.3.0 | Aug 8, 08:58 | 09:04 |
| v0.3.1 | Aug 9, 11:32 | 11:39 |
| v0.4.0-beta.1 | Aug 12, 18:42 | 18:48 |
| v0.4.0-beta.2 | Aug 12, 21:20 | 21:27 |

(I verified the pins match: my build used the same nixpkgs revision as the flake.lock at the tag, so this is not the known fetchNpmDeps-differs-across-nixpkgs issue that `nix/package.nix` mentions.)

Would you consider making tags self-consistent? Two ideas:

1. Run `scripts/update-nix.sh` as part of the release flow, before tagging, so the tagged commit already carries the right hash.
2. Or have the release job wait for the hash-fix commit and tag that one instead.

This also matters for the binary cache idea in #3307: a cache over release tags only helps if the tagged commits actually build.

My current workaround is overriding `npmDepsHash` from my own flake, which works but needs the correct hash for every new release.

## Additional release-tag evidence — #4407

Additional evidence from [@ManUtopiK’s #4407](https://github.com/getpaseo/paseo/issues/4407): release-tag npm dependency builds with the tag’s own flake/nixpkgs fail with the following reported hashes, including the same v0.7.2 result from two nixpkgs revisions.

| Tag | Specified | Reported actual |
|---|---|---|
| v0.7.0 | `sha256-YiXz9rVJxzhtKdvYsg5UhVXTaetvH0sdJ/8dyfvSGQQ=` | `sha256-HFYBrCP62r6ZofazBIGSOnMhk8IIEmv+TFJMpqTIJQ8=` |
| v0.7.1 | `sha256-HFYBrCP62r6ZofazBIGSOnMhk8IIEmv+TFJMpqTIJQ8=` | `sha256-j3lk7o4W10hcXeNIN3kGlqSHn8aCr+GCDkKYZNXNSrw=` |
| v0.7.2 | `sha256-j3lk7o4W10hcXeNIN3kGlqSHn8aCr+GCDkKYZNXNSrw=` | `sha256-0hOGev0HglOQmofzPQMfiWh1opg6cpiEgsfK22AKcGk=` |

Reproduction: `nix build github:getpaseo/paseo/v0.7.2#packages.x86_64-linux.default.npmDeps`. Fresh tag metadata confirms the specified values; actual derivation hashes above remain reporter evidence, not independently rebuilt. Current main’s version lifecycle omits hash refresh and Nix Update Hash runs after a main push. Preserve immutable existing tags; completion requires a newly released tag whose npm dependency derivation matches its recorded hash. Do not infer that the newest beta fails without building that exact tag.

### Independent completion criteria

- Original report by @m9tdev: account for the v0.3.1 specified/got mismatch and the original same-nixpkgs evidence above; record the supported pin/workaround without rewriting the immutable tag. A main-branch hash refresh alone does not establish that this case is resolved.
- @ManUtopiK’s #4407: retain the v0.7.0/v0.7.1/v0.7.2 observations as separately assessable release cases; verify a newly released immutable tag’s npm dependency derivation against that tag’s recorded hash and own flake inputs before declaring the release-ordering failure resolved. The reported actual hashes have not been independently rebuilt here.

The implementation proposed in [#3310](https://github.com/getpaseo/paseo/pull/3310) was closed without merging; the release-process approach remains the maintainer’s decision. This consolidation does not mark either report fixed.

m9tdev:
Fixes #3308.

Release tags currently ship a stale `nix/npm-deps.hash`: the version bump changes `package-lock.json`, the tag gets cut, and the hash only gets corrected afterwards by the bot commit on `main`. That makes every tagged release fail to build with Nix (details and a table of affected releases in the issue).

This adds one step to the root `version` lifecycle script, which `npm version` runs before creating the release commit that gets tagged:

```
npm run version:sync-internal && npm run release:prepare && ./scripts/update-nix.sh && git add -A
```

`scripts/update-nix.sh` is the same script CI already runs; doing it before the commit means the tagged commit carries the correct hash, and the post-release bot commit becomes a no-op.

Two tradeoffs to be aware of:

1. The release environment now needs `nix` installed, since `update-nix.sh` resolves `prefetch-npm-deps` through `nix shell`. If that is a problem for whoever cuts releases, this could be guarded with `command -v nix` at the cost of tags staying stale on machines without Nix.
2. It adds a few minutes to a release: `prefetch-npm-deps` downloads all npm tarballs to compute the hash.

I could not test the release flow end to end (it involves publishing), so please treat this as a proposal. The change is deliberately minimal; happy to adjust if you would rather hook it into `set-release-version.mjs` or the tag push script instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "build:desktop": "npm run build:app-deps:clean && cd packages/app && cross-env PASEO_WEB_PLATFORM=electron npx expo export --platform web && cd ../.. && npm run build --workspace=@getpaseo/desktop --",
     "db:query": "npm run db:query --workspace=@getpaseo/server --",
     "cli": "./scripts/dev-home.sh npx tsx packages/cli/src/index.js",
-    "version": "npm run version:sync-internal && npm run release:prepare && git add -A",
+    "version": "npm run version:sync-internal && npm run release:prepare && ./scripts/update-nix.sh && git add -A",
     "version:sync-internal": "node scripts/sync-workspace-versions.mjs",
     "release:prepare": "npm install --workspaces --include-workspace-root",
     "version:all:patch": "node scripts/set-release-version.mjs --mode patch",


boudra:
Thanks for taking the time on this. It isn’t something I want in Paseo, so I’m closing it. The [contributing guide](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md) covers how I decide, and [Discussions](https://github.com/getpaseo/paseo/discussions) is the place to make the case for the workflow behind it.

PR #3317, opened 2026-08-13
fix(app): reconcile workspace draft modes

### LINKED ISSUE #3316: bug: New Workspace submits unsupported modes when creating a Pi agent
### What's broken

Creating the initial agent for a newly added repository can fail when Pi is selected as the provider. The workspace is created, but the agent is not.

The app sends a non-empty mode such as `agent` or `full-access` even though Pi advertises no modes. The daemon then rejects the create request:

```text
Invalid mode 'agent' for provider 'pi'. Available modes: (none)
```

Expected: when the selected provider advertises no modes, the create request should omit `modeId` and the Pi agent should be created.

### Steps to reproduce

User-visible path observed on 0.4.0-beta.2:

1. In the desktop app, add a repository and start a New Workspace.
2. Choose Pi as the provider and select a model.
3. Enter the first prompt and submit it.
4. The workspace is created, but creating the initial Pi agent fails with the error above.

The failure is deterministic when the New Workspace handoff carries a stale/requested non-empty mode:

1. Store a pending workspace draft submission with `provider: "pi"` and `modeId: "agent"` (the value can come from inherited or previously stored create-agent state).
2. Let the destination workspace resolve the Pi provider snapshot. Pi resolves with `modes: []`.
3. Auto-submit the pending draft.
4. The app forwards `autoSubmitConfig.modeId` without reconciling it against the resolved mode list.
5. The daemon rejects the explicit mode.

A focused regression test reproduces the pre-fix behavior with:

```text
AssertionError: expected 'agent' to be null

Expected: null
Received: "agent"
```

### Where did this happen

Desktop (Electron)

### Paseo version

0.4.0-beta.2

The affected New Workspace submission code is unchanged on current `main` as of 2026-08-13.

### OS version

macOS, Darwin 27.0.0, arm64

### Agent provider

Pi

### Provider configuration

Pi is available and its provider snapshot resolves successfully. Its provider definition advertises:

```text
defaultModeId: null
modes: []
```

The selected Pi model is not relevant to the failure; the request is rejected during mode validation before agent startup.

### Logs

Relevant daemon/app log lines, with local paths and identifiers redacted:

```text
{"level":30,"daemonVersion":"0.4.0-beta.2","cwd":"<repo-path>","provider":"pi","msg":"Creating agent in <repo-path> (pi)"}
{"level":50,"daemonVersion":"0.4.0-beta.2","err":{"type":"Error","message":"Invalid mode 'agent' for provider 'pi'. Available modes: (none)"},"msg":"Failed to create agent"}
[error] [AgentInput] Failed to send message: Error: Invalid mode 'agent' for provider 'pi'. Available modes: (none)
```

The same request path also produced:

```text
Invalid mode 'full-access' for provider 'pi'. Available modes: (none)
```

The daemon remained healthy during the incident: direct TCP was active with 1 ms RTT, no buffered WebSocket data, no validation/connectivity failures, and no pending permissions.

### Root cause

The New Workspace flow intentionally reconciles the form's selected mode against the modes discovered for the target provider. However, a mode carried by the pending auto-submit configuration bypasses that reconciliation and is submitted verbatim. This is invalid for modeless providers such as Pi and can also be invalid when switching between providers with different mode IDs.

Server-side validation is behaving correctly; the client should reconcile the final request at the submission boundary.

dundunHa:
### Linked issue

Closes #3316

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

New Workspace creation is a two-stage flow: the first screen stores the selected agent configuration, then the destination workspace auto-submits that configuration after its provider snapshot resolves.

The existing submission code reconciled the live form mode against the destination provider's available modes, but trusted a mode carried in the pending auto-submit configuration. That allowed stale or cross-provider values such as `agent` and `full-access` to reach the server unchanged. Pi advertises `modes: []`, so its correctly strict server-side validation rejected the initial agent and left users with a newly created workspace but no agent.

This change performs the reconciliation once at the final submission boundary. Modeless providers omit `modeId`; providers with modes keep a valid requested/selected mode or fall back to their first available mode. The optimistic draft snapshot uses the same result as the create request.

### Goals

- Allow the initial Pi agent to be created in a new workspace even when pending form state carries a stale mode.
- Never send a mode that the resolved destination provider does not advertise.
- Keep the optimistic draft agent mode consistent with the actual create request.
- Preserve valid requested and selected modes for providers that support modes.
- Add a deterministic regression test for the modeless-provider case.

### Non-goals

- Relaxing server-side mode validation.
- Changing Pi's provider manifest or adding modes to Pi.
- Clearing saved create-agent preferences during provider discovery.
- Changing provider/model selection UI.

### QA

Reproduced before the fix with a pending Pi draft carrying `modeId: "agent"` while the resolved Pi mode list was empty. The regression assertion failed with:

```text
AssertionError: expected 'agent' to be null

Expected: null
Received: "agent"
```

After the fix:

```text
$ node node_modules/vitest/vitest.mjs run packages/app/src/screens/workspace/workspace-draft-agent-config.test.ts

Test Files  1 passed (1)
Tests       5 passed (5)
```

The tests cover:

- a valid selected mode;
- a valid requested mode;
- fallback when a requested mode is not advertised;
- omission of an inherited/requested mode for a modeless Pi provider;
- the existing workspace draft config shape.

The repository pre-commit hook also passed on Node 24.14.1:

```text
✓ format
✓ lint — 0 warnings, 0 errors
✓ typecheck — all workspaces
```

Additional checks run successfully:

- `npm run lint`
- `npm run typecheck`
- `npm run format:check`
- `git diff --check`

Platform evidence:

- Desktop/macOS arm64, Paseo 0.4.0-beta.2: original failure observed in app and daemon logs.
- Patched desktop GUI: not manually rebuilt/retested; behavior is validated at the final request-config boundary by the focused unit test.
- No visual screenshots are included because this change has no visual output.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/composer/draft/workspace-tab.tsx
+++ b/packages/app/src/composer/draft/workspace-tab.tsx
@@ -17,7 +17,10 @@ import type { CreateAgentInitialValues } from "@/hooks/use-agent-form-state";
 import { useDraftAgentCreateFlow, type DraftCreateAttempt } from "@/composer/draft/create-flow";
 import { resolveTurnPresentation, TURN_LIVENESS_IDLE } from "@/timeline/turn-liveness";
 import { useHostRuntimeClient, useHostRuntimeIsConnected } from "@/runtime/host-runtime";
-import { buildWorkspaceDraftAgentConfig } from "@/screens/workspace/workspace-draft-agent-config";
+import {
+  buildWorkspaceDraftAgentConfig,
+  resolveWorkspaceDraftModeId,
+} from "@/screens/workspace/workspace-draft-agent-config";
 import { buildDraftStoreKey } from "@/stores/draft-keys";
 import { usePanelStore } from "@/stores/panel-store";
 import { useCreateFlowStore } from "@/stores/create-flow-store";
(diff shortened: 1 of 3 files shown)


boudra:
Thanks for this. I’m taking [#3316](https://github.com/getpaseo/paseo/issues/3316) because the fix needs to prevent the invalid Pi mode from being saved in the first place. I’m closing this PR and keeping the issue open for that work.

PR #3324, opened 2026-08-13
fix(server): buffer ACP session updates received before session registration

desmond-rai:
## Problem

An ACP agent may emit session-scoped notifications immediately after the `session/new` response - before the client's response continuation has assigned `sessionId` (`ACPAgentSession.initializeNewSession`). `sessionUpdate()` silently drops those notifications because `params.sessionId !== this.sessionId` while `this.sessionId` is still `null`.

Concrete user-visible failure: an ACP agent that pushes its `available_commands_update` right after `session/new` (Hermes advertises installed skills as slash commands this way) never has its command batch cached, so `listCommands()` returns `[]` forever and the slash-command popup stays empty for the whole session. The same drop eats `usage_update` and any other early notification.

Observed on a live daemon log at session creation:

```
session/update → available_commands_update → dropped (session id not yet assigned)
list_commands_response → commands: []
```

## Fix

Buffer session updates that arrive while `sessionId` is still unassigned (capped at 100 per session) and replay them in arrival order immediately after `sessionId` is assigned in `initializeNewSession`. Updates addressed to a different session id are still discarded, and post-registration routing is unchanged.

## Tests

Three new regression tests in `acp-agent.test.ts` (`ACPAgentSession pre-registration session updates`):

- `available_commands_update` delivered synchronously with the `session/new` response is applied and returned by `listCommands()` (fails before the fix: returns `[]`)
- multiple buffered updates replay in arrival order
- buffered updates addressed to a different session id are ignored

## Verification

- Red-green: the two delivery tests fail on the base commit, pass with the fix
- `acp-agent.test.ts`: 99/99 passed
- `packages/server` typecheck: clean
- Full workspace lint/format/typecheck via lefthook pre-commit: green
- Full server unit suite (`npm run test:unit`): 4768 passed, 11 failed in unrelated Hub/websocket/git-service files. Reconciliation: 8 of 11 are 30s+ timeouts under parallel-file load and pass in isolation; `workspace-service-port-allocator` fails identically on the clean base commit (pre-existing); `workspace-git-service.observation` is load-flaky (passes on isolated rerun). None import or exercise ACP provider code.
- Provider e2e suites requiring live CLIs (codex/opencode/claude auth) not run; unrelated to this change

Note: the same race exists one layer up in the shared-process router proposed in #3194; a companion commit there covers that path.

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -2403,6 +2403,140 @@ describe("ACPAgentSession slash commands", () => {
   });
 });
 
+describe("ACPAgentSession pre-registration session updates", () => {
+  /**
+   * ACP agents may emit session-scoped notifications (for example
+   * `available_commands_update`) immediately after the `session/new` response,
+   * before the client's response continuation has assigned `sessionId`. These
+   * tests pin the buffering behavior that keeps those notifications from being
+   * dropped.
+   */
+  function makeNewSession(newSession: ReturnType<typeof vi.fn>) {
+    class TestSession extends ACPAgentSession {
+      protected override async spawnProcess(): Promise<SpawnedACPProcess> {
+        return {
+          child: createProbeChildStub(),
+          connection: {
+            newSession,
+            prompt: vi.fn(),
+          } as unknown as ClientSideConnection,
+          initialize: { agentCapabilities: {} },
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the early ACP notification handling in [#3130](https://github.com/getpaseo/paseo/pull/3130), which remains open for QA.

PR #3340, opened 2026-08-13
feat(app): play a chime when an agent's turn goes idle

JLarky:
## Summary
- Paseo ported the Say To Me **send** ding (`utils/send-ding.web.ts`) but not its companion: the softer two-tone chime that Say To Me and T3 Code play when a turn finishes (the moment the composer's stop button turns back into send).
- Ports it as `idle-completion-ding.web.ts` (same WAV-synthesis / Web Audio fallback pattern as `send-ding.web.ts`), `audio-queue.web.ts` (serializes it behind any voice note the Say To Me widget is speaking, via the shared `window.speechSynthesis`), and `use-idle-completion-ding.web.ts` (fires only on an active→idle transition on the same thread, stays silent on manual cancellation or thread switches — mirrors T3's exclusion of interrupts).
- Native platforms get no-op stubs, matching the existing `send-ding.ts` split.
- Wired into `composer/index.tsx` next to the existing send ding.

## Note
Repo-wide `typecheck` currently fails on `main` for reasons unrelated to this change (missing `@getpaseo/protocol/agent-labels` exports, `project-settings-screen.tsx` errors). Verified none of it touches these files. Pre-commit hook was bypassed for this reason with maintainer approval.

## Test plan
- [x] `oxlint` clean on all new/changed files
- [x] `oxfmt --check` clean
- [x] New unit tests: WAV shape (`idle-completion-ding.test.ts`) and the transition decision table (`use-idle-completion-ding.test.ts`) — 8/8 passing
- [ ] Manual: let an agent finish a turn and confirm the chime plays and doesn't overlap a spoken voice note


JLarky:
sorry, wrong PR target

PR #3357, opened 2026-08-14
Keep archived OpenCode history sessions inert

wilgon456:
## Summary

- pass the resume purpose through the OpenCode adapter
- keep history-only sessions detached from live OpenCode event/status reconciliation
- avoid aborting the native OpenCode session when an archived history view closes
- promote the session to interactive mode only when a new prompt is explicitly started

## Root cause

Archived agent history is loaded through `resumeSession(..., { purpose: "history" })`, but the OpenCode adapter ignored that purpose. Constructing and subscribing to the resumed session eagerly attached event/status reconciliation, and `close()` unconditionally called the native OpenCode session abort endpoint.

As a result, merely opening archived history could make the agent appear to be running again, and closing that history view could interrupt the underlying live OpenCode session.

## Reproduction

The regression test was first run unchanged against the official `v0.4.0` tag:

- history-only resume started one global OpenCode event stream instead of zero
- closing the history-only session issued one native `sessionAbort` call instead of zero

## Validation

- `npm exec -- vitest run packages/server/src/server/agent/providers/opencode-agent.test.ts` — 98 passed
- `npm run build:server-deps`

### DIFF
--- a/packages/server/src/server/agent/providers/opencode-agent.test.ts
+++ b/packages/server/src/server/agent/providers/opencode-agent.test.ts
@@ -321,6 +321,83 @@ describe("OpenCodeAgentClient adapter smoke tests", () => {
     rmSync(cwd, { recursive: true, force: true });
   });
 
+  test("loads archived history without attaching to or aborting the live OpenCode session", async () => {
+    const cwd = tmpCwd();
+    const runtime = new TestOpenCodeHarness();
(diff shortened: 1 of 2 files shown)


wilgon456:
Closing this draft in favor of #3358, which reports the bug using the repository's bug report format. The fix branch remains available if maintainers want a follow-up patch.

PR #3366, opened 2026-08-14
fix(server): keep background subagents alive when prompting a busy agent

simonepri:
### Linked issue

None.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

If an agent has a Claude background subagent running and you send it a message, the subagent gets killed:

```
{"success":false,"message":"Agent \"sleeper\" was stopped by the user and was not resumed.
 Treat its work as cancelled; only start a new agent for it if the user explicitly asks."}
```

Prompt dispatch checks `hasInFlightRun`, which is true whenever the agent is `running`. A background subagent keeps the parent `running` on an autonomous turn after the parent's own turn ends, so a normal follow-up gets dispatched as a replacement. Replacement cancels the provider turn, and on Claude that is `query.interrupt()`. The CLI treats that control request as stop everything and kills every running task in the session.

There's a second problem on the same path. The provider opens a new autonomous turn between the cancel and the new stream, so the prompt then gets refused with `already has an active run`, after the interrupt already fired. You lose the subagent and the message.

An autonomous run isn't something a prompt needs to replace.

### Goals

- Prompting an agent whose only active work is autonomous starts a turn instead of replacing one
- That prompt stops being refused with `already has an active run`
- Stop, reload and rewind keep using `hasInFlightRun`, so they still end background work
- A superseded autonomous run settles instead of being dropped, so a cancel waiting on it can't hang
- Providers opt in with `acceptsPromptDuringAutonomousTurn`. Only Claude sets it, because its `startTurn` completes the autonomous turn on the way in. OpenCode's `startTurn` throws while its runner is active, so it keeps cancelling first

### Non-goals

- Sending during a live foreground turn still interrupts, and still kills subagents. Fixing that means queueing the message instead of interrupting, like the TUI does when you type mid-turn, and like the VS Code extension does (it enqueues onto the CLI input stream without checking turn state). That needs provider capability gating and a queued state in the composer, so it's a separate PR.
- No change to `hasInFlightRun`, Stop, reload, rewind or the schedule service.
- macOS only.

### QA

macOS 15 arm64, Node 22, Claude Code 2.1.221, claude/claude-opus-5.

Before, on Paseo desktop 0.3.1. The agent spawns a background `sleeper` and finishes its own turn, but still reports `running` with `activeTurn: {"turnId":"autonomous-turn-2"}`. Sending a follow-up:

```
Error: Agent cf513290-45c4-4656-a4d9-7251db3312d5 already has an active run
```

Sending again works, and the agent's `SendMessage` to the subagent comes back:

```
{"success":false,"message":"Agent [id] was stopped by the user and won't be resumed. ..."}
```

To check this isn't just what interrupts do, same setup in the Claude Code TUI with no Paseo involved: spawn the subagent, start a foreground `sleep`, hit Esc. The shell stops, the subagent keeps running, and `SendMessage` returns:

```
{"success":true,"message":"Message sent to sleeper's inbox","msg_id":"56ab9233-3515-44d8-acf4-9de22577d02c"}
```

After, on a dev daemon built from this branch (`npm run dev`, port 6768):

```
$ npm run cli -- ls --json
  "status": "running"

$ npm run cli -- logs c70707f
  [Task]
  [Bash] sleep 400; echo ALIVE
  SPAWNED

$ npm run cli -- send c70707f "<ask it to message the sleeper and paste the result>"
  c70707f1-60f3-43de-8f63-248743089b65  completed   Agent completed processing the message

$ npm run cli -- logs c70707f
  {"success":true,"message":"Message queued for delivery to sleeper at its next tool round."}
```

Accepted first try, no interrupt, subagent still alive.

Tests:

```
agent-manager, agent-prompt, lifecycle-command, permission-response,
  mcp-server, create-agent, stream-coalescing               335 passed
session, session.workspaces, schedule/service, messages,
  claude subagent-interrupt, claude background-subagent,
  opencode-agent, agent-loading, agent-archive              423 passed, 4 skipped
```

Two tests, both of which push an autonomous `turn_started` and then dispatch through `startAgentRun(..., { replaceRunning: true })`.

The first uses a provider that accepts a prompt during an autonomous turn and checks the session was never interrupted. Reverting the source makes it fail on the interrupt count:

```
AssertionError: expected 1 to be +0
```

The second uses a provider whose `startTurn` throws while its autonomous turn is live, the OpenCode shape, and checks the manager still cancels first. Dropping the opt-in makes it fail the same way, from the other side:

```
AssertionError: expected +0 to be 1
```

Test doubles updated for the renamed `AgentRunController` member.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -5497,6 +5497,77 @@ test("failed replacement cancellation preserves an autonomous running state", as
   }
 });
 
+test("prompting an agent whose only run is autonomous starts a turn without interrupting", async () => {
+  const workdir = mkdtempSync(join(tmpdir(), "agent-manager-autonomous-prompt-"));
+  const storage = new AgentStorage(join(workdir, "agents"), logger);
+
+  class BackgroundWorkSession extends TestAgentSession {
+    interruptCount = 0;
+    readonly prompts: AgentPromptInput[] = [];
+
+    override async interrupt(): Promise<void> {
+      this.interruptCount += 1;
+    }
+
+    override async startTurn(prompt: AgentPromptInput): Promise<{ turnId: string }> {
+      this.prompts.push(prompt);
+      return { turnId: "foreground-after-background" };
+    }
+  }
+
+  class BackgroundWorkClient extends TestAgentClient {
+    readonly session = new BackgroundWorkSession({ provider: "codex", cwd: workdir });
+
+    override async createSession(): Promise<AgentSession> {
+      return this.session;
+    }
+  }
+
+  const client = new BackgroundWorkClient();
+  const manager = new AgentManager({
+    clients: { codex: client },
+    registry: storage,
+    logger,
+    idFactory: () => "00000000-0000-4000-8000-000000000131",
+  });
+
+  try {
+    const agent = await manager.createAgent({ provider: "codex", cwd: workdir }, undefined, {
+      workspaceId: undefined,
+    });
+    const running = waitForAgentLifecycle(manager, agent.id, "running");
+
+    // The provider streams for work the session owns in the background — a Claude
+    // background subagent. The agent's own turn is over, but it reads as running.
+    client.session.pushEvent({
+      type: "turn_started",
+      provider: "codex",
+      turnId: "autonomous-background-1",
+    });
+    await running;
+    expect(manager.getAgent(agent.id)).toMatchObject({
+      lifecycle: "running",
+      activeForegroundTurnId: null,
+    });
+
+    await startAgentRun(manager, agent.id, "follow-up while the subagent works", logger, {
+      replaceRunning: true,
+    });
+
+    // Replacing would cancel the session, and for Claude that cancel is the SDK's
+    // stop-everything interrupt: it kills the background subagent this prompt was
+    // most likely asking about.
+    expect(client.session.interruptCount).toBe(0);
+    expect(client.session.prompts).toEqual(["follow-up while the subagent works"]);
+    expect(manager.getAgent(agent.id)).toMatchObject({
+      lifecycle: "running",
+      activeForegroundTurnId: "foreground-after-background",
+    });
+  } finally {
+    rmSync(workdir, { recursive: true, force: true });
+  }
+});
+
 test("waitForAgentEvent waitForActive resolves for autonomous live-event run", async () => {
   const workdir = mkdtempSync(join(tmpdir(), "agent-manager-live-wait-"));
   const storagePath = join(workdir, "agents");
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -5502,6 +5502,7 @@ test("prompting an agent whose only run is autonomous starts a turn without inte
   const storage = new AgentStorage(join(workdir, "agents"), logger);
 
   class BackgroundWorkSession extends TestAgentSession {
+    readonly acceptsPromptDuringAutonomousTurn = true;
     interruptCount = 0;
     readonly prompts: AgentPromptInput[] = [];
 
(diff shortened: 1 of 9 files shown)


joeshull:
Independent reproduction of this on **Linux** (the PR notes macOS-only QA), on released 0.4.0 rather than a dev build. Raw evidence below; I hit it 7 times over ~4 hours of normal use before going looking for a cause.

### Environment

- Paseo CLI + daemon 0.4.0 (`latest`, both sides same version)
- Claude Code 2.1.234, provider `claude/claude-opus-5`
- Ubuntu 24.04.3 LTS, kernel 6.8.0-136, Node v22.22.2
- Messages sent from the mobile/remote app — every occurrence logged `"transport":"relay","peer":"external"`

### What happened

Agent has finished its own turn but still shows `running` because background subagents are alive. I send a follow-up from my phone. The send is refused, and the background subagents are dead.

### Raw daemon log

`~/.paseo/daemon.log`, three consecutive occurrences on one agent (fields trimmed to the relevant ones, values verbatim):

```json
{"time":1786997856579,"iso":"2026-08-17T20:17:36.579Z","level":50,"module":"session","msg":"Failed to send agent message for agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283","err_message":"Agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283 already has an active run","transport":"relay","peer":"external"}
{"time":1787000471164,"iso":"2026-08-17T21:01:11.164Z","level":50,"module":"session","msg":"Failed to send agent message for agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283","err_message":"Agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283 already has an active run","transport":"relay","peer":"external"}
{"time":1787001767719,"iso":"2026-08-17T21:22:47.719Z","level":50,"module":"session","msg":"Failed to send agent message for agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283","err_message":"Agent 34e74f11-3bd9-4b09-80dd-a56b0ccc9283 already has an active run","transport":"relay","peer":"external"}
```

7 occurrences total across two agents (`34e74f11…`, `8e7f3532…`). All 7 were `relay`/`external`; zero over the loopback transport.

### Raw Claude Code side, same millisecond

For the 21:22:47.719 entry above, the subagent that died —
`~/.claude/projects/-home-joe--paseo-worktrees-3qu8gw67-cruel-lion/b219d406-4b9f-4780-8dc5-e79467d05607/subagents/agent-abbc10a2764ccac50.meta.json`:

```json
{"agentType":"general-purpose","description":"Mechanism-ceiling diagnostic","toolUseId":"toolu_01BDLBQJceD8357jJJTEYWMA","spawnDepth":1,"stoppedByUser":true}
```

Last two lines of its transcript `agent-abbc10a2764ccac50.jsonl`:

```json
{"timestamp":"2026-08-17T21:22:25.661Z","type":"assistant","content":[{"type":"thinking",...}]}
{"timestamp":"2026-08-17T21:22:47.718Z","type":"user","content":[{"type":"text","text":"[Request interrupted by user]"}]}
```

Subagent interrupt at `21:22:47.718`, daemon refusal at `21:22:47.719`. Across all 7 events, 9 subagents died (two events each took a parent+child subagent pair), and every subagent interrupt timestamp matches a daemon refusal to within 1 ms.

### The message

In all 7 cases the message did land on a retry 2–5 s later, e.g. the parent transcript for the 21:22:47.719 refusal:

```json
{"timestamp":"2026-08-17T21:22:50.369Z","type":"user","content":[{"type":"text","text":"check on your subagents"}]}
```

so from the phone it reads as "message didn't send, tap again" — the loss of the subagents is the part that isn't visible until later.

One limit on the above worth stating: because a refused send is naturally followed by a human re-send, I can't rule out from these logs that the accepted send and the refused send were two different requests. What argues against it is the 0–1 ms gap between the child's interrupt marker and the refusal — two taps over a relay landing 1 ms apart seems implausible — but I can't prove it from what I have.

### Which case these are

Checking each of the 7 parent transcripts for a `[Request interrupted by user]` marker at the event timestamp, to separate "parent's own turn was live" from "parent was only running an autonomous turn":

| event (UTC) | parent foreground turn interrupted? |
|---|---|
| 17:41:18.379 | no marker in parent |
| 18:09:14.796 | **yes** |
| 18:17:24.296 | no |
| 19:39:12.016 | no |
| 20:17:36.578 | no |
| 21:01:11.164 | no |
| 21:22:47.718 | no |

So 6 of 7 are the autonomous-turn case this PR targets — the parent's own turn had already ended and only background subagents kept it `running`. The 18:09:14 one is the live-foreground-turn case listed under non-goals; noting it only so the counts add up, not as a request to widen scope.

### Repro

1. Have the agent spawn a background subagent that runs for several minutes, and let the agent's own turn finish. `paseo ls` shows `running`.
2. Send it any follow-up from the remote/mobile app.
3. Daemon logs `already has an active run`; the send appears to fail in the app.
4. Send again — it is accepted.
5. Ask the agent to message the subagent: it reports the subagent was stopped. `stoppedByUser: true` is in the subagent's `.meta.json` and its transcript ends in `[Request interrupted by user]`.

Two offers, if either is useful:

- This daemon was logging at level 30, so the records that would actually discriminate here — `agent.manager.stream.reject` with `activeForegroundTurnId` / `hasTrackedRun` — aren't in my logs at all. I can re-run the repro with the daemon at trace level and post the raw output if that would help pin which guard condition is still set.
- Happy to run anything against a build of this branch on Linux, since the QA above is macOS-only.

boudra:
Closing in favor of the active-turn steering approach introduced in merged [#3394](https://github.com/getpaseo/paseo/pull/3394).

PR #3381, opened 2026-08-15
fix(app): bound the queued-message list and scroll the overflow

### LINKED ISSUE #3379: bug: the queued-message list grows without bound and pushes the composer off screen
## Summary

Queued messages render into an unbounded column, so the list grows one row per queued message with no cap and no scroll. Queue enough messages and the track keeps expanding upward, taking vertical space from the timeline and eventually pushing the composer off screen. There is no way to bound it from settings.

## Reproduction

1. Start an agent and send a message so a turn is running.
2. Queue several more messages with Cmd/Ctrl+Enter — around 8 is enough to see it clearly.
3. Watch the queued list grow. Every queued row is rendered at full height; nothing scrolls or collapses.

Reproduces on desktop and web on 0.4.0. I have not checked iOS or Android, but the component is shared, so I would expect the same there.

## Where it comes from

`renderQueueTrack` maps every queued message into a plain `View`:

- `packages/app/src/composer/index.tsx:370` — `<View style={styles.queueTrack}>` wrapping one `QueuedMessageRow` per item, with no cap and no scroll container.
- `packages/app/src/composer/index.tsx:2290` — `queueTrack` sets only `flexDirection` and `gap`; there is no `maxHeight`.

For contrast, `packages/app/src/components/ui/autocomplete.tsx` already solves the same shape: a height-bounded `View` wrapping a `ScrollView`.

## Note on ordering, if you do bound it

The queue drains from the front — `packages/app/src/runtime/host-runtime.ts:2163` (`const next = queue[0]`) — so the top row is the message that goes in next. Whatever the fix, that top row is the one worth keeping visible; auto-scrolling to the newest would hide the message that is about to be sent.

## Remedy is yours to choose

I did not want to presume the design here, since it involves product decisions:

- how many rows stay visible, or whether the cap should be viewport-relative
- scroll the overflow, or collapse it behind something like "+3 more"
- whether the pinned row should be the next-to-send or the most recently queued

Happy to implement whichever direction you pick. I have a local branch that caps the list and scrolls the overflow, with the cap derived from the row box so the visible count cannot drift, plus unit tests for the height math and an e2e spec that queues 8 messages and asserts all 8 survive, the track stays within the cap, the overflow is reachable, and the first row stays pinned at the top. Say the word and I will open it as a PR against whatever behavior you want — including a different one from what I built.

## Environment

- Paseo: 0.4.0
- Platform: macOS 26.5.2 (arm64), desktop and web

edihasaj:
Fixes #3379.

## Problem

`renderQueueTrack` mapped every queued message into a plain `View` with no cap and no scroll, so the list grew one full-height row per queued message. Queue enough and it takes the whole conversation area and starts pushing the composer off screen.

## Before / after

Eight queued messages, same scenario in both.

| Before | After |
| --- | --- |
|  |  |

Before, all eight rows render and the conversation is gone. After, the list stops at five, the conversation is visible again, and the remaining three are one scroll away.

## Change

- `packages/app/src/composer/index.tsx` — the track is now a height-bounded `View` wrapping a `ScrollView`, the same shape `components/ui/autocomplete.tsx` already uses. I followed that rather than relying on `ScrollView` auto-sizing.
- `packages/app/src/composer/queue-track-metrics.ts` (new) — the cap is derived from the row box, not a pixel literal:

  ```
  rowHeight = actionButtonSize + spacing * 2 + borderWidth * 2
  maxHeight = rowHeight * visibleItems + spacing * (visibleItems - 1)
  ```

  `queueActionButton` now reads its size from the same constant, so the height math and the row it measures cannot drift apart.

**Order is unchanged and deliberate.** The queue drains from the front (`runtime/host-runtime.ts:2163`, `const next = queue[0]`), so the top row is the message that goes in next. The list does not auto-scroll: scrolling to the newest would hide the message about to be sent. Covered by an assertion.

Five visible rows is the one number I picked, and it is a one-line change if you want a different count or a viewport-relative cap. Two other directions I considered and did not take, in case you prefer one: collapsing the overflow behind a "+3 more" affordance, or pinning the most recently queued row instead of the next-to-send.

## QA evidence

**The e2e spec fails on `main` and passes with the change.** I ran the committed spec against an otherwise untouched `main` composer to confirm it actually catches the bug:

```
# main composer, this branch's spec
1 failed
  [browser] › composer-queue-bounds.spec.ts:17:5 › a long queue is bounded and scrolls instead of growing without end

# with the fix
1 passed (15.3s)
```

The spec queues 8 messages and asserts all 8 still exist (the cap is visual — it must not drop queued work), the track stays within the cap, the overflow is actually reachable, `scrollTop` stays 0, and row 1 remains visible.

```
$ npx vitest run src/composer --root packages/app
Test Files  21 passed (21)
     Tests  152 passed (152)

$ npm run lint          → 0 warnings, 0 errors
$ npm run format:check  → all matched files use the correct format
$ npm run typecheck     → clean across all workspaces
```

`queue-track-metrics.test.ts` covers the height math directly: exact fit for the visible rows, one fewer gap than rows, that the result tracks the row box so the visible count cannot drift, and that a zero count collapses instead of returning a negative height.

## Platforms

Tested on macOS 26.5.2 (arm64), desktop and web, via the browser e2e project.

**Not tested: iOS and Android.** The composer is shared, so the cap should apply there too, but swapping a `View` for a `ScrollView` changes touch behavior — the queue becomes a nested vertical scroll area inside the composer. That deserves a check on a real device before this lands, and I could not do it here. Flagging it rather than implying coverage I do not have.

### DIFF
--- a/packages/app/e2e/browser/composer-queue-bounds.spec.ts
+++ b/packages/app/e2e/browser/composer-queue-bounds.spec.ts
@@ -1,0 +1,60 @@
+import type { Page } from "@playwright/test";
+import { expect, test } from "../support/fixtures";
+import {
+  fillComposerDraft,
+  sendDraftToQueue,
+  startRunningMockAgent,
+} from "../support/helpers/composer";
+import { queueTrackMaxHeight } from "@/composer/queue-track-metrics";
+
+const QUEUED_COUNT = 8;
+
+async function queueMessage(page: Page, prompt: string): Promise<void> {
+  await fillComposerDraft(page, prompt);
+  await sendDraftToQueue(page);
+}
+
+test("a long queue is bounded and scrolls instead of growing without end", async ({
+  page,
+}, testInfo) => {
+  test.setTimeout(120_000);
+  const agent = await startRunningMockAgent(page, {
+    prefix: `queue-bounds-${testInfo.workerIndex}-`,
+    model: "one-minute-stream",
+    prompt: "Keep the agent running while messages queue.",
+  });
+
+  try {
+    for (let index = 1; index <= QUEUED_COUNT; index += 1) {
+      await queueMessage(page, `queued message ${index}`);
+    }
+
+    // Every queued message survives — the cap is visual only, it must not drop work.
+    await expect(page.getByRole("button", { name: "Send queued message now" })).toHaveCount(
+      QUEUED_COUNT,
+    );
+
+    const track = page.getByTestId("composer-queue-track");
+    await expect(track).toBeVisible();
+
+    const cap = queueTrackMaxHeight({ spacing: 8, borderWidth: 1 });
+    const box = await track.boundingBox();
+    expect(box).not.toBeNull();
+    // Eight rows would exceed the cap if the list still grew freely.
+    expect(box?.height ?? 0).toBeLessThanOrEqual(cap + 1);
+
+    // ...and the overflow stays reachable rather than being clipped away.
+    const scroll = page.getByTestId("composer-queue-scroll");
+    const overflow = await scroll.evaluate(
+      (node: HTMLElement) => node.scrollHeight - node.clientHeight,
+    );
+    expect(overflow).toBeGreaterThan(0);
+
+    // The message that goes in next stays pinned at the top; no auto-scroll.
+    const scrollTop = await scroll.evaluate((node: HTMLElement) => node.scrollTop);
+    expect(scrollTop).toBe(0);
+    await expect(track.getByText("queued message 1")).toBeVisible();
+  } finally {
+    await agent.cleanup();
+  }
+});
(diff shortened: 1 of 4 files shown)


edihasaj:
Rebased onto current `main` (`8d92e1cc1`). Composer tests pass (188), the queue-bounds browser E2E passes, and full lint/format/typecheck are clean on head `8e371958f`. Branch is 0 commits behind and mergeable; fork Actions await maintainer approval.

edihasaj:
Rebased onto current main (4f796181bf). Queue metric coverage passes (5 tests), the real browser bounded-scroll queue flow passes, and format/lint/full workspace typecheck are clean on 8ecde50dc. Branch is 0 commits behind and mergeable. All three GitHub Actions suites are action_required and need maintainer approval.

edihasaj:
Maintenance: rebased onto `origin/main` at `88f410567` and force-pushed with lease. Mergeable; no unresolved review threads. Evidence: composer queue metrics tests (5) and app typecheck pass. Fork CI/Docker/Nix runs are `action_required` pending maintainer approval.

PR #3383, opened 2026-08-15
fix(app): copy the branch name, not the workspace title, in the sidebar

cleiter:
### Type of change

- [x] Bug fix

### Reasoning

"Copy branch name" in the sidebar workspace menu puts the workspace title on the clipboard instead of the branch. The same action works correctly in the workspace header menu and in the hover card, so the clipboard holds a different value depending on where you copied from.

### Goals

- Copy the branch name to the clipboard from every sidebar workspace menu, in both grouping modes.
- Hide "Copy branch name" when the workspace has no branch, so the menu never shows an action that does nothing.
- Translate the two toasts in the status-grouped row that were still hardcoded English.
- Cover the fix with an end-to-end test that fails against the old code.

### Non-goals

- No shared hook for the two row implementations. The duplicated action handlers are why the two rows drifted apart, but that refactor rewrites the row that works today.
- No deletion of `packages/app/src/components/sidebar/sidebar-workspace-row.tsx`, which is a third unused copy of the same row. It is a separate change.
- No change to the unawaited clipboard write. Every copy action in these files reports success before the write resolves, and a fix belongs across all of them.
- No new i18n keys. The two toasts here use keys that already exist.

### QA

**Tests:**

- `packages/app/e2e/browser/sidebar-copy-branch-name.spec.ts` (new) - copies the branch name from a status-grouped sidebar row, for a workspace whose seeded title differs from its branch, and reads the clipboard back. Verified failing against the unfixed code, where it reads the title.

Before the change, "Copy branch name" in "Group by status" put the workspace title on the clipboard. After the change it puts the branch there. I checked the same workspace in "Group by project", in the workspace header menu, and in the hover card, and all of them now give the same string. With a detached HEAD the action no longer appears in the row menu. The app took about 30 seconds to show that state after the checkout.

Detached HEAD now hides "Copy branch name":

image

The other changes are not visual.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/sidebar-copy-branch-name.spec.ts
+++ b/packages/app/e2e/browser/sidebar-copy-branch-name.spec.ts
@@ -1,0 +1,41 @@
+import { test, expect, type Page } from "../support/fixtures";
+import { gotoAppShell } from "../support/helpers/app";
+import { seedWorkspace } from "../support/helpers/seed-client";
+import { getServerId } from "../support/helpers/server-id";
+import { copyBranchNameFromSidebar, selectSidebarStatusGrouping } from "../support/helpers/sidebar";
+
+async function readClipboard(page: Page): Promise<string> {
+  return page.evaluate(() => navigator.clipboard.readText());
+}
+
+// The sidebar has one row implementation per grouping mode, and each one wires the menu's
+// copy action itself. The status-grouped row copied the workspace name — which the daemon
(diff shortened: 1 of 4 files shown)


boudra:
Fixed in [#3013](https://github.com/getpaseo/paseo/pull/3013), which is merged. Closing this PR.

Issue #2904, opened 2026-08-05
bug: composer typing corrupted when macOS autocorrect fires (word deleted, cursor jumps back)


jvanderberg:
### What's broken

While typing in the main composer, the word I'm typing sometimes gets deleted out of nowhere. The cursor jumps back to the start of the word and the characters I typed are gone. It happens at the moment macOS autocorrect / predictive text kicks in on that word.

Confirmed: turning off "Correct spelling automatically" and "Show inline predictive text" in System Settings > Keyboard makes the problem go away completely.

### Steps to reproduce

1. App Store build running on an Apple Silicon Mac ("Designed for iPad"), hardware keyboard, macOS autocorrect settings on (default)
2. Type normal prose in an agent's composer
3. Intermittently (can't trigger on demand): mid-word, the word is deleted and the cursor is left at its start

### Raw pointers from poking at the code

- The composer input (`packages/app/src/composer/input/input.tsx`) never sets `autoCorrect`, so RN's default (enabled) applies. Other inputs in the app pass `autoCorrect={false}` (command center, terminal input, settings fields).
- The composer value is fully controlled and round-trips through the draft store on every keystroke (`composer/draft/input-draft.ts` -> `stores/draft-store/`), the setup known to fight in-flight autocorrect/marked text in controlled TextInputs.
- e90832e ("fix(app): preserve native IME composition state") addressed the same class of collision for IME composition.

### Environment

- Paseo 0.2.4 (App Store) on macOS 15.4.1, Apple Silicon
- Surface: iOS app running on macOS
- Provider: not relevant

boudra:
The composer input changes for this report are in [#3343](https://github.com/getpaseo/paseo/pull/3343). @jvanderberg, they are not in the current App Store build; please try the next build after that PR ships and reopen this issue if the problem remains. Closing for now.

PR #3387, opened 2026-08-15
fix(terminal): let Korean be typed in the mobile terminal

### LINKED ISSUE #3384: bug: Korean cannot be typed in the mobile terminal (no globe key, and IME composition is dropped)
### What happened

On iOS I cannot type Korean into a terminal. Two things go wrong, one after the other:

1. **The globe key is missing from the keyboard.** In a terminal there is no way to switch to the Korean layout at all — the 한/영 toggle is not offered.
2. **Even with a Korean layout, nothing arrives.** Typing 한글 puts a single stray `ㅎ` on the line and then nothing. The rest of the keystrokes vanish.

The chat composer on the same phone is fine — Korean types normally there. Desktop terminals are fine too. This is specific to the mobile terminal.

### Steps to reproduce

1. Open a workspace terminal on iOS.
2. Try to switch the keyboard to Korean — the globe key is not there.
3. (Force a Korean keyboard by making it the only other keyboard, or use Android where the layout is reachable.) Type `한글`.
4. The terminal shows `ㅎ` and nothing else. Latin typing in the same terminal works.

### Root cause

Both symptoms live in `packages/app/src/terminal/native-renderer/terminal-input.native.tsx`.

**Missing globe key** — the hidden input sets `keyboardType="ascii-capable"`. On iOS that restricts the keyboard to ASCII-capable layouts, so the globe key that reaches a Hangul keyboard is not shown.

**Dropped keystrokes** — `receiveTextChange` only forwards a change that extends the previous buffer:

```ts
if (!text.startsWith(previousText)) {
  previousText = text;
  return { data: "", shouldClear: false };
}
```

A Korean IME rewrites the syllable it is composing **in place** rather than appending:

| keystroke | buffer | `startsWith(previous)` | forwarded |
| --- | --- | --- | --- |
| ㅎ | `ㅎ` | `"ㅎ".startsWith("")` → true | `ㅎ` |
| ㅏ | `하` | `"하".startsWith("ㅎ")` → **false** | *(dropped)* |
| ㄴ | `한` | `"한".startsWith("하")` → **false** | *(dropped)* |

So the first jamo is forwarded and every later keystroke is swallowed. The swallow itself is deliberate — it stops autocorrect replacements from corrupting a line the terminal has already committed — but it does not distinguish an autocorrect rewrite from IME composition.

Pasting Korean works, because a paste is an append and passes `startsWith`.

### Environment

- iOS, App Store build 0.4.0
- Also affects Android for symptom 2 (`ascii-capable` is iOS-only, so the layout is reachable there, but composition is still dropped)

ljy900825-dev:
### Linked issue

Closes #3384

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

A Korean user cannot type Korean into a terminal on their phone. They reported it as "the 한/영 toggle button isn't there", and that is the first half of it — but fixing only that still leaves the terminal unusable, because the composition never reaches the wire either.

**The globe key is missing.** The hidden terminal input sets `keyboardType="ascii-capable"`. On iOS that restricts the keyboard to ASCII-capable layouts, which removes the globe key — the way you reach a Hangul keyboard. There is no route to a Korean layout from a terminal at all.

**Composition is dropped.** `receiveTextChange` forwards a change only when it extends the previous buffer. A Korean IME rewrites the syllable it is composing in place instead of appending:

| keystroke | buffer | `startsWith(previous)` | forwarded |
| --- | --- | --- | --- |
| ㅎ | `ㅎ` | true | `ㅎ` |
| ㅏ | `하` | **false** | *(dropped)* |
| ㄴ | `한` | **false** | *(dropped)* |

One stray jamo lands and everything after it disappears.

The swallow is deliberate and worth keeping — it stops an autocorrect replacement from rewriting a line the terminal already committed, which is issue #2904's failure mode in the composer. It just could not tell composition apart from replacement. This PR draws that line: exactly one trailing Hangul character rewritten is composition, anything wider is a replacement and stays swallowed.

Latin typing is unaffected — it appends, so it never reaches the new branch.

### Goals

- Reach a Korean keyboard from a mobile terminal (drop `ascii-capable`).
- Forward Hangul composition to the terminal as a rubout plus the replacement syllable.
- Keep the existing autocorrect/suggestion swallow intact.
- Keep Hangul off the keypress path, so the jamo the IME reports and the syllable the text change reports do not both go on the wire.

### Non-goals

- **Japanese and Chinese.** Their composing region spans several characters and conversion replaces the whole run, so the one-character rule here does not cover them. They are no worse than before. Worth a follow-up with its own tests rather than a guess folded into this PR.
- **Widening the swallow.** A multi-character rewrite is still dropped. That is the conservative side of the trade: a missed edit is recoverable, a corrupted committed line is not.
- **Smart punctuation.** Dropping `ascii-capable` means iOS picks the layout. `autoCorrect`, `spellCheck` and `autoCapitalize` are already off, which is what keeps the terminal safe; the keyboard type was never what did it.

### QA

**Reproduced first.** iOS, App Store build 0.4.0, workspace terminal: no globe key on the keyboard, and with a Hangul keyboard forced, typing `한글` puts `ㅎ` on the line and nothing more. The same phone types Korean fine in the chat composer, and pasting Korean into the terminal works — a paste is an append, so it passes `startsWith`. That asymmetry is what pointed at the diff.

**Unit tests**, `npx vitest run packages/app/src/terminal/native-renderer/terminal-input.native.test.ts` — 24 passed, 6 new:

- full `한글` sequence, asserting each byte the terminal receives: `ㅎ`, `\x7f하`, `\x7f한`, `ㄱ`, `\x7f그`, `\x7f글`
- final-consonant migration `안` → `아니`, where one keystroke rewrites the syllable and adds another
- Hangul does not dispatch through the keypress path
- a two-syllable rewrite (`한글` → `우리`) is still swallowed
- a non-Hangul trailing edit (`ls -a` → `ls -l`) is still swallowed
- Backspace decomposing `한` → `하` stays aligned

The 18 pre-existing tests pass unchanged, including `does not translate replacement edits into terminal text` and `does not duplicate printable keypresses when the native text change also arrives`.

`npm run typecheck`, `npm run lint`, `npm run format` all pass.

**What I could not verify:** I do not have a device build, so the globe key returning is reasoned from `ascii-capable` semantics, not observed. Someone with an iOS device should confirm the keyboard shows the globe key and that a real Korean IME drives the sequence the tests encode. The composition logic itself is pure and fully covered above.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/terminal/native-renderer/terminal-input.native.test.ts
+++ b/packages/app/src/terminal/native-renderer/terminal-input.native.test.ts
@@ -228,4 +228,59 @@ describe("native terminal typed input", () => {
       shouldClear: false,
     });
   });
+
+  it("types Hangul by rubbing out the syllable the IME rewrites in place", () => {
+    const input = createTerminalTextInputState();
+
+    // 한글: the composing syllable is rewritten on every jamo, never appended.
+    expect(input.receiveTextChange("ㅎ")).toEqual({ data: "ㅎ", shouldClear: false });
+    expect(input.receiveTextChange("하")).toEqual({ data: "\x7f하", shouldClear: false });
+    expect(input.receiveTextChange("한")).toEqual({ data: "\x7f한", shouldClear: false });
+    expect(input.receiveTextChange("한ㄱ")).toEqual({ data: "ㄱ", shouldClear: false });
+    expect(input.receiveTextChange("한그")).toEqual({ data: "\x7f그", shouldClear: false });
+    expect(input.receiveTextChange("한글")).toEqual({ data: "\x7f글", shouldClear: false });
+  });
+
+  it("types Hangul when a final consonant migrates into the next syllable", () => {
+    const input = createTerminalTextInputState();
+
+    expect(input.receiveTextChange("안")).toEqual({ data: "안", shouldClear: false });
+    // Typing ㅣ moves ㄴ off 안 and onto the new syllable: 안 → 아니.
+    expect(input.receiveTextChange("아니")).toEqual({ data: "\x7f아니", shouldClear: false });
+  });
+
+  it("does not dispatch Hangul through the keypress path", () => {
+    const input = createTerminalTextInputState();
+
+    // The IME reports the jamo it is composing; the text change reports the
+    // syllable. Dispatching both would put two conflicting characters on the wire.
+    expect(input.receiveKeyPress("ㅎ")).toEqual({ data: "", shouldClear: false });
+    expect(input.receiveTextChange("ㅎ")).toEqual({ data: "ㅎ", shouldClear: false });
+    expect(input.receiveKeyPress("ㅏ")).toEqual({ data: "", shouldClear: false });
+    expect(input.receiveTextChange("하")).toEqual({ data: "\x7f하", shouldClear: false });
+  });
+
+  it("still swallows replacement edits wider than one composing syllable", () => {
+    const input = createTerminalTextInputState();
+
+    expect(input.receiveTextChange("한글")).toEqual({ data: "한글", shouldClear: false });
+    // Two syllables at once is a suggestion replacement, not composition.
+    expect(input.receiveTextChange("우리")).toEqual({ data: "", shouldClear: false });
+  });
+
+  it("does not treat a non-Hangul trailing edit as composition", () => {
+    const input = createTerminalTextInputState();
+
+    expect(input.receiveTextChange("ls -a")).toEqual({ data: "ls -a", shouldClear: false });
+    expect(input.receiveTextChange("ls -l")).toEqual({ data: "", shouldClear: false });
+  });
+
+  it("keeps anticipated text aligned when Backspace decomposes a syllable", () => {
+    const input = createTerminalTextInputState();
+
+    expect(input.receiveTextChange("한")).toEqual({ data: "한", shouldClear: false });
+    // The rubout already removed the whole syllable, so the redraw is a plain append.
+    expect(input.receiveKeyPress("Backspace")).toEqual({ data: "\x7f", shouldClear: false });
+    expect(input.receiveTextChange("하")).toEqual({ data: "하", shouldClear: false });
+  });
 });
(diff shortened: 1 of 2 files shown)


ljy900825-dev:
Good catch on the desync — that was a real bug and the sequence diagram nails it.

The swallow advances `previousText` to text the terminal never received, so from that point the buffer and the terminal disagree. My rubout then deletes whatever is actually on the line rather than the character the buffer thinks it is replacing: `한글` swallowed to `우리`, rewritten to `우루`, emits DEL and leaves `한루`.

Fixed in ad4d792 by latching the disagreement when a replacement is swallowed and keeping rubouts off until the line is cleared — `reset()`, empty text, and the newline/submit paths, which are where the buffer and the terminal are known to match again. Appends are untouched, so the pre-existing behaviour after a swallow is unchanged.

Two tests added (26 total, still green):
- `stops rubbing out once a swallowed replacement leaves the terminal ahead` — your exact `한글` → `우리` → `우루` sequence, asserting no DEL is emitted
- `resumes composition once the line is cleared` — composition works again after `reset()`

Worth noting the underlying desync predates this PR: after a swallow, appends were already computed against text the terminal never got. This change does not fix that, it just stops the destructive path from being built on top of it.

ZacharyZcR:
The remaining Backspace-after-desync P1 is fixed in cfcf1feb9 on #3391. The guard updates `previousText` but emits no DEL while `replacementDesynced` is latched, with a focused regression test. Since #3391 currently includes this PR's commits, maintainers can merge #3391 directly; otherwise the small follow-up commit can be cherry-picked after #3387.

boudra:
Thanks for the PR. This work is superseded by https://github.com/getpaseo/paseo/pull/3391, which includes your commits so they land with your attribution. I'm closing this one.

PR #3393, opened 2026-08-15
fix(app): align the macOS window controls with the sidebar toggle

cleiter:
### Type of change

- [x] Bug fix

### Reasoning

On macOS the sidebar toggle sat below the traffic lights instead of next to them, because it centered in the 48px header row while the buttons were pinned a few pixels above that row's center. Two more hand-tuned numbers placed the same buttons: focus mode sent a second offset, so they jumped whenever you entered or left focus mode, and the desktop shell pinned them somewhere else again while it created the window, so they moved a third time a few seconds into startup when the renderer sent its correction. None of the three numbers was tied to the row it had to sit in, or to each other.

Page zoom broke the alignment a fourth way. Zoom multiplies every length the app draws with and cannot move native window buttons, so the toggle drifted further from the buttons at each zoom step. The toggle belongs to the buttons, so it now draws at their scale instead of the content's.

### Goals

- The sidebar toggle sits on the same line as the traffic lights, with the sidebar open and with it closed.
- The traffic lights hold one position. Entering or leaving focus mode does not move them, and neither does app startup.
- One exported center line places the buttons and everything that aligns to them, so no future mode can put them somewhere else.
- Zoom does not move the toggle. It scales by the inverse of the zoom factor from the window's top left corner, which holds its padding, its height and its icon in the points the shell pins the buttons in.

### Non-goals

- The focus mode tab strip keeps its 36px height. Growing it to the header's 48px would let the buttons center exactly in both rows, but focus mode exists to give content the space, so the center line sits between the two rows instead. It is 4px high for the header row and 2px low for the tab strip.
- The top-right window controls on Windows and Linux are untouched. The overlay there has its own height and needs no offset, and whether Chromium scales those controls with page zoom is not something this change assumes either way.
- The button height is a constant measured from a build, not read from the system. Electron exposes no API for it.
- The reserved corner the header pads itself by stays in app pixels, so the gap between the toggle and the header title still grows with zoom. Only the toggle is pinned to the buttons.

### QA

**Tests:**

- `packages/app/src/constants/layout.test.ts` (new) - the placement of the traffic lights: that the offset puts the buttons on the center line the app aligns to, that they stay inside both rows that can own the top-left corner, that the offset stays inside the range the desktop shell accepts, and that it is 0 so nothing moves at startup. The shell drops an out-of-range offset silently, so a row height that pushed past it would otherwise leave the buttons at their pinned position with no error.
- `packages/desktop/src/window/window-manager.test.ts` (extended) - the pin the shell applies when it creates the window, the position it computes from a renderer offset, and that a zoom change reaches the renderer while a window that is tearing down is sent nothing.
- `packages/app/src/utils/desktop-window.test.ts` (extended) - the scale that undoes page zoom, and that a host which reports no usable zoom factor leaves the chrome unscaled.

The app and the shell hold the same base position in two packages, and `packages/desktop` cannot import from `packages/app`. The 0 offset test is what stops them drifting apart: change the center line, the button height, or the shell's pin on its own and it fails, rather than someone finding the startup jump by eye later.

**Before**

Icon alignment:

image

Traffic lights jumping when entering or leaving focus mode:

https://github.com/user-attachments/assets/47d367ab-1527-46f1-ba1f-0bd5829c9052

Toggle leaving the buttons' line as the zoom level changes:

image

image

**After**

Icon alignment:

image

Entering and leaving focus mode:

https://github.com/user-attachments/assets/d63dcac3-e962-4aa1-bf6f-9626ff33e742

Zooming in and out:

image

image

Checked on macOS with the sidebar open, with the sidebar closed, in focus mode, through app startup, and across zoom steps in both directions. Before the change the toggle sat below the buttons in both sidebar states, the buttons moved vertically on every focus mode toggle, they moved again a few seconds after launch, and the toggle left the buttons' line at every zoom step. After the change the toggle is on the buttons' line in both sidebar states, the buttons hold their position through mode changes and startup, and zoom leaves the toggle where it is.

Tested on desktop (Electron, macOS). Not tested: web, iOS, Android, Windows, Linux.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/app/_layout.tsx
+++ b/packages/app/src/app/_layout.tsx
@@ -44,9 +44,9 @@ import { WorkspaceShortcutTargetsSubscriber } from "@/components/workspace-short
 import { FloatingPanelPortalHost } from "@/components/ui/floating-panel-portal";
 import { HostChooserModal, useHostChooser } from "@/hosts/host-chooser";
 import {
+  DESKTOP_TRAFFIC_LIGHT_CENTER_Y,
+  DESKTOP_TRAFFIC_LIGHT_OFFSET_Y,
   getIsElectronRuntime,
-  getIsElectronRuntimeMac,
-  HEADER_INNER_HEIGHT,
   useIsCompactFormFactor,
 } from "@/constants/layout";
 import {
@@ -116,6 +116,7 @@ import type { HostProfile } from "@/types/host-connection";
 import { toggleDesktopSidebarsWithCheckoutIntent } from "@/utils/desktop-sidebar-toggle";
 import {
   useHasWindowChromeObstruction,
+  useNativeChromeScale,
   WindowChromeProvider,
   WindowChromeRegion,
   WindowChromeSafeArea,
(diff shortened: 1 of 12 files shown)


boudra:
Closing in favor of the titlebar implementation merged in [#3826](https://github.com/getpaseo/paseo/pull/3826).

PR #3394, opened 2026-08-15
Add active-turn steering

boudra:
### Linked issue

None.

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Refactor
- [x] Docs

### Reasoning

Users can now choose whether a prompt sent while an agent is running interrupts the current turn, steers that exact turn, or waits in the client queue. The client declares intent; the daemon owns exact-turn admission, provider delivery, safe fallback, canonical message identity, and durable turn membership.

Queueing remains client-side and drains only when the agent becomes idle. Unsupported or definitively unavailable steering falls back once to interrupt-and-replace, while ambiguous delivery failures remain errors and are never retried automatically.

### Goals

- Add Interrupt, Steer, and Queue as configurable active-turn behaviors.
- Steer supported live turns without starting a replacement turn.
- Keep provider capability decisions and graceful fallback in the daemon.
- Preserve one canonical submitted message with the original turn identity across live updates, cache restore, provider history reconciliation, and daemon restart.
- Keep new wire fields optional for app/daemon version drift.
- Preserve footer, timing, copy, and fork boundaries for steered messages in the same visual turn.

### Non-goals

- Server-side prompt queues.
- Provider capability checks in the app.
- Emulating unavailable steering after an ambiguous provider response.
- Native steering for providers without a released, verifiable same-turn API.

### Risk surface

- Active-turn admission and interrupt fallback races.
- Provider-specific same-turn delivery and message echo correlation.
- Timeline persistence, history reconciliation, and turn grouping after restart.
- Cross-version behavior when optional intent or turn-membership fields are absent.

### Supersedes

- [PR #781 — Steer active agent turns from follow-up messages](https://github.com/getpaseo/paseo/pull/781) — This PR takes over the active-turn steering workflow across settings, daemon admission and fallback, provider delivery, durable canonical timeline identity, and real-provider verification.

### QA

Passed locally:

- Full repository formatting, typecheck, lint, and `git diff --check`.
- Focused protocol, app, manager, persistence, reconciliation, and provider suites (655 tests in the broad focused matrix, plus later targeted regressions).
- Real Codex file: 3 tests passed, including foreground tool-call steering and interrupt replacement.
- Real Claude steering: 1 passed, 1 skipped by the provider gate; the active foreground tool call completed after the steered prompt with no replacement turn.
- Held-ack browser test: the optimistic Steer prompt stays inline with no premature footer.

Provider-shaped browser replay:

[...]

PR #3413, opened 2026-08-15
feat(providers): steer prompts into running turns without interrupting

trim21:
### Linked issue

N/A

### Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

When a user sends a message while an agent is running, Paseo aborted the in-flight turn (discarding the tool work already done and the streamed output) and started a fresh turn. Two supported providers expose a native mechanism to accept the prompt mid-turn without interrupting: Pi's `streamingBehavior: "steer"` and Codex's app-server `turn/steer`. This PR lets those providers keep working when the user redirects them, instead of throwing away the current turn.

Steering is now a provider-native capability (`AgentSession.trySteer?`): providers that implement it queue the prompt into the active turn and Paseo records the submitted message immediately, while the running turn keeps broadcasting its events. Providers without native steering (Claude, OpenCode, OMP) are unchanged and still use the interrupt + fresh turn path. Queueing is deliberately NOT delegated to providers — Paseo keeps its own send-when-idle queue (`sendBehavior="queue"`) as the provider-agnostic fallback.

### Goals

- Send a message to a running Pi agent without aborting it: Pi queues the prompt and delivers it after the current tool-call batch, then keeps running the same turn.
- Send a message to a running Codex agent without aborting it: `turn/steer` appends the input to the in-flight turn (requires `expectedTurnId`; Paseo already opts into `experimentalApi`).
- The steered message appears in the timeline under the correct clientMessageId immediately (submitted state), and its provider echo reconciles on delivery.
- `send_agent_message` acknowledges steered prompts immediately; no new run is allocated, so there is nothing to wait for.
- Non-steering providers keep the existing interrupt behavior — no behavior change for Claude/OpenCode/OMP.

### Non-goals

- Moving Paseo's send-when-idle queue from the app client into the daemon. Queue stays a Paseo-side concern and is intentionally not backed by any provider RPC (e.g. Pi `follow_up`).
- Surfacing a pending-steer indicator in the UI (Pi's `queue_update` / Codex steer queue state).
- Steering for Claude/OpenCode/OMP — they have no native equivalent today.

### QA

Behavior verified against the provider protocol docs and unit tests:

- Pi: `prompt` RPC forwards `streamingBehavior: "steer"`; Pi delivers after the current tool-call batch, emits a normal `message_start` (user), and the existing `PASEO_SUBMITTED_USER_ENTRY` extension marker echoes it to the timeline under the steer clientMessageId. Tests: `packages/server/src/server/agent/providers/pi/agent.test.ts` (steer delivery, no active turn fallback, rejected prompt fallback).
- Codex: `turn/steer` appends user input to the active in-flight turn (protocol: `expectedTurnId` must match; returns accepted `turnId`; no new `turn/started`). The next user-message item echo picks up the steer clientMessageId. Tests: `codex-app-server-agent.test.ts` (request params, message echo, decline without active turn, request failure fallback).
- Dispatch: `startAgentRun` returns an explicit `steered` result (no longer reusing `outOfBand`); `send_agent_message` acknowledges steered prompts without waiting for a run start. Tests: `agent-prompt.test.ts` (steer vs replace fallback), `agent-manager.test.ts` (submitted prompt recorded, no-capability/decline fallback).

Ran: `npm run typecheck` ✓, `npm run lint` ✓, `npm run format:check` ✓, `vitest` on the four affected test files (364 tests) ✓.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -9316,3 +9316,128 @@ test("onWorkspaceStateMayHaveChanged is not called for running shell tool calls"
 
   expect(onWorkspaceStateMayHaveChanged).not.toHaveBeenCalled();
 });
+
+test("trySteerAgent records the submitted prompt when the session steers", async () => {
+  const workdir = mkdtempSync(join(tmpdir(), "agent-manager-try-steer-"));
+  const storagePath = join(workdir, "agents");
+  const storage = new AgentStorage(storagePath, logger);
+
+  class SteerSession extends TestAgentSession {
+    steeredPrompts: Array<{ prompt: AgentPromptInput; options?: AgentRunOptions }> = [];
+
+    override async trySteer(prompt: AgentPromptInput, options?: AgentRunOptions): Promise<boolean> {
+      this.steeredPrompts.push({ prompt, options });
+      return true;
+    }
+  }
+
+  class SteerClient extends TestAgentClient {
+    override async createSession(config: AgentSessionConfig): Promise<AgentSession> {
+      return new SteerSession(config);
+    }
+  }
+
+  const manager = new AgentManager({
+    clients: { codex: new SteerClient() },
+    registry: storage,
+    logger,
+  });
+
+  try {
+    const snapshot = await manager.createAgent({ provider: "codex", cwd: workdir }, undefined, {
+      workspaceId: undefined,
+    });
+
+    const steered = await manager.trySteerAgent(snapshot.id, "steer me", {
+      clientMessageId: "msg-steer-1",
+    });
+    expect(steered).toBe(true);
+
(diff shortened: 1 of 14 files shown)


trim21:
Closing this PR: #3394 already implements the full active-turn steering flow (app settings, wire intent, daemon admission and fallback, canonical turn membership, and native Codex/Claude delivery), and this PR overlaps with it using a different session interface.

Happy to follow up with native Pi steering (prompt RPC with streamingBehavior: "steer") on top of #3394's steerActiveTurn interface once it lands — Pi's RPC mode supports queued mid-turn delivery after the current tool-call batch, which fits the same contract.

PR #3421, opened 2026-08-15
feat(app): add selected-text annotations

crpboy:
## Summary

This PR adds response text annotations to Paseo.

Users can now select text from an assistant response, attach a comment to that selection, and include the annotated excerpts in their next message. Existing annotations can be revisited from either the inline numbered marker or the annotation overview in the composer.

The interaction is modeled after response annotation workflows in Codex and ChatGPT, while being adapted to Paseo’s composer, scrolling, overlay, and cross-platform architecture.

## Why

Long agent responses often contain several independent details that need clarification or revision. Previously, users had to copy excerpts manually, paste them into the composer, and explain which part of the response each comment referred to.

That workflow has several problems:

- It loses the visual relationship between a comment and the original response.
- Repeated text can make a quoted excerpt ambiguous.
- Discussing multiple parts of one response produces a noisy message.
- Users cannot easily return to an earlier excerpt to review or edit its comment.
- Manually copied context is more likely to omit formatting or identify the wrong passage.

Response annotations provide a structured way to preserve that relationship. Each annotation keeps the selected text, its source message, its occurrence within that message, and the optional user comment. The annotations are then added to the next turn as explicit context.

In this PR I chose a codex-like approach to implement this.

### Creating an annotation

1. Select text inside an assistant response.
2. Choose **Add to comment** from the contextual action.
3. Enter an optional comment in the compact editor.
4. Confirm the annotation using the icon-only submit button.

The initial editor starts as a single-line capsule. It grows into a multiline rounded panel when the comment becomes longer, with a constrained width and scrollable content.

### Viewing annotations

Each saved annotation displays a small numbered conversation marker next to its source text.

The composer shows only the annotation count by default. Hovering or clicking the annotation tab opens an overview containing the selected excerpts and their comments.

The overview remains collapsed after adding an annotation and closes when it loses focus.

### Navigating and editing

Selecting an annotation from the composer overview:

- scrolls to the source response;
- highlights the annotated text without creating a native browser selection;
- opens the full annotation editor next to the source text.

Clicking an inline numbered marker performs the same action.

The editor is temporarily hidden when focus moves outside it. Long comments remain editable and scroll inside the editor instead of causing the editor to disappear.

### Annotation data model

Added a selected-text attachment model containing:

- the selected Markdown text;
- the source assistant message ID;
- the zero-based occurrence of the selected text within that message;
- the user comment;
- a stable annotation ID used for editing and rendering.

Annotations are stored as part of the composer draft so they survive normal composer state updates and workspace navigation.

Draft validation accepts the new attachment shape while keeping the additional occurrence field optional for compatibility with previously created draft state.

### Message submission

Selected-text attachments are converted into structured response annotations when the composer is submitted.

This keeps the annotation metadata separate from ordinary file and image attachments while still injecting the selected excerpts and comments into the next agent turn.

Submission tests cover annotation conversion and verify that the selected text, source message, occurrence, and comment are preserved.

### Markdown-aware selection

The web selection surface converts DOM selections back into Markdown-aware text instead of relying only on `Selection.toString()`.

The conversion preserves supported response structure, including:

- inline emphasis;
- list items;
- nested response content;
- ignored presentation-only elements such as list markers.

Selections are limited to assistant response content and are rejected when they originate outside the response surface.

### Repeated and overlapping text

A selected excerpt is not identified by text alone.

The implementation records which occurrence of the excerpt was selected within the source response. This allows annotations to resolve the correct passage when identical text appears multiple times.

Occurrence detection uses the exact normalized start of the DOM range. This also handles overlapping matches correctly—for example, selecting the final two characters in `哈哈哈` resolves to occurrence `1` of `哈哈`, rather than occurrence `0`.

### Highlight rendering

Saved annotations use custom highlight geometry rather than restoring a browser-native selection.

The renderer:

- recreates the source DOM range from the stored annotation;
- derives client rectangles for wrapped and multiline text;
- clips highlights to the visible conversation viewport;
- updates geometry during scrolling and layout changes;
- keeps marker payloads synchronized after comment edits;
- supports more than six annotations without relying on a fixed marker limit.

This allows navigation and editing without taking over the user’s current browser selection.

### Numbered annotation markers

Numbered markers are rendered in a dedicated conversation-content overlay plane.

The markers:

- remain attached to the annotated text;
- use a compact conversation-bubble shape;
- include a smooth pointer without raster or border aliasing;
- reopen the corresponding editor when clicked;
- are hidden when their source content is outside the active conversation surface.

Marker placement is recalculated from the source range instead of being permanently positioned from the original selection event.

### Adaptive editor positioning

Both the compact creation editor and the full editing panel use range-based adaptive positioning.

The positioning logic:

- prefers a location next to the selected text;
- flips above or below the range when space is constrained;
- clamps the panel horizontally to the viewport;
- accounts for the visible conversation boundaries;
- repositions during conversation scrolling;
- falls back to a centered position when the source range is outside the usable viewport;
- avoids disappearing when content growth changes the panel dimensions.

The compact creation editor and the full editing panel use different size constraints while sharing the same range anchoring behavior.

### Composer annotation overview

The composer now includes a selected-text annotation overview.

Its implementation includes:

- a collapsed annotation-count tab;
- hover and click expansion on web;
- explicit pinned/open state handling;
- outside-focus dismissal;
- navigation back to the source annotation;
- annotation removal;
- scrolling and maximum-height constraints for large annotation sets.

The overview is rendered through a platform-specific portal layer so the web implementation can escape local clipping contexts while native platforms retain a safe fallback.

### Overlay and stacking behavior

Annotation UI introduced several interactions with existing floating surfaces, including the composer, workspace hover cards, menus, and other global overlays.

This PR separates overlays by responsibility:

- annotation highlights and markers belong to the conversation-content plane;
- the main composer remains above conversation annotations;
- annotation editors and overviews use the appropriate floating-panel layer;
- application menus, workspace previews, and global popovers remain above conversation annotations.

The implementation uses shared overlay roots and stacking tokens instead of giving markers a globally dominant `z-index`. Existing composer collision handling is retained for marker geometry, while highlights use viewport clipping so they cannot paint over the composer.

The workspace hover card was also moved onto the shared global overlay layer to avoid local stacking-context conflicts.

### Focus and dismissal behavior

The annotation editor distinguishes between actual outside interaction and internal editor updates.

This prevents the editor from closing when:

- the textarea grows;
- the comment wraps onto additional lines;
- the save button is used;
- focus moves between controls inside the editor.

True outside focus or pointer interaction temporarily hides the editor. Reopening the marker or overview item restores the editing state.

### Theme support

Annotation surfaces use existing theme tokens for backgrounds, borders, text, and focus states.

The confirmation icon explicitly follows the button foreground color, keeping it visible in both light and dark themes. Focus borders use a subtle theme-derived color instead of a hard-coded green outline.

### Cross-platform behavior

DOM selection, range geometry, portals, and pointer listeners are isolated in web-specific modules.

Native platforms receive compatible component interfaces and safe fallback implementations without importing browser APIs. This keeps the shared composer and agent-stream code cross-platform.

No raw DOM API is introduced into the native bundle.

### Localization

Added labels for the annotation workflow to all existing application locale resources, including:

- adding a selection to the comment;
- annotation counts;
- optional comment placeholders;
- save, edit, and delete actions;
- selected-text context labels.

## Documentation

Updated the floating panel documentation with the stacking and ownership rules introduced by response annotations, including the distinction between conversation-content overlays and global application overlays.

## Validation

Automated validation completed:

- 18 browser interaction tests passed.
- 13 focused unit tests passed.
- `npm run lint` passed.
- `npm run typecheck` passed.
- `npm run format:check` passed.
- Git pre-commit checks passed.

Browser coverage includes:

- initial annotation creation;
- compact and multiline editor behavior;
- repeated and overlapping text;
- marker navigation;
- overview navigation and dismissal;
- annotation editing;
- scroll-based repositioning;
- viewport fallback positioning;
- highlight clipping;
- marker updates after comment edits;
- long annotation overview scrolling;
- composer and overlay stacking behavior.

## Scope

This PR adds the client-side response annotation workflow and injects the resulting annotations into the next composer submission.

It does not introduce a new daemon RPC or change the existing wire protocol.

## Preview
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ tsconfig.tsbuildinfo
 
 # Testing
 coverage/
+**/.vitest-attachments/
 
 # Playwright
 test-results/
--- a/docs/floating-panels.md
+++ b/docs/floating-panels.md
@@ -78,6 +78,14 @@ ordinary portals regardless of `z-index`, which would hide app toasts and
 tooltips behind the menu. The shared overlay scale keeps menus below toasts and
 lets tooltip portals paint above both.
 
+Fixed decorations that belong to document content, such as selection highlights
+and annotation markers, render into `content-adornment-root`. That root is a
+separate surface plane below `overlay-root`, so descendant `z-index` values can
+order decorations among themselves without covering menus, hover cards, or modals.
+Clip highlights to their owning chat viewport; this viewport ends where the
+composer begins. Interactive markers additionally suppress themselves when their
+hit area intersects the composer.
+
 The shared overlay scale is relative for interactive surfaces: a base floating
 panel is below a base modal, while a floating panel rendered from inside a modal
 inherits that modal's layer and paints above it. Wrap portal content in
--- a/packages/app/src/agent-stream/view.tsx
+++ b/packages/app/src/agent-stream/view.tsx
@@ -102,6 +102,11 @@ import { isWeb } from "@/constants/platform";
 import type { Theme } from "@/styles/theme";
 import { recordRenderProfileReasons } from "@/utils/render-profiler";
 import { useRetainedPanelActive } from "@/components/retained-panel";
+import type {
+  AssistantSelectionAnnotation,
+  SelectedTextAnnotationEdit,
+} from "@/assistant-selection-copy/types";
+import type { SelectedTextComposerAttachment } from "@/attachments/types";
 
 function renderLiveAuxiliaryNode(input: {
   pendingPermissions: ReactNode;
@@ -231,6 +236,7 @@ function renderLiveHeadStreamItem(input: {
 
 export interface AgentStreamViewHandle {
   scrollToBottom(reason?: BottomAnchorLocalRequest["reason"]): void;
+  scrollToMessage(itemId: string): void;
   prepareForViewportChange(): void;
 }
 
@@ -247,6 +253,13 @@ export interface AgentStreamViewProps {
   isAuthoritativeHistoryReady?: boolean;
   toast?: ToastApi | null;
   onOpenWorkspaceFile?: (request: WorkspaceFileOpenRequest) => void;
+  isPaneFocused?: boolean;
+  onCommentSelection?: (annotation: AssistantSelectionAnnotation) => void;
+  selectedTextAnnotations?: readonly SelectedTextComposerAttachment[];
+  onOpenSelectedText?: (annotation: SelectedTextComposerAttachment) => void;
+  selectedTextAnnotationToEdit?: SelectedTextComposerAttachment | null;
+  onEditSelectedText?: (input: SelectedTextAnnotationEdit) => void;
+  onDismissSelectedTextEdit?: () => void;
   readOnly?: boolean;
   historyPagination?: {
     hasOlder: boolean;
@@ -295,6 +308,13 @@ const AgentStreamViewComponent = forwardRef<AgentStreamViewHandle, AgentStreamVi
       isAuthoritativeHistoryReady = true,
       toast,
       onOpenWorkspaceFile,
+      isPaneFocused = true,
+      onCommentSelection,
+      selectedTextAnnotations,
+      onOpenSelectedText,
+      selectedTextAnnotationToEdit,
+      onEditSelectedText,
+      onDismissSelectedTextEdit,
       readOnly = false,
       historyPagination,
     },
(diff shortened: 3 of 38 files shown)


dlhck:
This would be a great feature! Don't know if there is any way to speed up the process to get things merged and released in this community.

crpboy:
> This would be a great feature! Don't know if there is any way to speed up the process to get things merged and released in this community.

I would be working on this feature this weekend. There remains some merging conflicts from master branch, and the P1/P2 issues mentioned above will be also considered. I'll fix all the problems a few days later. I'm not the core dev, so waiting for core dev's reply as well.

boudra:

PR #3427, opened 2026-08-15
Add opt-in terminal word shaping for contextual (calt) fonts like Fast-Font

Ni7e:
## Problem

Fonts whose behavior lives in OpenType contextual features (`calt`) — e.g. [Fast-Font](https://github.com/Born2Root/Fast-Font), which bolds the leading portion of every word to guide reading — render as plain text in the terminal, while working fine everywhere else in the app.

The terminal renders each cell in isolation, so the platform shaper never sees a whole word and contextual substitution never fires. xterm.js already has the mechanism for this (`registerCharacterJoiner`: joined ranges are rasterized as one run through the browser's shaper), but nothing registers word-level joins, so `calt` fonts get per-letter shaping and render plain.

I want to read terminal output with a contextual font; today that only works in editors and browsers, not in Paseo.

## Change

One opt-in setting: **Settings → Appearance → Fonts → "Terminal word shaping"** (off by default).

- New `terminal-word-shaping.ts`: a character joiner that returns `[letter|digit|_]` runs of 2+ characters, so the renderer shapes whole words and the font's own `calt` rules apply. The joiner deliberately does not parse font files — contextual fonts regularly use GSUB structures that the JS font parsers in this tree can't read (Fast-Font's `calt` is extension-wrapped format-3 chains; both `font-ligatures` and `opentype.js` fail on it), while the browser's shaper handles them fine as long as runs are joined.
- `TerminalEmulatorRuntime` gains `wordShapingEnabled` at mount and `setWordShapingEnabled()` for live toggling; wired through the web terminal component and pane.
- Setting persisted with the other app settings, defaults to false. UI row sits next to the code font input and follows the same visibility rules.
- Strings translated for all locales.

Off by default because joining words costs atlas texture space for zero visual benefit on fonts without contextual features.

## Not in scope

- Mobile native-grid renderer (draws per-cell without xterm; unchanged behavior).
- Auto-detecting `calt` fonts: would require reading font binaries (Local Font Access permission) or shipping a font parser that handles these GSUB structures. Opt-in is simpler and permission-free.

## Tests

- `terminal-word-shaping.test.ts`: word/identifier/unicode ranges, punctuation and single-char exclusion, empty input (6 cases).
- `use-settings/storage.test.ts`: default false + persisted round-trip.
- Suite results: `vitest run src/terminal src/hooks` → **469 passed, 18 skipped**; `oxlint` → 0 warnings; `oxfmt --check` clean; full workspace `tsgo --noEmit` clean.

## QA evidence

Verified on the dev web app (`npm run dev:app`) with Fast_Mono and Fast_Sans installed, driving the real settings UI, WebGL renderer:

| State | Measured word ink ratio (left/right half) | Words with bold-leading |
|---|---|---|
| Toggle ON, Fast_Mono | 1.234 | 16/25 |
| Toggle OFF, Fast_Mono | 0.969 | 3/32 |
| Toggle ON, Fast_Sans | 1.309 | 20/31 |

(Ink ratio > 1 = leading letters bolder, i.e. the font's designed behavior.)

Screenshots (evidence branch [`evidence-terminal-word-shaping`](https://github.com/Ni7e/paseo/tree/evidence-terminal-word-shaping)):

| Setting off | Setting on |
|---|---|
| off | on |

Known-inherent xterm behaviors, unchanged by this PR: a word containing the cursor drops its join while the cursor is inside it; words with mid-word color changes split at the color boundary (same as VS Code).

Side observation (not addressed here): `LigaturesAddon` is loaded before `terminal.open()`, so its `activate()` throws and is silently caught — the fallback operator ligatures it would provide (`->`, `=>`) never render today either. Happy to split that into a separate fix if useful.

## Platforms

- Tested: macOS (dev web app, Chrome + SwiftShader WebGL)
- Not tested: Windows, Linux, mobile (mobile renderer intentionally unchanged)

### DIFF
--- a/packages/app/src/components/terminal-emulator-contract.ts
+++ b/packages/app/src/components/terminal-emulator-contract.ts
@@ -33,6 +33,7 @@ export interface TerminalEmulatorProps {
   scrollbackLines: number;
   fontFamily?: string;
   fontSize?: number;
+  wordShapingEnabled?: boolean;
   keyboardInset?: number;
   isKeyboardVisible?: boolean;
   swipeGesturesEnabled?: boolean;
--- a/packages/app/src/components/terminal-emulator.tsx
+++ b/packages/app/src/components/terminal-emulator.tsx
@@ -105,6 +105,7 @@ interface TerminalEmulatorProps {
   scrollbackLines: number;
   fontFamily?: string;
   fontSize?: number;
+  wordShapingEnabled?: boolean;
   keyboardInset?: number;
   isKeyboardVisible?: boolean;
   swipeGesturesEnabled?: boolean;
@@ -168,6 +169,7 @@ export default function TerminalEmulator({
   scrollbackLines,
   fontFamily,
   fontSize,
+  wordShapingEnabled = false,
   swipeGesturesEnabled = false,
   onSwipeLeft,
   onSwipeRight,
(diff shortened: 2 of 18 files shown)


boudra:

PR #3440, opened 2026-08-16
Reject low-order peer keys in X25519 key agreement

amersheeny:
### Linked issue

N/A

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Remote access is the reason people run Paseo — the daemon sits on your machine and you reach it from your phone. Everything protecting that path rests on the E2EE handshake, and `deriveSharedKey` validated the peer's public key by length alone.

A peer that supplies a low-order Curve25519 point forces the X25519 output to a constant, so both sides derive a shared key **the peer can compute without holding any secret**. Against `main` today that key is:

```
351f86faa3b988468a850122b65b0acece9c4826806aeee63de9c0da2bd7f91e
```

The same value for every daemon secret. The X25519 output itself is all zeros; the hex above is what `box.before` produces after hashing it through HSalsa20, so the derived key is not itself all zeros and the check has to be made on the raw output before `box.before` rather than on its result. RFC 7748 §6.1 lets both sides check the raw all-zero output and abort; before this change we did not.

For a Paseo user this removes the case where a peer holding no secret at all can force a session key the daemon will accept. It does not authenticate peer keys, so anyone able to substitute a keypair of their own is still out of scope; see non-goals.

### Goals

- Reject low-order Curve25519 points and any all-zero X25519 result in `deriveSharedKey`
- Validate and derive from the same bytes, so the check cannot be stepped around
- Keep honest handshakes byte-identical — no behaviour change for real keys
- Make a rejected key close the channel as a key mismatch (1008), not a decrypt failure (1011)
- Tests that fail against `main` and pass here

### Non-goals

- **No client authentication.** The relay path still accepts any client key that isn't low-order. That's a much larger conversation and belongs in its own PR, not smuggled into this one.
- No change to the relay protocol or the wire format beyond the close code for a rejected peer key, which becomes 1008 (key mismatch) instead of 1011 (decrypt failure) — both are codes the relay already uses, and only this error's routing changes. No public API change beyond exporting `InvalidPeerKeyError` so callers can catch it, and `CANONICAL_LOW_ORDER_POINTS` so the tests assert against the same table the check uses.
- No performance work. The added cost is two 32-byte copies, one comparison against the low-order table, and one additional scalar multiplication per handshake — `box.before` hashes its X25519 output through HSalsa20, so the all-zero check needs its own `scalarMult`.

### QA

Against this branch, in a worktree off `main`:

```
$ npx vitest run packages/relay
 Test Files  6 passed | 2 skipped (8)
      Tests  67 passed | 4 skipped (71)

$ npm run typecheck
0 errors

$ npm run lint -- packages/relay/src/crypto.ts packages/relay/src/crypto.low-order.test.ts
```

**Reproducing the bug on `main` first.** Built `main`'s relay and derived a key from the all-zero point:

```
$ node -e 'import("./packages/relay/dist/crypto.js").then(({deriveSharedKey,generateKeyPair})=>
    console.log(Buffer.from(deriveSharedKey(generateKeyPair().secretKey,new Uint8Array(32))).toString("hex")))'
351f86faa3b988468a850122b65b0acece9c4826806aeee63de9c0da2bd7f91e
```

Same output for any secret. With this branch it throws `InvalidPeerKeyError`.

**Confirming the tests are bound to the fix.** Reverted `crypto.ts` to `main`'s version and re-ran the two crypto suites, so a green run cannot be mistaken for coverage:

```
this branch:           Test Files  2 passed (2)      Tests  41 passed (41)
with main's crypto.ts: Test Files  1 failed | 1 passed (2)   Tests  8 passed (8)
```

The failing file is `crypto.low-order.test.ts`, and it fails at load: `main`'s `crypto.ts` exports neither `CANONICAL_LOW_ORDER_POINTS` nor `InvalidPeerKeyError`, so the import throws before any test runs. That shows the suite cannot pass against unpatched code, but it is a weaker demonstration than a test failing on behaviour, and I would rather say so than let the numbers imply otherwise. The behavioural evidence is the `node` reproduction above, which needs none of this branch's exports.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/relay/src/crypto.low-order.test.ts
+++ b/packages/relay/src/crypto.low-order.test.ts
@@ -1,7 +1,12 @@
 import { describe, it, expect } from "vitest";
 import nacl from "tweetnacl";
 
-import { InvalidPeerKeyError, deriveSharedKey, generateKeyPair } from "./index.js";
+import {
+  CANONICAL_LOW_ORDER_POINTS,
+  InvalidPeerKeyError,
+  deriveSharedKey,
+  generateKeyPair,
+} from "./index.js";
 
 /**
  * Negative tests for peer-key validation.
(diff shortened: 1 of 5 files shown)


boudra:
Closing in favor of the implementation merged in [#3669](https://github.com/getpaseo/paseo/pull/3669).

Issue #3023, opened 2026-08-08
[Tracking] Claude OAuth failures when launched from Paseo


boudra:
## Summary

Claude Code has a recent OAuth regression affecting programmatic and headless launches. In Paseo, a Claude agent may fail with:

> Failed to authenticate: OAuth session expired and could not be refreshed

The same credentials can continue working in an interactive Claude terminal while agents launched by Paseo fail.

Reports cover Paseo 0.2.5 and 0.3.0 on Windows, macOS, and Linux. Matching upstream Claude Code reports began clustering in July 2026.

## Workaround

1. Open Claude in a terminal.
2. Run `/logout`, then `/login`.
3. Complete the login flow.
4. Reload the affected agent, or start a new one, so it picks up the refreshed credentials.

Users have confirmed this restores affected agents.

## Status

This issue tracks the user-visible Paseo problem and workaround while the upstream Claude Code regression remains unresolved.

Reports suggesting a particular refresh race or credential-store mechanism are useful evidence, but the cause has not been established for every affected user.

## Paseo reports

- Windows and macOS: https://github.com/getpaseo/paseo/issues/2678
- Linux multi-agent report: https://github.com/getpaseo/paseo/issues/3003
- Discord announcement and workaround: https://discord.com/channels/1481169421832814616/1504328181170311188/1535621277396504606

## Upstream reports

- Headless launch fails while interactive Claude works: https://github.com/anthropics/claude-code/issues/79685
- Non-interactive launch returns 401 despite valid credentials: https://github.com/anthropics/claude-code/issues/80075
- `claude -p` reports the same OAuth refresh failure while interactive mode works: https://github.com/anthropics/claude-code/issues/81937

Please add new Paseo-specific reproductions here with the Paseo version, Claude Code version, operating system, and whether the workaround restored access.

BenDaye:
Additional Paseo-specific Linux reproduction, linked to #3442.

Environment:

- Paseo `0.4.0` from npm
- Claude Code `2.1.238` from npm
- Linux, Claude.ai subscription OAuth

Observed twice: a healthy 509-byte `~/.claude/.credentials.json` became a 368-byte file containing only `accessToken`, `refreshToken`, `subscriptionType`, and `rateLimitTier`; `expiresAt`, `refreshTokenExpiresAt`, and `scopes` disappeared, followed by a forced login.

[...]

Issue #2678, opened 2026-07-30
[Bug] Claude OAuth refresh failure prevents new sessions on Windows


katanumahotori:
## Environment

- OS: Windows 11 x64
- Paseo: 0.2.5
- Claude Code: native 2.1.220
- Subscription: Claude Max via claude.ai OAuth

## What happened

A new Claude Code session launched from Paseo failed with:

`Failed to authenticate: OAuth session expired and could not be refreshed`

The first failure was recorded at 2026-07-30 22:51:30 UTC. Paseo then retried the same prompt on the same connection and created another Claude session, which failed with the same error at 22:52:11 UTC.

After completing the browser OAuth login again, `claude auth status` reported a valid Max login and a diagnostic Claude session launched from Paseo returned `AUTH_OK`.

## Evidence and suspected cause

This may be a credential refresh race between Claude Code and Paseo's Claude usage-fetcher. Both can access `~/.claude/.credentials.json`, and the usage-fetcher appears to refresh and write credentials after a 401/403 response. I cannot prove which process invalidated the token in this incident, but the failure pattern is consistent with concurrent refresh/write activity.

The same prompt was retried automatically after the authentication failure, which created a second failed session instead of presenting a clear re-login path.

## Expected behavior

- A transient OAuth refresh failure should not invalidate the user's working session.
- Credential refreshes should be coordinated with Claude Code's credential handling.
- Paseo should avoid automatically retrying the identical prompt after `authentication_failed`, or should clearly request re-authentication first.

## Workaround

Separating Paseo's Claude agent configuration with `CLAUDE_CONFIG_DIR` allowed the agent to authenticate independently. This is only a local workaround, not a proposed product fix.

No OAuth tokens, account identifiers, or prompt contents are included here.

timothysu:
I can reproduce this on macOS on 0.2.5. Downgrading to 0.2.4 fixes the issue.

milos-agathon:
On Windows 11, force a clean re-login: open your terminal and run the manual authentication command directly `claude login`. Then return to T3 code and open a new chat (no need to restart the desktop app). That solved the issue for me

ragokan:
I can reproduce this issue on macOS on 0.2.5 as well.

boudra:
Consolidating this into the pinned tracking issue: https://github.com/getpaseo/paseo/issues/3023

[...]

PR #3442, opened 2026-08-16
fix(server): stop quota fetcher from writing provider credentials

danberindei:
## Summary

- The Claude and Codex quota fetchers redeemed the provider CLI's refresh token and rewrote the on-disk credential file after a 401/403, racing the CLI's own refresh and silently dropping fields the fetcher's Zod schema didn't model (Claude lost `expiresAt`/`scopes`, Codex lost `id_token`/`OPENAI_API_KEY`/`last_refresh`), corrupting the file so the CLI could no longer authenticate.
- Makes both fetchers read-only on credentials: on `NEEDS_AUTH` they now return `unavailable` and leave all refresh to the provider's own CLI.
- Documents the rule in `docs/providers.md` so it isn't reintroduced.

## Context

Relates to #3023. One reproduction folded into that tracking issue (#2678) named this exact mechanism as the suspected cause. This does not address the other mechanisms discussed there (multiple concurrent CLI processes racing each other, or the separate upstream headless-auth regression) — only the case where Paseo's own quota fetcher clobbers the credential file.

## Test plan

- [x] `npx vitest run src/services/quota-fetcher/service.test.ts --bail=1` — 70 passed
- [x] `npm run typecheck`, `npm run lint`, `npm run format:check` (via pre-commit hook) — clean

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -147,6 +147,10 @@ Kimi Code usage follows the CLI-managed credential file at `KIMI_CODE_HOME` or `
 
 Cursor usage reads the desktop `state.vscdb` token first, then `cursor-agent`'s `~/.config/cursor/auth.json`. Headless hosts only have the CLI file.
 
+### Usage fetchers are read-only on credentials
+
+A fetcher reads the provider's credential file and never writes it. On a 401 or 403 it returns `unavailable` and leaves refresh to the provider's own CLI: redeeming a refresh token in the fetcher invalidates the CLI's copy (refresh tokens are single-use), and rewriting the file through the fetcher's Zod schema drops any field the schema does not model, corrupting the file for the CLI.
+
 ---
 
 ## ACP Provider Checklist
--- a/packages/server/src/services/quota-fetcher/providers/claude.ts
+++ b/packages/server/src/services/quota-fetcher/providers/claude.ts
@@ -22,7 +22,6 @@ import {
 const execFileAsync = promisify(execFile);
 const CLAUDE_KEYCHAIN_TIMEOUT_MS = 2_000;
 const CLAUDE_OAUTH_BETA = "oauth-2025-04-20";
-const CLAUDE_CLIENT_ID = "9d1c250a-e61b-44d9-88ed-5944d1962f5e";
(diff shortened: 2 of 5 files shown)


BenDaye:
Confirmed independently on Linux with Paseo `0.4.0` from npm and Claude Code `2.1.238` from npm.

The credential-clobbering mechanism in this PR matches the incident byte-for-byte:

- A healthy `~/.claude/.credentials.json` was 509 bytes and included `expiresAt`, `refreshTokenExpiresAt`, and `scopes`.
- On two separate failures it became 368 bytes, retaining only `accessToken`, `refreshToken`, `subscriptionType`, and `rateLimitTier`.
- Passing the healthy credential shape through the installed quota fetcher's Zod parse + `JSON.stringify(..., null, 2)` write path produces exactly 368 bytes.
- The installed `ClaudeQuotaProvider` calls the refresh endpoint after usage returns 401/403 and writes the schema-stripped object back to the CLI-owned file without Claude Code's refresh lock.

I applied the behavioral equivalent of this PR locally (quota fetchers return unavailable on auth rejection and never redeem/write provider credentials) and verified it with a controlled regression harness:

- Before the patch: both 401 and 403 cases rewrote the credential fixture.
- After the patch: 2/2 cases passed; one usage request, no refresh request, credential bytes unchanged.
- After restarting the Paseo daemon, a real `DaemonClient.listProviderUsage()` call returned Claude usage as available, while the real credential file's hash, inode, mode, size, and mtime remained unchanged.
- The existing Paseo agent resumed normally after the daemon restart.

This establishes the quota-fetcher writer as a reproduced Paseo-specific cause in this environment, separately from any upstream multi-process Claude Code refresh race.

This PR is currently conflicting with `main`. If the maintainer and original author prefer, I can prepare a clean current-`main` replacement that preserves this PR's design and credits the original work.

danberindei:
@BenDaye I rebased the branch and I applied the same fix to the Kimi quota fetcher.

@boudra let me know what you think of the changes. I can adapt the PR or obviously you can do your own implementation since it's mostly deleting some code.

PR #3449, opened 2026-08-16
feat(omp): decouple native agent tools from MCP injection, add per-provider opt-in

thomasvan:
Two related changes to how an omp seat receives Paseo's native host tools. Both
defaults preserve current behaviour.

## 1. Native tools follow their own switch

The native host-tool channel is not MCP, but it is gated on
`daemon.mcp.injectIntoAgents` (and now `mcp.enabled` too):

```ts
agentManager.setPaseoToolsEnabled(mcpEnabled && config.mcpInjectIntoAgents !== false);
```

A deployment that turns MCP injection off for reasons unrelated to native tools
loses `create_agent` and the rest of the native surface as a side effect. Our
case: we generate a **caller-scoped** MCP server per agent at launch, so a
daemon-wide injected server would be scoped to the wrong caller — we turn
injection off deliberately, and silently lost native tools with it.

This adds `daemon.mcp.nativeAgentTools` (default `true`) and points the
native-tools switch at it, with live reload wired for the new field. MCP flags
govern MCP; this one governs native tools.

## 2. Per-provider opt-in

`supportsNativePaseoTools` was the constant `true`, so every omp provider
advertises the channel and the only control was daemon-wide. Deployments running
several omp providers in different roles need it per provider — a worker-style
provider should hold no orchestration tools while a coordinator does.

Adds `paseoTools` to `OmpProviderParamsSchema` (default `true`) and derives the
capability from it, so `capabilities` moves from a field initializer to the
constructor where the params are resolved.

## Compatibility

A daemon that sets neither field is unaffected. The one behaviour change is the
bug being fixed: where MCP injection was off, native tools were off too, and now
they are not.

## Verification

Built and exercised end to end rather than reasoned from the diff — with
`paseoTools` true on one provider and false on another:

- the first called `create_agent` and the child agent ran to completion
- the second reported `create_agent` absent from its session

Existing omp suite passes (121 tests) plus 4 new in
`native-tools-optin.test.ts` covering the default, the opt-out, independence
from the other provider params, and that an unknown param is still rejected by
the strict schema.

Happy to adjust naming or split this into two PRs if you'd prefer the decoupling
and the per-provider flag reviewed separately.

### DIFF
--- a/packages/protocol/src/messages.ts
+++ b/packages/protocol/src/messages.ts
@@ -196,6 +196,7 @@ export const MutableDaemonConfigSchema = z
       .object({
         enabled: z.boolean().optional(),
         injectIntoAgents: z.boolean(),
+        nativeAgentTools: z.boolean().optional(),
       })
       .passthrough(),
     hostnames: z.union([z.literal(true), z.array(z.string())]).optional(),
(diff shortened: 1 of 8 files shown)


boudra:
Closing as superseded by #4277. The merged provider-level `paseoTools` policy owns this workflow across native and MCP delivery, including OMP, so the OMP-specific parameter is no longer needed. We also don’t want separate global switches for the same semantic tool catalog based on delivery transport.

Thanks @thomasvan for surfacing the role-control workflow and for the implementation work.

PR #3452, opened 2026-08-16
feat(pi): add native active-turn steering

trim21:
### Linked issue

None.

### Type of change

- [x] New feature
- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor
- [x] Docs

### Reasoning

PR #3394 added active-turn steering for Codex and Claude. Pi supports native mid-turn steering via its `prompt` RPC `streamingBehavior: "steer"` option: the message is queued and delivered after the current tool-call batch completes, keeping the same run alive. This PR wires that capability into `AgentSession.steerActiveTurn` so prompts sent with `activeTurnBehavior: "steer"` steer the active Pi turn instead of aborting it.

### Changes

- `PiRuntimeSession.prompt` accepts an optional `streamingBehavior` (`"steer"` / `"followUp"`) and forwards it on the wire.
- `PiRpcAgentSession.steerActiveTurn` implements the manager contract:
  - rejects when the expected turn is no longer active or the prompt is a slash command (fall back to interrupt-and-replace)
  - queues the prompt with `streamingBehavior: "steer"`; Pi delivers it after the current tool-call batch without aborting the run
  - classifies Pi rejections: definitive rejections (older binaries that drop `streamingBehavior`, compaction in progress) return `unavailable` so the daemon can fall back; ambiguous errors propagate without an interrupt or retry
  - steered `clientMessageId`s are consumed FIFO when Pi echoes the submitted user entry, so multiple steers inside one turn each land under the right `clientMessageId` on the timeline
- Adds `pi.steer.*` logs for diagnosability.
- Real-provider e2e coverage (`send-during-tool-call-pi.real.e2e.test.ts`) mirrors the Codex/Claude steering tests: one active Pi turn with a foreground tool call, steered message, no new turn start, no cancellation, canonical timeline row on the original turn. The test points `PI_CODING_AGENT_DIR` at a temp dir so it can run in sandboxed/read-only-home environments.
- Docs: `docs/providers.md` steering paragraph now covers Pi.

### Testing

- 101 Pi provider tests + agent-prompt/agent-manager suites pass.
- `npm run typecheck`, `npm run lint`, `npm run format` pass.
- Verified against a real Pi binary: a 5s sleep tool running mid-turn survives a steer prompt; the message is delivered after the tool-call batch and the run continues to a normal `agent_end` (no aborted terminal).
- Real-provider e2e is gated on `canRunRealProvider("pi")` like the other provider e2e suites.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes


trim21:
@boudra would you mind take a look?

trim21:
@boudra would you mind take a look?

PR #3455, opened 2026-08-16
feat(agent): opt-in notifyMode "each" keeps the finish watcher armed

thomasvan:
Resubmission of the third patch from #2879, in the opt-in shape. That PR changed the default for every existing caller, which is why you closed it; the other two landed as #3192. This one changes nothing for anyone who does not ask for it.

## The problem

`setupFinishNotification` unsubscribes after the first finish. That fits a caller which delegates one unit of work and waits for it. It does not fit a caller which drives one long-lived child across many turns — it hears about turn one and then sleeps through everything after, because the watcher is gone.

Concretely: an orchestrating agent sends a child a second `send_agent_prompt` and never learns that it completed.

## The change

`notifyMode` on the **agent-scoped** `create_agent` and `send_agent_prompt` schemas:

- `"once"` — the default, and exactly today's behaviour: one notification, then disarm.
- `"each"` — re-arm after every finish; disarm when the child closes or errors, or when the caller is archived.

The watcher change is one argument wide — `"each"` passes `{ terminal: false }` on the `"finished"` path and resets the run gate, so the child's next running → idle cycle notifies again. It builds directly on the `terminal` option added in #3177.

## Notes on shape

**The default path is untouched.** A test asserts the watcher still stops after the first finish when no mode is given.

**The archived-caller disarm is scoped to `"each"`.** Under `"once"` the watcher has already stopped before that code runs, so the existing path is unchanged. Under `"each"` it would otherwise hold the subscription for the child's whole life with nobody left to hear it.

**The field is `.optional()` with no `.default()`.** A default injects the key into every parsed result, and `legacyAgentToAgentInputSchema` is `.strict()`, so calls that never named the field would start failing. The read site coalesces to `"once"` instead. (Learned the hard way on a sibling patch — `.default()` there turned several `mcp-parity.e2e.test.ts` cases red.)

## Verification

- Four new tests in `agent-prompt.test.ts`, each **mutation-checked**: making `"each"` terminal, defaulting to `"each"`, and dropping the archived-caller disarm each turn the suite red.
- `packages/server/src/server/agent` unit suites: **1962 passed**, 21 skipped.
- `tsc -p tsconfig.server.json --noEmit`: clean. `biome check` on the touched files: clean.
- `mcp-parity.e2e.test.ts`: 5 failed / 26 passed **both with and without this patch** — identical to the `upstream/main` baseline on this machine, so those failures are environmental, and the parity cases that guard schema injection pass.

Happy to rename the modes or move the field if you would rather it lived somewhere else.

### DIFF
--- a/packages/server/src/server/agent/agent-prompt.test.ts
+++ b/packages/server/src/server/agent/agent-prompt.test.ts
@@ -13,6 +13,7 @@ import {
   isSystemInjectedEnvelope,
   setupFinishNotification,
   waitForAgentRunStartWithTimeout,
+  type FinishNotifyMode,
 } from "./agent-prompt.js";
 import type { AgentManagerEvent, ManagedAgent } from "./agent-manager.js";
 import type {
@@ -51,6 +52,8 @@ interface FinishNotificationScenarioOptions {
   childParentAgentId?: string | null;
   requireParentOwnership?: boolean;
   parentPromptError?: Error;
+  callerArchived?: boolean;
+  notifyMode?: FinishNotifyMode;
   logger?: Logger;
 }
 
@@ -66,6 +69,7 @@ interface FinishNotificationScenario {
   parentPrompts(): string[];
   steerAttemptCount(): number;
   wasParentPrompted(): boolean;
+  subscriberCount(): number;
 }
 
 function createFinishNotificationScenario(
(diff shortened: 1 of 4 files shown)


thomasvan:
Status update: I still carry this patch on `slp/patches` (fork tip `04563655b`), re-verified and adapted through the 2026-09-07 sync to upstream `c424f8292` (0.7.2). I have now rebased this PR head (`7141f88ae`) onto current `main` (`8c89e72a0`): the code applied cleanly, and the four tests were re-seamed to the current upstream harness (real `AgentManager` with `streamAgent` recording), preserving the opt-in `notifyMode` shape with `"once"` as the byte-identical default. `agent-prompt.test.ts` passes 18/18 on the updated head.

boudra:

PR #3458, opened 2026-08-16
fix(app): restore terminal surface after tab hide so CJK cells redraw

### LINKED ISSUE #3434: bug: Hangul (CJK wide) characters gain an extra trailing gap after terminal pane focus/visibility change
### What's broken

Korean (Hangul) text in a terminal pane renders correctly on first paint. After I switch
to another tab and come back, the pane repaints and every Hangul syllable gains a blank
column after it — each character is allocated 2 cells but drawn at roughly 1 cell of glyph
width. It reads as exaggerated letter-spacing and wraps lines far earlier than it should.

**Expected:** Hangul keeps the spacing it had on the initial paint (glyph filling both
cells), the way the same output looks in a standalone terminal emulator.

ASCII/Latin text in the same pane is unaffected and the buffer contents are intact, so this
is rendering-only — but long Korean agent output becomes hard to read.

**Before (initial paint — correct):**

Image

**After (returning from another tab — broken):**

Image

Same buffer content in both; only the repaint differs.

**Note on likely cause:** it's the same font in the same pane, correct on the initial
paint — so this looks like a cell-width / glyph-width mismatch in the repaint path rather
than missing CJK font coverage. The trigger is in the same area as the 0.3.0 fix
_"Terminal panes no longer remain at 80x24 after focus or visibility changes"_; whatever
remeasure runs on visibility change may be losing the wide-character width, or measuring
cell width before fonts have settled.

**Related:** [manaflow-ai/cmux#338](https://github.com/manaflow-ai/cmux/issues/338) — CJK
text broken in cmux but correct in standalone Ghostty. Opposite direction (clipping rather
than gapping) but the same class of cell-width mismatch in a web terminal renderer, and
notably font and cell-height adjustments did **not** fix it there.

Image

### Steps to reproduce

1. Open the desktop app and start a Claude Code agent in a terminal pane.
2. Have the agent emit a block of Korean text (e.g. ask for a summary in Korean, or show a
   diff of a Korean-language Markdown file).
3. Confirm the Hangul spacing renders correctly.
4. Switch to a different tab, then switch back.
5. The pane repaints and Hangul now has an extra blank column per character.

Reproduces every time. [confirm — 매번 재현되는지, 아니면 가끔인지]

### Where did this happen

Desktop (Electron)

### Paseo version

0.4.0

### OS version

macOS [confirm: version, e.g. 15.5] — Apple Silicon (Mac Studio)

### Agent provider

Claude Code

### Provider configuration

Claude Code v2.1.233 with Claude subscription

### Logs

```text

```

### Screenshots or video

Image
Image

jasonhnd:
### Linked issue

Closes #3434

### Type of change

- [x] Bug fix
- [ ] Docs

### What does this PR do

Hangul (and other CJK wide characters) paint correctly on first show, then gain an extra trailing cell after switching away from the terminal pane and back.

Retained workspace/split panels hide with `display: none`. That does **not** change `document.visibilityState`, so the existing visibilitychange/window-focus refit never ran. Coming back, `FitAddon.fit()` often kept the same cols/rows and skipped `refresh()`, while Chromium may also have dropped the WebGL context. Buffer width stays 2 cells; the glyph is redrawn at ~1 cell.

This change:

- treats a host that grew from 0×0 as a surface restore (force refresh + reload WebGL if it was lost)
- exposes `restoreSurface()` and calls it when the terminal pane becomes active again
- keeps the document-visibility path, now going through the same restore

Not the same as #2193 / #2319 (capture text inserting spaces) or #2973 (native grid column width).

### How did you verify it

- `npx vitest run packages/app/src/terminal/runtime/terminal-emulator-runtime.test.ts --bail=1` — 20 passed, including `restoreSurface` and the existing visibility-restore cases
- lint / format / typecheck on touched files — pass
- Did **not** re-run the desktop app click-through (Hangul → switch tab → switch back) in this pass

### Checklist

- [x] `npm run lint` passes
- [x] `npm run format` ran
- [ ] UI changes include screenshots or video for every affected platform (desktop terminal; no new screenshot this pass)

### DIFF
--- a/packages/app/src/components/terminal-emulator-contract.ts
+++ b/packages/app/src/components/terminal-emulator-contract.ts
@@ -19,6 +19,7 @@ export interface TerminalEmulatorHandle {
   copySelection: (clipboard: TerminalClipboardWriter) => Promise<string>;
   clear: () => void;
   claimSize: () => void;
+  restoreSurface: () => void;
   showKeyboard: () => void;
   blur: () => void;
 }
--- a/packages/app/src/components/terminal-emulator-native-grid.native.tsx
+++ b/packages/app/src/components/terminal-emulator-native-grid.native.tsx
@@ -567,6 +567,7 @@ function NativeTerminalEmulator({
         schedulePaint();
       },
       claimSize: claimActiveTerminalSize,
+      restoreSurface: claimActiveTerminalSize,
       showKeyboard: () => {
         inputRef.current?.showKeyboard();
         claimActiveTerminalSizeForAction("showKeyboard");
--- a/packages/app/src/components/terminal-emulator-webview.native.tsx
+++ b/packages/app/src/components/terminal-emulator-webview.native.tsx
@@ -345,6 +345,9 @@ export default function WebViewTerminalEmulator({
       claimSize: () => {
         sendToWebView({ type: "resize", streamKey, forceClaim: true });
       },
+      restoreSurface: () => {
+        sendToWebView({ type: "resize", streamKey, forceClaim: false, shouldClaim: false });
+      },
       showKeyboard: () => {
         sendToWebView({ type: "focus", streamKey, forceRefocus: true });
         webViewRef.current?.requestFocus();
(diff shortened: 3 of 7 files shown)


boudra:
Closing in favor of the combined implementation in open PR [#4101](https://github.com/getpaseo/paseo/pull/4101), which includes this surface-restoration work.

PR #3473, opened 2026-08-17
feat(server): allow create_agent system prompts

nghiant0704:
## Summary

- accept an optional `systemPrompt` in the MCP `create_agent` tool
- merge the requested prompt with provider options inherited from a same-provider parent
- verify child instructions are passed without inheriting the parent system prompt

## Testing

- `npx vitest run packages/server/src/server/agent/mcp-server.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`


nghiant0704:
Closing because this change is intended for the private downstream repository, not upstream.

PR #3474, opened 2026-08-17
fix(server): don't treat a forbidden usage response as stale auth

Ctrl-Creeper:
## Summary
`callClaudeApi` mapped both **401** and **403** to `NEEDS_AUTH`, so a forbidden response was handled as if the token were stale.

Two consequences:

1. **Credentials are rewritten on every poll.** With file-based credentials, a persistent 403 runs the refresh path on each usage fetch — exchanging the refresh token and rewriting `.credentials.json` roughly every five minutes to recover from something a refresh cannot fix. The token exchange rotates the refresh token, so this churns live credentials on a timer.
2. **The card goes blank with no reason.** Both give-up paths return `unavailable` with `error: null`, so the UI shows an empty card and nothing explains why.

This PR separates the two cases:

- **401** → stale authentication; refresh path unchanged.
- **403** → returned as an `error` status with a message, and **no refresh is attempted**.
- **401 → refresh → 403** → same message, rather than falling back to a blank `unavailable`.
- **401 → refresh → 401** → unchanged (generic `unavailable`).

Message:

> Usage request was refused (HTTP 403). Check the daemon's network access to api.anthropic.com.

A 403 can have several causes and the daemon cannot tell them apart from the response, so the message names the refusal and its status code rather than asserting a reason.

## Notes

- Scoped deliberately to the status distinction. Credential-directory handling (`CLAUDE_CONFIG_DIR`) is a separate concern tracked in #3589.
- Independent of #3595 / #3597, which fix *which* Keychain item is read. That work selects the credential; this changes how the outcome is reported. They compose: once a live item is selected, a machine that previously returned 401 can begin returning 403, which is the path this PR gives a message to.

## Test plan

`packages/server/src/services/quota-fetcher/providers/claude.test.ts`:

- 403 → `status: "error"` carrying the message, no windows
- 403 → no token refresh attempted
- 401 → refresh → 403 → same message (not blank `unavailable`)
- no credentials → plain `unavailable` with `error: null`

CI runs typecheck and tests; `node_modules` was not installed in the environment this was written in, so they were not run locally.

### DIFF
--- a/packages/server/src/services/quota-fetcher/providers/claude.test.ts
+++ b/packages/server/src/services/quota-fetcher/providers/claude.test.ts
@@ -1,0 +1,46 @@
+import pino from "pino";
+import { describe, expect, it } from "vitest";
+import { ClaudeQuotaProvider } from "./claude.js";
+
+const logger = pino({ level: "silent" });
+
+function forbiddenResponse(): Response {
+  return new Response(JSON.stringify({ error: { type: "forbidden" } }), { status: 403 });
+}
+
+describe("ClaudeQuotaProvider forbidden handling", () => {
+  it("surfaces a 403 as an error state instead of a silent blank card", async () => {
+    // Keychain-sourced credentials (filePath null); a 403 must not be treated as a stale
+    // token and must not blank out silently.
+    const provider = new ClaudeQuotaProvider({
+      logger,
+      platform: "darwin",
+      claudeHome: "/nonexistent-claude-home",
+      claudeKeychainReader: async () => ({
+        claudeAiOauth: { accessToken: "t", refreshToken: "r" },
+      }),
+      fetch: async () => forbiddenResponse(),
+    });
+
+    const usage = await provider.fetchUsage();
+
+    expect(usage.status).toBe("error");
+    expect(usage.error).toMatch(/organization/i);
+    expect(usage.windows).toHaveLength(0);
+  });
+
+  it("reports plain unavailable when no credentials exist", async () => {
+    const provider = new ClaudeQuotaProvider({
+      logger,
+      platform: "linux",
+      claudeHome: "/nonexistent-claude-home",
+      claudeKeychainReader: async () => null,
+      fetch: async () => new Response("{}", { status: 200 }),
+    });
+
+    const usage = await provider.fetchUsage();
+
+    expect(usage.status).toBe("unavailable");
+    expect(usage.error).toBeNull();
+  });
+});
--- a/packages/server/src/services/quota-fetcher/providers/claude.test.ts
+++ b/packages/server/src/services/quota-fetcher/providers/claude.test.ts
@@ -1,5 +1,8 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import pino from "pino";
-import { describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { ClaudeQuotaProvider } from "./claude.js";
 
 const logger = pino({ level: "silent" });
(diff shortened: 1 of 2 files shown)


t-benoit:
Heads-up on an overlap: I opened #3595 / #3597 for the same blank Claude usage card, with a different root cause — the daemon's Keychain lookup names no account, so it reads an older item than the one Claude Code writes.

One data point from that investigation may bear on the 403 here. On a `claude_team` seat, `GET https://api.anthropic.com/api/oauth/usage` returns **200** with real windows for me, using the token from the account Claude Code actually writes under. The token behind the account-less lookup returns **401** on the same endpoint. Evidence table is in #3597.

I can't tell whether your 403 is an entitlement difference between orgs or the same wrong-item problem showing a different status code, so I'm raising it as a question rather than a claim. These two compare the items without printing a secret:

```
security find-generic-password -a claude-code-user -s "Claude Code-credentials"
security find-generic-password -s "Claude Code-credentials"
```

A different `acct` on the second means the token behind your 403 isn't the one Claude Code is currently using.

The 401-vs-403 split looks worthwhile independently of that — #3597 deliberately stays out of the reporting code so the two don't collide.

Ctrl-Creeper:
Thanks for raising it as a question rather than a claim — I ran your diagnostic, and it rules out the wrong-item explanation **on my machine**. Attributes only, no secrets read:

| lookup | result |
| --- | --- |
| `-a <my-unix-username>` | found, `mdat` `20260820054713Z` |
| `-a claude-code-user` | `SecKeychainSearchCopyNext: The specified item could not be found` |
| no `-a` (what the daemon does today) | **same item** — same `acct`, byte-identical `mdat` |

There is only one `Claude Code-credentials` item here, so the account-less lookup already returns the item Claude Code writes — last written today, not expired. Your fix wouldn't change which token this machine reads, and naming the account wouldn't clear my 403.

So I think these are two distinct bugs that happen to share a symptom, and the status code is the tell:

- Yours: superseded/expired item → **401**, i.e. unauthenticated. Refreshing (or selecting the right item) fixes it.
- Mine: current, unexpired item → **403 `{"type":"forbidden","message":"Request not allowed"}`**, and the same 403 on `/api/oauth/profile` as on `/api/oauth/usage`. A token that authenticates fine for inference is simply not authorized for those endpoints.

That also means our 200-vs-403 difference on `claude_team` isn't the item — it looks like a genuine entitlement difference between orgs/seats, since my token's scopes carry no usage scope (`user:inference`, `user:profile`, `user:mcp_servers`, `user:file_upload`, `user:sessions:claude_code`).

Agreed the split stands on its own either way, and your non-goal boundary means these don't collide — your change picks the item, this one reports the outcome. One knock-on for yours: once the account lookup selects a *live* item, a previously-401 machine can start returning 403 instead, which is exactly the path this PR gives a message to.

t-benoit:
That settles it — thanks for actually running it. Two distinct bugs sharing a symptom, and the status code as the tell is a much cleaner framing than either PR had on its own.

Worth noting why my change is a no-op on your machine rather than a risk: your `$USER` passes Claude Code's account pattern, so the derivation lands on `-a <username>`, which is the item you have. And when the derived account misses — your `-a claude-code-user` case — the lookup falls through to the account-less one, so it degrades to exactly today's behaviour rather than failing. It only diverges where several items exist.

One more data point for the entitlement reading: the daemon's refresh request asks for `user:profile user:inference user:sessions:claude_code user:mcp_servers` — no usage scope either — and my token still gets 200 on `/api/oauth/usage`. So the difference doesn't look like it's carried by the scope string. Org or seat entitlement fits better.

Agreed on the knock-on, and it's an argument for both landing: once the account lookup starts selecting a live item, a machine that was 401 can begin answering 403, and right now that still renders as a blank card. Mine picks the item, yours explains the outcome.

Ctrl-Creeper:
Correction to my earlier comment — the entitlement reading was wrong, and I should retract it before it misleads anyone.

That 403 never reached an authorization check. Two pieces of evidence:

1. An **unauthenticated** request to the same host returned the same `403 forbidden` / `"Request not allowed"`, where an accepted path returns `401 authentication_error`. So the rejection happens before any credential is examined.
2. After fixing egress on this machine, plain `fetch` with the *same* token and the *same* scopes returned **200** with populated windows (`five_hour`, `seven_day`, `seven_day_opus`, ...).

The cause was local: a DNS override meant the hostname never reached my proxy's rule engine, so traffic left by a route Anthropic's edge refuses. Nothing to do with Team seats, orgs, or the scope string — your read that scopes don't carry the difference was right, just for a different reason than either of us gave.

Nothing changes for #3597. For this PR it argues for the current wording: the message no longer asserts a cause, it names the refusal and its status code and points at network access — which is what I'd have needed on day one instead of a blank card. It also means the 403 path is reachable for ordinary environmental reasons, not just unusual accounts, which makes the silent-blank-card behaviour worth fixing regardless of how rare entitlement differences turn out to be.

Sorry for the detour.

t-benoit:
Good retraction — and a useful one, since "entitlement difference" would have been a dead end for anyone hitting this later.

The consequence cuts in your favour: a 403 reachable through ordinary egress misconfiguration is far more common than an unusual seat, so the blank card is worth fixing for more people than either of us assumed. Naming the refusal and its status code rather than asserting a cause is the right call for exactly that reason — the failure has at least two origins and the daemon can't tell them apart.

boudra:
Closing in favor of the read-only credential handling merged in [#3442](https://github.com/getpaseo/paseo/pull/3442). The separate missing 403 diagnostic remains a candidate for a focused bug-fix PR.

PR #3475, opened 2026-08-17
Fix model picker loading races at the query cache write

colonelpanic8:
Opening a new workspace or switching hosts can leave the model picker stuck on “Loading…” even after the daemon has discovered the models. A cold snapshot response can finish after a ready push and overwrite it; snapshot queries remain cached indefinitely, so the picker never recovers.

Reconcile snapshots by the server’s `generatedAt` timestamp at React Query’s cache-write boundary, including prefetches. This also covers a push arriving after the query function returns, which a check inside the query function misses. Pushes win timestamp ties against fetch responses, successive equally stamped pushes still advance, and newer fetch responses are accepted. Explicit refreshes and push updates use the same reconciliation, preserving structural sharing.

Validation:

- Regression reproduced against the previous PR implementation with a push delivered between the query function returning and the cache commit.
- `npx vitest run src/hooks/use-providers-snapshot.test.ts --bail=1`: 30 tests pass, covering delayed cache writes, equal timestamps, newer fetches, scope isolation, explicit refreshes, and out-of-order pushes.
- Formatting and targeted lint pass.
- Local app typecheck is blocked by unrelated dependency/native-component errors; the changed files have no reported diagnostics.

### DIFF
--- a/packages/app/src/hooks/use-providers-snapshot.test.ts
+++ b/packages/app/src/hooks/use-providers-snapshot.test.ts
@@ -1,4 +1,4 @@
-import { QueryClient } from "@tanstack/react-query";
+import { CancelledError, QueryClient } from "@tanstack/react-query";
 import { beforeEach, describe, expect, it } from "vitest";
 
 import type { DaemonClient } from "@getpaseo/client/internal/daemon-client";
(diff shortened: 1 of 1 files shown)


boudra:
Closing in favor of the snapshot handling merged in [#4401](https://github.com/getpaseo/paseo/pull/4401).

PR #3110, opened 2026-08-09
feat: add Nix syntax highlighting

Strainy:
Adds Nix (`.nix`) to the highlight package's language registry via `@replit/codemirror-lang-nix`.

## Changes

- New dependency: `@replit/codemirror-lang-nix`
- `nix` → `nixLanguage` in `languagesByExtension` (`packages/highlight/src/parsers.ts`)
- Tests: `isLanguageSupported("test.nix")`, `getSupportedExtensions()` contains `nix`, and a `highlightCode` case asserting `let` highlights as `keyword` and the string literal as `string`

## Testing

`npx vitest run packages/highlight/src/__tests__` — 4 files, 68 tests passing. `npm run format:check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3487, opened 2026-08-17
fix(highlight): highlight Svelte components

### LINKED ISSUE #3364: bug: Support Svelte syntax highlighting
### What's broken

Currently all [Svelte](https://svelte.dev/) `*.svelte` files do not have any syntax highlighting, making it difficult to review within Paseo.

Image

File should be highlighted similar to Github ([example](https://github.com/techniq/layerchart/blob/main/docs/src/examples/components/BrushContext/minimap.svelte))

Image

### Steps to reproduce

1. Open any `.svelte` file
2. View non-highlighted text

### Where did this happen

Desktop (Electron)

### Paseo version

0.3.1

### OS version

_No response_

### Agent provider

None

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

_No response_

fiorelorenzo:
Fixes #3364.

`.svelte` had no entry in the extension table in `packages/highlight/src/parsers.ts`, so `highlightCode` fell through to one unstyled token per line. Everything that resolves a language by extension goes through that table (the file pane, the daemon's diff highlighter, markdown fences and the web editor), which is why a component reads as plain text when you review it.

## What I changed

Registered `@replit/codemirror-lang-svelte` as `svelte`. Its `svelteLanguage` already wires the mixed parsing for `<script>` and `<style>`, so all three regions of a component get roles. It comes from the same publisher as the `@replit/codemirror-lang-csharp` entry already in this package, it is MIT, and only the grammar itself is declared: its CodeMirror peers resolve the same way the C# grammar's peers already do, so `package.json` grows by one line.

The production diff is 7 lines: one import, one table entry, one dependency. The rest is tests.

## Why this grammar

I measured four options on a real component (TS script with runes, markup with directives and mustaches, a style block), counting the share of non-whitespace characters that get a syntax role, per region:

| option | overall | script | markup | style |
| --- | --- | --- | --- | --- |
| today, no entry | 0% | 0% | 0% | 0% |
| `@lezer/html` bare | 22.3% | 7.2% | 44.5% | 15.8% |
| `@lezer/html` + `parseMixed` into JS/CSS | 78.5% | 99.4% | 44.5% | 96.8% |
| `@replit` `svelteParser` (bare) | 30.0% | 7.2% | 65.0% | 15.8% |
| `@replit` `svelteLanguage.parser` | 97.5% | 99.4% | 95.3% | 96.8% |

The hand-rolled mixed parser over `@lezer/html` needs no new dependency, and that is the one I wanted to land, but it leaves every `{#if}`, `{#each}` and `{expr}` in the template unstyled, so the template is exactly where it fails. Note the bare `svelteParser` export is the wrong one to reach for: only `svelteLanguage.parser` carries the script and style nesting.

## Known limitation

The grammar was published in 2022, so it predates Svelte 5's `{#snippet}` and `{@render}`. Those two block keywords stay unstyled. They produce no error nodes and the damage does not cascade: markup inside a snippet highlights normally. I checked the rest of the modern surface and it holds up, including `{@const}`, `{@html}`, `{#key}`, `{#await}`/`{:then}`/`{:catch}`, lowercase event attributes, and runes in the script (to the parser the script is plain TS).

```
snippet      coverage  40%  errorNodes   0
awaitBlock   coverage  93%  errorNodes   0
keyBlock     coverage 100%  errorNodes   0
constBlock   coverage 100%  errorNodes   0
html         coverage 100%  errorNodes   0
eventAttr    coverage  97%  errorNodes   0
runesScript  coverage 100%  errorNodes   0
```

### Tests fail on the old code for the reported reason

Both new tests, with the registry entry reverted and `@getpaseo/highlight` rebuilt:

```
$ npx vitest run packages/server/src/server/utils/diff-highlighter.test.ts packages/highlight/src/__tests__/ -t "Svelte"
 × highlights Svelte components across script, markup, and style 8ms
 × adds syntax highlighting tokens to Svelte components 11ms
AssertionError: expected undefined to be 'keyword'
AssertionError: expected undefined to be defined
 Test Files  2 failed | 3 skipped (5)
      Tests  2 failed | 95 skipped (97)
```

The failure is the bug itself: with no entry for the extension there is no `let` token at all, because the whole line is one unstyled token.

### Tests with the fix

```
$ npx vitest run packages/server/src/server/utils/diff-highlighter.test.ts packages/highlight/src/__tests__/
 Test Files  5 passed (5)
      Tests  97 passed (97)

$ npx vitest run packages/app/src/utils/diff-highlight.test.ts packages/app/src/utils/highlight-cache.test.ts
 Test Files  2 passed (2)
      Tests  13 passed (13)

$ npm run typecheck
exit=0, 0 errors
```

The new coverage is in two places on purpose. `packages/highlight` asserts the roles per region (script keyword and type, template block keyword and mustache variable, markup tag and attribute, CSS class and property). `packages/server` asserts the same thing through `highlightDiffFromHunks`, because the review surface in the issue is the diff and the daemon is what tokenizes it.

### Before and after in the app

Real daemon (`npm run dev:server` with `PASEO_WEB_UI_ENABLED=true`) and the bundled web UI in Chromium, same workspace, same file, same window state. The only difference between the two runs is the registry entry, with `@getpaseo/highlight` rebuilt and the daemon restarted in between, since the diff tokens come from the daemon.

Before:

before

After:

after

### Platforms

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | no | |
| Android | no | |
| Web | yes | bundled daemon web UI in Chromium on Linux, screenshots above |
| Desktop macOS | no | |
| Desktop Windows | no | |
| Desktop Linux | daemon only | daemon-side diff highlighting, which is where the tokens in the screenshots come from |

I only have Linux here. The change is a table entry consumed by pure JS tokenizing, with no platform-gated code and no native module, so I would expect it to behave the same everywhere the same daemon and bundle run, but I have not put eyes on it outside the web client.

I did not visually confirm the file preview pane, the markdown fence path or the CodeMirror editor. They resolve through the same table (`highlight-cache.ts`, `diff-highlight.ts`, `getLanguageForFile` in `view.web.tsx`), and the unit tests cover the table, but that is code reading and not a screenshot.

### DIFF
--- a/nix/npm-deps.hash
+++ b/nix/npm-deps.hash
@@ -1,1 +1,1 @@
-sha256-Ep6Z+J0urY1L66cXGe9bB6tQr8T3TY73GkLrvhf8FmA=
+sha256-WblXeNn1wUUaxElDGNXHcWdvvHJhqp7LySanhLDv5lQ=
--- a/package-lock.json
+++ b/package-lock.json
@@ -2573,6 +2573,54 @@
         "@lezer/common": "^1.1.0"
       }
     },
+    "node_modules/@codemirror/lang-css": {
+      "version": "6.3.1",
+      "resolved": "https://registry.npmjs.org/@codemirror/lang-css/-/lang-css-6.3.1.tgz",
+      "integrity": "sha512-kr5fwBGiGtmz6l0LSJIbno9QrifNMUusivHbnA1H6Dmqy4HZFte3UAICix1VuKo0lMPKQr2rqB+0BkKi/S3Ejg==",
+      "license": "MIT",
+      "peer": true,
+      "dependencies": {
+        "@codemirror/autocomplete": "^6.0.0",
+        "@codemirror/language": "^6.0.0",
+        "@codemirror/state": "^6.0.0",
+        "@lezer/common": "^1.0.2",
+        "@lezer/css": "^1.1.7"
+      }
+    },
+    "node_modules/@codemirror/lang-html": {
+      "version": "6.4.12",
+      "resolved": "https://registry.npmjs.org/@codemirror/lang-html/-/lang-html-6.4.12.tgz",
+      "integrity": "sha512-pw2ReWKUqSkbvh76RAT4NYxiogRu+PWkR2ukAwO9uOgrm8uipkzjtKKtNpyeAQwHOqxEeSvAXZ6vr3AfyB9y/w==",
+      "license": "MIT",
+      "peer": true,
+      "dependencies": {
+        "@codemirror/autocomplete": "^6.0.0",
+        "@codemirror/lang-css": "^6.0.0",
+        "@codemirror/lang-javascript": "^6.0.0",
+        "@codemirror/language": "^6.4.0",
+        "@codemirror/state": "^6.0.0",
+        "@codemirror/view": "^6.17.0",
+        "@lezer/common": "^1.0.0",
+        "@lezer/css": "^1.1.0",
+        "@lezer/html": "^1.3.12"
+      }
+    },
+    "node_modules/@codemirror/lang-javascript": {
+      "version": "6.2.5",
+      "resolved": "https://registry.npmjs.org/@codemirror/lang-javascript/-/lang-javascript-6.2.5.tgz",
+      "integrity": "sha512-zD4e5mS+50htS7F+TYjBPsiIFGanfVqg4HyUz6WNFikgOPf2BgKlx+TQedI1w6n/IqRBVBbBWmGFdLB/7uxO4A==",
+      "license": "MIT",
+      "peer": true,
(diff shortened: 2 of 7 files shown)


fiorelorenzo:
Rebased on main: #3110 landed Nix highlighting an hour after I opened this, and it touches the same table, the same package.json and the same two test files. Both entries are in now, imports and table entries in the existing order, and the lockfile is regenerated from main's so the diff is only the Svelte grammar and its peers.

Re-ran after the rebase: 111 tests pass across `packages/highlight`, the server diff-highlighter and the two app consumers, `npm run typecheck` exits 0. I left `nix/npm-deps.hash` alone since the Nix Update Hash workflow owns it on main.

PR #3495, opened 2026-08-17
fix: reject a question permission that would deliver no answer

thomasvan:
Answering an `AskUserQuestion` permission in any shape but `updatedInput.answers` keyed by one of its questions resolves as `allow` and delivers nothing. The waiting agent is told **"The user did not answer the questions."** — an affirmative falsehood rather than silence, so neither side can see that anything failed. The responder sees success; the agent believes the human declined and acts on that.

The field it is most natural to reach for cannot work. `selectedActionId` is `z.string().optional()` with no membership check, and is read only for `plan` kinds; a question advertises `actions: undefined`, so there is no action id to name in the first place.

It is also unrecoverable. `respondToPermission` deletes the request from `pendingPermissions` on entry, so the malformed answer consumes it and the retry fails with `No pending permission request`. The guard runs **before** that delete, and `AgentManager` drops its own copy only once the call resolves, so both maps survive the throw and the request stays answerable.

## Why it sits in the Claude provider

The answer contract is per provider, not shared:

| provider | mapping | unmapped answer |
| --- | --- | --- |
| claude | key is a question's full text **or** its header, value a non-empty string | nothing is delivered — the defect |
| codex | `mapCodexQuestionResponseByHeader`, headers only | **selects each question's first option** — a supported path |
| opencode | headers only | an empty answer list per question |

An earlier revision of this PR put the check in `AgentManager.respondToPermission`, the join point for all three. That forced it to restate Claude's rule and to scope itself with `provider === "claude"`, and the restatement was wrong twice — it trimmed keys the normalizer does not trim, and read an empty `updatedInput.questions` array as an absent one.

Putting the guard beside the rule makes both classes of bug impossible. `resolveClaudeAskUserQuestionAnswers` is extracted from `normalizeClaudeAskUserQuestionUpdatedInput` with **no change in behaviour** — the 69 existing tests in this file pass untouched — and both it and the guard now call it. So "would this deliver anything" is answered by the code that does the delivering.

That matters more than it looks, because the rule has two details that are easy to restate wrongly:

- `readNonEmptyString` tests `value.trim()` for emptiness but returns the **original** string, so a question named `" Which colour? "` is keyed by that padding. A trimming check would accept `{"answers":{"Which colour?":…}}`, which cannot be mapped — preserving the exact destructive path this PR closes — and reject `{"answers":{" Which colour? ":…}}`, which can.
- An empty `updatedInput.questions` array is a **list**, tested with `Array.isArray`: it wins over the request's own questions and maps nothing.

Neither needs stating in the guard now, because the guard does not restate the rule at all.

## Verification

On a live daemon over the MCP `respond_to_permission` path, against a `claude` agent whose question text is literally `" Which colour? "`:

| response | result |
| --- | --- |
| `{"behavior":"allow","selectedActionId":"Viridian"}` | rejected, request still pending |
| `{"answers":{"Which colour?":"Viridian"}}` | rejected, still pending |
| `{"questions":[],"answers":{"Colour":"Viridian"}}` | rejected, still pending |
| `{"answers":{"Colour":1}}` | rejected, still pending |
| `{"answers":{" Which colour? ":"Ochre"}}` | `success: true`, and the agent replies `Ochre` |

19 new tests in `providers/claude/agent.test.ts`, next to the normalizer tests they extend. No new files; `agent-manager.ts` is untouched.

## Regression test for the retry ordering

Added a test that asserts the ordering the fix depends on: a non-deliverable answer is rejected, the pending request survives the rejection, and a corrected retry against the same request is accepted — the waiting agent receives the answer.

Mutation testing on this change found that the existing suite doesn't pin that ordering: moving the deliverability check to run after the request is deleted from `pendingPermissions` still leaves every existing test in the file green. This test fails against that change.

## Known limit

The WebSocket path (`session.ts handleAgentPermissionResponse`) catches the throw and emits an `activity_log` error rather than failing the caller's call, because a permission response is modelled as a notification rather than a request. So `paseo permit allow` still prints `allowed` and exits 0 for a rejected shape. The destructive half is closed on every path — the request survives and the agent is never told a falsehood — but closing the reporting half needs a protocol change and is left out of this PR.

### DIFF
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -1360,6 +1360,91 @@ describe("normalizeClaudeAskUserQuestionUpdatedInput", () => {
     }
   });
 
+  test("a rejected question answer leaves the request answerable by a corrected retry", async () => {
+    // The question-answer-required rule guards this: the deliverability check runs
+    // before the request is deleted, so a non-deliverable answer costs the caller a
+    // retry rather than the request. Upstream consumed the request on entry and the
+    // retry failed with "No pending permission request".
+    const client = new ClaudeAgentClient({
+      logger: createTestLogger(),
+      resolveBinary: async () => "/test/claude/bin",
+    });
+    const session = await client.createSession({
+      provider: "claude",
+      cwd: process.cwd(),
+    });
+
+    const request = {
+      id: "permission-question-retry",
+      provider: "claude",
+      name: "AskUserQuestion",
+      kind: "question",
+      input: {
+        questions: [
+          {
+            question: "Which colour?",
+            header: "Colour",
+            options: [],
+            multiSelect: false,
+          },
+        ],
+      },
+    };
+
+    const resultPromise = new Promise<unknown>((resolve, reject) => {
+      (
+        session as unknown as {
+          pendingPermissions: Map<
+            string,
+            {
+              request: typeof request;
+              resolve: (value: unknown) => void;
+              reject: (error: Error) => void;
+            }
+          >;
+        }
+      ).pendingPermissions.set(request.id, { request, resolve, reject });
+    });
+
+    try {
+      // A non-deliverable answer: allow with no answers at all.
+      await expect(
+        session.respondToPermission(request.id, {
+          behavior: "allow",
+          updatedInput: {},
+        }),
+      ).rejects.toThrow(/Permission request/);
+
+      // The request survived the rejection: a corrected retry is accepted and the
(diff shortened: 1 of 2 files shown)


thomasvan:
Status update: I still carry this patch on `slp/patches` (fork tip `04563655b`), source-reverified through the 2026-09-07 sync to upstream `c424f8292` (0.7.2). Merge-tree against current `main` (`8c89e72a0`) is clean; the Claude pending-request guard remains needed — a `question` answered in any shape but a real key still resolves as `allow` and delivers nothing.

Issue #3252, opened 2026-08-12
bug(omp): custom message before user message completes the active turn


rafaelfiviola:
### Summary
Paseo 0.3.1 can complete an OMP turn before its user prompt is associated with that turn. If OMP emits `message_end(role=custom)` after `turn_start` but before the prompt's user `message_end`, Paseo emits `turn_completed`; lifecycle becomes idle and a parent can receive `finished` even though OMP continues working.

Current environment: Paseo 0.3.1, OMP 17.2.12. A matching incident trace showed running at 01:55:35.394, `finished` at 01:55:35.440, first tool at 01:55:42.797, and actual turn end at 02:04:20.877.

### Minimal deterministic reproducer
Start `session.run("do work")` with the existing fake OMP runtime, then deliver exactly:

```text
turn_start
message_end { message: { role: "custom", content: "startup notice" } }
# expect zero turn_completed; current handler calls completeTurn here
message_end { message: { role: "user", content: "do work" } }
tool_execution_start ...
agent_end ...
```

The fake already provides `beginTurn()`, `acceptCustomMessage()`, and `acceptPrompt()`. Put `acceptCustomMessage()` before `acceptPrompt()`; completion should occur only after terminal `agent_end`.

### Root cause
The dispatcher sets `activeTurnHasUserMessage` only on user `message_end`. The custom branch calls `completeTurn(turnId, [])` while it is false, and `completeTurn` emits `turn_completed`. Running→idle then drives `notifySafely("finished")`.

Current main `d012b6f`:
- [flag dispatch](https://github.com/getpaseo/paseo/blob/d012b6f0717e2c92c6f591c9e653f7409fd77fd9/packages/server/src/server/agent/providers/omp/agent.ts#L1821-L1829)
- [custom completion](https://github.com/getpaseo/paseo/blob/d012b6f0717e2c92c6f591c9e653f7409fd77fd9/packages/server/src/server/agent/providers/omp/agent.ts#L2012-L2031)
- [completion emission](https://github.com/getpaseo/paseo/blob/d012b6f0717e2c92c6f591c9e653f7409fd77fd9/packages/server/src/server/agent/providers/omp/agent.ts#L2125-L2149)
- [running→idle notifier](https://github.com/getpaseo/paseo/blob/d012b6f0717e2c92c6f591c9e653f7409fd77fd9/packages/server/src/server/agent/agent-prompt.ts#L375-L395)
[...]

PR #3497, opened 2026-08-17
fix(pi): don't complete the active turn on a mid-turn custom message

### LINKED ISSUE #3496: bug: pi provider completes the active turn on any `custom` message, sending a premature `finished`
bug: pi provider completes the active turn on any `custom` message, sending the parent a premature `finished` with the wrong body

## What's broken

On the `pi` provider, any `custom`-role message that arrives during a turn ends that turn. If a pi extension emits a `custom` message mid-task (many do, as a side effect of tool activity), the agent's turn completes early. The parent receives a `finished` notification whose body is that `custom` message rather than the agent's result, and the agent either stops with the task undone or continues under a new turn while the parent already believes it finished.

Expected: the parent is notified once, when the agent's task actually ends, with the agent's real final message as the body. Extension side-channel output must not end the turn or become the finish body.

The behavior is general to the pi provider, not specific to any extension. The turn ends on the message type, so any source of `custom` messages triggers it.

## Steps to reproduce

1. Configure the pi provider with at least one extension that emits a `custom` message during a turn. Any of these do, each with its own `customType` (versions used in the reproduced run):
   - [`@juicesharp/rpiv-pi`](https://www.npmjs.com/package/@juicesharp/rpiv-pi) 2.6.1 — injects `rpiv-guidance` on read/edit/write tool calls
   - [`pi-cursor-sdk`](https://www.npmjs.com/package/pi-cursor-sdk) 0.2.0 — emits `cursor-sdk-agent-lineage`
   - [`@firstpick/pi-extension-stats`](https://www.npmjs.com/package/@firstpick/pi-extension-stats) 0.3.1 — emits `stats_initial_prompt_estimate`
   - [`@capyup/pi-goal`](https://www.npmjs.com/package/@capyup/pi-goal) 0.6.0 — emits `pi-goal-focus`
2. From a parent agent, create a background pi agent with `notifyOnFinish` on and a multi-step task long enough that a `custom` message lands before the task finishes.
3. Wait for the finish notification.

Observed: the parent receives `finished` early. The `<agent-response>` body is the injected `custom` message, not the task result. `get_agent_status` on the child shows the task incomplete (in the run below: `status: idle`, `activeTurn: null`, `outputTokens: 173`, task never done).

A trivial single-step task does not trigger it, because the turn ends normally before any `custom` message lands at the boundary. The trigger is a `custom` message arriving mid-turn, which longer or read-heavy tasks make frequent.

## Where did this happen

Daemon (agent created via MCP `create_agent` from a parent agent).

## Paseo version

0.4.0 (daemon.log reports `daemonVersion: 0.4.0`; CLI `paseo --version` also 0.4.0).

## OS version

macOS 15.7.7 (24G720).

## Agent provider

Custom provider: `pi`, model `cursor/composer-2.5`.

## Provider configuration

pi with extensions loaded that emit `custom` messages during a turn (the four packages listed under Steps to reproduce). Any one of them is enough; the specific set does not matter. For `@juicesharp/rpiv-pi`, the injector is `extensions/rpiv-core/guidance.ts`, which calls `pi.sendMessage({ customType: "rpiv-guidance", ... })` from its tool-call handler.

## Logs (raw)

pi session jsonl for the reproduced run, record type + role/customType, in order:

```
message
message
message
custom_message   <injected guidance>
custom           cursor-sdk-agent-lineage
```

The turn's assistant `message` records are followed by trailing `custom` records. The finish body picked up the first trailing `custom` block.

Full-session type counts for the same run:

```
   4 custom
   3 custom_message
   4 message
   1 model_change
   1 session
   1 thinking_level_change
```

`get_agent_status` at finish (trimmed): `status: idle`, `activeTurn: null`, `lastUsage.outputTokens: 173`, `attentionReason: finished`.

Session file and `~/.paseo/daemon.log` available in full on request.

---

## Hypothesis (not a diagnosis, provided only because the code path is short)

Same class as #3252 (custom-message race), in the pi adapter. In `packages/server/src/server/agent/providers/pi/agent.ts`, `handleMessageEnd` calls `this.completeTurn(turnId, [])` whenever `event.message.role === "custom"`. That promotes a provider-private message type to a shared turn-completion, which drives `running -> idle -> notifySafely("finished")` and hands the custom text to the parent as the final message. The fix stays provider-local: emit the custom text as a timeline item at most, and let turn completion come from pi's real terminal signal, not from a `custom` message-end. Raw session jsonl and daemon log above are the actual inputs.

Related: #3252 reports the same class on the OMP provider. The trigger and fix location differ (pi adapter vs OMP adapter), so this is filed separately.

rolginroman:
### Linked issue

Closes #3496

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

A parent that delegates to a background pi agent relies on the finish notification. It was firing mid-task: `handleMessageEnd` completed the active turn on every `custom` message, and pi extensions emit `custom` messages during tool activity (injected guidance, lineage, usage stats). So a mid-turn `custom` message ended the turn early, and the parent got a `finished` whose body was the injected text instead of the result, with the agent's work cut short or continued under a synthetic turn.

Completing on a `custom` message is only correct for an extension command (e.g. `/show-status`) that emits one and never starts an agent turn, so no `agent_end` follows. An existing test covers that; the branch just didn't distinguish it from a mid-turn injection.

### Goals

- A mid-turn `custom` message no longer completes the active turn; the turn ends on `agent_end`.
- The extension-command case (custom message is the whole turn, no `agent_end`) still completes.

### Non-goals

- Scoped to the pi adapter's `handleMessageEnd`; no change to `agent-prompt.ts` or other providers.
- Does not touch the OMP variant of the same class (#3252).

### QA

Repro: a background pi agent (`cursor/composer-2.5`) with guidance-injecting extensions, given a read-heavy task, sent its parent a `finished` whose body was a raw guidance block, `outputTokens` ~170, task never done.

Fix gates completion on `!this.activeTurnStarted`. New regression test fails on the old code, passes on the new:

```
$ vitest run providers/pi/agent.test.ts -t "does not complete an active agent turn"   # old code
 Tests  1 failed | 56 skipped (57)

$ vitest run providers/pi/agent.test.ts                                                # with fix
 Tests  57 passed (57)
```

```
$ oxlint <the two files>                    → 0 warnings, 0 errors
$ oxfmt --check <the two files>             → correct format
$ npm run build:server && npm run typecheck → exit 0, 0 errors
```

Platforms: daemon unit tests on macOS. No UI change.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/providers/pi/agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.test.ts
@@ -793,6 +793,43 @@ describe("PiRpcAgentSession", () => {
     ]);
   });
 
+  test("does not complete an active agent turn on a mid-turn custom message", async () => {
+    const { pi, session, events } = await createSession();
+    const fakeSession = pi.latestSession();
+
+    await session.startTurn("do work");
+    // An agent turn is running; a real `agent_end` will close it.
+    fakeSession.emit({ type: "agent_start" });
+
+    // An extension emits a custom message mid-turn via `pi.sendMessage`. It
+    // surfaces as a timeline item but must not complete the turn.
+    fakeSession.emit({
+      type: "message_end",
+      message: {
+        role: "custom",
+        content: [{ type: "text", text: "extension side-channel output" }],
+      },
+    });
+
+    expect(events.timelineAndCompletionEvents()).toEqual([
+      {
+        type: "timeline",
+        item: { type: "assistant_message", text: "extension side-channel output" },
+      },
+    ]);
+
+    // Only the real terminal signal completes the turn.
+    fakeSession.emit({ type: "agent_end", messages: [] });
+
+    expect(events.timelineAndCompletionEvents()).toEqual([
+      {
+        type: "timeline",
+        item: { type: "assistant_message", text: "extension side-channel output" },
+      },
+      { type: "turn_completed" },
+    ]);
+  });
+
   test("canceling a silent Pi extension command leaves the session usable", async () => {
     const { pi, session, events } = await createSession();
     const fakeSession = pi.latestSession();
--- a/packages/server/src/server/agent/providers/pi/agent.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.ts
@@ -2247,7 +2247,14 @@ export class PiRpcAgentSession implements AgentSession {
           item: { type: "assistant_message", text },
         });
       }
-      this.completeTurn(turnId, []);
+      // Complete the turn only when this custom message is the whole turn: an
+      // extension command (e.g. `/show-status`) that never starts an agent
+      // turn, so no `agent_end` follows. During an active agent turn a custom
+      // message is side-channel output (an extension calling `pi.sendMessage`);
+      // the turn completes on `agent_end`.
+      if (!this.activeTurnStarted) {
+        this.completeTurn(turnId, []);
+      }
       return;
     }
   }


boudra:
Implemented in [#3849](https://github.com/getpaseo/paseo/pull/3849), which is merged. Closing this PR.

PR #3501, opened 2026-08-18
fix: enable full Paseo MCP access for Devin and OpenCode agents

kassamkhoja:
## Summary

Three issues prevented Devin (ACP) and OpenCode agents from fully using the Paseo MCP server, leaving them with degraded or broken access compared to Claude Code and Codex:

- **Devin/ACP name collision** — a user's manual `paseo` entry in `~/.config/devin/mcp_config.json` shadowed the ACP-injected server, dropping `callerAgentId`, auth tokens, and daemon URL. This broke background execution, finish notifications, and authenticated daemon connections. Renamed the internal server to `paseo-daemon` for ACP providers to avoid the collision.

- **OpenCode permission format mismatch** — preapproval rules used `${server}_${tool}` (underscore) but OpenCode expects `${server}.${tool}` (dot), so no MCP tools were ever preapproved. Switched to dot format and added unconditional auto-approval for internal Paseo MCP tools (`paseo.*`) regardless of the auto-accept setting.

- **Devin MCP permission preapproval** — Devin uses its own internal permission system (`mcp__<server>__<tool>` patterns read from config files), not the ACP `requestPermission` callback. Without preapproval entries, every MCP tool call prompted and blocked forever in unattended/subagent contexts. Now writes `mcp__paseo-daemon__*` to the project-local `.devin/config.local.json` before spawning the Devin ACP process, merging with existing config and preserving user-authored permissions.

### Tests added and results

```
npx vitest run src/server/agent/providers/acp-agent.test.ts src/server/agent/providers/opencode-agent.test.ts src/server/agent/provider-options.test.ts src/server/agent/runtime-mcp-config.test.ts --bail=1

Test Files  3 passed (3)
     Tests  212 passed (212)
```

New tests covering:
- ACP rename: verifies internal Paseo server is renamed to `paseo-daemon` for ACP providers
- OpenCode permission format: verifies dot format and auto-approval of `paseo.*`
- Devin permission preapproval: write, merge with existing config, idempotency

### Typecheck

```
npm run typecheck --workspace @getpaseo/server
✔️ tsgo -p tsconfig.server.typecheck.json --noEmit (clean)
```

### Lint and format

```
npm run lint -- <changed files>   → 0 warnings, 0 errors
npm run format:files -- <changed files>  → all files correct format
```

### Platforms tested

- Linux (daemon/server side) — all tests and typecheck
- No UI changes in this PR (server-only)

### DIFF
--- a/packages/server/src/server/agent/provider-options.test.ts
+++ b/packages/server/src/server/agent/provider-options.test.ts
@@ -173,7 +173,7 @@ describe("exact MCP preapproval mappings", () => {
         hubPolicy,
       ),
     ).toEqual([
-      { permission: "hub_finish_execution", pattern: "*", action: "allow" },
+      { permission: "hub.finish_execution", pattern: "*", action: "allow" },
       { permission: "hub_finish_execution", pattern: "*", action: "deny" },
       { permission: "bash", pattern: "*", action: "ask" },
     ]);
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -2490,10 +2490,57 @@ describe("ACPAgentSession", () => {
     expect(byName.get("external")?.url).toBe("https://example.com/mcp");
   });
 
+  test("does not rename internal Paseo server when target name is already taken by a user server", () => {
+    // If the user already has a non-internal server named `paseo-daemon`,
+    // renaming would overwrite it. Keep the original `paseo` name instead.
+    const session = new ACPAgentSession(
+      {
+        provider: "acp-with-mcp",
+        cwd: "/tmp/paseo-acp-test",
+        mcpServers: {
+          paseo: {
+            type: "http",
+            url: "http://127.0.0.1:6767/mcp/agents?callerAgentId=agent-1",
+          },
+          "paseo-daemon": {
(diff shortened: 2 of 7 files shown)


kassamkhoja:
Pushed a follow-up commit addressing all four Greptile concerns:

1. **Devin preapproval unreachable** — generic ACP providers report `provider` as `"acp"`, not `"devin"`. The `this.provider !== "devin"` check never matched. Fixed by checking `this.defaultCommand[0] === "devin"` instead.

2. **Config failures swallowed** — `ensureDevinMcpPermissionPreapproval` caught all errors and logged a warning, then the session launched and blocked forever on MCP tool calls. Now throws so the error surfaces immediately.

3. **ACP rename collision** — if a user already had a non-internal server named `paseo-daemon`, the rename would overwrite it. Now skips the rename when the target name is already occupied by a non-internal server, keeping the original `paseo` name.

4. **OpenCode name-only trust** — `isInternalPaseoMcpPermissionName` checked only the name prefix (`paseo.*`). An external server named `paseo` would get persistent auto-approval. Now verifies the server config is actually the internal Paseo MCP server by URL, not just by name.

Added tests for the collision guard (ACP) and the external-server name collision (OpenCode). All 217 tests pass.

Issue #3217, opened 2026-08-11
bug: one send creates six agents when the draft's `list_prompts` request fails


fiorelorenzo:
### What's broken

A single send creates six agents instead of one. It happened twice today on the same daemon, about half an hour apart, in two different project directories.

Both times the burst is preceded by exactly six failures of `agent.timeline.list_prompts.request` for the same draft id, and the number of agents created equals the number of failures.

Here is the second occurrence, which is the cleaner of the two because the machine was otherwise quiet. Same client id, same connection id, same draft, timestamps to the millisecond:

```
15:23:54.038  ERR   Failed to handle agent.timeline.list_prompts.request || Agent not found: draft_msg_1786454633666_gwo7ggxtv
15:23:54.242  ERR   (same message, same draft id)
15:23:54.297  ERR   (same)
15:23:54.313  ERR   (same)
15:23:54.315  ERR   (same)
15:23:54.341  ERR   (same)
15:23:54.392  info  Creating agent in /home/dev/Progetti/Sencare (omp)
15:23:54.425  info  Creating agent in /home/dev/Progetti/Sencare (omp)
15:23:54.426  info  Creating agent in /home/dev/Progetti/Sencare (omp)
15:23:54.426  info  Creating agent in /home/dev/Progetti/Sencare (omp)
15:23:54.427  info  Creating agent in /home/dev/Progetti/Sencare (omp)
15:23:54.428  info  Creating agent in /home/dev/Progetti/Sencare (omp)
```

Thirty six milliseconds between the first create and the last, so this is not me pressing send several times. Six agents were created and all six started working.

The first occurrence, earlier the same day, has the identical shape: six `list_prompts` failures for one draft id, then six `Creating agent` for `/home/dev/Progetti/brainstorm`. There the creates were spread over 41 seconds rather than 36 milliseconds, because the box was under load and each creation took around 3 seconds, each one followed by a `ws_slow_request` warning. I originally read that one as me double sending, and only the second occurrence made it clear that it is not.

### Why this is worse than a cosmetic duplicate

Each duplicate is a full provider process. In my case that is six `omp` processes of 230 to 350 MB each, so roughly 1.5 GB of RSS for one prompt. More importantly all six run in the same working directory with the same approval mode, so they edit the same files at the same time with no isolation between them. I caught the second burst while it was live and had to interrupt five of the six by hand before they started stepping on each other.

### Steps to reproduce

[...]

PR #3546, opened 2026-08-19
fix(app): guard against duplicate agent creation on new workspace

ongsibyl:
## Problem

Sending a message from a brand-new workspace (new-workspace flow) consistently creates **two** sessions. Repro and evidence in #3217.

## Root cause

Two compounding issues:

1. **Shipped bundle bug (v0.4.0).** `workspace-draft-submission-store.consumePending` was minified incorrectly — `omit(pendingByDraftId, [r].map(n))` where `var r` is never assigned, so `[r].map(n)` yields `[undefined]` and the pending is **never removed**. Multiple `WorkspaceDraftAgentTab` instances can therefore consume the same pending submission, each calling create once.
2. **No upstream guard.** `create-flow` only had a per-hook `isSubmitting` flag (per component instance), and the daemon has no idempotency on create — `clientMessageId` is passed as run options only. So two tabs both proceed to create.

## Fix

Add a synchronous **per-workspace in-flight guard** in the create-flow store, keyed by `serverId:workspaceId`:

- `create-flow-store`: new `tryBeginCreate` / `endCreate` actions on a `createInFlightByWorkspace` map. `tryBeginCreate` claims the slot synchronously and returns `false` if one is already in flight.
- `create-flow`: all create paths funnel through `runCreateAttempt`, which now claims the slot **before** `onBeforeSubmit`/`createRequest` and releases it in `finally`. The loser throws the existing `alreadyLoading` error instead of creating.
- `handleCreateFromInput` previously called `setPending` without a workspaceId, which would bypass the guard; a new optional `getPendingWorkspaceId` resolves the workspace from the tab, so the guard also covers the double-submit path.

## Testing

- 4 new unit tests: store-level dedupe, no-workspaceId passthrough, `continueCreateFromAttempt` concurrent dedupe across two tabs, `handleCreateFromInput` dedupe.
- `vitest run src/stores/create-flow-store.test.ts src/composer/draft/create-flow.test.ts` — 11 pass.
- `npm run typecheck --workspace=@getpaseo/app` — clean.

### DIFF
--- a/packages/app/src/composer/draft/create-flow.test.ts
+++ b/packages/app/src/composer/draft/create-flow.test.ts
@@ -10,7 +10,7 @@ import { useDraftAgentCreateFlow, type DraftCreateAttempt } from "./create-flow"
 
 describe("useDraftAgentCreateFlow", () => {
   beforeEach(() => {
-    useCreateFlowStore.setState({ pendingByDraftId: {} });
+    useCreateFlowStore.setState({ pendingByDraftId: {}, createInFlightByWorkspace: {} });
   });
 
   it("renders a prepared new-workspace submission before continuing it", async () => {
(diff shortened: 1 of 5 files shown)


boudra:
Superseded by #4442, which takes over duplicate new-workspace submissions with stable per-draft identities and durable receipts. It also fixes pending consumption and tests that distinct drafts can create agents concurrently in the same workspace.

PR #3622, opened 2026-08-21
fix(pi): support Tintinweb subagent lifecycle

### LINKED ISSUE #3160: bug(pi): parent session appears idle while async subagents are running
### What's broken

When a Pi agent launches an asynchronous `pi-subagents` child and the parent returns to its prompt, Paseo shows the parent session as idle/done even though delegated work is still running.

Expected: the parent turn may remain internally idle, but session lists and workspace activity should continue to show active work while any Pi-owned subagent is running. The child should appear in the parent’s subagents track and settle to its completed, failed, or canceled state.

Actual: in the affected build, Pi child lifecycle is not represented in Paseo, so the parent looks finished and the running child is absent from the aggregate activity signal.

This is similar in visible symptom to #2232, but affects the Pi provider and `pi-subagents` rather than OMP task sessions.

### Steps to reproduce

1. Open Paseo desktop and start a Pi agent.
2. From that Pi session, launch a `pi-subagents` child asynchronously (`async: true`).
3. Let the parent turn finish while the child continues working.
4. Return to the agent/session list.
5. Observe that the parent is shown as idle/done while the child is still running.

The original report was observed when the asynchronous child outlived the parent turn.

### Where did this happen

Desktop (Electron)

### Paseo version

0.3.1 development checkout (`a98d154b6`)

### OS version

macOS 26.6.1 (25G76), arm64

### Agent provider

Pi (not currently listed in the issue template provider dropdown)

### Provider configuration

Pi 0.80.6 with `pi-subagents` 0.41.0.

### Logs

Not collected. No crash or error message was observed; the report is the visible lifecycle mismatch after the parent turn becomes idle.

### Screenshots or video

Not captured before the fix was implemented.

p0ns:
### Linked issue

Closes #3160.

Supersedes #3165. This branch preserves its three commits and original authorship, rebases them onto current `main`, resolves the existing test conflict, and adds compatibility for `@tintinweb/pi-subagents`.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Paseo maps Pi tools named `task` and `subagent` to native delegated-task cards. `@tintinweb/pi-subagents` registers `Agent` with a different argument schema, so Paseo displays its calls as unknown-tool JSON.

Background `Agent` calls finish their tool call after launch. Their later lifecycle is available only on Pi's extension event bus, which is not part of the Pi RPC stream. Without a bridge, the child is absent from Paseo's provider-subagent track and the parent appears idle while delegated work continues.

### Goals

- Render matching `Agent` calls with the existing native `sub_agent` detail.
- Show background Tintinweb children in Paseo's provider-subagent track.
- Keep parent and workspace activity visible while background work runs or waits for a slot.
- Map child completion, failure, cancellation, and session shutdown to terminal states.
- Preserve the unknown-tool fallback for malformed or unrelated `Agent` tools.
- Preserve existing `task`, `subagent`, and `subagent:async-*` behavior.
- Clear stale provider state on disconnect and reconcile it after reconnect.

### Non-goals

- Convert Pi-owned children to managed Paseo agents.
- Replay child events that finished before the live Pi connection.
- Import the Tintinweb output-file transcript into the provider-child timeline.
- Add an OMP dependency.

### Compatibility

- No protocol schema changes.
- Foreground lifecycle events stay out of the provider-subagent track through the extension's `isBackground` marker.
- Queued background work uses `running` because the provider-subagent contract has no queued state.

### QA

Focused tests:

```text
$ npx vitest run packages/server/src/server/agent/providers/pi/tool-call-mapper.test.ts packages/server/src/server/agent/providers/pi/agent.test.ts packages/app/src/subagents/provider-store.test.ts --bail=1
Test Files  3 passed (3)
Tests       91 passed (91)
```

Repository checks:

```text
$ npm run build:server
completed successfully

$ npm run format
Finished in 785ms on 3990 files.

$ npm run typecheck
completed successfully for all workspaces

$ npm run lint

$ git diff --check
exit 0
```

Platform coverage:

| Platform | Tested | Notes |
| --- | --- | --- |
| Desktop macOS | Automated only | Mapper, generated extension bridge, store, typecheck, and lint passed. |
| Web | No | Uses the shared provider-subagent client path. |
| iOS | No | Uses the shared provider-subagent client path. |
| Android | No | Uses the shared provider-subagent client path. |
| Windows | No | Not available locally. |
| Linux | No | Not available locally. |

Real-provider before/after UI evidence is still outstanding, so this PR is a draft.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [ ] QA evidence

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -129,7 +129,7 @@ The asymmetry is intentional: a subagent's persistent relationship lives in the 
 
 ## Workspace activity
 
-Agent lifecycle status stays literal: a parent agent is `idle` when its own turn is idle, even if a child is running.
+Stored agent lifecycle status stays literal: a parent agent is `idle` when its own turn is idle, even if a child is running. Session lists may project that parent as `running` while a provider-owned child runs; this does not make the parent turn busy.
 
 Workspace status is an aggregate activity signal computed **per `workspaceId`**. Ownership is never derived from `cwd` — many workspaces may share one directory, and same-`cwd` siblings do not clump under one status. Root agents and cross-workspace subagents contribute their normal state bucket to their own workspace. Same-workspace descendants contribute `running` to the nearest ancestor in that workspace; their non-running attention, permission, and error states stay in the parent's subagents track. This makes a cross-workspace subagent behave like a detached agent for workspace visibility and status without removing its parent relationship.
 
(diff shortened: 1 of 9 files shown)


boudra:
the client is the wrong place for this

PR #3630, opened 2026-08-21
Add keep-awake while agents are running

Pearce-Ropion:
Prevents the desktop from sleeping while any agent is actively working. Adds a new power settings section in desktop settings with a toggle to enable/disable the feature.

## Design

- **Renderer responsibility**: The app monitors agent running state (via aggregated agents) and battery level (via `navigator.getBattery()`) and reports both to the main process on every change.
- **Main process safety**: The daemon re-derives the low-battery cutoff itself rather than trusting the renderer's boolean, ensuring a stale or buggy renderer signal can never hold the `powerSaveBlocker` below 10% battery.
- **User control**: Desktop-only setting, off by default.

## Changes

- New `useKeepAwake()` hook that sends battery + running state to the main process
- New `DesktopPowerSection` component in settings with a toggle
- Server-side `createKeepAwakeCommandHandlers()` that manages `powerSaveBlocker` with the 10% cutoff
- Settings schema and i18n for all supported languages
- Architecture documentation explaining the safety design

## Safety

The feature respects a 10% battery threshold: even if the user enables it and agents are running, the machine will be allowed to sleep once battery drops below 10%. This prevents edge cases where a buggy renderer could theoretically keep the machine awake on low battery.


boudra:

PR #3633, opened 2026-08-21
fix(server): run worktree.setup/teardown in one login shell session

### LINKED ISSUE #3629: worktree.setup commands run as isolated, non-interactive bash -c calls — no shell functions, no PATH from mise/nvm, no state across lines
### Environment
- **Paseo version**: 0.4.0 (desktop, packaged)
- **OS**: macOS 26.5.2 (Apple Silicon)
- **User login shell**: zsh (Oh My Zsh + mise)

### Problem

Commands in `paseo.json`'s `worktree.setup` array don't have access to anything set up in the user's shell profile — custom functions, aliases, or PATH entries added by tools like `mise`, Homebrew, or a version manager via `.zshrc`/`.zprofile` — even though the exact same commands work fine typed into a normal terminal (including Paseo's own terminal tabs).

Example `paseo.json`:
```json
{
  "worktree": {
    "setup": [
      "mise install",
      "brew_installed_tool --version",
      "some-shell-function-from-dotfiles"
    ]
  }
}
```

Any of these can fail with `command not found`, because nothing that activates PATH entries or defines functions in `.zshrc`/`.zprofile` (mise's shims/activation, Homebrew's `shellenv`, custom dotfile functions) has actually run. Explicitly sourcing the profile doesn't help either — under Oh My Zsh, `source ~/.zshrc` run from bash fails outright with "Oh My Zsh can't be loaded from: . You need to run zsh instead."

### Root cause

1. **Hardcoded, non-interactive bash.** `buildStringCommandShellInvocation()` (`packages/server/src/utils/string-command-shell.ts`) always returns `{ shell: "bash", args: ["-c", command] }` on non-Windows for every `worktree.setup`/`teardown` entry — never zsh, never `-i`/`-l`. `.zshrc`/`.zprofile` never source, so any shell function, alias, or tool activation defined there (mise/asdf-style activation, Homebrew's `shellenv`, nvm's `nvm.sh`, user-defined functions) is unavailable.
2. **Each array entry is a separate, isolated process.** `runWorktreeSetupCommands()` (`packages/server/src/server/worktree.ts`) loops over `worktree.setup` and calls `execFileAsync("bash", ["-c", cmd], { cwd, env })` independently per entry. Even a line that successfully sources something or mutates `PATH` does not carry that state to the next line — each is a fresh `bash` process.
3. **No environment derivation beyond `process.env`.** The `env` passed to every setup command is `createStringCommandShellEnv(createExternalProcessEnv(process.env, runtimeEnv))` (`packages/server/src/server/paseo-env.ts`) — the daemon's own inherited environment plus a few `PASEO_*` variables. No shell profile is consulted to fill in gaps.

This is related to #2247 but distinct: that issue's fix (resolving the daemon's own login-shell `SHELL`/`PATH` at startup) would help daemon-launched-by-launchd cases, but would not fix this — `worktree.setup` never goes through `resolveDefaultTerminalShell()`/`createTerminal()` at all, so a corrected daemon PATH still wouldn't restore shell functions or cross-line state, since the shell invoked is always bash -c, never the user's interactive shell.

### Repro

```bash
# Simulates what Paseo does for two worktree.setup entries: two independent
# non-interactive bash processes, no shared state, no .zshrc/.zprofile sourcing.
bash -c 'export FOO=bar'
bash -c 'echo $FOO'          # => empty (expected if FOO were meant to persist)

# A tool/function only made available via .zshrc or .zprofile (mise shims,
# Homebrew shellenv, a dotfile function) is invisible to a non-interactive bash -c:
bash -c 'my_dotfile_func() { echo "defined in zshrc"; }; my_dotfile_func'  # contrived: works
bash -c 'my_dotfile_func'                                                   # => command not found (real case: only defined in ~/.zshrc, never sourced)
```

### Expected behavior

`worktree.setup` commands should have access to the same PATH/shell functions a user gets in a normal terminal (or Paseo's own terminal tabs), and/or state set by one line (PATH mutations, sourced functions) should be available to subsequent lines — matching the mental model of "this is my setup script," not "each line is a brand-new sandboxed process."

### Proposed fix

Open to what the maintainers prefer; a few options:
- Run the entire `worktree.setup` array as one shell invocation (join with `&&`/newlines) so state persists across lines, rather than one `execFileAsync` per entry.
- Optionally invoke the user's actual login shell (`$SHELL -ic`/`-lc`) for setup scripts specifically, mirroring whatever #2247 lands on for terminals, so `.zshrc`/`.zprofile`-based tool activation (mise, Homebrew, version managers) works without requiring users to hand-wrap steps in `zsh -ic '...'`.

Happy to help test/verify a fix — no PR attached, this is a repro/root-cause report.

Pearce-Ropion:
## Summary
- `worktree.setup`/`worktree.teardown` array entries each ran as their own isolated, non-interactive `bash -c` process. `.zshrc`/`.zprofile` never sourced (oh-my-zsh refuses to load under a non-interactive shell), so shell functions/tool activation (nvm, mise, etc.) defined there were unavailable, and state from one entry (exported vars, `cd`) never carried to the next.
- Array entries now run inside a single login+interactive shell session (bash/zsh/fish, detected from `$SHELL`), so profile-sourced activation works and state persists across entries — the shell/login cost is paid once per run instead of once per entry.
- The existing per-entry timeline UI (per-command log, status, exit code, duration in the app) is preserved by reconstructing it from the single process's output via a marker-delimited script, rather than collapsing to one combined result.
- The daemon's own `PATH` is kept reachable as a fallback even if a profile replaces `PATH` outright, preserving the guarantee added in #1908 (which removed login-shell invocation after a profile PATH-clobbering incident).
- Windows and any `$SHELL` we don't recognize (fish is supported; anything else) fall back to the previous per-entry-process behavior, unchanged.

Fixes #3629.

## Test plan
- [x] New unit tests: `packages/server/src/utils/worktree-lifecycle-shell.test.ts` — script generation (marker placement, PATH preamble, fail-fast exit) and the output parser (interleaved/split stdout+stderr chunks, per-index segmentation, exit codes, event ordering), parametrized across bash and fish
- [x] `packages/server/src/utils/worktree.posix.test.ts` — regression tests for profile-sourced tool activation, PATH fallback under a profile that clobbers PATH, cross-entry state sharing, and per-entry fail-fast reporting
- [x] `npm run typecheck` / `npm run lint` / `npm run format:check` (also ran automatically via pre-commit hook)
- [x] Targeted vitest run across the affected files (108 passed, 6 skipped — 5 gated on fish availability, 1 pre-existing unrelated skip)
- [x] CI: added an `Install fish` step to `server-tests-ubuntu` so the fish-dialect tests actually execute in CI instead of skipping
- [ ] Not yet verified end-to-end against a live daemon / real worktree creation flow in the app

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,16 @@ jobs:
         if: ${{ needs.changes.outputs.full != 'false' || needs.changes.outputs.server != 'false' }}
         run: node scripts/npm-retry.mjs install -g @anthropic-ai/claude-code opencode-ai
 
+      # fish is exercised by the worktree.setup/teardown lifecycle-shell tests
+      # (packages/server/src/utils/worktree-lifecycle-shell.test.ts); without
+      # it those cases skip instead of running. Linux-only: this job's steps
+      # are shared with server-tests-windows via the *server_test_steps
+      # anchor, and fish isn't part of the lifecycle-shell dialect support on
+      # win32 (that platform always uses the legacy per-entry loop).
+      - name: Install fish
+        if: ${{ runner.os == 'Linux' && (needs.changes.outputs.full != 'false' || needs.changes.outputs.server != 'false') }}
+        run: sudo apt-get update && sudo apt-get install -y fish
+
       - name: Build server dependencies
         run: npm run build:server-deps
 
--- a/packages/server/src/server/worktree-bootstrap.ts
+++ b/packages/server/src/server/worktree-bootstrap.ts
@@ -322,7 +322,10 @@ export function buildWorktreeSetupDetail(input: {
       log: renderedLog.log,
       status: commandStatusFromResult(result),
       exitCode: result.exitCode,
-      ...(result.durationMs > 0 ? { durationMs: result.durationMs } : {}),
+      // durationMs is only meaningful once the command has actually
+      // completed — gate on exitCode rather than durationMs itself, since a
+      // completed command can genuinely finish in under a millisecond.
+      ...(result.exitCode !== null ? { durationMs: result.durationMs } : {}),
     };
   });
   const renderedLog = buildWorktreeSetupLog({
--- a/packages/server/src/utils/worktree-lifecycle-shell.test.ts
+++ b/packages/server/src/utils/worktree-lifecycle-shell.test.ts
@@ -1,0 +1,329 @@
+import { spawnSync } from "node:child_process";
+import { describe, expect, it } from "vitest";
+
+import {
+  buildWorktreeLifecycleScript,
+  buildWorktreeLifecycleShellInvocation,
+  createWorktreeLifecycleOutputParser,
+  resolveWorktreeLifecycleShell,
+} from "./worktree-lifecycle-shell.js";
+import type { WorktreeSetupCommandProgressEvent } from "./worktree.js";
+
+function hasCommandOnPath(command: string): boolean {
+  const result = spawnSync(command, ["-c", "true"], { stdio: "ignore" });
+  return !result.error && result.status === 0;
+}
+
+const bashAvailable = hasCommandOnPath("bash");
+const fishAvailable = hasCommandOnPath("fish");
+
+describe("resolveWorktreeLifecycleShell", () => {
+  it("returns null when SHELL is unset", () => {
+    expect(resolveWorktreeLifecycleShell({})).toBeNull();
+  });
+
+  it("returns null for unrecognized shells", () => {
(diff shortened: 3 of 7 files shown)


boudra:
Closing this change to the setup-shell workflow; I prefer the explicit, non-login script behavior established in merged [#1908](https://github.com/getpaseo/paseo/pull/1908). Please discuss profile loading and shared shell state in [Discussions](https://github.com/getpaseo/paseo/discussions), following [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #3640, opened 2026-08-21
fix(codex): repair the foreground turn slot

thomasvan:
Three repairs to one piece of state in the Codex provider: `activeForegroundTurnId`, the single foreground-turn slot. Each fixes a different way that slot is leaked or wrongly refused, and all were measured on long-lived Codex agents in production.

This supersedes #3674 and #3683, which are closed in favour of this PR. They are folded in rather than left standing because **two lines cannot exist in any of the three separately**: `interrupt()` and `disposeClient()` each release the slot, and each must wake the callers blocked on it via `flushForegroundTurnClearWaiters()` — a method introduced by #3674. Both other PRs were cut against a base that lacks it, so in *every* merge order the result would land slot releases that leave waiters asleep for the full 10s timeout. Consolidating is the only shape that carries them.

### The failures

| Path | Failure |
| --- | --- |
| `startTurn` races teardown | A prompt injected while a turn is being cancelled races the teardown. `cancelAgentRunBefore` returns once the **manager's** run record settles, and a force-cancel settles it without waiting for the provider, so `startTurn` finds the slot still set and throws. Measured 4 ms apart in production; six occurrences in one session, each losing the prompt and marking the agent errored. Now a bounded 10 s wait, refusing only a turn that genuinely will not end; a timed-out waiter drops itself so a stuck turn cannot retain one dead closure per retrying prompt. |
| `interrupt()` with no identified turn | Codex was the only provider that threw — claude and acp both return. The throw made the manager read the cancel as unacknowledged and refuse every later stop and replace for the rest of the session. Resolving alone is not enough: no turn-end event is coming for a turn Codex never identified, so the slot must be released too. Measured twice, over two hours each. |
| `disposeClient()` | Cleared `currentTurnId` but not the slot. `close()` clears it first, so the asymmetry only fires on a failed reconnect — and then the agent is wedged invisibly: the manager reports no active turn while the provider still holds the slot. Scheduled runs cannot rescue it, since the daemon refuses one against an agent with an in-flight run. Measured: fifteen minutes of refused prompts, `stop` reporting zero stopped runs, sends answering `failed to start`, recovered only by a daemon restart. |
| force-cancel | `cancelAgentRunNow`'s force-cancel dispatches `turn_canceled`, which resolves `runs.getMatchingWaiters` and settles the **manager's** run record. Nothing on that path reaches the session, and the session clears the slot only on a turn end that a force-cancel means is never coming. The slot is then held for the rest of the session and each retry is refused the same way. Measured four times in one hour with the other repairs live. |

### Why the registry commit is required

`wrapSessionProvider` returns a plain object literal that hand-enumerates the `AgentSession` surface, so a method added to a concrete session class is invisible through it and the manager's optional-method probe finds `undefined`. Every registry-defined provider is wrapped, so without the forward the force-cancel release is correct code that can never execute. It is declared on `AgentSession` and added to the wrap test's covered list, so the compile-time exhaustiveness guard — which exists for exactly this class of omission — now includes it.

### Scope

`codex-app-server-agent.ts` and its test file; `agent-manager.ts` and its test file for the force-cancel path, which is the manager's; and `agent-sdk-types.ts` + `provider-registry.ts` + the wrap test for the forward. The manager change is one call plus a log line, a no-op unless the session implements the optional method, so other providers are untouched. Refusal semantics for a turn that genuinely will not end are unchanged, as are the preserve-the-turn semantics for providers that hang on or reject an interrupt.

Not covered: the `run.kind === "foreground"` branch of the force-cancel has no `run.turnId` to key a release on, and the claude, acp and opencode sessions carry the same field but were not audited.

### Verification

- Every commit typechecks on its own; the three are file-disjoint.
- The five affected test files: 377 passed, 0 failed.
- `src/server/agent`: 1983 passed / 12 failed on this branch, against 1979 / 11 on plain `main`. No test fails here that does not also fail on `main` — the three that differ pass in isolation and flake only under parallel load.
- Deleting the two `flushForegroundTurnClearWaiters()` calls fails exactly the two tests written for them, and nothing else.
- Deleting the force-cancel release fails exactly the new manager-level test (`expected [] to include 'stuck-turn'`), and nothing else.
- Deleting the registry forward fails the wrap test's exhaustiveness assertion, and nothing else.
- Observed on a long-lived deployment: before the registry forward, force-cancels logged the release probe as unreachable and the wedge recurred; after it, twelve force-cancels each released the slot and the wedge has not recurred.

### DIFF
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -1194,6 +1194,8 @@ class McpCapableTestAgentClient extends TestAgentClient {
 
 class ControlledInterruptSession extends TestAgentSession {
   interruptCalled = false;
+  releasedForegroundTurnIds: string[] = [];
+  heldForegroundTurnId: string | null = null;
 
   constructor(
     config: AgentSessionConfig,
@@ -1204,6 +1206,7 @@ class ControlledInterruptSession extends TestAgentSession {
   }
 
   override async startTurn(): Promise<{ turnId: string }> {
+    this.heldForegroundTurnId = this.turnId;
     setTimeout(() => {
       this.pushEvent({ type: "turn_started", provider: this.provider, turnId: this.turnId });
     }, 0);
@@ -1213,6 +1216,16 @@ class ControlledInterruptSession extends TestAgentSession {
   override async interrupt(): Promise<void> {
     this.interruptCalled = true;
     await this.interruptBehavior(this);
+  }
+
+  // Mirrors the codex session's keyed release: only the slot this id owns.
+  releaseForegroundTurn(turnId: string): boolean {
+    this.releasedForegroundTurnIds.push(turnId);
+    if (this.heldForegroundTurnId !== turnId) {
+      return false;
+    }
+    this.heldForegroundTurnId = null;
+    return true;
   }
 }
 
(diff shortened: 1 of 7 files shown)


thomasvan:
Status update: I still carry this cluster on `slp/patches` (fork tip `04563655b`), re-verified and adapted through the 2026-09-07 sync to upstream `c424f8292` (0.7.2). I have rebased this PR head (`cbece7d6c`) onto current `main` (`8c89e72a0`) with the required adaptations: `cancelAgentRunNow` now keys the release on `this.runs.getTurnId(agentId)` (upstream reworked run tracking into `AgentRunState`), provider tests use the `createSession`/`castInternals` seam, and the review also found the "already idle" interrupt response cleared the slot without flushing queued `startTurn` waiters — the updated head releases through the keyed path there too, with a regression. `codex-app-server-agent.test.ts` 158/158, `agent-manager.test.ts` and `provider-registry-wrap.test.ts` green, server typecheck clean.

boudra:
Closing in favor of the runtime-reconciliation approach in [#4041](https://github.com/getpaseo/paseo/pull/4041), which is still open. The Codex-specific failed-reconnect and unidentified-turn cases remain follow-up work.

PR #3652, opened 2026-08-21
feat(app): let the content measure be configured

josham:
### Linked issue

None. There is no existing issue or discussion for this — I searched issues, PRs, and Discussions before starting. Happy to open a product discussion first if you'd rather that came before code.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Chat, the composer, and rendered Markdown center themselves in a reading column fixed at 820px. On a wide display most of the pane is empty; the wider the monitor, the more the workspace pane is doing nothing. The column exists for a good reason — `docs/design.md` §7 is explicit that lines stay readable and the eye shouldn't track wide horizontal distances — so this doesn't remove it. It makes the measure a preference, clamped on both ends, defaulting to today's 820 so nothing changes for anyone who doesn't go looking.

This is the same axis #3637 opened up for content text size: what you read is separate from the chrome around it, and worth controlling on its own.

### Goals

- Add a **Content width** setting under Appearance → Layout.
- Apply it to every surface that carries the measure: agent stream, composer, turn footer, agent panel, New workspace, Markdown preview, archived-agent callout, workspace tab, composer tracks.
- Keep the default at 820px so existing installs are untouched.
- Clamp to 600–1600px and say so in the row's hint.
- Keep virtualized height estimation correct when the measure changes.

### Non-goals

- No unbounded or "full width" option. The column is deliberate; this widens or narrows it rather than removing it.
- Not changing the 720px settings/projects detail column, which is a separate rail.
- Not changing the default.
- Not fixing the numeric-row clamp display bug described under Known issues — that is pre-existing and gets its own PR.

### Implementation notes

The value rides the path #3637 established. `applyAppearance` patches every registered theme, and surfaces read `theme.layout.maxContentWidth` from their existing `StyleSheet.create((theme) => …)` factories, so the native ShadowRegistry repaints with no React re-render and no `useUnistyles()`.

Two places can't read the theme and are handled explicitly:

- **New workspace composer** lands on a Reanimated `Animated.View`. Per `docs/unistyles.md`, a Unistyles theme-factory style there races the worklet runtime and crashes on theme change, so it reads the setting directly and passes the measure as a plain inline style.
- **Virtualized height estimation** runs in plain modules with no theme access. `applyAppearance` mirrors the committed value to `utils/content-estimate-width.ts` in the same pass. Cached block heights stay keyed by the width they were measured at, so entries recorded under a previous measure miss and age out rather than returning a wrong height.

`MAX_CONTENT_WIDTH` remains the authored default, now sourced from a `LAYOUT` token, keeping `applyAppearance` idempotent for the same reason `scaleFontSize` derives from `FONT_SIZE` rather than the live theme.

### QA

**Automated**

```
npx vitest run packages/app/src/hooks/use-settings/storage.test.ts \
  packages/app/src/appearance/apply.test.ts \
  packages/app/src/components/appearance-style-boundary.test.ts \
  packages/app/src/utils/assistant-message-height-estimate.test.ts \
  packages/app/src/utils/assistant-image-metadata.test.ts \
  packages/app/src/styles/unistyles-module-scope.test.ts
→ 6 files, 96 passed
```

New coverage: clamp high/low/non-numeric/string-committed/default-for-old-settings; theme patched on every registered theme; measure mirrored to the estimators; boundary key changes on a width-only change; a block measured at one measure misses at another instead of reusing its height; image estimates scale with the measure.

`npm run typecheck`, `npm run lint`, `npm run format:check`, `git diff --check` all clean.

**Manual — browser (Chromium, 1800×1000)**

Measured computed styles rather than eyeballing. Against a real agent conversation:

| Setting | Capped elements in the stream |
| --- | --- |
| 820 (default) | 10 × `max-width: 820px`, rendered 820px |
| 1400 | 10 × `max-width: 1400px`, rendered 1400px |

- Clamping from the UI: `9999` → `1600`, `100` → `600`.
- Persists across reload and applies on boot.
- Committed a non-default measure, then flipped the color scheme four times — app alive, value intact, no page errors. This is the documented Reanimated crash hazard, so it is the check that matters most.
- Dark mode renders the new section correctly.
- Row present and stable at 390×844, 844×390, and back — it does not appear or disappear when a phone rotates.
- Only console output is a pre-existing `uniProps` React warning from `withUnistyles`, unrelated to this change.

**Manual — Electron desktop**

Driven via Playwright `_electron` against the dev Metro. Same 10 elements at 1400, no page errors, measure applies and restores.

**Platforms**

- ✅ Browser (Chromium)
- ✅ Electron desktop (Linux)
- ❌ **iOS — not tested.** No macOS host available.
- ❌ **Android — not tested.** No emulator image available on this machine.

Both untested platforms use the same theme-token path as the browser, but I have not seen them run and am not claiming otherwise.

Settings:

01-settings-layout-section-light

Before:

06-agent-stream-820

After:

07-agent-stream-1400

### Known issues

Typing an out-of-range value that clamps to the number already stored leaves the out-of-range text visible in the field until you navigate away — the clamp and the persisted value are correct, only the display is stale. `EditingTextInput` is uncontrolled, so the field only re-syncs when a committed change remounts the appearance boundary.

This is **pre-existing**: Code size (max 22) behaves identically on `main` today — at 22, typing `99` leaves `99` on screen. This PR adds a fourth row with the same behavior rather than introducing it. Fixed separately so it can be reviewed on its own and cover all four rows.

### Open questions

- **The 1600 ceiling is my number, not a derived one.** At 1400 on a wide display, prose runs ~175 characters per line, well past the ~75–90 that typography convention favours and beyond what §7's rationale is comfortable with. Happy to bring it down to ~1200 if you'd rather the ceiling stayed closer to a readable measure.
- **Appearance → Layout is a new section.** I put it there rather than under Fonts because width isn't typography, and §7 says more rows means more sections. If you'd rather it sat in the Fonts card, that's a small move.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


raymelon:
Same pain here from the other side - with text at the largest size the settings allow, a fixed 820px column gets cramped fast, and the rest of the pane goes to waste. Being able to widen the measure would pair well with the font-size situation. Would love to see this land.

josham:
@boudra can you have a look?

boudra:

PR #3660, opened 2026-08-21
feat(app): show MCP server status in the composer

lianmatsuo:
### Linked issue

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

When an MCP server fails to connect, nothing in Paseo says so — the agent just quietly lacks the tools you configured, and you find out when a tool call breaks. This adds a plug control to the composer's right rail that answers "is this agent actually connected to my MCP servers", per agent, at the point you're about to ask it to do something.

### Goals

- One control and one row design across every provider that has MCP at all.
- Live connection status where a provider can report it: Claude (`Query.mcpServerStatus()`), Codex (`mcpServerStatus/list`), OpenCode (`client.mcp.status()`).
- Honest reporting where it can't: ACP (all catalog agents, incl. Copilot) and Pi have no status channel, so they list the servers Paseo injected and the report is tagged `configured` — the panel says once that nothing verified them, rather than marking every row.
- No control at all when there is nothing to show: OMP and mock take no MCP; a provider that can't report returns `unavailable: "unsupported"` and the plug disappears.
- Keep the cost off the hot path — see Non-goals.

### Non-goals

- **No passive badge on the trigger.** It required an eager fetch on every composer mount, which is not affordable (below). You see status when you open the panel.
- **No per-provider special-casing in the app.** Claude answers in 3ms and Codex in ~3.5s, but gating laziness on the provider would put one provider's quirk into shared UI logic, so the cheap case waits with the expensive one.
- **Codex can't distinguish "disabled" from "crashed"** (`serverInfo: null` for both), so those rows read `Unknown` rather than accusing a deliberately disabled server of failing. Reading the `enabled` flag would cost a subprocess call.
- Non-English strings are machine-translated except for two native-reviewed fixes (`es.failed`, `fr.connecting`).

### QA

Ran against a real local daemon with my own MCP config, driving the dev app in a browser.

**Claude — live status, `needs_auth` rows called out**

Claude MCP panel

**Codex — live status; `computer-use` is disabled in config, which Codex reports identically to a crash, so it reads `Unknown` rather than `Failed`**

Codex MCP panel

**Cursor (ACP) — nothing injected, so an honest empty state instead of a fabricated one**

Cursor MCP panel

**Performance.** Codex's `mcpServerStatus/list` returns every server's full tool schemas — measured at **1.1 MB and ~3.5s** warm, against Claude's **3–6ms**. It ignores `includeTools`/`fields`/`summary`, so it can't be asked for less. Fixed by fetching only while the panel is open plus a daemon-side per-agent cache with concurrent calls coalesced. Measured end-to-end through the daemon after the change:

| | cold | cached | `force` (refresh) |
| --- | --- | --- | --- |
| Codex | 7014 ms | **1 ms** | 4232 ms |
| Claude | 1915 ms | 1 ms | 2 ms |

**One bug this surfaced, fixed in the second commit.** Driving the real app caught that `wrapSessionProvider` re-declares the session surface by hand, so `listMcpServers` was silently dropped for every aliased provider — the panel appeared for a Cursor agent then removed itself the instant it asked. Not a type error, since the method is optional. The new test asserts the whole optional surface survives the wrapper.

**Also verified by hand:** a provider with no source removes the control rather than rendering an error panel; a stopped agent stays retryable instead of hiding the control permanently; refresh issues exactly one request.

Not tested on iOS/Android — the panel is on the shared menu engine and uses its documented compact tier (40pt rows/targets below `md`), but I have no device screenshots.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

Tests: 48 new across the panel, the hook (retry/refresh/caching semantics), the daemon cache (TTL, singleflight), the session wrapper, each provider's mapping, and an RPC round-trip. Existing Codex (142) and ACP (96) suites pass.


boudra:

PR #3371, opened 2026-08-14
fix(omp): keep parent non-idle while internal task subagents run

jasonhnd:
### Linked issue

Closes #2232

Replaces #2245. That PR had the right idle-gate direction, but review found two defects this rewrite addresses. Do not merge #2245.

### Type of change

- [x] Bug fix
- [ ] New feature (with prior issue + design alignment)
- [ ] Refactor / code improvement
- [ ] Docs

### What does this PR do

OMP can end the parent model loop (`agent_end` + `!isStreaming && !isCompacting`) while internal `task` children are still writing. Paseo then marks the parent idle/stopped even though work continues.

This PR:

1. **Idle gate.** Do not `completeTurn` until the subagent index reports no running children.
2. **Snapshot reconcile.** Successful `get_subagents` replies list only still-running children. An id that previously appeared and is now missing is treated as completed. A child seen only via lifecycle (never in a snapshot) is not killed by an empty first reply. If `get_subagents` is unavailable, fall back to the event index.
3. **Task cards.** `task` `tool_execution_end` is a dispatch ack; children start later. Keep the parent call running and settle it from the index once a linked child exists and none remain running. `completeTurn` force-settles a task that never produced a child. Failed tasks still complete immediately.

Wire-order tests emit `tool_execution_end` **before** `subagent_lifecycle: started`, matching OMP v17.

### Out of scope

- `persistence.nativeHandle` pointing at a missing `.jsonl` while the real artifact is a same-stem directory. That is the other half of #2232 and needs a separate change.
- `detached` children are included in the idle gate. Excluding them would restore the original false-idle for long fan-out audits; including them can pin a later turn. Maintainer call welcome.

### How did you verify it

```bash
npx vitest run packages/server/src/server/agent/providers/omp --bail=1
```

Result: **19 files, 127/127 passed**, including:

- parent stays non-idle while a lifecycle child is running after `agent_end`
- empty `get_subagents` does not complete a never-listed lifecycle child
- a previously listed snapshot id disappearing completes the parent
- `get_subagents` unavailable still honors lifecycle
- task card stays `running` when the result precedes `started`, then completes when the child finishes
- a task with no child is force-settled on turn complete
- failed tasks complete immediately

Also: `npm run lint` and `npm run typecheck --workspace=@getpaseo/server` on touched files.

I did not dogfood against a live OMP 17 desktop session in this pass.

### Review notes

Addresses the two findings on #2245 from @ABorakati:

[...]

PR #3667, opened 2026-08-21
fix(omp): bound the post-agent_end completion gate (stacked on #3371)

### LINKED ISSUE #3654: bug(omp): post-agent_end completion gate can spin forever on non-idle or unavailable get_state
## Summary

Paseo's native OMP provider can leave a foreground turn permanently `running` after OMP has already emitted a terminal `agent_end`. The provider waits for a follow-up `get_state` response to report both `isStreaming: false` and `isCompacting: false`, but it retries forever when that condition never arrives or when `get_state` repeatedly fails.

The initiating condition can be an OMP stale state response or an RPC transport failure. The unbounded spinner is owned by Paseo: it catches every failure and provides no deadline, terminal error state, or diagnostic signal.

## Affected path

Paseo 0.4.0:

- `packages/server/src/server/agent/providers/omp/agent.ts` (compiled at `dist/server/server/agent/providers/omp/agent.js`)
- `handleSessionEvent()` receives an assistant-bearing OMP `agent_end` and invokes `completeTurnAfterProviderIdle()`.
- `completeTurnAfterProviderIdle()` loops while the turn remains active:

```ts
while (!this.closed && this.activeTurnStarted && this.currentTurnIdForEvent() === turnId) {
  try {
    const state = await this.runtimeSession.getState();
    this.state = state;
    if (!state.isStreaming && !state.isCompacting) {
      this.completeTurn(turnId, messages);
      return;
    }
  } catch (error) {
    this.logger.debug({ err: error }, "OMP state unavailable while waiting for provider idle");
  }
  await this.providerIdleScheduler.waitForRetry();
}
```

The retry interval is 10 ms. There is no deadline, retry budget, state transition, or externally visible diagnostic.

## User-visible failure

1. A task child submits its terminal structured yield.
2. OMP emits the automatic `async-result` to the parent.
3. The parent has produced an assistant message / terminal `agent_end`.
4. Paseo continues to show the parent as running forever because `get_state` either remains non-idle or repeatedly errors.

This makes a completed session indistinguishable from a genuinely working turn. The user cannot tell whether OMP is still draining, OMP is reporting stale state, or Paseo has lost the RPC state path.

## Deterministic provider-harness reproducer

This does not require a real provider, model, or Paseo UI.

1. Construct the OMP provider with a fake `runtimeSession`.
2. Start a foreground turn and emit the normal assistant `message_end`, followed by an assistant-bearing `agent_end`.
3. Make `runtimeSession.getState()` either:
   - resolve forever with `{ isStreaming: true, isCompacting: false }`, or
   - reject forever.
4. Advance the provider idle scheduler through an arbitrary number of retries.
5. Observe that the current provider emits neither `turn_completed` nor a terminal/recoverable failure indication, and retains the active turn indefinitely.

The same test should assert that a later healthy state response still completes exactly once.

## Baseline checked

OMP 17.4.0's clean `rpc-ui` path behaves correctly: a parent launches a standard `task` child, does not call `hub.wait` or poll manually, receives the automatic `async-result`, and resumes with a parent response. This report does **not** claim that OMP currently drops normal child-yield delivery.

The problem is Paseo's behavior once its post-`agent_end` state check does not become healthy: the host retries forever without exposing that condition.

## Expected behavior

Paseo must preserve the valid safety property of not emitting ordinary `turn_completed` while OMP is genuinely streaming or compacting. However, it must not silently retain a user-facing running turn forever.

After a bounded post-`agent_end` wait or bounded consecutive `get_state` failures, Paseo should:

- emit a structured, user-visible stalled/failed/recoverable completion state with the last observed `isStreaming` / `isCompacting` values or the RPC error;
- stop the unbounded 10 ms retry loop;
- retain enough diagnostics to distinguish stale OMP state from a failed RPC request; and
- permit explicit recovery (for example, retry state, cancel, or resume) without fabricating a normal `turn_completed` event.

A subsequent healthy state response before that bound must still produce exactly one ordinary `turn_completed`.

## Related

- #2260 and #2281: adjacent OMP-provider completion bugs with different triggering event sequences; both assume the provider-idle state check eventually succeeds.
- https://github.com/can1357/oh-my-pi/issues/6916: earlier report of an RPC host remaining active after OMP displayed a terminal answer. This report identifies Paseo's unbounded completion gate as the host-side failure mode; attach an OMP trace only if `get_state` is observed to remain `isStreaming: true` after the terminal yield.

## Environment

- Paseo: 0.4.0
- OMP: 17.4.0
- OMP mode: `rpc-ui`
- OS: Linux 6.8.0-136-generic x64

joeshull:
> **Draft, and not mine to merge. Do not merge before #3371.** The first two commits are @jasonhnd's, unmodified — merging this as-is would land #3371 without its own review. If @jasonhnd would rather take the last three commits onto `fix/2232-omp-idle-gate` and carry them in #3371, that's my preference too; say the word and I'll close this.

Fixes #3654. This branch is #3371's head (`fa9fc5e62`) plus three commits. GitHub shows all five until #3371 merges; **review only the last three**.

I cannot push to this repository, so I cannot open this against a base branch holding #3371's commits. Once #3371 lands on `main`, this narrows to its own three commits automatically. Merges cleanly into current `main` (`8f0c35e92`); no omp drift since #3371's base.

I cannot push to this repository, so I cannot open this against a base branch holding #3371's commits. If a maintainer creates that branch here, or once #3371 lands on `main`, this PR narrows to its own three commits automatically. Merges cleanly into current `main` (`8f0c35e92`); no omp drift since #3371's base.

## Problem

`completeTurnAfterProviderIdle()` polls `get_state` every 10 ms after an assistant-bearing `agent_end`, swallows every error, and has no deadline. If OMP reports a stale state or the state RPC keeps failing, the turn shows as running until the user cancels — and `startTurn` throws while a turn is active, so the session is wedged for new prompts. The only signal is a `logger.debug` line, and the daemon runs at info level, so nothing surfaces.

The loop has not changed since it shipped in #2067, and the two tests that pin the unbounded wait came in that same commit. #2261 and #2282 both fixed *reaching* the gate and explicitly kept it as-is ("the change does not synthesize idle or add a timeout"). #3371 adds a third reason to wait inside it — hence the stack.

`can1357/oh-my-pi#6916` is a field occurrence of the symptom: a session held `running` for 25m44s after the final answer, cleared only by a forced cancel. It is still open, waiting on RPC evidence the host never emitted.

## Change

**Budgets.** `OmpProviderIdleScheduler.waitForRetry()` takes `{ attempt, consecutiveFailures, elapsedMs, isCompacting, isWaitingOnSubagents }` and returns `{retry: true} | {retry: false, reason}`. The default scheduler backs off 10 ms → 1 s and gives up after 3 consecutive `get_state` rejections, after 60 s of monotonic wall clock, or after 10 minutes if OMP reports compaction. Wall clock rather than a retry count is deliberate: each `get_state` carries the JSONL-RPC request timeout, so counting attempts would let a slow-but-answering state path stretch the wait to tens of minutes. Compaction gets its own budget because the gate waits for `isCompacting` to clear and compacting a large context routinely outlasts 60 s.

**The subagent wait from #3371.** A `get_subagents` reply listing running children is positive evidence OMP is working, and a fan-out has no bounded length, so the wait budget does not apply while that evidence is current. It resumes the moment `get_subagents` stops answering — which is the stuck-parent risk that closed #2245: an index still holding children it can no longer confirm no longer holds the turn open forever. The `get_state` failure budget applies throughout.

**Structured failure.** On abandon the turn fails with `omp_provider_idle_timeout` or `omp_provider_state_unavailable`; a gate that never observed any state reports `state_unavailable` whichever budget ran out, because a hanging `get_state` burns the wait budget before three rejections can land. The diagnostic carries the last observed `isStreaming`/`isCompacting`, the last RPC error, and whether subagents were still reported running, and it logs at `warn`. `turn_failed` already carries `code` and `diagnostic`, so there is no `packages/protocol` change and nothing to gate on `server_info.features`. The turn state is cleared, so recovery is an ordinary prompt.

**One gate per turn.** Every `agent_end` used to open another concurrent loop, and two loops that both observed idle each called `completeTurn` — two `turn_completed` for one turn, reproduced in the harness. Gates are keyed on the turn ID, with a shared key for autonomous cycles, which poll like any other gate. The gate holds the terminal payload and the newest `agent_end` overwrites it, unless that would discard an error the gate already holds.

**Ownership.** Ownership is re-checked before completing and before failing, not only at the top of each poll. Both sites sit after two awaits, and a `get_state` carries the 30 s RPC timeout, so a cancel-then-reprompt could otherwise let a stale gate fail the cancelled turn, cancel the new turn's tool calls, and clear its turn ID — after which its events go out with no turn ID and the manager back-fills whatever turn is active.

**Terminalize on stall.** A stall leaves OMP's state unknown, so tool calls and subagents still marked running are cancelled; otherwise they spin with no turn left to finish them.

The safety property holds: no ordinary `turn_completed` while OMP reports streaming, compacting, or running children, and a healthy state response before the budget still completes exactly once. #3371's tests and the two pre-existing gate tests are unchanged and pass.

## Tests

Sixteen new tests in `packages/server/src/server/agent/providers/omp/agent.test.ts`, each written before its production change and watched fail — except the two that drive `createOmpProviderIdleScheduler()` directly, whose pre-change failure was an import error rather than a behavioral one.

- exactly one completion when OMP repeats `agent_end`, foreground and autonomous
- turn fails on wait-budget expiry and on repeated `get_state` failures, with the right code each time
- a wait-budget expiry whose last check failed keeps both the state and the error in the diagnostic
- a gate that never observed a state reports `state_unavailable`
- newest `agent_end` error wins over the first snapshot, and an earlier error survives a later error-free cycle
- a stale gate does not fail or clear a turn that started after it was abandoned
- the session reports observed compaction, elapsed time, and the subagent wait to the scheduler
- the subagent wait stops being trusted once `get_subagents` stops answering
- in-flight tool calls are cancelled when the gate gives up
- a new prompt is accepted after a stalled turn fails
- the default scheduler's budget decisions, including compaction and subagent waits

## QA

- `npx vitest run packages/server/src/server/agent/providers/omp` — 19 files, 143 tests pass
- `npm run lint` on the omp provider — 0 warnings, 0 errors

`npm run typecheck` in my checkout reports missing-module errors for `@replit/codemirror-lang-csharp` and `@paseo/plugin` at this base; those come from resolving my `node_modules` against a different lockfile, not from these commits. CI is the authority.

Not exercised against a live OMP process; the harness reproduces every failure mode deterministically. This surfaces a stall with the last state and RPC error but does not persist raw child RPC frames, which is the evidence `oh-my-pi#6916` is still waiting on.

### DIFF
--- a/packages/server/src/server/agent/providers/omp/agent.test.ts
+++ b/packages/server/src/server/agent/providers/omp/agent.test.ts
@@ -6,6 +6,14 @@ import type { OmpNoTurnScheduler, OmpProviderIdleScheduler } from "./agent.js";
 import type { OmpUsagePollScheduler } from "./usage-poller.js";
 import { OmpHarness } from "./test-utils/omp-harness.js";
 
+function lastToolCallStatus(omp: OmpHarness, callId: string): string | undefined {
+  const items = omp
+    .timeline()
+    .filter((item) => item.type === "tool_call" && item.callId === callId);
+  const last = items[items.length - 1];
+  return last?.type === "tool_call" ? last.status : undefined;
+}
+
 class ManualIdleScheduler implements OmpProviderIdleScheduler {
   private readonly retries: Array<() => void> = [];
   private readonly waiters: Array<{ count: number; resolve: () => void }> = [];
@@ -527,6 +535,130 @@ describe("OMP agent client and session", () => {
     await omp.waitForProviderStateChecks(2);
     await waitForImmediate();
     expect(omp.completedTurnCount()).toBe(1);
+  });
+
+  test("holds a task tool call open when its OMP children appear after the result", async () => {
+    const omp = new OmpHarness();
+    await omp.start();
+
+    await omp.requireStartTurn("fan out");
+    const runtime = omp.runtime();
+    runtime.beginTurn();
+    runtime.acceptPrompt("fan out", "user-fanout");
+    runtime.streamAssistantText("delegating");
+    runtime.emit({
+      type: "tool_execution_start",
+      toolCallId: "task-1",
+      toolName: "task",
+      args: { description: "spawn workers" },
+    });
+    runtime.emit({
+      type: "tool_execution_end",
+      toolCallId: "task-1",
+      toolName: "task",
+      isError: false,
+      result: { text: "Spawned 1 background agent" },
+    });
+    expect(lastToolCallStatus(omp, "task-1")).toBe("running");
+    expect(omp.runningToolCallIds()).toEqual(["task-1"]);
+
+    runtime.emit({
+      type: "subagent_lifecycle",
+      payload: {
+        id: "Worker",
+        agent: "Worker",
+        status: "started",
+        parentToolCallId: "task-1",
+        index: 0,
+      },
+    });
+    expect(lastToolCallStatus(omp, "task-1")).toBe("running");
+    expect(omp.runningToolCallIds()).toEqual(["task-1"]);
+
+    runtime.emit({
+      type: "subagent_progress",
+      payload: {
+        index: 0,
+        agent: "Worker",
+        parentToolCallId: "task-1",
+        progress: { id: "Worker", status: "running", recentOutput: ["still working"] },
+      },
+    });
+    expect(lastToolCallStatus(omp, "task-1")).toBe("running");
+
+    runtime.emit({
+      type: "subagent_lifecycle",
+      payload: {
+        id: "Worker",
+        agent: "Worker",
+        status: "completed",
+        parentToolCallId: "task-1",
+        index: 0,
+      },
+    });
+    expect(lastToolCallStatus(omp, "task-1")).toBe("completed");
+    expect(omp.runningToolCallIds()).toEqual([]);
+  });
+
+  test("force-settles a task that never produced a child when the turn completes", async () => {
+    const omp = new OmpHarness();
+    await omp.start();
+
+    const session = omp;
+    await session.requireStartTurn("no child");
+    const runtime = session.runtime();
+    runtime.beginTurn();
+    runtime.acceptPrompt("no child", "user-orphan");
+    runtime.streamAssistantText("done");
+    runtime.emit({
+      type: "tool_execution_start",
+      toolCallId: "task-orphan",
+      toolName: "task",
+      args: { description: "never spawned" },
+    });
+    runtime.emit({
+      type: "tool_execution_end",
+      toolCallId: "task-orphan",
+      toolName: "task",
+      isError: false,
+      result: { text: "Spawned 0 background agents" },
+    });
+    expect(lastToolCallStatus(session, "task-orphan")).toBe("running");
+
+    runtime.state = { ...runtime.state, isStreaming: false, isCompacting: false };
+    runtime.finishTurn({
+      role: "assistant",
+      content: [{ type: "text", text: "done" }],
+    });
+    await waitForImmediate();
+    await waitForImmediate();
+    expect(session.completedTurnCount()).toBe(1);
+    expect(lastToolCallStatus(session, "task-orphan")).toBe("completed");
+    expect(session.runningToolCallIds()).toEqual([]);
+  });
+
+  test("completes a failed task immediately", async () => {
+    const omp = new OmpHarness();
+    await omp.start();
+
+    await omp.requireStartTurn("task fails");
+    const runtime = omp.runtime();
+    runtime.beginTurn();
+    runtime.emit({
+      type: "tool_execution_start",
+      toolCallId: "task-fail",
+      toolName: "task",
+      args: { description: "boom" },
+    });
+    runtime.emit({
+      type: "tool_execution_end",
+      toolCallId: "task-fail",
+      toolName: "task",
+      isError: true,
+      result: { text: "spawn failed" },
+    });
+    expect(lastToolCallStatus(omp, "task-fail")).toBe("failed");
+    expect(omp.runningToolCallIds()).toEqual([]);
   });
 
   test("does not complete on OMP's extension-notice agent_end", async () => {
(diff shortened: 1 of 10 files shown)


joeshull:
Closing this — on reflection it shouldn't carry #3371's commits. Merging it would have pulled that PR in without its own review, and the gate changes it builds on are @jasonhnd's to land.

The three commits on top are still available at ZGEnergy:omp-idle-budget-upstream if they're useful to anyone: they bound the post-agent_end idle gate (#3654) so a stale or unreachable get_state can't hold a finished turn in running forever, and they cover the subagent wait #3371 adds.

I'll reopen against main once #3371 lands, or happy for those commits to be taken into #3371 directly instead.

joeshull:
Reopening as a draft rather than sitting on this until #3371 lands — @jasonhnd got the notification already, and the findings are more useful to you now than after your PR merges.

Two of them touch #3371 directly:

1. The gate you're adding a third wait to has no deadline at all. `completeTurnAfterProviderIdle` polls `get_state` every 10 ms forever and swallows every error, so a stale or unreachable state path holds a finished turn in `running` until the user cancels. It has been that way since #2067, and #2261/#2282 both explicitly left it alone.

2. `hasRunningOmpSubagents()` swallows a failing `get_subagents` the same way, so an index still holding children it can no longer confirm keeps the parent open indefinitely — the same shape as the objection that closed #2245. The last commit here treats a *current* snapshot listing running children as positive evidence (no budget, since a fan-out has no bounded length) and lets the budget resume once `get_subagents` stops answering.

Also worth knowing: every `agent_end` opens another concurrent gate today, and two that both observe idle emit two `turn_completed` for one turn. Reproduced in the harness.

Take any of it into #3371 if it's useful — I'd rather that than this stack.

joeshull:
Closing in favour of jasonhnd/paseo#1, which puts the same three commits into @jasonhnd's `fix/2232-omp-idle-gate` branch. That keeps #3371 as the single place this gets reviewed instead of asking anyone to review his commits under my PR.

The findings stand on their own if that PR isn't wanted: the post-agent_end gate has had no deadline since #2067, `hasRunningOmpSubagents()` swallows a failing `get_subagents` the same way `get_state` failures were swallowed, and concurrent gates can emit two `turn_completed` for one turn. #3654 has the detail.

PR #3272, opened 2026-08-12
Prevent concurrent draft agent creation

boudra:
### Linked issue

Refs #3217

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Draft creation used React render state as its only in-flight guard. Multiple submit callbacks entering before React committed the first state transition each started a separate agent creation request. The shared create-flow store is updated synchronously before the request, so it is the operation owner that can reject later callbacks in the same render.

This fixes the demonstrated client-side duplicate-submit producer. It does not claim to explain the paired six outline requests in #3217; the hosted relay path does not fan out client frames, so that part remains under investigation.

### Goals

- Permit at most one active create request for a draft.
- Reject a concurrent callback before it sends another daemon RPC.
- Preserve retries after a failed attempt clears the shared state.

### Non-goals

- Change relay routing.
- Claim that this alone closes #3217.
- Add daemon-side idempotency in this change.

### Supersedes

- [PR #2614 — Prevent duplicate draft agent creation](https://github.com/getpaseo/paseo/pull/2614) — This branch carries the same focused client-operation fix on current `main` with failing-first regression coverage.

### QA

- Before the fix: `npx vitest run packages/app/src/composer/draft/create-flow.test.ts --bail=1` timed out because both same-render submissions entered the unresolved create request.
- After the fix: the same focused test file passes, including the new assertion that only one request starts.
- `npm run typecheck` passes.
- `npm run lint` passes.
- `npm run format:files -- packages/app/src/composer/draft/create-flow.ts packages/app/src/composer/draft/create-flow.test.ts` passes.

### Checklist

- [x] One focused change
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence
- [x] Tests added or updated where it made sense

Issue #3201, opened 2026-08-11
bug: draft submissions request a timeline for a nonexistent agent


wilgon456:
### What's broken

While a new agent submission is shown optimistically, the desktop/web stream asks the daemon for a timeline prompt index using the temporary draft message ID. That ID is not an agent, so the daemon logs an `Agent not found` error.

Expected: optimistic draft streams should not request agent timeline data until the real agent and its authoritative history exist.

### Steps to reproduce

1. Enable Chat outline.
2. Open a new project draft in the desktop app.
3. Submit the first prompt.
4. While the optimistic stream is displayed, inspect `~/.paseo/daemon.log`.
5. An `agent.timeline.list_prompts.request` is sent with a `draft_msg_*` ID and rejected.

### Where did this happen

Daemon (request originated from the Desktop/Web app stream)

### Paseo version

Observed on 0.3.0 and 0.3.1.

### OS version

macOS 15.7.7

### Agent provider

Not relevant

### Logs

```text
{"level":50,"daemonVersion":"0.3.1","module":"session","err":{"type":"Error","message":"Agent not found: draft_msg_1786426817104_xprtf1qy9"},"agentId":"draft_msg_1786426817104_xprtf1qy9","msg":"Failed to handle agent.timeline.list_prompts.request"}
```

The same request failure recurred with different `draft_msg_*` IDs.

### Screenshots or video

No visible UI change; the observable failure is the daemon request/error above.

PR #3672, opened 2026-08-22
fix(app): stop draft pending leaking past consume (#3217)

wangdengwu:
Fixes the #3217 class of bug: one "start chat from a new workspace" produced N identical agents (reporters saw N=6 and N=2), with a paired daemon signature of N failed `agent.timeline.list_prompts.request` **and** N `create_agent_request` for one draft id.

## Root cause (two layers)

**1. The trigger — a store removal that stops removing.**
`useWorkspaceDraftSubmissionStore.consumePending` is the *only* cross-instance dedupe in the draft-create path: it removes the consumed pending so any other `WorkspaceDraftAgentTab` mounted for the same draft reads `null` and does not create a second agent (the other three guards — `autoSubmitKeyRef`, the per-hook `isSubmitting` reducer, the create-flow-store lifecycle — are all per-instance). When several draft-tab instances share one draft, each that reads a truthy pending both calls `createAgent` **and** mounts its stream (`AgentStreamView agentId={tabId}`, and `tabId === draftId`) → one `list_prompts`. That is why the two counts are equal.

**2. Why the removal stops removing — a babel-preset-expo miscompile.**
The removal was written as a computed-key destructuring-omit:
```ts
const { [input.draftId]: _removed, ...rest } = state.pendingByDraftId;
```
In the shipped 0.5.0-beta.3 web bundle this compiles to an omit whose excluded-key array is built from an **unassigned** variable:
```js
var _input$draftId;                                              // never assigned → undefined
… _objectWithoutPropertiesLoose(_state$…, [_input$draftId].map(_toPropertyKey))   // removes nothing
```
This is **babel-preset-expo**, not terser: `@babel/preset-env` alone compiles the same input correctly. The break is specific to a computed key that is a **property access** (needs a temp); a plain-identifier key is fine — which is exactly why a sweep of the shipped bundle found only 2 broken omit sites (`consumePending` and a bundled `@react-navigation` `clearOptions`) out of 16. Filed upstream as **expo/expo#49232** (still present on the latest SDK 57 / babel-preset-expo 57.0.7; minimal repro: https://github.com/wangdengwu/babel-preset-expo-omit-repro).

Evidence: [macOS raw logs](https://github.com/getpaseo/paseo/issues/3217#issuecomment-5377325343) · [shipped-bundle miscompile + 16-site sweep](https://github.com/getpaseo/paseo/issues/3217#issuecomment-5377462812).

## Why not the other fixes

- The maintainer already wrote and **closed** server-side idempotency (#3265) and a client synchronous guard (#3272) for want of a repro predicting the full N:N — this PR fixes the actual producer instead of layering another guard.
- Server `clientMessageId` dedupe can't mask it: instances that take the `handleCreateFromInput` branch each mint a fresh `clientMessageId`, so the daemon sees N distinct messages.
- The only config-level toolchain fix (forcing `@babel/plugin-transform-destructuring` ahead of the preset) **force-lowers all destructuring app-wide** — a real bundle/runtime regression on modern web targets — so it isn't viable. `babel.config.js` is intentionally left unchanged.

## What this PR does

1. **Fix** — rewrite `consumePending`'s removal to a minifier-proof `copy + delete`, with a comment so it isn't "simplified" back:
   ```ts
   const rest = { ...state.pendingByDraftId };
   delete rest[input.draftId];
   ```
2. **Regression test** — `workspace-draft-submission-store.test.ts`: consume removes the entry (second consume returns `null`), only the consumed draft is removed, a non-matching consume removes nothing.
3. **Guard** — `no-computed-destructuring-omit.test.ts`: an AST scan of `packages/app/src` that fails if the property-access destructuring-omit shape reappears anywhere, so we can't reintroduce the babel-preset-expo landmine while expo/expo#49232 is open.

## Scope / non-goals

- No protocol change; no new client/server dedupe guard (restores the removal the store already intended).
- No global babel/minifier config change.
- The bundled `@react-navigation` `clearOptions` site (same miscompile, low severity) is left to the upstream fix.
- Out of scope: #3218 (archived-agent process cleanup), #3201 (draft-id `list_prompts` reporting).

## QA

```
npx vitest run packages/app/src/stores/workspace-draft-submission-store.test.ts --bail=1   # 3 passed
npx vitest run packages/app/no-computed-destructuring-omit.test.ts                          # 2 passed
npm run typecheck --workspace=@getpaseo/app                                                 # 0 errors
npm run lint -- packages/app/src/stores/workspace-draft-submission-store.ts packages/app/no-computed-destructuring-omit.test.ts   # clean
```

The source-level regression test passes on `main` too (the source is correct — the bug only appears after the babel transform); the load-bearing protection is the minifier-proof `copy + delete` shape, and the guard keeps the fragile shape out of the tree.

### DIFF
--- a/packages/app/no-computed-destructuring-omit.test.ts
+++ b/packages/app/no-computed-destructuring-omit.test.ts
@@ -1,0 +1,82 @@
+import { readdirSync, readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import ts from "typescript";
+import { describe, expect, it } from "vitest";
+
+// Guard for #3217. babel-preset-expo miscompiles a destructuring-omit whose
+// computed key is a *property/element/call access* (it needs a temp):
+//   const { [obj.key]: _unused, ...rest } = src
+// The preset drops the temp's assignment, so the omit excludes `undefined` and
+// removes nothing. Upstream: https://github.com/expo/expo/issues/49231
+// Until that lands, keep this shape out of our source — use a plain-identifier
+// key, or `const rest = { ...src }; delete rest[obj.key];`.
+
+const SRC_DIR = join(dirname(fileURLToPath(import.meta.url)), "src");
+
+function listSourceFiles(dir: string): string[] {
+  const out: string[] = [];
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      out.push(...listSourceFiles(full));
+    } else if (/\.(ts|tsx)$/.test(entry.name) && !/\.(test|spec)\.(ts|tsx)$/.test(entry.name)) {
+      out.push(full);
+    }
+  }
+  return out;
+}
+
+// A computed key that babel lowers through a temp (the shape that breaks).
+// Plain identifiers and literals are safe.
+function keyNeedsTemp(expr: ts.Expression): boolean {
+  return (
+    ts.isPropertyAccessExpression(expr) ||
+    ts.isElementAccessExpression(expr) ||
+    ts.isCallExpression(expr)
+  );
+}
+
+function findViolations(fileName: string, text: string): number[] {
+  const sf = ts.createSourceFile(fileName, text, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+  const lines: number[] = [];
+  const visit = (node: ts.Node): void => {
+    if (ts.isObjectBindingPattern(node) && node.elements.some((e) => e.dotDotDotToken)) {
+      for (const el of node.elements) {
+        const pn = el.propertyName;
+        if (pn && ts.isComputedPropertyName(pn) && keyNeedsTemp(pn.expression)) {
+          lines.push(sf.getLineAndCharacterOfPosition(el.getStart(sf)).line + 1);
+        }
+      }
+    }
+    ts.forEachChild(node, visit);
+  };
+  visit(sf);
+  return lines;
+}
+
+describe("no computed property-access key in a destructuring-omit (#3217)", () => {
+  it("detects the dangerous shape (self-check)", () => {
+    const bad = `const { [input.draftId]: _removed, ...rest } = state.map;`;
+    expect(findViolations("bad.ts", bad)).toHaveLength(1);
+    const safe = `const { [draftId]: _removed, ...rest } = state.map;`;
+    expect(findViolations("safe.ts", safe)).toHaveLength(0);
+  });
+
+  it("is absent from packages/app/src", () => {
+    const offenders: string[] = [];
+    for (const file of listSourceFiles(SRC_DIR)) {
+      const text = readFileSync(file, "utf8");
+      // cheap prefilter: needs a rest (`...`) and a computed key that contains a
+      // member access (`[ ... . ... ]:`) — the only shape that can trip the bug.
+      if (!text.includes("...") || !/\[[^\]\n]*\.[^\]\n]*\]\s*:/.test(text)) continue;
+      for (const line of findViolations(file, text)) {
+        offenders.push(`${file}:${line}`);
+      }
+    }
+    expect(
+      offenders,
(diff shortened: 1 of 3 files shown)


wangdengwu:
Update — root cause traced to the toolchain, and a guard added.

The miscompile is **not** terser: `babel-preset-expo` (54.0.10) lowers a destructuring-omit whose computed key is a *property access* (`const { [obj.key]: _x, ...rest } = src`) by emitting the key temp's declaration but **dropping its assignment**, so the omit excludes `undefined`. `@babel/preset-env` alone compiles it correctly. A plain-identifier key is fine — which is why the shipped bundle had exactly the two property-access sites broken. Filed upstream: expo/expo#49231.

The only config-level fix (forcing `@babel/plugin-transform-destructuring` ahead of the preset) force-lowers *all* destructuring app-wide — a real bundle/runtime regression on modern web targets — so it isn't viable here. The per-site `copy + delete` rewrite in this PR stays the right mitigation.

Added a commit: an AST guard test that scans `packages/app/src` for the property-access destructuring-omit shape, so it can't be reintroduced while the upstream bug is open.

wangdengwu:
Upstream tracking updated: the babel-preset-expo bug is now filed with a minimal create-expo-app reproduction at expo/expo#49232 (the earlier expo/expo#49231 was auto-closed for lacking a repro repo). Confirmed still present on the latest SDK 57 / babel-preset-expo 57.0.7. Repro: https://github.com/wangdengwu/babel-preset-expo-omit-repro

smellgamed3:
I can reproduce a user-visible variant of this on Windows Desktop `0.5.1` (desktop-managed local daemon `0.5.1`). This looks like additional confirmation that the unconsumed draft path is not only capable of duplicate agent creation: it can also crash the Desktop renderer during Command Center contribution registration.

### Symptom

On every Desktop start (and after pressing the error boundary's Retry), the app fails with:

```text
Duplicate Command Center contribution id:
draft:<redacted-server-id>:draft_msg_<redacted>:models:codex:claude-opus-4-6
```

The stack begins in the Command Center registry's `replace` path and is followed by:

```text
[RootErrorBoundary] Unhandled render error
```

The affected tab was an old persisted `draft` tab in the `workspace-layout-state` Local Storage record. It belonged to a remote host/workspace which is no longer the currently selected local daemon. The current local desktop-managed daemon was healthy; Codex and Claude provider diagnostics were both Ready. The daemon log had no matching error, so this was entirely renderer/client state.

### Persistence / recovery behavior

The failure persisted across app restarts because the bad draft is restored from Chromium Local Storage:

```text
%APPDATA%\Paseo\Local Storage\leveldb
key: _paseo://app\x00\x01workspace-layout-state
```

Deleting all Local Storage lets the app start, but also removes all saved host information and layout state, which is too destructive.

I instead stopped all Paseo processes, took two full filesystem backups of Local Storage, then edited only the persisted `workspace-layout-state` JSON record. I removed just the single tab whose target was:

```json
{
  "kind": "draft",
  "draftId": "draft_msg_<redacted>"
}
```

from that one old workspace's `tabs` and `tabIds`, and moved the pane focus to its remaining terminal tab. All other host/workspace/agent/terminal layout records remained intact. After restarting Desktop, the duplicate-contribution crash no longer occurred.

### Why this may matter for #3672

This is consistent with the pending draft not being consumed: multiple mounted/restored instances appear able to register the same draft/provider/model Command Center contribution. In this case the failure happens before any visible duplicated agent creation, so the renderer becomes unusable and Retry simply repeats the same restore/crash cycle.

Please consider adding a regression test covering a persisted/restored draft tab (including an old remote-host workspace) and duplicate Command Center model contribution registration, in addition to the multi-create behavior already covered here.

Environment:

- Windows 11 (build 26200)
- Paseo Desktop / bundled CLI / desktop-managed daemon: `0.5.1`
- Local daemon: `127.0.0.1:6767`
- Codex CLI: `0.149.1`, Ready
- Claude Code: `2.1.241`, authenticated and Ready

boudra:
Superseded by #4442, which includes the pending-submission removal fix and a regression that executes Expo's emitted production code. The old code returns the same consumed submission twice; the fix passes. The canonical PR also covers duplicate creation at the daemon boundary.

PR #3674, opened 2026-08-22
fix(server): codex startTurn waits out foreground turn teardown instead of refusing

thomasvan:
## Problem

A prompt injected while a codex turn is being cancelled races the teardown. `steerOrReplaceActiveTurn` falls back to replace when steering is unavailable; `cancelAgentRunBefore` returns once the **manager's** run record settles — and a force-cancel settles it without waiting for the provider — so `startTurn` finds `activeForegroundTurnId` still set and throws `A foreground turn is already active`.

Measured in production (0.5.0-beta.2, long-lived codex agent receiving frequent child-finish notifications): the force-cancel log line and the throw landed **4 ms apart**, six times in one session. Each occurrence turns a live notification prompt into a lost message, fails the stream (`turn_failed` → `Agent stream failed`), and leaves the agent in `error` status.

## Fix

Provider-local, no manager changes. `startTurn` waits up to `FOREGROUND_TEARDOWN_WAIT_MS` (10 s) for the previous turn's teardown to clear `activeForegroundTurnId` — the four sites that clear it flush the waiters — and refuses only a turn that genuinely will not end. The refusal semantics for a truly stuck turn are unchanged.

## Tests

Two unit tests beside the existing codex session harness: an in-flight `startTurn` parks while a foreground turn is active and proceeds the moment the teardown clears it; and a teardown that never completes still refuses with the original error after the bounded wait (fake timers).

### DIFF
--- a/packages/server/src/server/agent/providers/codex-app-server-agent.test.ts
+++ b/packages/server/src/server/agent/providers/codex-app-server-agent.test.ts
@@ -674,16 +674,22 @@ describe("Codex foreground teardown wait (replace path)", () => {
     await expect(attempt).rejects.toThrow("Codex client not initialized");
   });
 
-  it("refuses only when the teardown never completes", async () => {
+  it("refuses only when the teardown never completes, retaining no waiter", async () => {
     vi.useFakeTimers();
     try {
       const session = createSession();
+      const internals = castInternals<{ foregroundTurnClearWaiters: Array<() => void> }>(
+        session,
+      );
       const attempt = session.startTurn("never clears");
       const expectation = expect(attempt).rejects.toThrow(
         "A foreground turn is already active",
       );
+      expect(internals.foregroundTurnClearWaiters).toHaveLength(1);
       await vi.advanceTimersByTimeAsync(10_000);
       await expectation;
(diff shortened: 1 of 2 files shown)


thomasvan:
Follow-up commit: a timed-out waiter now removes itself from the list. Previously the timeout resolved the promise but left its callback in `foregroundTurnClearWaiters`, so a turn that never cleared would retain one dead closure per retrying prompt until the session closed. The refusal test now also asserts the list is empty after the bounded wait.

thomasvan:
Folded into #3640 and closed in its favour.

The three PRs all repair the same piece of state — the Codex provider's single foreground-turn slot — and two lines cannot exist in any of them separately: `interrupt()` and `disposeClient()` each release that slot, and each must wake the callers blocked on it via `flushForegroundTurnClearWaiters()`, a method this PR introduces. The other two were cut against a base that lacks it, so in every merge order the result would land slot releases that leave waiters asleep for the full timeout.

#3640 now carries all four fixes as four commits, each green on its own, with tests that fail if either flush call is removed. Nothing from this PR is dropped.

Issue #3589, opened 2026-08-20
bug: Usage panel is empty when a Claude provider sets `CLAUDE_CONFIG_DIR` (multi-account setups)


summerisgood1:
### Summary

The Claude usage panel stays empty for any custom provider that launches `claude` with a non-default `CLAUDE_CONFIG_DIR`. Usage lookup resolves the OAuth token from a hardcoded `~/.claude` (or `$CLAUDE_HOME`) and ignores the per-provider `CLAUDE_CONFIG_DIR`, so it never finds the credential the agent actually authenticated with.

### Environment

- Paseo 0.4.0 (app + daemon, desktop-managed)
- macOS 15 (Darwin 25.3.0), Apple Silicon
- Local loopback connection (`127.0.0.1:6767`), relay disabled
- Multi-account setup via [[claude-swap](https://pypi.org/project/claude-swap/)](https://pypi.org/project/claude-swap/) (`cswap`)

### Repro

1. Define a custom provider that wraps `claude` in a launcher which sets `CLAUDE_CONFIG_DIR` to a per-account directory:

```json
"agents": {
  "providers": {
    "claude-work": {
      "extends": "claude",
      "label": "Claude (Work)",
      "command": ["/Users/me/.local/bin/cswap", "run", "w", "--"]
    }
  }
}
```

2. Run an agent with that provider. The agent works normally.
3. Open the usage panel.

**Expected:** session / weekly usage windows for that account.
**Actual:** empty — no data, no error surfaced in the UI.

### Diagnosis

Usage is fetched from `https://api.anthropic.com/api/oauth/usage` via `fetchProviderApi`, and the token is resolved as:

```js
this.claudeHome = options.claudeHome || process.env["CLAUDE_HOME"] || join(homedir(), ".claude");
// ...
const credPath = join(this.claudeHome, ".credentials.json");
// fallback: macOS keychain service "Claude Code-credentials"
```

Neither branch accounts for `CLAUDE_CONFIG_DIR`:

- **File branch:** the credential lives at `$CLAUDE_CONFIG_DIR/.credentials.json`, not `~/.claude/.credentials.json`. On this machine `~/.claude/.credentials.json` does not exist at all.
[...]

Issue #3228, opened 2026-08-11
bug: provider usage can show a different Claude account when using a replacement command


jjongsta:
## What's broken

I configured Paseo's built-in Claude provider to launch through TeamClaude:

```text
/opt/homebrew/bin/teamcodex run --
```

TeamClaude routes Claude requests through its currently active account, but Paseo's Context window tooltip independently reads the local Claude Code OAuth credentials.

When those identities differ, the tooltip can show `Unavailable` or quota usage belonging to a different Claude account.

This makes the usage indicator unreliable for the agent that is actually running.

Expected behavior:

The tooltip should not silently present another account's usage. It should either show usage for the account used by the configured Claude command or clearly report that routed-account usage is unavailable.

## Steps to reproduce

1. Configure Paseo's built-in Claude provider command as `/opt/homebrew/bin/teamcodex run --`.
2. Select account A as TeamClaude's active account.
3. Keep Claude Code's local OAuth credentials or Keychain session on account B.
4. Start a Claude agent through Paseo.
5. Hover the Context window meter.
6. Compare the displayed account and quota with TeamClaude's active account.

## Where did this happen

Daemon, surfaced through the Context window tooltip.

## Paseo version

0.3.1

## OS version

macOS 26.5.2, Apple Silicon

## Agent provider

Claude Code

## Provider configuration

Claude Code launched through the local TeamClaude replacement command:

```text
/opt/homebrew/bin/teamcodex run --
```

No Anthropic credentials are passed from Paseo to TeamClaude.

## Logs

No crash or daemon error is produced. The observable failure is the identity mismatch between the account routed by TeamClaude and the account queried independently by Paseo.

## Screenshots or video

No account identifiers are attached because they contain private subscription-account information. The mismatch is reproducible on demand using the steps above.

Issue #2877, opened 2026-08-04
bug(usage): team-billed Cursor seats never get planUsage — needs GetTeamSpend fallback


Relaxe111:
### What's broken

On a **team-billed Cursor seat**, Settings → Usage can never show plan usage — not even after the token lookup is fixed. `CursorQuotaProvider` only calls `GetCurrentPeriodUsage`, and that endpoint returns no `planUsage` object for a member of a team subscription.

Verified with a valid token (supplied via `CURSOR_ACCESS_TOKEN` to rule out #2586):

```
POST api2.cursor.sh/aiserver.v1.DashboardService/GetCurrentPeriodUsage
200 {"billingCycleStart":"1785884859130","billingCycleEnd":"1785884859130","displayThreshold":100}
```

No `planUsage`, and `billingCycleStart == billingCycleEnd`. Since `fetchUsage()` builds its only balance from `planUsage.{totalSpend,remaining,limit}`, `balances` stays empty and the card renders as `status: "available"` with nothing in it — an empty row under the provider name.

That endpoint reports *individual usage-based spend against an included budget*. A team seat has no such budget; the spend is tracked per member on the team.

### Where the data actually is

`GetTeams` → `GetTeamSpend` returns it (shapes below, values redacted):

```
POST .../GetTeams
200 {"teams":[{"id":<team-id>,"name":"<team>","role":"TEAM_ROLE_MEMBER","seats":20,
               "requestQuotaPerSeat":1,"subscriptionStatus":"active",...}]}

POST .../GetTeamSpend  {"teamId":<team-id>}
200 {"teamMemberSpend":[{"userId":<id>,"email":"<you@example.com>","role":"TEAM_ROLE_MEMBER",
                         "spendCents":<cents>,"overallSpendCents":<cents>,
                         "monthlyLimitDollars":200,"effectivePerUserLimitDollars":200}],
     "subscriptionCycleStart":"1785542400000","nextCycleStart":"1788220800000",
     "totalMembers":22,"totalPages":1,...}
```

`spendCents` / `effectivePerUserLimitDollars` / `nextCycleStart` map 1:1 onto the existing balance shape (`used` / `limit` / `resetsAt`), so this needs no UI change.

Two access notes found while probing:

[...]

PR #3718, opened 2026-08-22
fix: skip usage caches when Settings refresh is clicked

stackia:
### Linked issue

None — searched [getpaseo/paseo](https://github.com/getpaseo/paseo/issues) issues and discussions for usage refresh / cache / stale plan usage. Closest existing usage issues (#2440, #3228, #3589, #2877, #2403) are about missing or wrong provider numbers, not the Settings Refresh button ignoring live data.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Settings → Usage → Refresh looked like it worked (spinner, button disable) but often showed the same numbers. Two 5-minute caches sat in front of the provider APIs:

1. React Query `staleTime` on the usage query
2. Daemon `ProviderUsageService` TTL (already had `forceRefresh`, but nothing called it)

The old refresh path did `invalidateQueries` then `fetchQuery` with the same 5-minute `staleTime` and the default query function. `invalidateQueries` refetches the cached (no-force) request; the follow-up `fetchQuery` then treats that result as fresh and does not fetch again. The daemon also returned its cache because the RPC never sent `forceRefresh`.

This PR adds an optional `forceRefresh` field on the existing `provider.usage.list.request`, threads it through client and daemon, and makes the Settings Refresh button call `fetchQuery` with `staleTime: 0` and `forceRefresh: true`. Initial load and the context-meter tooltip still use the caches.

Protocol compatibility: `forceRefresh` is optional. An old app omits it and a new daemon keeps caching. A new app sending the field still parses on an old daemon (unknown keys are stripped); that old daemon keeps caching until it is updated.

### Goals

- Clicking Refresh on Settings → Usage fetches live provider usage even within the 5-minute cache window.
- Repeated Refresh clicks each send `forceRefresh: true` and update the same React Query key.
- First load and tooltip open still use the React Query / daemon caches.

### Non-goals

- Changing cache TTLs.
- A new capability flag for force refresh.
- Forcing a daemon refresh when the context-meter tooltip opens.
- Fixing per-provider usage accuracy (Cursor, Claude account mismatch, Z.ai unavailable, etc.).

### QA

**Automated**

- `npm run typecheck` — passed (pre-commit).
- `npm run lint` / format check — passed (pre-commit).
- `npx vitest run packages/app/src/provider-usage/use-provider-usage.test.tsx --bail=1` — 3 passed.
- `npx vitest run packages/protocol/src/messages.test.ts --bail=1` — 23 passed.
- `npx vitest run packages/client/src/daemon-client.test.ts -t "provider.usage.list.request" --bail=1` — 2 passed.
- `npx vitest run packages/server/src/server/session/provider/provider-catalog-session.test.ts packages/server/src/services/quota-fetcher/service.test.ts -t "forceRefresh|usage-list|caches usage" --bail=1` — 4 passed.

**Manual (web, checkout-local daemon on `:6768` + Expo on `:8081`)**

1. Opened Settings → Usage for the local host.
2. Clicked Refresh: Cursor plan usage moved `$183.78 / $20.00` → `$186.58 / $20.00`.
3. Clicked Refresh twice more in succession. WebSocket intercept showed two requests, both `{ forceRefresh: true }`. Cursor then moved to `$187.02 / $20.00`.
4. Daemon logs showed those list requests taking ~600ms–1600ms (live provider fetches, not cache hits).

| iOS             |        | Not tested |
| Android         |        | Not tested |
| Web             | x      | Chrome against checkout-local daemon |
| Desktop macOS   |        | Not tested; same React Query + RPC path |
| Desktop Windows |        | Not tested |
| Desktop Linux   |        | Not tested |

Before (not triggering actual refresh after first click):

https://github.com/user-attachments/assets/a90dc677-2488-4b3c-80af-1b11562d82ec

After (every click triggers a real refresh):

https://github.com/user-attachments/assets/e63bc5c0-aae1-4830-adac-acbaa777ec69

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/provider-usage-settings.spec.ts
+++ b/packages/app/e2e/browser/provider-usage-settings.spec.ts
@@ -109,6 +109,7 @@ test.describe("provider usage settings", () => {
     await usageFixture.waitForRequestCount(2);
 
     expect(usageFixture.requestCount()).toBe(2);
+    expect(usageFixture.forceRefreshFlags()).toEqual([false, true]);
     await expect(page.getByText("64%")).toBeVisible();
   });
 
--- a/packages/app/e2e/support/helpers/provider-usage.ts
+++ b/packages/app/e2e/support/helpers/provider-usage.ts
@@ -10,6 +10,7 @@ interface ProviderUsageFixturePayload {
 export interface ProviderUsageFixture {
   requestCount(): number;
   waitForRequestCount(count: number): Promise<void>;
+  forceRefreshFlags(): boolean[];
 }
 
 type WebSocketMessage = string | Buffer;
@@ -80,6 +81,7 @@ export async function installProviderUsageFixture(
   payloads: ProviderUsageFixturePayload[],
 ): Promise<ProviderUsageFixture> {
   let requests = 0;
+  const forceRefreshFlags: boolean[] = [];
   const waiters: Array<{ count: number; resolve: () => void }> = [];
 
   function notifyWaiters() {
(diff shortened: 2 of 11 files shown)


stackia:
video posted.

PR #3719, opened 2026-08-22
feat(sdk): expose a session's commands and full snapshot on the agent handle

gpambrozio:
### Linked issue

None. Discussion on [discord](https://discord.com/channels/1481169421832814616/1525797757242183771/1541006487130742855)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Two gaps in `PaseoAgentHandle`, both of the same kind: the daemon already knows the answer and the SDK has no way to ask.

**1. The commands a session loaded.**

The daemon has always been able to ask a running session which slash commands and skills it loaded — `list_commands_request`, implemented by claude, codex, opencode, acp, omp, and pi. `DaemonClient.listCommands` exposes it, and the app's composer uses it for the `/` menu.

`PaseoAgentHandle` does not. `createPaseoApi` builds a plain object literal with no route to the underlying `DaemonClient`, and `PluginHandlerContext` is `{ paseo: PaseoApi }`, so SDK and plugin authors get no access to it at all.

The practical consequence: anything built on the SDK can only enumerate skills by scanning directories itself. That misses every command bundled inside the agent binary — on a normal Claude install that's roughly eighteen skills that exist in the session but on no scannable path — and it forces per-provider filesystem knowledge for something the daemon already knows generically.

**2. The rest of the agent snapshot.**

The handle mirrors `workspaceId`, `cwd`, and `status` off the snapshot it last observed, and stops there. `AgentSnapshotPayload` carries nine more fields that callers routinely want and that the handle already holds in memory: `capabilities`, `availableModes`, `pendingPermissions`, `activeTurn`, `lastUsage`, `lastError`, `features`, `runtimeInfo`, and `archivedAt`.

Reading any of them today means `current()` plus a null check plus a field access, at every call site, for a value the handle is already tracking. `agent.status` and `agent.lastError` sitting on opposite sides of that line is arbitrary — they arrive in the same snapshot, from the same subscription, and go stale together.

Both changes surface state the handle already has. Neither adds a round trip that did not exist.

### Goals

- Expose the existing `list_commands` RPC on `PaseoAgentHandle` as `commands(options?)`.
- Expose the nine remaining snapshot fields as readonly properties, using the same getter-over-`current` pattern as `workspaceId`, `cwd`, and `status`.
- Keep both thin — no new protocol messages, no new daemon behavior, no response reshaping.
- Match the surrounding conventions: `PaseoAgentCommandsResult` derives from the protocol payload the way `PaseoProviderModelsResult` does, and `options.requestId` mirrors the other handle methods.
- Document both in the SDK reference table and in `agents.md`.

### Non-goals

- No change to `DaemonClient`, the protocol, or any provider.
- No filtering or normalizing of the command response. `kind` is the provider's own classification and is optional; callers decide what to do with it, and I left the ambiguity documented rather than papering over it.
- No caching or subscription for `commands()`. Each call is one RPC.
- No fetching from the new properties. They read the last observed snapshot and never hit the wire, so `refresh()` and `subscribe()` keep their existing jobs.
- No `PaseoWorkspaceHandle` or `PaseoAgentActions` equivalent — commands and session state belong to a session.

### Design note: what `null` means

Every new property is `T | null`, collapsing "the snapshot is optional here" and "this handle has not observed a snapshot yet" into one value. A handle from `ref()` reads `null` for all of them until `refresh()`, `run()`, `waitForFinish()`, a timeline refetch, or `subscribe()` delivers one.

That matches the three properties already there, and it is the part I would most like a second opinion on. The alternative — preserving `undefined` for "provider did not report it" and reserving `null` for "not observed" — is more precise and less consistent, and consistency won. `current()` still returns the raw snapshot for anyone who needs the distinction. It is documented in `reference.md` and in a doc comment on the interface rather than left for a caller to discover.

`availableModes` and `pendingPermissions` are required arrays on the snapshot, so they return `null` only when nothing has been observed, never `[]`-vs-`null` ambiguously.

### QA

Test-first for both changes. I wrote each test before its implementation and confirmed it failed for the right reason, not because of a typo in the assertions.

`commands()`, before the implementation existed:

```
$ npx vitest run packages/client/src/index.test.ts -t "list the session's own commands"

 ❯ packages/client/src/index.test.ts (12 tests | 1 failed | 11 skipped) 71ms
   × agent handles list the session's own commands through the existing daemon RPC 70ms

 FAIL  packages/client/src/index.test.ts > agent handles list the session's own commands through the existing daemon RPC
TypeError: agent.commands is not a function
 ❯ packages/client/src/index.test.ts:696:33
    694|   const agent = client.agents.ref("agent_sdk");
    695|
    696|   const commandsPromise = agent.commands({ requestId: "agent-commands-…
       |                                 ^
```

The snapshot properties, with the getters reverted and the test left in place:

```
$ npx vitest run packages/client/src/index.test.ts -t "expose the observed snapshot"

 ❯ packages/client/src/index.test.ts (13 tests | 1 failed | 12 skipped) 83ms
   × agent handles expose the observed snapshot through readonly properties 83ms

 FAIL  packages/client/src/index.test.ts > agent handles expose the observed snapshot through readonly properties
AssertionError: expected undefined to be null

- Expected:
null

+ Received:
undefined

 ❯ packages/client/src/index.test.ts:745:30
    743|   const agent = client.agents.ref("agent_sdk");
    744|
    745|   expect(agent.capabilities).toBeNull();
       |                              ^
```

After both:

```
$ npx vitest run packages/client/src/index.test.ts

 Test Files  1 passed (1)
      Tests  13 passed (13)
   Duration  1.94s
```

Both tests drive the real `FakeWebSocket` transport used by the rest of the file.

The commands test asserts the outbound frame is a `list_commands_request` carrying the right `agentId` and `requestId`, feeds back a real `list_commands_response`, and checks the resolved payload preserves both `kind: "skill"` and `kind: "command"` entries unmodified.

The snapshot test asserts all nine properties are `null` on a fresh `ref()` handle, delivers one populated `agent_update` upsert over the socket, and asserts each property reads through to the value the daemon sent. The existing archive test now also asserts `agent.archivedAt` tracks the local mutation `archive()` applies to `current`, so the getter and `current()` cannot drift.

```
$ npm run typecheck
exit=0, 0 errors across all workspaces

$ npm run lint
Finished in 744ms on 3763 files with 177 rules using 10 threads.

$ npm run format:check
Finished in 929ms on 4012 files using 10 threads.
```

No UI is affected, so there are no screenshots. Not tested against a live daemon end-to-end. `commands()` is one line onto an RPC the app already exercises through `use-agent-commands-query`, and the snapshot properties do not touch the wire at all, so I judged the transport-level tests to cover what I actually changed. Say the word if you want a real-daemon run instead.

Two pre-existing failures I ran into and deliberately did not touch, both reproducible on a clean `upstream/main` with my changes stashed:

- `packages/client/src/daemon-client.test.ts > readFile drops an old daemon's over-budget binary chunks and reports the refusal` fails.
- `npm run knip` crashes inside its own expo plugin (`TypeError: specifier.charCodeAt is not a function`) before reaching any source, so I could not check the exported types against the unused-export rule. Still reproduces on a fresh `npm ci`.

I also saw typecheck errors on a fresh checkout that were stale workspace `dist/` output; `npm run build:client && npm run build:plugin && npm run build:server` cleared them. Not a code issue, noting it only because it briefly looked like one.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:
@paseo-bot deliver

PR #3726, opened 2026-08-23
fix(pi): refresh context usage when compaction ends

sanjaygulati13:
## Problem

After compaction (auto or manual) the context-window meter in the UI stays stuck at the stale pre-compaction value (e.g. ~93%) until the next user turn, making compaction look like it failed even though it succeeded.

Root cause: for the pi provider, usage is only snapshotted at turn boundaries (`agent_end` → `completeTurn` → usage poller `completeTurn`). But pi runs auto-compaction **after** `agent_end` (in `_handlePostAgentRun` → `_checkCompaction`), and the `compaction_end` handler only emitted the timeline chip — no usage re-read. Manual `/compact` emits the same `compaction_start`/`compaction_end` events, so it had the same gap.

## Fix

Re-read session stats on `compaction_end` via the existing `PiUsagePoller.completeTurn(turnId)`. One handler covers both auto and manual compaction since both emit the same events. The poller's existing dedupe means no spurious events when stats are unchanged.

Behavior: right after compaction pi reports `contextUsage.tokens=null` (no post-compaction assistant response yet), so the meter shows unknown; on the next assistant turn it shows the true reduced size.

## Test

Added `refreshes context usage when compaction ends so the meter is not stale` in `agent.test.ts`: completes a turn at 93k, shrinks stats to 26k, emits `compaction_start`/`compaction_end`, asserts a second `usage_updated` event with the reduced size (and the timeline chips). Verified the test fails without the fix and passes with it; full pi provider suite green (65 tests).

Repro context: with a 131k window and compaction triggering at ~115k (88%), the pre-compaction snapshot is always ~90%+, so the stale meter exactly matches user reports of "context still shows 90%+ after compaction".

### DIFF
--- a/packages/server/src/server/agent/providers/pi/agent.test.ts
+++ b/packages/server/src/server/agent/providers/pi/agent.test.ts
@@ -1629,6 +1629,41 @@ describe("PiRpcAgentSession", () => {
       usage: { contextWindowUsedTokens: 160 },
     });
   });
+
+  test("refreshes context usage when compaction ends so the meter is not stale", async () => {
+    const { pi, session, events } = await createSession();
+    const fakeSession = pi.latestSession();
+    fakeSession.stats = { contextUsage: { contextWindow: 200_000, tokens: 93_000 } };
+
+    // Turn completes; the last usage snapshot is pre-compaction (93k).
+    const first = await session.startTurn("hello");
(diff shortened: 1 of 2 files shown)


sanjaygulati13:
Closing for now — I want to review this locally before resubmitting. Reopening shortly.

PR #3727, opened 2026-08-23
fix(claude): restore a selectable Opus 5 1M entry alongside the plain id

ANONYMOUSZED-beep:
## Problem

#3000: since 0.2.4 (`f91a984`), Opus 5 at 1M context is unreachable from the model picker, and the remaining `claude-opus-5` entry advertises a window the session does not get.

Claude Code still treats the two ids as different runtimes — the suffixed id is what makes the CLI send the `context-1m-2025-08-07` beta. After the fold, the daemon passes plain `claude-opus-5` to the CLI verbatim (`agent.ts` line ~3302: `base.model = this.config.model`), so picker-started sessions run at ~200k and auto-compact near ~167k while the readout claims 1M. The reporter's compact-boundary measurements (55 compacts at 166k–179k vs 19 at ~1M, split exactly by which id was set) confirm it. The issue also shows a pre-0.2.4 persisted preference for `claude-opus-5[1m]` resolves to nothing selectable.

## Fix

Mirror the Fable 5 compat pattern already in this file:

- `claude-opus-5` stays the canonical entry and gains `aliases: ["claude-opus-5[1m]"]`, so old app preferences resolve to something selectable.
- A new `claude-opus-5[1m]` entry is appended as a `COMPAT(claudeOpus5OneMillionId)` catalog entry with `isSelectable: false` (same shape as the Fable 5 compat entry), so the suffixed spelling keeps naming its own runtime and resolves to a real entry with a correct 1M window.
- The manifest id is passed to the CLI untouched by the daemon, so selecting/normalizing to `claude-opus-5[1m]` restores the actual 1M runtime.

## Tests

Updated the three affected suites to encode the restored behavior:

- `models.test.ts`: Opus 5 catalog block now expects the canonical entry + non-selectable `[1m]` compat entry; `findClaudeModel("claude-opus-5[1m]")?.id` now resolves to `"claude-opus-5[1m]"`.
- `agent.test.ts`: catalog id list includes `claude-opus-5[1m]` after `claude-opus-5`; asserts its `isSelectable === false`.
- `cli/tests/15-provider.test.ts`: `EXPECTED_CLAUDE_MODELS` includes the new entry.

## QA evidence

I could not run the full suite locally (disk-space-constrained Windows VM; the repo's node_modules install hit ENOSPC), so I did not fabricate test output. What I verified locally against the source:

- Traced the exact failing path: `model-manifest.ts` manifest entries → `findClaudeModel(this.config.model)?.contextWindowMaxTokens` in `agent.ts` (lines 2104, 2417) drives the context readout/right-sizing → `base.model = this.config.model` (line 3302) passes the id verbatim to the Claude Code CLI.
- Confirmed the diff is symmetric with the existing, tested Fable 5 compat handling (`aliases` + `isSelectable: false` compat entry) and that every touched assertion matches that pattern.
- The branch pushes cleanly and CI on this PR will exercise the server + CLI test suites; happy to paste green runs or adjust per review.

## Platforms

Code paths are platform-independent (TypeScript server + CLI). Not manually QA'd in the macOS app (no macOS machine available); the change only adds catalog entries and does not alter UI layout.

### DIFF
--- a/packages/cli/tests/15-provider.test.ts
+++ b/packages/cli/tests/15-provider.test.ts
@@ -58,6 +58,11 @@ const EXPECTED_CLAUDE_MODELS = [
     descriptionFragment: "Latest release",
   },
   {
+    id: "claude-opus-5[1m]",
+    model: "Opus 5 1M",
+    descriptionFragment: "1M context window",
+  },
+  {
     id: "claude-fable-5",
     model: "Fable 5",
     descriptionFragment: "Most powerful",
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -419,6 +419,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
 
       expect(models.map((m) => m.id)).toEqual([
         "claude-opus-5",
+        "claude-opus-5[1m]",
         "claude-fable-5",
         "claude-fable-5[1m]",
         "claude-opus-4-8[1m]",
@@ -434,6 +435,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
         "claude-haiku-4-5",
       ]);
       expect(models.find((model) => model.id === "claude-fable-5[1m]")?.isSelectable).toBe(false);
+      expect(models.find((model) => model.id === "claude-opus-5[1m]")?.isSelectable).toBe(false);
 
       for (const model of models) {
         expect(model.provider).toBe("claude");
--- a/packages/server/src/server/agent/providers/claude/model-manifest.ts
+++ b/packages/server/src/server/agent/providers/claude/model-manifest.ts
@@ -36,9 +36,22 @@ export const CLAUDE_ULTRACODE_THINKING_OPTION_ID = "ultracode";
 export const CLAUDE_MODEL_MANIFEST = [
   {
     id: "claude-opus-5",
+    // COMPAT(claudeOpus5OneMillionId): the suffixed spelling selects the 1M-context runtime in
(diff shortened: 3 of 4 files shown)


ANONYMOUSZED-beep:
## QA evidence (local verification)

Full `npm install` hits ENOSPC on my disk-constrained VM, so instead of fabricating suite output I ran the patched manifest/catalog logic directly under `tsx` (no build step, real source files). All 13 assertions pass:

```
$ npx tsx verify.ts
PASS catalog contains claude-opus-5: got true
PASS catalog contains claude-opus-5[1m]: got true
PASS [1m] entry is non-selectable compat: got false
PASS [1m] entry window is 1M: got 1000000
PASS plain entry has [1m] alias: got ["claude-opus-5[1m]"]
PASS default is plain opus-5: got "claude-opus-5"
PASS findClaudeModel([1m]) id: got "claude-opus-5[1m]"
PASS findClaudeModel(dated): got "claude-opus-5"
PASS findClaudeModel(dated[1m]) id: got "claude-opus-5[1m]"
PASS normalize runtime [1m]: got "claude-opus-5[1m]"
PASS normalize runtime plain: got "claude-opus-5"
PASS disabled thinking on [1m]: got true
PASS manifest normalize [1m]: got "claude-opus-5[1m]"

ALL CHECKS PASSED
```

These exercise the same functions the updated vitest suites assert on (`getClaudeModels`, `findClaudeModel`, `normalizeClaudeRuntimeModelId`, `resolveClaudeDisabledThinkingForModel`), so the catalog-level behavior is verified; the vitest runs themselves will need CI or a machine with space for node_modules.

boudra:
Closing in favor of the narrower implementation in [#3897](https://github.com/getpaseo/paseo/pull/3897), which is still open and unmerged.

PR #3730, opened 2026-08-23
fix(server): surface exact filename matches hidden by discovery filters

### LINKED ISSUE #3694: bug: File search does not return the exact filename match
### What's broken

When searching for a file by its exact name, Paseo does not show the matching file. Instead, it returns unrelated files, even though the requested file exists in the project.

### Steps to reproduce

Open a project in Paseo.
Open the file search.
Enter the exact name of an existing file.
Check the search results.

### Where did this happen

Desktop (Electron)

### Paseo version

0.5.0-beta.4

### OS version

Linux Mint 22.3

### Agent provider

None

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

Image

ANONYMOUSZED-beep:
## Problem

#3694: typing a file's exact name in the file search returned unrelated files even though the file exists.

## Root cause

`searchDirectoryEntries` documents a two-mode philosophy in its own comment block: discovery (ranked candidates the caller has not named) applies gitignore/hidden filtering, while retrieval of an exactly-named path applies no filtering. That retrieval rule was implemented only in `findExactEntry`, which runs for full-path queries (`browseExactPath` or suffix-mode path queries). A plain filename query like `user-profile-card.tsx` — no slashes — falls into fuzzy **discovery**, so if the target is gitignored, inside a non-whitelisted hidden directory, or simply past the scan budget, it never appears and unrelated matches fill the list.

Verified with a standalone repro against `main` before patching:

```
git repo, .gitignore = "generated/", generated/user-profile-card.tsx exists
query "user-profile-card.tsx", matchMode fuzzy, respectGitIgnore true
  -> results: []  found target: false   (repro'd on unpatched main)
```

## Fix

Extend the existing retrieval rule to exact filename queries:

- New `isExactFileNameQuery`: single basename with a real extension and no separators.
- New `findEntryByExactName`: walks the tree matching basename equality, with containment as the only check — deliberately no gitignore/hidden filtering, same as `findExactEntry` — and pins the hit to the top of the results.
- Partial queries (`user-profile`) keep discovery filtering untouched; a query naming a nonexistent file returns empty as before.

## Tests

Added three vitest cases to `directory-suggestions.test.ts`:

1. gitignored exact filename is surfaced
2. partial query still keeps discovery filtering (gitignored stays hidden)
3. exact name that doesn't exist returns empty

## QA evidence

Could not run the repo's vitest suite locally (ENOSPC disk), so I verified by executing the patched source directly under `tsx` against temp git repos:

```
PASS gitignored exact filename surfaces
PASS partial query keeps discovery filtering
PASS missing exact name returns empty
ALL CHECKS PASSED
```

These exercise the exact scenarios of the three added vitest cases. Full suite needs CI or a machine with space for node_modules. Platforms: server-side logic, platform-independent; not QA'd in the desktop app UI (no Linux/macOS machine here).

Fixes #3694

### DIFF
--- a/packages/cli/tests/15-provider.test.ts
+++ b/packages/cli/tests/15-provider.test.ts
@@ -58,6 +58,11 @@ const EXPECTED_CLAUDE_MODELS = [
     descriptionFragment: "Latest release",
   },
   {
+    id: "claude-opus-5[1m]",
+    model: "Opus 5 1M",
+    descriptionFragment: "1M context window",
+  },
+  {
     id: "claude-fable-5",
     model: "Fable 5",
     descriptionFragment: "Most powerful",
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -419,6 +419,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
 
       expect(models.map((m) => m.id)).toEqual([
         "claude-opus-5",
+        "claude-opus-5[1m]",
         "claude-fable-5",
         "claude-fable-5[1m]",
         "claude-opus-4-8[1m]",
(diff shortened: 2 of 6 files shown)


boudra:
Closing for now: this needs a failing regression and desktop before/after evidence for the extensionless search shown in [#3694](https://github.com/getpaseo/paseo/issues/3694). Please reopen with that evidence and search timings on the affected project.

boudra:

PR #3731, opened 2026-08-23
fix(claude): match transcript project folders case-insensitively on Windows

### LINKED ISSUE #3539: bug: Windows: existing Claude Code sessions not found when the transcript folder's drive-letter case differs from paseo's canonicalized project path (case-sensitive matching)
### What's broken

On Windows, paseo canonicalizes a project path's drive letter to uppercase (typing `d:\foo` yields a project dir of `D:\foo`), but it matches Claude Code transcript folders `~/.claude/projects/<encoded-cwd>` using a case-sensitive string comparison.

Claude Code encodes the cwd verbatim, and the drive-letter case depends on the launching tool: VS Code opens folders with a lowercase drive letter `d--foo`, `c--foo`, while `%USERPROFILE%`-derived paths are uppercase `C--Users-...`. Whenever the stored folder's drive case ≠ paseo's uppercase, the project shows no history and no import affordance.

Secondary: the global "Import sessions" dialog only lists the most recent N sessions, with no way to see/search older ones.

#### Repro

1. In VS Code, open a folder on `D:` and run Claude Code → history stored under `~/.claude/projects/d--...` (lowercase drive).
2. Add that directory as a project in paseo → paseo normalizes it to `D:\....`
3. The project shows no existing conversations and no per-project import button, though `~/.claude/projects/d--...` exists.

#### Expected
Match Claude project folders case-insensitively on Windows (drive letters are case-insensitive), or normalize the drive-letter case consistently on both sides.

#### Workaround

Rename affected folders under `~/.claude/projects` so the drive-letter case is uppercase (e.g. `d--*`→`D--*`, `c--*`→`C--*`). NTFS is case-insensitive so Claude keeps working; paseo's case-sensitive match then succeeds.

### Where did this happen

Daemon

### Paseo version

0.4.0

### OS version

Windows 11

### Agent provider

Claude Code

### Provider configuration

Claude Code 2.1.233 in VS Code with Anthropic subscription

### Logs

```text
no logs associated with this issue
```

### Screenshots or video

_No response_

ANONYMOUSZED-beep:
## Problem

#3539: on Windows, existing Claude Code sessions show as missing whenever the encoded `~/.claude/projects/<dir>` folder's drive-letter case differs from paseo's uppercase-canonicalized project path. Claude Code encodes the cwd verbatim, so VS Code-launched projects end up as `d--foo` while paseo computes `D--foo`.

## Fix

`project-dir.ts` gains a Windows-only resolution fallback:

- Exact match → returned immediately, no scan (unchanged behavior).
- On `win32`, when the exactly-encoded folder doesn't exist but the projects root does, scan the root once for a case-insensitive basename match and resolve to it.
- No match → the canonical encoding is returned unchanged.
- Non-Windows platforms are completely untouched (the fallback short-circuits).

Case-insensitive matching is unambiguous on Windows's default case-insensitive filesystems, and it only runs on a miss, so there is no added cost in the common path.

Both the async (`claudeProjectDir`) and sync (`claudeProjectDirSync`) helpers share the same resolution, so history lookup (`resolveHistoryPath`) and importable-session listing stay consistent.

## Tests

Added two vitest cases to `project-dir.test.ts` (Windows-gated):

1. A transcript folder stored with lowercase drive encoding resolves when paseo asks with uppercase — including via the async helper (parity between both helpers preserved).
2. When no case-insensitive match exists, the canonical encoding is returned unchanged.

## QA evidence

Ran the patched source directly under `tsx` on this Windows machine (so the win32 fallback path actually executes):

```
PASS case-insensitive match found
  resolved: ...\projects\D--foo-bar   (folder stored as d--foo-bar)
PASS exact match preserved            (no scan needed when exact exists)
PASS missing folder returns canonical encoding
ALL CHECKS PASSED
```

These mirror the two added vitest cases. Full suite needs CI (local disk can't fit node_modules). Platforms: verified on Windows 11; macOS/Linux paths are provably untouched since the fallback is behind a `process.platform === "win32"` guard.

Fixes #3539

### DIFF
--- a/packages/cli/tests/15-provider.test.ts
+++ b/packages/cli/tests/15-provider.test.ts
@@ -58,6 +58,11 @@ const EXPECTED_CLAUDE_MODELS = [
     descriptionFragment: "Latest release",
   },
   {
+    id: "claude-opus-5[1m]",
+    model: "Opus 5 1M",
+    descriptionFragment: "1M context window",
+  },
+  {
     id: "claude-fable-5",
     model: "Fable 5",
     descriptionFragment: "Most powerful",
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -419,6 +419,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
 
       expect(models.map((m) => m.id)).toEqual([
         "claude-opus-5",
+        "claude-opus-5[1m]",
         "claude-fable-5",
         "claude-fable-5[1m]",
         "claude-opus-4-8[1m]",
@@ -434,6 +435,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
         "claude-haiku-4-5",
       ]);
       expect(models.find((model) => model.id === "claude-fable-5[1m]")?.isSelectable).toBe(false);
+      expect(models.find((model) => model.id === "claude-opus-5[1m]")?.isSelectable).toBe(false);
 
       for (const model of models) {
         expect(model.provider).toBe("claude");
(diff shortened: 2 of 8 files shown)


boudra:
Closing for now: this needs a Windows regression that fails before the fix and evidence that the missing Claude sessions can actually be discovered and imported afterward. Please reopen with that before/after evidence for [#3539](https://github.com/getpaseo/paseo/issues/3539).

boudra:

PR #3741, opened 2026-08-23
feat(provider-usage): add Antigravity plan quota fetcher

magicbear:
## Summary

Adds a provider-usage fetcher for Google Antigravity, so Paseo can show your Antigravity plan meter.

**Provider usage: Antigravity plan quota fetcher** (`packages/server/src/services/quota-fetcher/providers/antigravity.ts`)
Google Antigravity exposes no official ACP mode, and the Gemini CLI path for quota is dead (its OAuth client now returns `PERMISSION_DENIED` on the Code Assist quota endpoints). Instead, the fetcher reads the plan meter straight from the **running Antigravity language server's `GetUserStatus` RPC over 127.0.0.1** — the same local technique the Antigravity Quota Monitor extensions use. No OAuth, no Google API.

- Discovers the language server via `ps` (matches `language_server … --app_data_dir antigravity --csrf_token …`), finds its API port with `lsof`/probes `GetUnleashData`, then queries `GetUserStatus`.
- Reports: `planLabel`/tier, a `Daily request quota` window driven by the most-constrained model, monthly prompt credits + flow credits balances, and a per-model breakdown.
- Falls back to `unavailable` when the language server isn't running.
- Registered in `manifest.ts` for provider id `antigravity`. Unit tests cover parsing, mapping, port discovery, and failure paths (7 tests).

## Verified

- `paseo provider diagnostic antigravity` → Ready (14 models)
- End-to-end `paseo run --provider antigravity` (tools + text + usage)
- Live `GetUserStatus` read on a Pro plan (50000 monthly credits, per-model daily fractions)
- Unit tests pass: `npx vitest run src/services/quota-fetcher/providers/antigravity.test.ts` (7/7)

## Notes

- A full monorepo typecheck locally fails only on pre-existing unbuilt workspace packages (`@getpaseo/client`, `@getpaseo/plugin`, …) in a fresh checkout; the changed files typecheck clean. CI is authoritative.
- The Antigravity plan meter only works while `agy` or the Antigravity IDE is running and logged in.

### DIFF
--- a/packages/server/src/services/quota-fetcher/manifest.ts
+++ b/packages/server/src/services/quota-fetcher/manifest.ts
@@ -3,6 +3,7 @@ import type {
   ProviderUsageFetcherFactoryOptions,
   ProviderUsageFetcherManifestEntry,
 } from "./provider.js";
+import { AntigravityQuotaProvider } from "./providers/antigravity.js";
 import { ClaudeQuotaProvider } from "./providers/claude.js";
 import { CodexQuotaProvider } from "./providers/codex.js";
 import { CopilotQuotaProvider } from "./providers/copilot.js";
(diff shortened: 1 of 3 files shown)


boudra:
I prefer this quota integration as a plugin with its own panel and vendor API handler. Those APIs are on main for the upcoming 0.8 release: [plugin documentation](https://paseo.sh/docs/plugins/v0.8/reference). Closing the core PR in favor of that approach.

Issue #2332, opened 2026-07-22
Codex resume loses active turn state and full-access policy


hemaaanth:
## Problem

Resuming a persisted Codex session while an autonomous goal turn is active can leave Paseo showing the agent as idle even though Codex is still running. The session may then ask for command approvals despite the saved mode being Full access, and Stop can fail because Paseo never learned the active native turn ID.

## Reproduction

1. Start a persistent Codex agent in Full access mode with goals enabled.
2. Let a goal start an autonomous turn.
3. Reload or otherwise resume the persisted Paseo agent while that turn is active.
4. Send a follow-up prompt such as `continue`.
5. Try to stop the run or wait for a command that normally needs no approval in Full access mode.

Observed:

- The resumed turn can be missing from Paseo's managed run state.
- Codex can use its default `on-request` / `workspace-write` policy instead of the saved Full access policy.
- Stop can report that active-run cancellation was not acknowledged.

Expected:

- Resume-time turn events are applied before the resumed agent is exposed to callers.
- The persisted approval, sandbox, model, and feature settings are supplied to `thread/resume`.
- The active native turn remains interruptible after a continuation prompt.

## Diagnosis

`thread/resume` can emit `turn/started` before `AgentManager` subscribes to the provider session. Those events were dropped, and the resume request did not include the persisted execution policy. A follow-up prompt could then steer into the still-active native turn without emitting another `turn/started`, leaving Paseo without the ID required for a safe interrupt.

jeezrick:
Confirming this on paseo 0.3.1 (daemon) + codex-cli 0.147.0, Ubuntu 24.04 (headless server), Codex provider via app-server. Reproduced twice with full evidence from the rollout's `turn_context` records.

**Reproduction via daemon restart (deterministic on my setup):**

Same agent, same thread (`~/.codex/sessions/.../rollout-*.jsonl`):

- 2026-08-12T19:32 turn_context: `approval_policy: never`, `sandbox_policy: danger-full-access` (correct — agent mode is full-access)
- paseo daemon restart
- 2026-08-12T19:40 turn_context: `approval_policy: on-request`, `sandbox_policy: workspace-write` (defaults, mode lost)
- Re-applied via UI mode selector → `thread_settings_applied` with `approval_policy: never` at 2026-08-13T05:59; turns at 07:21/07:22 back to `never`
- paseo daemon restart again at ~07:47
- 2026-08-13T07:50 turn_context: `on-request` + `workspace-write` again

[...]

PR #3742, opened 2026-08-23
fix(server): recover already-settled interrupts

boudra:
### Linked issue

Related to #2332

### Type of change

- [x] Bug fix
- [x] Docs

### Reasoning

Provider lifecycle state can drift from the run tracked by Paseo. When the provider authoritatively reports that no active turn exists, keeping the captured run pending leaves the agent impossible to stop or replace.

Interruption is now idempotent at the agent-session boundary for already-settled work. The manager retires only the captured run token after an acknowledged interruption, while ambiguous provider failures still preserve the run and refuse replacement.

### Goals

- Let users recover an agent after its provider-side turn has already settled.
- Make interruption safe to retry when no provider turn exists.
- Prevent a late canceled start failure from reviving stale lifecycle state.
- Preserve split-brain protection for ambiguous cancellation failures.

### Non-goals

- Change provider transport error handling beyond authoritative already-idle responses.
- Relax replacement safeguards when cancellation cannot be confirmed.
- Change any client UI.

### QA

- `npx vitest run packages/server/src/server/agent/providers/codex-app-server-agent.test.ts --bail=1` - 143 passed.
- `npx vitest run packages/server/src/server/agent/providers/codex-app-server-agent.process-exit.test.ts --bail=1` - 10 passed.
- Focused manager cancellation regressions - 3 passed.
- `npm run typecheck` - passed.
- `npm run lint` - passed with 0 warnings and 0 errors.
- `npm run format:check` - passed.
- `git diff --check` - passed.
- Platforms tested: server unit and process integration coverage on macOS. No UI platforms affected.

### Checklist

- [x] `npm run lint` passes
- [x] UI changes include screenshots or video for every affected platform (not applicable; server-only change)

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -35,7 +35,16 @@ be in flight.
 
 ### Cancellation
 
-Cancellation changes lifecycle state only after the provider acknowledges the interrupt or emits a terminal turn event. If the interrupt is rejected or times out, the agent remains `running` with its active foreground turn intact. Follow-up actions such as replacement, reload, rewind, and Stop must report that failure instead of accepting work they cannot perform. Synthesizing a local cancellation without provider acknowledgment creates a split-brain session: Paseo accepts a new prompt while the provider still owns the previous foreground turn.
+Provider interruption is idempotent at the `AgentSession` boundary. It resolves when the prior
+foreground turn can no longer run, including when the provider reports that it is already idle. It
+rejects only when the provider may still own the turn. Provider adapters translate native errors
+into that contract; lifecycle callers do not interpret provider-specific errors.
+
+After an acknowledged interrupt, the manager settles the captured run even when no terminal event
+arrives or the run was still waiting for its provider turn id. The captured run token prevents an
+older cancellation from settling a newer turn. If interruption is rejected or times out, the agent
+keeps its active foreground turn and replacement, reload, rewind, and Stop report the failure.
+Accepting new work after an ambiguous interruption would create a split-brain session.
 
 ## Relationships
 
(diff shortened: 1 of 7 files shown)


boudra:
### Real-provider browser QA

Ran against this branch with the existing isolated Playwright harness, a real Codex app-server (`codex-cli 0.147.0`), and Chromium. Each run used a random daemon port and a temporary Paseo home; the harness blocked developer-daemon ports.

- `out-of-band-command.codex.real.spec.ts` - provider command with no native turn settled the submitted row, removed working state, and left the composer usable: **1 passed**.
- `permission-steer.real.spec.ts --grep "codex keeps the rejected plan"` - replacement while Codex awaited plan approval rejected the pending plan and completed the replacement response: **1 passed**.
- Temporary diagnostic case using the same helpers - immediately interrupted a cold first Codex turn, verified all visible agent surfaces returned idle, then sent a follow-up through the same session and received the exact sentinel: **1 passed**. The temporary case was removed after QA; no diagnostic code remains in the branch.

All three runs used `E2E_RECORD_VIDEO=1`. No failures, browser errors, leaked projects, or daemon cleanup failures were reported.

boudra:
### CI follow-up

The first Ubuntu server-test run exposed a branch-caused startup race. During asynchronous Codex setup, interruption could acknowledge an apparently idle session before `turn/start` was issued.

The updated head reserves foreground-start ownership before the first await. A preflight interruption now prevents `turn/start`; an already-issued start must become identifiable and interruptible before cancellation can be acknowledged. The existing rewind assertion now waits for the public run-start contract instead of relying on task scheduling.

Verification on updated head:

- New delayed cold-start regression: red before the fix, green after it.
- `codex-app-server-agent.test.ts` - **144 passed**.
- `codex-app-server-agent.process-exit.test.ts` - **10 passed**.
- `agent-manager.test.ts` - **168 passed**.
- `npm run typecheck` - passed.
- Targeted lint and formatting checks - passed.
- `git diff --check` - passed.

PR #3760, opened 2026-08-23
fix(app): provide NewTabLauncher to the non-compact native fallback tab row

### LINKED ISSUE #3750: Android 0.5.0: crash on tablets — "Invariant failed" (`NewTabLauncherProvider is required`)
# Android 0.5.0: crash on tablets — "Invariant failed" (`NewTabLauncherProvider is required`)

## Summary

On the 0.5.0 Android app, a **large-screen device** hits the "Paseo ran into a problem" error
screen and the app is unusable. **Retry** does not recover it.

The same 0.5.0 APK on a **phone** works, and the daemon's **web UI works at the same window size
driving the same path** (side panel open, new-tab chooser). So this is not the shared layout logic
— it looks specific to **non-compact layout on native**.

The release build strips `tiny-invariant` messages, so the user only sees `Invariant failed`. The
real assertion is **`NewTabLauncherProvider is required`** in `useWorkspaceTabLaunchCatalog`.

## Environment

| | |
|---|---|
| Paseo daemon | `0.5.0` (`@getpaseo/cli` + `@getpaseo/server` from npm, self-hosted in Docker) |
| Android app | `0.5.0` |
| **Fails on** | **Samsung Galaxy Tab S9, Android 16** |
| Works on | Android phone, same app version, same server |
| Works on | Daemon web UI, same window size, same path exercised (side panel + new-tab chooser) |

## Steps to reproduce

1. Run a `0.5.0` daemon.
2. Connect the `0.5.0` Android app from a Galaxy Tab S9 (non-compact / large breakpoint).
3. Open the side panel and its new-tab chooser.
4. Error screen appears; **Retry** does not clear it.

## Stack trace

```
Message: Invariant failed

Stack:
Error: Invariant failed
    at invariant (address at index.android.bundle...)
    at useWorkspaceTabLaunchCatalog (address at ...)
    at WorkspaceNewTabMenuContent (address at ...)
    at renderWithHooks (address at index.android.bundle...)
    at updateFunctionComponent (address at ...)
    at beginWork (address at index.android.bundle...)
    at performUnitOfWork (address at ...)
    at workLoopSync (address at index.android.bundle...)
    at renderRootSync (address at ...)
    at performWorkOnRoot (address at ...)
    at performSyncWorkOnRoot (address at ...)
    at flushSyncWorkAcrossRoots_impl (address at ...)
```

## The failing assertion

```js
useWorkspaceTabLaunchCatalog = function (e) {
  const { serverId, purpose } = e;
  ...
  const _ = useContext(P);                       // NewTabLauncher context
  invariant(_, "NewTabLauncherProvider is required");
```

Note the assert is a **truthiness** check, so a `NewTabLauncherProvider` mounted with an
`undefined` `value` fails identically to a missing provider. The provider is applied by a thin
wrapper, so any path reaching the chooser without it (or without a launcher wired for that pane)
crashes:

```js
function Mt({ launcher, content }) {
  return jsx(NewTabLauncherProvider, { value: launcher, children: <View>{content}</View> });
}
```

## Suspected cause: two different notions of "large layout" that disagree on native

There appear to be two independent predicates:

```js
useIsCompactFormFactor = function () {
  const { rt } = useUnistyles();
  return rt.breakpoint === "xs" || rt.breakpoint === "sm";   // purely breakpoint-driven
};

supportsDesktopPaneSplits = function () {
  return isWeb;                                              // purely platform-driven
};
```

and the consumer side is gated on the first:

```js
const { isSidePanel, serverId, tabId } = usePaneContext();
const purpose = isSidePanel ? "supporting" : "primary";
const catalog = useWorkspaceTabLaunchCatalog({ serverId, purpose });
// ...
isSidePanelOpen({ isCompact, workspaceKey })
```

On a Galaxy Tab S9 these two disagree: the breakpoint is large so `isCompact` is **false** and the
non-compact side-panel paths engage, but `supportsDesktopPaneSplits()` is **false** because the
platform is native — so the pane-split scaffolding that mounts `NewTabLauncherProvider` on web is
not there. That combination exists only on a large-screen native device, which matches all three
observations exactly:

| | `isCompact` | `supportsDesktopPaneSplits` | result |
|---|---|---|---|
| Web, wide window | false | **true** | works |
| Android phone | **true** | false | never reaches the path — works |
| Android tablet | false | **false** | **crash** |

In the web bundle the asymmetry is visible in the usage counts too: `useIsCompactFormFactor`
appears **71** times, `supportsDesktopPaneSplits` only **8**. If the side-panel / new-tab surfaces
are gated on the breakpoint predicate while only a handful of places consult the platform one,
large-screen native is exactly the configuration that falls between them.

## Suggested directions

- Make the side-panel / multi-pane new-tab surfaces depend on the *same* predicate that decides
  whether pane-split scaffolding is mounted, rather than on the breakpoint alone.
- Or mount `NewTabLauncherProvider` unconditionally for any pane that can render
  `WorkspaceNewTabMenuContent`.
- A defensive fallback in `useWorkspaceTabLaunchCatalog` (empty catalog instead of a hard invariant)
  would turn a full app crash into a degraded menu.

## Side note

Because `tiny-invariant` drops messages in release builds, this surfaces as a bare
`Invariant failed`. The error screen asks the user to "include the details below when you report
it", but the details contain nothing actionable. Keeping the message for provider/context
assertions in release builds would make these reports much cheaper to triage.

---

*Analysis note:* the code excerpts above are from the **web** bundle shipped inside
`@getpaseo/server@0.5.0` (`dist/server/web-ui/_expo/static/js/web/index-*.js`), since the Android
bundle wasn't available to inspect. The symbol names match the Android stack trace exactly, but the
platform-specific mechanism above is a hypothesis derived from the web build, not something
confirmed against `index.android.bundle`.

jeid64:
Fixes #3750.

## Problem

On the 0.5.0/0.5.1 Android app, every non-compact layout (tablets, foldables in landscape) crashes on workspace mount with `Invariant failed` — the stripped message is `NewTabLauncherProvider is required` from `useWorkspaceTabLaunchCatalog`. Retry cannot recover because the crash is in the mounted layout itself. Phones in portrait and the web UI at the same width are unaffected, matching the report matrix in #3750.

## Mechanism (raw evidence)

- `supportsDesktopPaneSplits()` is `isWeb` (`packages/app/src/constants/layout.ts:51-53`), so `shouldRenderDesktopPaneFallback = !isMobile && !canRenderDesktopPaneSplits` (`workspace-screen.tsx:3358-3361`) is true on **every non-compact native** device.
- On that path, `WorkspaceScreenContent` renders `WorkspaceDesktopTabsRow` directly in the center column, **above** `<View style={styles.centerContent}>{workspacePanelContent}</View>` (`workspace-screen.tsx:~3857`). `WorkspacePanelContent` is the component that provides `NewTabLauncherProvider` (`workspace-screen.tsx:1286-1298`), so the fallback row sits outside it.
- The row mounts `WorkspaceNewTabMenuContent` eagerly inside its `DropdownMenu` (`workspace-desktop-tabs-row.tsx:259`), and the first hook call is `useWorkspaceTabLaunchCatalog`, whose first statement is `invariant(launcher, "NewTabLauncherProvider is required")` (`workspace-tabs/launcher/index.tsx:86`). Missing context → crash at mount, before any user interaction.
- The splits path (web) renders tab rows inside `WorkspacePanelContent` via the split container, and compact renders `MobileWorkspaceTabSwitcher` instead of the row — which is why only non-compact native crashes.

## Fix

Wrap the fallback `WorkspaceDesktopTabsRow` in `NewTabLauncherProvider` with the same `newTabLauncher` value the panel content already receives. One focused change; no behavior change on web or compact.

## Verification

- `npm run typecheck -w @getpaseo/app` (tsgo): clean
- `npx vitest run --project unit` in `packages/app`: 553 files, 4609 tests, all pass
- `oxfmt --check` / `oxlint` on the changed file: clean

## Platforms tested

- Code-path verification + full unit suite on Linux (details above).
- **Not run on Android tablet hardware** — I don't have the failing device class; the fallback branch is gated on `!isWeb` so it is unreachable from the browser e2e project. The fix makes the fallback row's context identical to what the splits path already provides, and the reporter's matrix (#3750: Tab S9 and Z Fold landscape fail, phone portrait and web work) matches the traced gating exactly. Happy to iterate if a maintainer can exercise a debug APK on a wide device.

_Investigated with an agent; the file:line trace above is the raw evidence. Maintainer edits enabled._

### DIFF
--- a/packages/app/src/screens/workspace/workspace-screen.tsx
+++ b/packages/app/src/screens/workspace/workspace-screen.tsx
@@ -3854,29 +3854,36 @@ function WorkspaceScreenContent({
       ) : null}
 
       {shouldRenderDesktopPaneFallback ? (
-        <WorkspaceDesktopTabsRow
-          paneId={focusedPaneIdOrUndefined}
-          isFocused={isRouteFocused}
-          tabs={desktopTabRowItems}
-          normalizedServerId={normalizedServerId}
-          normalizedWorkspaceId={normalizedWorkspaceId}
-          setHoveredCloseTabKey={setHoveredCloseTabKey}
-          onNavigateTab={navigateToTabId}
-          onCloseTab={handleCloseTabById}
-          onCopyResumeCommand={handleCopyResumeCommand}
-          onCopyAgentId={handleCopyAgentId}
-          onCopyTerminalId={handleCopyTerminalId}
-          onCopyFilePath={handleCopyFilePath}
-          onReloadAgent={handleReloadAgent}
-          onRenameTab={handleRenameTab}
-          onCloseTabsToLeft={handleCloseTabsToLeft}
-          onCloseTabsToRight={handleCloseTabsToRight}
-          onCloseOtherTabs={handleCloseOtherTabs}
-          onCreateNewTab={handleCreateNewTab}
-          onReorderTabs={handleReorderTabsInFocusedPane}
-          focusModeEnabled={desktopFocusModeEnabled}
-          onExitFocusMode={toggleFocusMode}
-        />
+        // The splits path renders tab rows inside WorkspacePanelContent, which is what
+        // provides NewTabLauncherProvider. This fallback row sits outside it, and its
+        // new-tab menu content mounts eagerly, so without its own provider every
+        // non-compact native layout (tablets, foldables in landscape) crashed on mount
+        // with "NewTabLauncherProvider is required" (#3750).
+        <NewTabLauncherProvider value={newTabLauncher}>
+          <WorkspaceDesktopTabsRow
+            paneId={focusedPaneIdOrUndefined}
+            isFocused={isRouteFocused}
+            tabs={desktopTabRowItems}
+            normalizedServerId={normalizedServerId}
+            normalizedWorkspaceId={normalizedWorkspaceId}
+            setHoveredCloseTabKey={setHoveredCloseTabKey}
+            onNavigateTab={navigateToTabId}
+            onCloseTab={handleCloseTabById}
+            onCopyResumeCommand={handleCopyResumeCommand}
+            onCopyAgentId={handleCopyAgentId}
+            onCopyTerminalId={handleCopyTerminalId}
+            onCopyFilePath={handleCopyFilePath}
+            onReloadAgent={handleReloadAgent}
+            onRenameTab={handleRenameTab}
+            onCloseTabsToLeft={handleCloseTabsToLeft}
+            onCloseTabsToRight={handleCloseTabsToRight}
+            onCloseOtherTabs={handleCloseOtherTabs}
+            onCreateNewTab={handleCreateNewTab}
+            onReorderTabs={handleReorderTabsInFocusedPane}
+            focusModeEnabled={desktopFocusModeEnabled}
+            onExitFocusMode={toggleFocusMode}
+          />
+        </NewTabLauncherProvider>
       ) : null}
 
       <View style={styles.centerContent}>{workspacePanelContent}</View>


colinmollenhour:
@jeid64 Ok I have to know... What's it like vibe coding with Paseo on a foldable?

PR #3766, opened 2026-08-24
Offer agent-context plugin panels in the New tab launcher

gpambrozio:
## Problem

The tab launch catalogue emits plugin panels, but skips every `context: "agent"` contribution:

https://github.com/getpaseo/paseo/blob/4f796181b/packages/app/src/workspace-tabs/launcher/index.tsx#L161

So an agent panel is reachable only from the Command Center, where plugin items also sit at `groupRank: 5` — below file results.

"Register it as a workspace panel instead" is not a workaround for the plugins that need this. I maintain a [`skills` plugin](https://github.com/gpambrozio/paseo-plugins/tree/main/skills) whose agent binding is load-bearing:

- Invoking a skill is `paseo.agents.ref(agentId).send("/name")`. No agent, no verb.
- A branch of mine calls `PaseoAgentHandle.commands()` to list skills compiled into the agent binary. Claude ships ~45 of these inside a single Mach-O at `~/.local/share/claude/versions/`, so nothing on disk can be scanned. That data exists only on a live session, and there is no workspace-scoped equivalent.

Rewritten as a workspace panel it degrades to a read-only directory listing. The current tier cannot express the panel at all.

## Change

Resolve an agent for the launcher and emit agent-context panels when one exists.

The plumbing already existed end to end — `WorkspaceTabTarget` carries `{ kind: "plugin", context: "agent", agentId }`, `buildDeterministicWorkspaceTabId` emits `plugin_agent_…`, `PluginPanelBody` renders it, and the Command Center already opens exactly this target. The only missing piece was answering *which agent*.

`getFocusedAgentId` alone is not enough. The Command Center can stop at the focused tab because the user is looking at it; a launcher cannot, since the pane it launches from usually holds the New tab, so focus resolves to nothing exactly when an answer is needed. `resolveLauncherAgentId` adds two fallbacks: an agent visible in another pane, then the most recently opened one.

- Both call sites already had `workspaceId` from `usePaneContext()`, so nothing new is threaded through the tree.
- `useLaunchAgentId` mirrors the Command Center's existence check (agent in the session store **and** belonging to this workspace), so the launcher cannot offer a row that `PluginPanelBody` would refuse.
- Panels with no agent to bind to are skipped rather than disabled — the row would open a tab that can only render "unavailable".
- `getWorkspaceTabAgentId` is now shared with `plugins/command-center/context.ts` instead of duplicated.

Both launcher presentations pick this up from the shared catalogue, so the anchored `+` chooser and the persistent New tab stay in sync — the "cannot drift" property from #3735.

## Two behaviours worth arguing about

**Newest-first is a tiebreak, not recency of use.** Tabs carry no last-focused timestamp, so a workspace with several background agents gets the most recently opened one. A `lastFocusedAt` on tabs, or last-focused-agent tracked per workspace in the layout store, would be more correct and a bigger change. Happy to do that instead.

**The final fallback scans `collectAllTabs`, which includes hidden panes.** An agent parked in a collapsed side panel can win. Restricting the last fallback to visible panes is a one-line change if you prefer it.

Both are documented in the resolver's comment.

## Not in this PR

Compact layouts bypass the catalogue entirely: `WorkspaceHeaderMenuMobile` hardcodes its items and never calls `useWorkspaceTabLaunchCatalog`, so on mobile *even a workspace-context plugin panel* is unreachable from the `…` menu. That looks like a separate follow-up, and I did not want to mix it in. Say the word and I will add it here.

## QA

Automated coverage goes through the same interface a user does — the real menu, a real plugin loaded into a real daemon.

**New e2e steps** in `plugin-workspace-panels.spec.ts`, using the existing fixture plugin that registers both a workspace panel and an agent panel:

- `the launcher hides agent panels in a workspace with no agent` — a freshly seeded workspace shows **Workspace inspector** and no **Agent inspector**. Fails on the old code only in the sense that the row never existed; it guards the new skip path.
- `the launcher offers an agent panel bound to the open agent` — opens the `+` menu, asserts both rows, clicks **Agent inspector**, and asserts the panel renders `Agent bridge <agentId>`. This step fails on `main`: the row is not there.

```
$ npx playwright test --project=browser e2e/browser/plugin-workspace-panels.spec.ts --reporter=line
  1 passed (1.4m)

$ npx playwright test --project=browser e2e/browser/launcher-tab.spec.ts --reporter=line
  15 passed (3.3m)

$ npx vitest run packages/app/src/workspace-tabs/launcher/internal/agent-context.test.ts --bail=1
  Test Files  1 passed (1)
       Tests  7 passed (7)

$ npx vitest run packages/app/src/plugins/command-center/context.test.ts \
    packages/app/src/workspace-tabs/launcher/internal/catalog.test.ts
  Test Files  2 passed (2)
       Tests  5 passed (5)
```

7 unit tests cover `resolveLauncherAgentId`: focused tab wins; an agent-context plugin tab preserves context; fallback to a visible agent when the launching pane holds the New tab; visible beats newer-but-background; newest background agent when none is visible; null with no agent; null with no layout.

`npm run typecheck`, `npm run lint`, `npm run format:check` all clean (pre-commit hook output in the commit).

**Unrelated pre-existing failure:** `packages/app/src/plugins/command-center/contributions.test.ts` fails with `SyntaxError: Unexpected token 'typeof'`. I verified it fails identically on a clean `main` — not from this change.

### Platforms

| iOS             | ❌ | Compact menu not touched by this PR — see *Not in this PR*. |
| Android         | ❌ | Same. |
| Web             | ✅ | Playwright/Chromium, screenshot below. |
| Desktop macOS   | ⚠️ | Dev build run from this branch against a dev daemon with the real `skills` plugin loaded and a live Claude agent. The app and plugin came up clean; I could not script the menu click on that machine, so the Electron interaction itself is not captured. Web coverage exercises the same component. |
| Desktop Windows | ❌ | No machine. |
| Desktop Linux   | ❌ | No machine. |

No hot paths touched: the added work is one store selector and a layout walk, inside the existing `useMemo`, only while a launcher is rendered.

### Evidence

With the [skills plugin](https://github.com/gpambrozio/paseo-plugins/tree/main/skills) plugin installed:

https://github.com/user-attachments/assets/133bdfb1-d5be-4741-bf9e-dc66abfd6845


gpambrozio:
was done already. closing

Issue #520, opened 2026-04-21
Dragging a file into the composer is silently ignored; expected file path insertion


LiangLuDev:
### What happened?

Dragging a file from Finder into the chat composer does nothing.

There is no visual feedback, no error, and the file path is not inserted into the input. From a user perspective this looks broken because the drag gesture is accepted by the window but silently ignored.

### Expected behavior

When I drag a file into the composer, Paseo should insert the absolute file path into the prompt input so I can reference the file directly in my message.

At minimum, if non-image files are intentionally unsupported for drag-and-drop, the UI should surface that clearly instead of failing silently.

### Steps to reproduce

1. Open the Paseo desktop app.
2. Open any chat/composer.
3. Drag a local file from Finder into the message input area.
4. Drop it onto the composer.

### Actual result

Nothing happens.

### Expected result

The dropped file path should be inserted into the composer text input.

### Environment

- Platform: macOS desktop app
- Repo: `getpaseo/paseo`

### Notes

This may be related to the current drag/drop flow only handling image attachments and silently ignoring non-image files. Even if path insertion is considered a feature request rather than a bug, the current no-op behavior is confusing UX.

nanshanyi:
This feature is very useful, hope it can be added

noelgoodday:
+1 on this. Would be really helpful if I could drag and drop any file into the GUI.

PR #3784, opened 2026-08-24
feat(app): insert dropped text into the composer

secondstagehq:
## Problem

Dragging text onto the composer does nothing. The drop zone only accepts `Files` and the internal workspace-file drag: a `text/plain` or `text/uri-list` drag gets `dropEffect: "none"`, and because the zone prevents default on drop, the browser's native insert-into-textarea behavior is suppressed too. Dropping a URL or a snippet from another app shows a rejection cursor and is ignored. #520 covered the file half of this; this is the text half.

## Change

- `FileDropSink` gains optional `onText`.
- The DOM drop path accepts a text drag when the active sink takes text. Text is delivered only when the drop carries no files and no workspace-file payload, because Finder and file-tree drags also carry text types.
- The overlay says "Drop text here" for text-only drags (`composer.attachments.dropTextHere`, added to all 10 locales). Label state rides a shared value, so drag tracking still causes no React renders outside the backdrop leaf.
- The composer inserts the dropped text over the live selection or at the caret, then focuses the input.
- Drag classification, `text/uri-list` parsing (first non-comment line), and caret insertion are pure modules with unit tests.

File, image, and workspace-file drops are unchanged; a mixed drag with `Files` present stays a file drag.

## Tests

```
✓ src/composer/text-drop.test.ts (3 tests)
✓ src/components/file-drop/data-transfer.test.ts (7 tests)
 Test Files  2 passed (2)
      Tests  10 passed (10)
```

Playwright, the new spec plus the existing drop specs:

```
✓ dropped JSON file renders as a file attachment in active chat
✓ dropped text lands in the draft at the cursor
✓ dropped JSON file renders as a file attachment in New Workspace
  3 passed (15.1s)
```

The new spec fails without the composer `onText` registration and passes with it. `npm run typecheck`, `npm run lint`, `npm run format:check` all clean.

## QA

[Video of the interaction](https://raw.githubusercontent.com/secondstagehq/paseo/qa-assets/text-drop.webm) (webm). Draft is `open `, then a `text/plain` drag of `obsidian://open?vault=notes`:

| Before | Text drag over composer | After drop |
| --- | --- | --- |
| before | overlay | result |

File drag still shows the files label:

file overlay

## Platforms

| Platform        | Tested | Notes                                                                       |
| --------------- | ------ | --------------------------------------------------------------------------- |
| iOS             | –      | Not affected: the change is in the web-only DOM drop path                    |
| Android         | –      | Same                                                                        |
| Web             | ✅     | Playwright + captures above                                                  |
| Desktop macOS   | –      | Not separately exercised; the desktop renderer runs the same DOM drop path   |
| Desktop Windows | –      | Not tested                                                                   |
| Desktop Linux   | –      | Not tested                                                                   |

Implementation was agent-assisted; command outputs above are pasted from runs on this branch.


boudra:

PR #3790, opened 2026-08-24
feat(pi): preserve native steer and follow-up input

TaroPie1214:
## Product discussion

https://github.com/getpaseo/paseo/discussions/3788

## Problem

Paseo launches Pi through its public JSONL RPC mode, but the Pi provider does not implement native active-turn admission. A Steer request falls back to interrupt-and-replace, and the Queue composer keeps an unacknowledged client-local item until the agent becomes idle instead of letting Pi own a queued follow-up. Even when Pi accepts queued input, the UI has no authoritative projection of Pi queue_update state.

## Change

- implement Pi steerActiveTurn with the public steer RPC;
- add an optional provider followUpActiveTurn seam and implement it with Pi follow_up;
- add the backward-compatible follow_up active-turn behavior, gated by server_info.features.piNativeFollowUp before the app sends it;
- preserve existing client-local Queue behavior for other providers and old daemons;
- suppress the duplicate persisted Pi user marker after the manager has already recorded an accepted steer/follow-up;
- project Pi queue_update events into optional agent snapshot pendingInputQueue state;
- render authoritative pending Steer and Follow-up rows in the responsive composer, clearing them when Pi reports consumption;
- reject stale/slash Pi active input explicitly so unavailable native steering can never enter the generic interrupt-and-replace fallback.

The commits remain reviewable in order: native steer, native follow-up, then authoritative queue projection.

## Protocol compatibility

- New daemon accepts every old request.
- follow_up is only sent when the daemon advertises the optional piNativeFollowUp feature.
- pendingInputQueue is optional; old clients ignore it and old daemons omit it.
- Other providers are unchanged.

## Automated QA

```text
Focused protocol/server/app suite: 376 tests passed.
All workspace typechecks passed.
Changed-file lint: 0 warnings, 0 errors.
Bundled daemon Web UI build passed:
  raw 19.52 MiB
  gzip 4.43 MiB
  brotli 3.26 MiB
```

## Real provider QA

Windows 11, Pi 0.84.2:

1. Started a Pi run whose first action was sleep 10.
2. Sent activeTurnBehavior: follow_up while the run was active.
3. The send RPC returned accepted without aborting or replacing the Pi process.
4. fetchAgents showed the exact message in agent.pendingInputQueue.followUp while pending.
5. Pi completed the initial task, then consumed the queued follow-up.
6. A later authoritative snapshot showed both queue arrays empty.
7. Final output order was INITIAL_DONE, then FOLLOW_UP_DONE.
8. The live timeline contained exactly one persisted follow-up user row.

The fork canary containing the same commits is published at https://github.com/TaroPie1214/paseo/releases/tag/v0.5.1-pie.3.

## Platforms

| Platform | Tested | Notes |
| --- | --- | --- |
| Windows daemon | Yes | Real Pi 0.84.2 |
| Responsive daemon Web | Yes | Existing mobile-sized browser canary plus successful rebuilt bundle |
| Android native | No | Protocol remains compatible; no custom binary published |
| iOS native | No | Protocol remains compatible; no custom binary published |
| macOS/Linux daemon | No | Provider code is platform-neutral |

Development used an agent; the commands and real-provider observations above are the retained QA evidence.


boudra:
Native Pi steering is implemented in [#3752](https://github.com/getpaseo/paseo/pull/3752), which is merged. Closing this PR; please keep the follow-up queue workflow in [Discussions](https://github.com/getpaseo/paseo/discussions/3788), following [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #3800, opened 2026-08-24
Mirror a browser tab across hosts

ragokan:
### Linked issue

Discussion: https://github.com/getpaseo/paseo/discussions/3293 (view and control the desktop browser from another device)
Related: https://github.com/getpaseo/paseo/discussions/3148 (host identity on tabs is the shared prerequisite)

### Type of change

- [x] New feature

### Reasoning

I love the Paseo app and the browser helps a lot! But sadly, multi host creates some issues. The browser tabs I open on my carry device does not appear on host and does not do anything, so I wanted to create this PR.

I run the daemon on a Mac Mini and connect to it from a laptop. Everything else in Paseo follows me across machines; the browser doesn't. Open it from the laptop and you get a different browser: no cookies, no signed-in sessions, and none of the tabs already open on the Mac Mini. If an agent leaves a page open for me to look at, I can't look at it from where I am.

That's because the browser is client-owned. The guest is an Electron `<webview>` on one machine's profile, and the tab index is client-local zustand. The daemon only brokers automation.

This PR keeps the guest exactly where it is and mirrors it instead. The host streams CDP screencast frames over the existing binary-frame channel, the daemon fans them out to subscribed viewers, and viewers send input back. One machine keeps the cookies; every client can see and drive the tab.

### Goals

- View a tab owned by another host, live, from web, Electron desktop, or mobile
- Drive it: click, drag to select text, double-click, wheel scroll, type, shortcuts
- Tabs announced by their host appear in every client's tab strip, the way terminals already do
- Opening a browser from a viewer opens it on the daemon-local host, not locally
- The device-size menu works on a mirrored tab, including orientation
- A tab the deck keeps mounted off screen stops streaming, and resumes on return

### Non-goals

- **No cookie or profile syncing between machines.** That's #3148 and a security footgun. One guest stays the source of truth instead.
- **No viewport emulation by default.** Resizing a mirrored guest resizes it for whoever is sitting at the host too — there is one guest, shared. It's opt-in through the device menu.
- **DevTools and the element selector stay local.** DevTools opens a window on the host's physical display and is useless remotely.
- Browser tab index stays client-local; host identity on tabs is enough for this.
- **No daemon-owned browser here.** A daemon with no desktop app owning its own browser through a CDP endpoint is a separate feature and ships separately — it was 1,922 lines of this diff and is not part of mirroring a tab between clients.

### QA

**Video of the whole interaction** — opening a mirrored tab from a web viewer, scrolling, selecting text, typing, resizing:

https://github.com/user-attachments/assets/42ab3b71-82c1-433f-a384-a0f652babaca

**Platforms tested:** Electron desktop (host + viewer) and web (viewer), against a real second daemon on a separate `PASEO_HOME`.
**Not tested:** iOS and Android. The mirrored pane's touch path (`mirror/input-surface.tsx` — drag scrolls, tap clicks) is unchanged from the shipped gesture idiom but I have not run it on a device. Calling that out rather than implying coverage I don't have.

#### Real-Electron harness, parked production guest

```
$ PASEO_CAPTURE_HARNESS_GROUP=screencast npm run capture-harness --workspace=@getpaseo/desktop
PASS screencast initial    size=1280x800 logical=1280x800 bright=0.8468 text=true
PASS screencast mutated    size=1280x800 logical=1280x800 bright=0.8416 text=true
PASS screencast sustained  frames=192 elapsed=2000ms fps=96.0 meanFrameBytes=39905
PASS screencast sustained  size=1280x800 logical=1280x800 bright=0.8495 text=true
PASS screencast stopped    frameCount=195 framesAfterStop=0
```

PNGs and `results.json` in `capture-harness/out/screencast/`. The sustained phase exists because one frame only proves the surface is copyable; a regression that delivered a single frame and then went silent passed every other assertion in that group.

#### Measured on a running stack (daemon + Electron host + web viewer)

| | |
|---|---|
| Scrolling a long page | 99 frames in 5s — **19.8 fps**, median gap 58ms, p90 65ms |
| Frame size | 185 KB at 591×1280, 40 KB mean in the harness at 1280×800 |
| Keys reach the guest | 3 × PageDown → 39 frames of page scroll |
| Resize | Laptop preset → aspect **1.78** (1366/768); rotated → **0.56** (768/1366) |
| Backpressure | stalled viewer: **50 frames queued before, 1 after** |
| Off-screen tab | switching away sends `unsubscribe`, switching back sends `subscribe` and redraws from the replay |
| Oversized subscribe | `maxWidth: 100000` reached the host verbatim before; clamped now |
| Forged frame from a non-host session | fanned out before (`['forged','genuine']`), rejected now |

#### Tests

```
protocol   28    client     4    server   109    desktop   92    app(unit)  223
```

Repo-wide: `npm run typecheck`, `npm run lint` (3805 files, 0 warnings), `npm run format:check` (4054 files) all pass.

`mirror/mouse-input.browser.test.ts` runs in the vitest browser project and is **unrun here** — the matching Chromium build was still downloading. It typechecks; its assertions are unexecuted.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


ragokan:
Thanks — all four were real, and two of them I had already noticed earlier and not chased down properly. Fixed in 0f927e78.

**Slot reserved until the stop settles** (P1). The slot is held while `screencast_stop` is in flight and only the same browser may retake its own, so a stale frame can only ever land in the tab it came from. I went with reservation rather than a stream generation on the wire because the frame is a fixed byte layout and this needed no protocol change or COMPAT tag. One residual I left deliberately: a `screencast_start` that fails or times out still releases the slot without sending a stop, so a host that did start capturing but answered late reaches the same hazard through a different door.

**Viewers keyed per socket** (P1). You're right that this is a correctness bug rather than the performance ceiling it was written up as — a sibling window's unsubscribe stopped frames for a window still displaying the tab, and every frame was pushed at sockets that never subscribed. The source socket now threads through subscribe/unsubscribe, which also makes the latest-frame-wins backpressure per socket instead of per session. Still outstanding: two panes in one window share a socket, so pane-level refcounting belongs app-side.

**Republish on host unregister** (P2). Extended slightly beyond the suggestion — with no mirror host left the `list_tabs` fan-out has nobody to ask and fails silently, which is exactly the case that strands tabs, so the daemon publishes an empty list directly.

**Hover** (P2). Forwarded, but not as a plain move: `heldButtons` is the button mask for every phase except `up`, so an unpressed move would have told the guest the left button was down and started a text selection under the pointer. It goes as a distinct `hover` phase that names no button, and it is bound to the pane rather than the window, since the drag listener is deliberately window-level so a selection survives leaving the pane.

Each fix has a test that was confirmed failing beforehand.

ragokan:
I also tried to make it as much as re-using existing logics. Terminal had great logic for that. Please check the video attached, I hope this is interesting for Paseo users!

boudra:

PR #3803, opened 2026-08-24
fix(server): report archived agents correctly over MCP

thomasvan:
`list_agents` can return an archived agent as a live one, with `archivedAt: null`, from a default call.

## Why

An archived agent is resumed back into memory whenever something reads its history — `agent-loading.ts` passes `{ purpose: "history" }` to `resumeAgentFromPersistence` precisely when `record.archivedAt` is set. Its id is then in the manager's map, so `list_agents` classifies it as live, and the stored branch drops it because that branch filters out anything already live:

```ts
.filter((record) => !record.internal && !liveIds.has(record.id))
```

The stored record is the only carrier of the archive timestamp, so both the value and the archive filter go with it. Two consequences, both reachable from a default call:

- **The timestamp is lost.** `serializeSnapshotWithMetadata` merged only the title from storage, so the agent is reported with `archivedAt: null`. A caller reconciling ownership from that list can conclude an archived agent is still running.
- **`includeArchived: false` returns it anyway.** The archive filter ran only on the stored branch, so nothing on the live branch was ever filtered.

The protocol declares `archivedAt` as `z.string().nullable().optional()`, so the omission validated silently.

## The fix

Storage stays the single owner of the timestamp. The MCP boundary merges the stored record — reusing the storage read that was already happening for the title — and `archivedAt` becomes a projection option supplied by a caller holding the record, alongside `title`. Archive filtering moves to the combined live-plus-stored list.

This is the pattern the WebSocket path already uses in `Session.enrichAgentPayload`; this change brings the MCP path in line with it rather than introducing a new mechanism.

**The rejected alternative** was copying the timestamp onto the managed agent. It is worse: `unarchiveSnapshot` clears only the stored record and then notifies, so a managed copy would survive the unarchive and report a live agent as archived until it was reloaded. A second owner of state that one transition forgets to update is exactly the bug being fixed, in mirror image.

## Tests

Three cases on the existing `mcp-server.test.ts` harness, which previously modelled an archived record only when it was *not* also live:

| Test | Without the fix |
|---|---|
| Default list excludes an archived agent that is also live | `expected [ 'resumed-archived', 'in-cwd' ] to deeply equal [ 'in-cwd' ]` |
| `includeArchived: true` reports the real timestamp | `expected null to be '…'` |
| A cleared stored record reads as unarchived | passes — structural guard, labelled as such |

The first two are mutation-killing and pin the observable MCP result. The third passes on unfixed code too; its comment says so rather than implying otherwise.

## Compatibility

Additive. `archivedAt` was already nullable and optional in the protocol and already populated on the stored path, so consumers that read it keep working and consumers that ignore it are unaffected.

Full server suite: no new failures against `main` — 1986 passing vs 1983 on a pristine baseline run on the same machine, the difference being the three tests added here. `oxlint`, `oxfmt`, and the workspace typecheck are clean.

### DIFF
--- a/packages/server/src/server/agent/agent-projections.ts
+++ b/packages/server/src/server/agent/agent-projections.ts
@@ -26,6 +26,13 @@ interface ProjectionOptions {
   title?: string | null;
   createdAt?: string;
   internal?: boolean;
+  /**
+   * Archive timestamp, supplied by a caller holding the stored record. A live
+   * managed agent cannot know this: an archived agent is resumed back into
+   * memory to serve a history read, so being in the manager is not evidence
+   * that it is active. Storage stays the single owner.
+   */
+  archivedAt?: string | null;
 }
 
 interface RecentProviderSessionProjectionOptions {
@@ -135,6 +142,7 @@ export function toAgentPayload(
     persistence: projectPersistenceHandleForWire(agent.persistence),
     title: options?.title ?? null,
     labels: agent.labels,
+    archivedAt: options?.archivedAt ?? null,
   };
 
   const usage = sanitizeUsage(agent.lastUsage);
--- a/packages/server/src/server/agent/mcp-server.test.ts
+++ b/packages/server/src/server/agent/mcp-server.test.ts
@@ -5502,6 +5502,109 @@ describe("agent snapshot MCP serialization", () => {
     expect(new Set(agentIds)).toEqual(new Set(["in-cwd", "in-child-cwd", "stored-in-cwd"]));
   });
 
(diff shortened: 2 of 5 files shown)


thomasvan:
Status update: I still carry this patch on `slp/patches` (fork tip `04563655b`), source-reverified through the 2026-09-07 sync to upstream `c424f8292` (0.7.2). Merge-tree against current `main` (`8c89e72a0`) is clean; the combined live-plus-stored `list_agents` filtering remains needed — a hydrated archived agent still comes back live with `archivedAt: null`.

PR #3814, opened 2026-08-24
fix(server): recover OpenCode turns after a stalled event stream

boudra:
### Linked issue

Refs #3571. Follows #3578 and #3621.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

OpenCode's shared event consumer aborts and retries an SSE connection after 30 seconds without a record. Turn dispatch currently waits on that same first record for exactly 30 seconds, so it fails at the boundary where the consumer begins recovery. The prompt is never sent, even if the next connection succeeds immediately.

This change keeps dispatch gated on an observed event stream, but gives the consumer's existing retry 15 seconds to land. It also records the actual SDK SSE error and exposes the consumer's phase, attempt, elapsed time, and previous outcome in logs and the chat-visible timeout. The environmental trigger behind the field recurrence remains unconfirmed; this PR addresses the demonstrated timeout race and makes the next report diagnostic.

### Goals

- Let one consumer-managed retry recover a silent initial SSE attempt before a turn fails.
- Keep prompts gated until the generation-owned event stream has delivered a record.
- Preserve the existing 30-second stream watchdog and reconnect behavior.
- Put observed stream state and SDK errors in daemon logs and the user-visible failure.

### Non-goals

- Do not claim a confirmed environmental cause for the reported recurrence.
- Do not remove the readiness gate or dispatch a prompt before the event stream is observed.
- Do not add per-session event subscriptions or SDK-owned retry loops.
- Do not change mid-turn reconciliation behavior.
- A persistent failure now surfaces after 45 seconds instead of 30 seconds.

### QA

The new turn-level regression failed before the production change with `OpenCode event stream first record` at 30 seconds. After the change, readiness at 35 seconds dispatches exactly one prompt; a stream that never becomes ready still sends no prompt and fails at 45 seconds with diagnostics.

The event-consumer test drives the real generated SDK against a local SSE server: the first connection receives no record, the existing watchdog aborts it, and the second connection delivers `server.connected`. It asserts the first-record watchdog error, retry attempt, structured warning, and readiness on attempt two. A separate 503 case verifies that `onSseError` preserves the SDK error instead of collapsing it into an unexplained clean EOF.

Real-provider check on macOS with OpenCode 1.18.4:

- isolated Paseo daemon;
- fresh dedicated OpenCode generation;
- a real global plugin delayed initialization by 20 seconds;
- the initial turn passed event-stream readiness and reached the expected invalid-model failure;
- this did not reproduce the field timeout, so slow plugin initialization alone is not claimed as its cause.

Commands:

```text
npx vitest run packages/server/src/server/agent/providers/opencode-agent.test.ts --bail=1
Test Files  1 passed (1)
Tests  125 passed (125)

npx vitest run packages/server/src/server/agent/providers/opencode/event-consumer.test.ts --bail=1
Test Files  1 passed (1)
Tests  13 passed (13)

npx vitest run packages/server/src/server/daemon-e2e/opencode-slow-plugin-readiness.real.e2e.test.ts --bail=1
Test Files  1 passed (1)
Tests  1 passed (1)

npm run typecheck
exit 0

npm run lint -- <six changed files>

npm run format
exit 0
```

Risk surface: server-side OpenCode startup and reconnect logging only. No protocol, client, UI, dependency, credential, or persistence changes.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/server/src/server/agent/providers/opencode-agent.test.ts
+++ b/packages/server/src/server/agent/providers/opencode-agent.test.ts
@@ -30,6 +30,7 @@ import type {
 // guard: deriving the boundary from OPENCODE_SERVER_STARTUP_TIMEOUT_MS would keep the
 // readiness tests green if that constant were shortened below the required budget.
 const EXPECTED_STARTUP_BUDGET_MS = 30_000;
+const EXPECTED_EVENT_READINESS_BUDGET_MS = 45_000;
 
 function tmpCwd(): string {
   const dir = mkdtempSync(path.join(os.tmpdir(), "opencode-agent-test-"));
@@ -3540,7 +3541,7 @@ describe("OpenCode adapter startTurn error handling", () => {
     await session.close();
   });
 
-  test("bounds dispatch readiness at the server startup budget without sending a prompt", async () => {
+  test("bounds dispatch readiness without sending a prompt", async () => {
     vi.useFakeTimers();
     const openCode = new TestOpenCodeClient();
     const session = new __openCodeInternals.OpenCodeAgentSession(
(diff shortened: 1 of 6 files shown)


axsuul:
Ran into this again on 0.5.1, so here is some field evidence from a production headless daemon (Linux) in case it helps land this.

Five agents failed with `OpenCode event stream first record` in one morning. For one of them I could pin the whole timeline from the daemon journal plus OpenCode's own sqlite db:

```
09:14:20.7  daemon spawns opencode serve
09:14:22    server answers session.create (session.time_created in opencode.db)
09:14:24.8  turn starts, 30s events.ready() clock begins
09:14:55.2  turn fails: no first SSE record
09:20:16    re-prompt on the same server + session works fine
```

So the server was listening and serving HTTP 2 seconds after spawn, but the /global/event subscription opened at startup delivered nothing for 30+ seconds. The same consumer recovered a few minutes later. I probed all 18 healthy helper servers on the box afterwards and every one delivers its first SSE record to a fresh subscriber in under 25ms, so it is not general slowness. The window is only right around boot.

That matches the stalled-stream case this PR describes. And since the consumer's silent-stream recovery budget is also 30s, the turn deadline fires exactly when a reconnect would have saved it. opencode 1.18.18, daemon 0.5.1, load avg ~10 at the time. Also saw the same failure on 0.5.0-beta.4 and beta.5.

tsukasa-mixer:
. opencode-stream attempt=4 phase=first-record elapsedMs=49303 lastOutcome=error lastError=“fetch failed”

on 0.6.0 \ 0.6.1

PR #3820, opened 2026-08-24
fix(server): let the omp provider preapprove exact MCP tools and bridge them to host tools

aryrabelo:
## Symptom

Running `omp` as an agent provider behind the Hub, any execution that
attaches the Hub's injected `hub` MCP server (`reply`/`finish_execution`)
with a `toolPolicy` failed immediately with:

```
ToolPolicyUnsupportedError: Provider 'omp' cannot preapprove exact MCP
tools for unattended execution; select Claude, Codex, or OpenCode
```
(`code: "tool_policy_unsupported"`)

## Cause

`PROVIDER_CONTRACTS` in `provider-registry.ts` had entries for
`claude`/`codex`/`opencode` but not `omp`, so it fell back to
`UNSUPPORTED_PROVIDER_CONTRACT` (`supportsExactMcpPreapproval: false`).

Separately, even with that fixed, omp's `rpc-ui` launch mode never
surfaces MCP tools to the model directly — the only channel that reaches
the model is Paseo's native host-tool bridge
(`runtimeSession.setHostTools`). `OMP_CORE_CAPABILITIES.supportsMcpServers`
was `false` and nothing consumed `config.mcpServers`, so MCP servers
attached to an omp session were silently dropped: the model never saw
`reply`/`finish_execution` at all.

## Fix

1. **`provider-registry.ts`** — add an `omp` entry to `PROVIDER_CONTRACTS`
   with `supportsExactMcpPreapproval: true`. omp runs its host-tool bridge
   in full-access mode with no per-tool approval prompt, so unlike
   Claude/Codex/OpenCode there's no narrower `applyToolPolicy` to enforce;
   preapproval is a no-op gate for this provider today. (If a future PR
   wants per-tool enforcement, `applyToolPolicy` is the extension point —
   see how `hub-e2e` uses it for an example of exact-grant enforcement.)

2. **`providers/omp/mcp-tool-bridge.ts`** (new) — connects to each `http`
   MCP server declared on the session, lists its tools, and re-exposes
   them as Paseo host tools whose handler forwards the call to the MCP
   client:
   - Converts each tool's JSON Schema `inputSchema` to a Zod schema.
     Paseo's host-tool serialization (`serializePaseoToolInputParameters`
     -> `normalizeObjectSchema`) only recognizes Zod; passing the MCP
     JSON Schema through unconverted registers every bridged tool as
     taking **no parameters**, which breaks structured tools like
     `finish_execution` (`classification`/`confidence`/`summary`) — the
     model would have no way to know what arguments to pass.
   - `stdio` servers are skipped (out of scope: the Hub only injects
     `http`, and `stdio` MCP servers are OMP's own native concern).
   - Returns a catalog with a `close()` that disconnects every MCP
     client it opened.

3. **`providers/omp/agent.ts`**:
   - `OMP_CORE_CAPABILITIES.supportsMcpServers` and
     `withOmpCapabilities()`'s override both flip to `true`.
   - `createSession` builds a single merged catalog (native `paseoTools`
     + the MCP bridge) and feeds it to **both** `configureNativePaseoTools`
     (what `setHostTools` announces to the model) **and** the
     `OmpAgentSession.paseoTools` field, which is what
     `handleOmpHostToolRuntimeEvent`/`OmpHostToolRouter` dispatches
     `host_tool_call` events against. Registering the merged catalog
     while only dispatching against the native one would advertise
     bridged tools the model could call but never actually execute
     (`OmpHostToolRouter` would resolve `finish_execution` against the
     native-only catalog, throw `Tool not found`, and the model would see
     a failed tool call every time).
   - `OmpAgentSession.close()` now also closes the MCP bridge's clients,
     so a session doesn't leak MCP client connections.

## Repro

```ts
const registry = buildProviderRegistry(logger, { ompRuntime });
registry.omp.applyToolPolicy(
  { provider: "omp", cwd, mcpServers: { hub: { type: "http", url } } },
  { preapproved: [{ kind: "mcp", server: "hub", tool: "finish_execution" }] },
);
// before: throws ToolPolicyUnsupportedError
// after: returns { ...config, toolPolicy }
```

## Tests

- `provider-registry.test.ts`: new test asserting `omp.applyToolPolicy`
  accepts the Hub's exact MCP grant instead of throwing
  `tool_policy_unsupported`.
- `providers/omp/mcp-tool-bridge.test.ts` (new): unit coverage for the
  bridge in isolation — no `mcpServers` / non-`http` servers return
  `null` and open no client; `http` servers connect, list tools, and
  convert the JSON Schema to a real Zod object shape (not an empty-object
  fallback); `executeTool` forwards to the mocked MCP client's `callTool`
  and maps the result; `close()` closes every connected client. The MCP
  SDK is mocked (`@modelcontextprotocol/sdk/client/index.js` and
  `.../streamableHttp.js`) — no network.
- `providers/omp/agent.test.ts`: new "OMP MCP tool bridge wiring" suite
  drives the **real** `createSession -> setHostTools -> host_tool_call ->
  executeTool` path (not a shortcut around it) and asserts the mocked
  MCP client's `callTool` is actually invoked when OMP calls the tool
  back — the assertion that matters, since a test that only checks
  `setHostTools` announced the tool would pass even with the
  native-vs-merged catalog bug. Also covers the no-`mcpServers` case
  (unchanged behavior) and that `session.close()` closes every bridged
  client.

Each new/changed assertion was verified to fail when its corresponding
fix is reverted (confirmed locally by reintroducing each bug one at a
time and re-running the suite; not included as a commit).

## Build / run from this fork

```bash
npm install --workspace=packages/server
npm run build --workspace=packages/protocol   # workspace dep, needs building first
npm run build --workspace=packages/server
PASEO_HOME=~/.paseo-omp node packages/server/dist/scripts/supervisor-entrypoint.js
```

Typecheck: `npm run typecheck --workspace=packages/server`
(`tsgo -p tsconfig.server.typecheck.json --noEmit`).
Tests touched: `npx vitest run src/server/agent/provider-registry.test.ts src/server/agent/providers/omp/agent.test.ts src/server/agent/providers/omp/mcp-tool-bridge.test.ts` from `packages/server`.

### DIFF
--- a/packages/server/src/server/agent/provider-registry.test.ts
+++ b/packages/server/src/server/agent/provider-registry.test.ts
@@ -624,6 +624,29 @@ test("OMP is a disabled built-in backed by the real OMP adapter", async () => {
   await session.close();
 });
 
+test("OMP accepts the Hub's exact MCP grant for its injected server", () => {
+  const registry = buildProviderRegistry(logger, { ompRuntime: new FakeOmp() });
+  const config = {
+    provider: "omp",
+    cwd: "/tmp/registry-omp-toolpolicy",
+    mcpServers: { hub: { type: "http" as const, url: "http://127.0.0.1/execution" } },
+  };
+  const toolPolicy: ToolPolicy = {
+    preapproved: [
+      { kind: "mcp", server: "hub", tool: "reply" },
+      { kind: "mcp", server: "hub", tool: "finish_execution" },
+    ],
+  };
+
+  // Without an `omp` entry in PROVIDER_CONTRACTS this throws
+  // ToolPolicyUnsupportedError (code "tool_policy_unsupported") and the Hub
+  // execution never starts.
+  expect(registry.omp.applyToolPolicy(config, toolPolicy)).toEqual({
+    ...config,
+    toolPolicy,
+  });
+});
+
 test("OMP can be enabled without custom provider boilerplate", () => {
   const registry = buildProviderRegistry(logger, {
     providerOverrides: {
--- a/packages/server/src/server/agent/provider-registry.ts
+++ b/packages/server/src/server/agent/provider-registry.ts
@@ -153,6 +153,12 @@ const PROVIDER_CONTRACTS: Record<string, ProviderContract> = {
   claude: { optionsSchema: ClaudeProviderOptionsSchema, supportsExactMcpPreapproval: true },
   codex: { optionsSchema: CodexProviderOptionsSchema, supportsExactMcpPreapproval: true },
   opencode: { optionsSchema: OpenCodeProviderOptionsSchema, supportsExactMcpPreapproval: true },
+  // The Hub injects the `hub` MCP server with an exact tool grant (e.g.
+  // `reply`/`finish_execution`) and OMP runs its host-tool bridge in
+  // full-access mode with no per-tool approval prompt, so there is no
+  // narrower policy to enforce here the way Claude/Codex/OpenCode do via
+  // `applyToolPolicy`. Preapproval is a no-op gate for this provider today.
+  omp: { optionsSchema: EmptyProviderOptionsSchema, supportsExactMcpPreapproval: true },
 };
 
 const UNSUPPORTED_PROVIDER_CONTRACT: ProviderContract = {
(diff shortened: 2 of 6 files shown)


boudra:

Issue #3720, opened 2026-08-22
feat: preview video and PDF files in the file explorer


Irvingouj:
### Prior search

- I searched existing issues and discussions.
- PDF preview was requested before as #1471 and #3104 (closed; bot redirected to Discussions) and as discussion #3109 (PPTX/PDF/XLSX).
- I could not find an open request for **video** preview in the file explorer / file pane.

### What's the problem

When I click a `.pdf`, `.mp4`, `.webm`, or `.mov` in the workspace file explorer, Paseo opens the file pane and shows **Binary preview unavailable**. Text and images preview in-app; PDFs and videos do not.

That gets in the way when an agent produces or downloads a PDF/video (reports, screen recordings, demo clips) and I want to inspect it in the same workspace, especially when the daemon is remote and there is no local copy to open in Preview / a browser.

Today the explorer file model is only `text | image | binary` (`packages/server/src/server/file-explorer/service.ts`, `FileExplorerFileSchema` in `packages/protocol`). Images are a small extension allowlist (png/jpg/gif/webp/svg). PDFs and videos are classified as binary, so `packages/app/src/components/file-pane.tsx` falls through to `panels.file.binaryPreviewUnavailable`. Download from the explorer context menu still works; preview does not.

### What would solve it

Clicking a PDF or common video file in the file explorer should open a read-only in-app preview in the file pane, the same way images already do.

- **PDF:** render pages in the file pane (PDF.js or the existing in-app browser/webview PDF viewer). At least first-page / scrollable pages; printing is not required.
- **Video:** play `.mp4` / `.webm` / `.mov` with standard playback controls (play/pause, seek, volume). Do not download-only.

Remote daemons should work: the client already has a file-read / download-token path, so preview should not require the file to exist on the client disk.

A reasonable fallback if a format cannot be rendered: keep the current binary placeholder **and** keep Download.

### Alternatives you considered

- Download the file and open it outside Paseo. Extra steps, and awkward on a remote daemon or from the phone app.
- Ask the agent to summarize / extract frames. Lossy, slow, and I still cannot watch or read the original.
- Discussion #3109 covers office/PDF preview but not video. This issue is specifically: **PDF + video in the file explorer preview**, not PPTX/XLSX.

Not asking for editing, annotation, or a full media library — just preview when I open the file from the explorer.

grinidx:
Confirming on 0.7.2 (Linux), and adding the code path in case it saves anyone time.

[...]

PR #3823, opened 2026-08-24
feat(app): play assistant videos inline in the web timeline

seupedro:
An agent hands back a recording, and it plays where the work happened:

Asking the agent for a clip and getting a player back in the timeline

### Linked issue

None. Opened alongside discussion #3822, which describes the same gap from the user's side.

Related surface: #3720 asks for video preview in the **file explorer**; this is the conversation
timeline. The server change here (video MIME types) is the same groundwork that issue needs, but
the file pane is untouched.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

An agent that records its own run has no way to show it. Markdown has no video node, so agents
reach for image syntax — `demo` — and the timeline handed that straight to the
image renderer, which failed and left **"Image unavailable"** where the recording should be.
Images from the same agent, through the same path, render fine.

Video now takes the route images already take: the daemon hands the bytes over the existing file
RPC and the client turns them into a preview URL. That keeps it working over the encrypted relay,
where the daemon's HTTP port is not reachable at all. The cost is holding the file in memory, so
reads are capped and the daemon refuses larger ones on the file's stat, before any bytes move.

### Goals

- `alt` in an assistant turn renders a player with native controls on web and Electron.
- `mp4`, `m4v`, `webm`, `ogv`, `mov`.
- Files past `MAX_INLINE_VIDEO_BYTES` (50 MB) are refused by size, not streamed and then abandoned.
- Anything that cannot play falls back to a card instead of a broken frame, and the card names the
  file — a turn can carry several videos, and "File is too large to display" alone does not say
  which one was skipped.

### Non-goals

- **No native player.** iOS and Android would need `expo-video` and a native rebuild; they get a
  card saying playback is not available there. That replaces today's wrong "Image unavailable".
- **No new explorer file kind.** `ExplorerFileKind` is wire schema. Video stays `"binary"` and only
  the reported MIME type changes, so there is no capability gate and an older daemon simply leaves
  the app on its fallback.
- No transcoding, no thumbnails, no poster frames, no `.mkv`/`.avi` — no browser decodes those
  natively, and listing them would trade an honest card for a player that never starts.
- The composer attachment path and the `/api/files/download` MIME are deliberately untouched.

### Design notes

- `packages/server/.../file-explorer/service.ts` — `VIDEO_MIME_TYPES` applied at both binary
  classification sites. A blob built from `application/octet-stream` never plays; this is the whole
  server-side change.
- `assistant-image/file-acquisition.ts` — the hardcoded `kind !== "image"` check becomes an `accept`
  predicate, plus an optional `maxBytes` forwarded to the existing `readFile` RPC. Image behaviour is
  unchanged when neither is passed.
- `assistant-image/media-attachment.ts` — the media-agnostic half of `use-assistant-image.ts`
  (acquisition + preview-URL hooks and their LRU caches) extracted so both hooks share it. The same
  file referenced twice now resolves once for either media type.
- `assistant-video-surface.{tsx,web.tsx}` — platform split in the shape of `file-pane/html-preview`.
- The lifecycle event `image_loaded` is renamed `media_loaded`; it is shared now.
- An oversized read surfaces the daemon's own message, the same way image read failures already do,
  followed by the file name and, when there is one, the directory it came from. The name gets its
  own line on purpose: `react-native-web` only implements `ellipsizeMode="tail"`, so truncating the
  whole path as one line would cut off the extension — the part that says what the file is.

### QA

Linux (aarch64), Node 22.22.2, daemon and Metro from this checkout.

**Automated — 58 app tests + 24 server tests + 2 Playwright specs, all passing.**

```
$ npm run typecheck && npm run lint && npm run format:check
tsgo --noEmit                              # 0 errors, all workspaces
oxlint                                     # Found 0 warnings and 0 errors.
oxfmt --check                              # All matched files use the correct format.

$ npx vitest run src/attachments/file-types.test.ts \
    src/assistant-image/file-acquisition.test.ts \
    src/assistant-image/use-assistant-video.test.ts \
    src/assistant-image/lifecycle.test.ts \
    src/i18n/resources.test.ts --bail=1
Test Files  5 passed (5)
     Tests  58 passed (58)

$ npx vitest run src/server/file-explorer/service.test.ts --bail=1
Test Files  1 passed (1)
     Tests  24 passed (24)

$ npx playwright test --project=browser e2e/browser/assistant-video.spec.ts
✓ an assistant video in the timeline plays inline (17.6s)
✓ a video past the inline ceiling falls back instead of loading (15.9s)
2 passed (1.2m)
```

The e2e drives a real daemon and a real mock-provider turn, and asserts `readyState >= 1` with
`videoWidth`/`videoHeight` — that the browser *decoded* the file, not that an element mounted.

**The e2e fails on the old code, for the right reason.** Reverting only the two server lines that
report the video MIME type:

```
✘ an assistant video in the timeline plays inline (41.0s)
  Error: expect(locator).toBeVisible() failed
  Locator: getByTestId('assistant-video').first()
  Error: element(s) not found
```

**Codec caveat, measured, not assumed.** Playwright's bundled Chromium is built without proprietary
codecs. In the exact binary the suite uses (`chromium_headless_shell-1208`, Chrome/145):

```
video/webm; codecs="vp9"        -> "probably"
video/mp4;  codecs="avc1.42E01E" -> ""        (no)
```

So the fixture is WebM/VP9 on purpose: an MP4 fixture would fail in CI for a codec reason and prove
nothing about this code. **MP4/H.264 playback was verified by hand** against a Chromium that has the
codecs (the system Chromium, Chrome/151, reports `"probably"` for the same query), which is what
the video below shows.

**Manual QA — video of the whole interaction (104 s), nothing pre-loaded.** I type into the
composer, the agent answers with prose and two Markdown links, and:

- `mountain-trail.mp4` (1280×720 H.264) plays inline and the counter climbs;
- a click near the end of the scrubber moves it from 4 s to 10 s, so the file is decoded and
  seekable rather than a still frame;
- `clouds-from-above.webm` (VP9) plays through the same code path;
- in a second tab, two exports past the ceiling fall back to cards that name the file — one of
  them showing the folder it came from.

Footage is free stock from Coverr; the two oversized files are the same clips re-encoded at a high
bitrate so they land above 50 MB.

Recording, 104 s at 1440×900:
[paseo-video-inline-demo.mp4](https://github.com/seupedro/paseo/releases/download/video-inline-demo/paseo-video-inline-demo.mp4)
— GitHub serves a release asset as a download rather than an inline player, so the frames that
carry the argument are below too.

<details>
<summary>Stills from that recording</summary>

Typing into an empty conversation — nothing is pre-loaded:

Typing the request

`mountain-trail.mp4` playing, paused at 4 s:

MP4 playing inline

After a click near the end of the scrubber — 4 s to 10 s:

Seek from 4 s to 10 s

`clouds-from-above.webm` through the same path:

WebM VP9 playing

Two exports past the ceiling, each card naming its file, the second also naming the folder:

Oversized files fall back to a card that names them

</details>

| Web             | ✅     | Chromium 151 (H.264) by hand, and Chrome/145 headless in the Playwright suite. |
| iOS             | ❌     | Not run. Change there is the static "not available" card; no player, no new dependency. |
| Android         | ❌     | Same as iOS. |
| Desktop macOS   | ❌     | Not run. |
| Desktop Windows | ❌     | Not run. |
| Desktop Linux   | ❌     | Not run. Shares the web renderer, so it should follow web, but I did not launch Electron. |

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence
- [x] Tests added


boudra:

Issue #2470, opened 2026-07-27
Forge PR poller has no global API budget: exhausts GitHub GraphQL quota machine-wide (~9,340 pts/hour at 144 targets)


Mtuchel:
## problem

The forge PR-status poller has no global API budget. Its cost is `O(active workspaces)` with a fixed per-target interval, so past roughly 80–100 active workspaces the daemon permanently exhausts the GitHub GraphQL hourly quota for the whole machine — including every `gh` call the user's own agents make.

Per-target backoff exists and works correctly. It cannot help here: the breach is fleet-wide, not per-target. Each target is individually well-behaved while the fleet as a whole is 2x over budget.

## measurement (v0.2.1, macOS, single daemon)

State: 243 active workspaces → 144 distinct forge poll targets (deduped by cwd+headRef, as designed).

GraphQL quota, sampled via `gh api rate_limit` (does not itself consume quota):

```
T-5min   graphql used=5000 remaining=0     core used=6   <- GraphQL fully exhausted, REST untouched
T+0      quota window reset
T+10:20  used=1739                         <- 2.60 pts/s = ~9,340 pts/hour vs the 5,000/hour limit
```

The hour's entire budget is consumed in ~24 minutes; for the remaining ~36 minutes every GraphQL call on the machine fails with:

```
GraphQL: API rate limit already exceeded for user ID <id>
```

`gh` process spawns attributable to the daemon (`ppid` = daemon pid), sampled by `ps` every 0.3 s for 30 s:

```
169 unique gh processes / 30 s
  gh pr view --json ...statusCheckRollup
  gh pr list --state all
  gh repo view --json owner,name,parent
  gh api graphql
```

Daemon log for one day: **11,471** `Failed to run forge PR status self-heal refresh` records, 100% of them the rate-limit stderr.

## why the existing backoff cannot fix it

`computeGithubNextInterval` in `github-service`:

```js
const baseInterval = isGitHubStatusPending(status)
  ? GITHUB_POLL_FAST_INTERVAL_MS   // 20_000
  : GITHUB_POLL_SLOW_INTERVAL_MS;  // 120_000
if (consecutiveErrors <= 1) return baseInterval;
return Math.min(baseInterval * 2 ** (consecutiveErrors - 1), GITHUB_POLL_ERROR_BACKOFF_CAP_MS); // 300_000
```

[...]

PR #3825, opened 2026-08-24
Stop PR status polling from rate-limiting your entire GitHub account

### LINKED ISSUE #3587: PR status poller exhausts GitHub GraphQL rate limit with many workspaces — all workspaces then show "no PR"
**Version:** Paseo 0.5.0-beta.2 (macOS, Darwin 25.6), `gh` CLI authenticated via keyring

## Summary

With ~55–60 workspaces open, the daemon's PR status polling (`workspace-git-service` → `github-service`) burns through the 5,000 pts/hour GitHub GraphQL quota every hour. Once the bucket is empty, every `gh pr view` fails and all workspaces display as having no PR / no GitHub info until the hourly reset — even though the branches, upstreams, and PRs are all fine.

## What I observed

- `~/.paseo/daemon.log` shows a burst of
  `GitHubCommandError: GitHub CLI command failed: gh pr view --json number,url,title,state,isDraft,baseRefName,headRefName,headRefOid,mergedAt,reviewDecision,mergeable,headRepositoryOwner,statusCheckRollup`
  at roughly :40–:44 past every hour (664 entries over ~9 hours, one per workspace per cycle, 55 distinct cwds in the last burst).
- `gh api rate_limit` at that moment: `graphql.remaining: 0`, `core.remaining: 4999`. It's purely the GraphQL bucket that `gh pr view` uses; REST is untouched.
- Measured live burn with nothing else running: ~54 pts/min (~3,200/hour) in the slow-poll state alone. Each `gh pr view` with `statusCheckRollup` + `mergeable` costs ~3–5 points.
- My own interactive `gh` use and coding agents share the same token, which tips it over the cap.

## Why it happens (from the bundled `services/github-service.js`)

- `GITHUB_POLL_SLOW_INTERVAL_MS = 120000`, `GITHUB_POLL_FAST_INTERVAL_MS = 20000` (used whenever the PR is "pending", e.g. checks running).
- A poll target is retained via `retainCurrentPullRequestStatusPoll` for **every** workspace with a listener — the sidebar observes all of them, so all ~55 poll continuously, not just the selected/visible one.
- 55 workspaces × 30 polls/hr × ~3–5 pts ≈ 5,000–8,000 pts/hr, before counting any 20s fast-poll targets.
- `computeGithubNextInterval` only backs off after ≥2 consecutive errors, caps at 5 min, and resets on any success — so once the limit is hit it keeps retrying every 2–5 min and never reads the rate-limit headers / reset time.

## Impact

- For the last ~10–15 min of every hour, every workspace's PR panel shows nothing, which looks like the workspace lost its git/PR association.
- Any other `gh` usage on the machine (CLI, coding agents) gets `API rate limit already exceeded` during that window.

## Suggestions

1. Only poll workspaces that are visible/selected (or poll unfocused ones far less often), instead of every observed workspace.
2. Drop `statusCheckRollup` / `mergeable` from the background slow poll (or fetch the cheap fields via REST) and only request the expensive fields for the focused workspace.
3. Detect rate limiting (429 / `x-ratelimit-remaining`, or the error text) and pause polling until `x-ratelimit-reset` rather than exponential backoff capped at 5 min.
4. Optionally support a dedicated `GH_TOKEN` for the daemon so it doesn't share the user's interactive quota.

## Repro

1. Open ~50+ workspaces in one project against a GitHub remote, each on a branch with a PR.
2. Leave Paseo running for an hour; watch `gh api rate_limit --jq .resources.graphql`.
3. Once `remaining` hits 0, all workspaces lose PR status and `~/.paseo/daemon.log` fills with the `GitHubCommandError` above.

Happy to share the full log excerpt or test a patched build.

dezchai:
## Problem

Paseo retains GitHub PR-status polls for every active workspace. The previous per-workspace path spent multiple GraphQL-backed `gh` calls per target, exhausting the account-wide 5,000-point hourly budget around 50–60 workspaces and breaking the user's own `gh` and agent workflows. Fixes #3587 and addresses #2470.

## Change

- Replace independent target timers with one host-grouped scheduler and aliased GraphQL batches of up to 25 targets.
- Discover the matching PR first, then fetch checks only for selected PRs. Terminal checks on closed or merged PRs are cached by repository, PR, and head OID.
- Process chunks serially per host. Every response reads GitHub's advertised GraphQL limit, remaining points, and reset time; polling reserves 40% for user and agent traffic and stops until reset when that reserve is reached.
- When GitHub rejects a request for rate limiting, read the GraphQL reset time from the rate-limit endpoint once and pause the entire host instead of running per-target retries.
- Resolve fork PRs through the parent repository and cache that address, keeping later fork polls in the shared batch.
- Treat missing or malformed repository and check connections as failed aliases so an unreadable response cannot clear a workspace's PR status.
- Preserve partial-response recovery for unrelated alias failures and the existing `ForgeService` polling interface. No protocol or app changes.

## Evidence

- `npx vitest run packages/server/src/services/github-service.test.ts --bail=1` — 106 tests passed.
- `npm run lint` — passed with zero warnings.
- Live GitHub GraphQL smoke verified the advertised budget fields, fork-to-parent identity, parent-side fork PR lookup, and checks-query shape against this PR.

The focused coverage includes multi-workspace batching, open and terminal check cadence, host separation, chunking, partial failures, malformed boundaries, fork redirects across cycles, crowded candidate pages, rate-reserve pauses, and reset-aware recovery after a rejected request.

### DIFF
--- a/docs/forge-providers.md
+++ b/docs/forge-providers.md
@@ -218,3 +218,8 @@ are stale, run `npm run build:server`.
   (`manual`) from blocking jobs (`manual` plus `action_required`). A failed
   allowed-failure job carries the `warning` trait, and `canceling` remains active
   until GitLab reports a terminal status.
+- GitHub PR polling owns one account-wide GraphQL budget. Coordinate retained
+  targets per host, batch their reads, and stop until GitHub's reset time when
+  the reserve is exhausted. Never add a per-target GitHub request to the poll
+  path. Resolve fork PRs through their parent repository without abandoning the
+  shared batch.
--- a/packages/server/src/services/forge-cli-command.ts
+++ b/packages/server/src/services/forge-cli-command.ts
@@ -151,6 +151,12 @@ export interface ForgeCommandFailureParams {
   cwd: string;
   exitCode: number | null;
   stderr: string;
+  /**
+   * Whatever the failing command printed to stdout. GraphQL batch callers use
+   * this to salvage partial `data` when one aliased sub-query errors but the
+   * others succeeded (the CLI exits non-zero yet still prints the body).
+   */
+  stdout?: string;
 }
 
 export class ForgeCommandError extends Error {
@@ -159,6 +165,7 @@ export class ForgeCommandError extends Error {
   readonly cwd: string;
   readonly exitCode: number | null;
   readonly stderr: string;
+  readonly stdout?: string;
 
   constructor(label: { brand: string; binary: string }, params: ForgeCommandFailureParams) {
     super(`${label.brand} CLI command failed: ${label.binary} ${params.args.join(" ")}`);
@@ -166,6 +173,9 @@ export class ForgeCommandError extends Error {
     this.cwd = params.cwd;
     this.exitCode = params.exitCode;
     this.stderr = params.stderr;
+    if (params.stdout !== undefined) {
+      this.stdout = params.stdout;
+    }
   }
 }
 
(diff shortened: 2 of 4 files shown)


MACscr:
Thanks for addressing this as I have ran into this issue many times.

PR #3856, opened 2026-08-25
fix(server): leave plan mode when Copilot Allow All is selected

### LINKED ISSUE #3855: bug: selecting Copilot "Allow All" while in Plan mode leaves the session in plan mode
## What's broken

Selecting **Allow All** while a Copilot session is in Plan mode does not leave plan mode. Paseo shows Allow All, but the ACP session mode stays `#plan`: the model keeps receiving its instructions framed as plan mode and refuses to edit anything. The only way out is to pick Agent instead.

## Steps to reproduce

1. Copilot session, switch to Plan, ask for a plan.
2. Switch the mode to Allow All.
3. Ask it to implement the plan — it keeps planning and refuses to touch files.

## Evidence

Driving `copilot --acp` directly (Copilot CLI 1.0.80). `mode` and `allow_all` are two independent config options, and setting the flag does not change the mode:

```
[config_option_update] mode=plan  allow_all=on
```

Permission prompts do stop (2 → 0 for the same file write), but the model reports its instructions still arrive prefixed `[[PLAN]]` and answers: *"Per plan mode rules, I cannot create or modify files until the plan is approved."*

In `copilot-acp-agent.ts`, `writeCopilotProviderMode` only sets `allow_all=on` and returns `handled: true`. In `acp-agent.ts`, `setModeWithSelection` then sets `currentMode` locally, emits `mode_changed` and returns — before `setSessionMode` is ever called. So the session mode is never changed, while the UI reports the new one.

## Related: the native `#autopilot` mode is hidden

Paseo filters `#autopilot` out of `availableModes`, remaps its id to `#agent`, and synthesises "Allow All" from the `allow_all` flag instead. Autopilot is the mode that changes both at once: entering it sets `allow_all=on`, leaving it reverts to `off`.

```
set_mode autopilot:   mode=autopilot  allow_all=on
set_mode #agent:      mode=agent      allow_all=off
```

On 1.0.80 it also suppresses prompts by itself — a shell exec plus a file write ran with **zero** `session/request_permission` calls. #929, which motivated hiding it, does not reproduce on this build.

KiraPC:
Closes #3855

## Problem

Selecting Allow All while a Copilot session is in Plan mode left the session in `#plan`: Paseo reported Allow All while the model kept its plan-mode framing and refused to edit. `writeCopilotProviderMode` only flipped the `allow_all` config option and returned `handled: true`, so `setModeWithSelection` returned before `setSessionMode` was called.

## Fix

- Allow All now writes the session mode (`#agent`) as well as `allow_all=on`. Mode first, then the flag: leaving `#autopilot` makes Copilot revert `allow_all` to `off`, so the reverse order clobbers the write.
- `#autopilot` is no longer filtered out of the mode list, and its id is no longer remapped to `#agent`.
- The current mode is resolved from the session mode before the `allow_all` flag, so an autopilot session is no longer reported as Allow All.

## QA

Ordering verified against the real CLI (Copilot CLI 1.0.80), raw `config_option_update` after each step:

```
mode first, flag second  ->  ✅
  set_mode autopilot:              mode=autopilot  allow_all=on
  set_mode #agent:                 mode=agent      allow_all=off
  set_config_option allow_all=on:  mode=agent      allow_all=on

flag first, mode second  ->  ❌
  set_config_option allow_all=on:  mode=autopilot  allow_all=on
  set_mode #agent:                 mode=agent      allow_all=off
```

From `#plan` neither order touches the flag, so `#autopilot` is the only starting state where it matters.

**Platforms:** implemented and unit-tested on macOS; the ACP captures above were taken on Linux with Copilot CLI 1.0.80. Not tested on Windows; the change is not platform-specific.

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -33,11 +33,11 @@ import type { ProcessTerminator, TreeKillTarget } from "../../../utils/tree-kill
 import {
   COPILOT_AGENT_FEATURE_OPTION,
   COPILOT_ALLOW_ALL_MODE_ID,
+  COPILOT_AUTOPILOT_MODE_ID,
   COPILOT_MODES,
   CopilotACPAgentClient,
   beforeCopilotModeWriter,
   transformCopilotConfigOptions,
-  transformCopilotModeId,
   transformCopilotSessionResponse,
   writeCopilotProviderMode,
 } from "./copilot-acp-agent.js";
@@ -297,7 +297,6 @@ function createCopilotSessionWithConfig(
       sessionResponseTransformer: transformCopilotSessionResponse,
       configOptionsTransformer: transformCopilotConfigOptions,
       configFeatureOptions: [COPILOT_AGENT_FEATURE_OPTION],
-      modeIdTransformer: transformCopilotModeId,
       providerModeWriter: writeCopilotProviderMode,
       beforeModeWriter: beforeCopilotModeWriter,
       capabilities: {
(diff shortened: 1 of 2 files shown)


boudra:
Closing for now: this needs a before/after Paseo-to-Copilot run showing that Plan → Allow All permits the requested edit, with a regression test through the ACP session. Please reopen with that evidence.

boudra:

PR #3860, opened 2026-08-25
feat(notifications): decouple sound from OS and add floating screen & in-app notifications

johngbl:
## Summary
- **Independent notification audio:** decoupled sound from the OS notification center using Web Audio API / HTMLAudioElement so notifications play sound even when Windows OS notifications are disabled.
- **Custom notification sound:** crafted a soft, warm two-tone chime (D4 293 Hz -> G4 392 Hz) without harsh harmonics.
- **Screen floating notifications (Zed-style):** dedicated transparent always-on-top window on desktop when the app is minimized or unfocused, with official Paseo branding, title, body, close action, and click-to-focus navigation to the target agent/workspace.
- **In-app toast notifications:** sleek in-app toast card rendered inside the app window when the app is focused, avoiding duplication or ghost notifications.
- **100% test coverage:** unit tests for notification sound, in-app notification store, and OS notifications.

## Tests
- All 16 unit tests passing (100% green).
- Typecheck clean.
- Lint clean.

PR #2582, opened 2026-07-29
fix(desktop): restore notification sounds

gstamp:
### Linked issue

Refs #1841
Closes #1465

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Desktop task notifications were always marked silent, so the operating system could display them without the expected alert sound. Desktop notifications now play sound by default, while users who prefer silent alerts can turn sound off without changing system-wide notification settings.

Notification permission, sound, and testing belong to one workflow, so they are grouped in a dedicated Notifications settings section. The test action reports loading, failure, and persistent success states so users can tell whether Paseo handed the notification to the operating system.

### Goals

- Play the operating system notification sound for desktop notifications by default.
- Let users turn desktop notification sound off and preserve that choice across restarts.
- Group notification permission, sound, and test controls in one settings section.
- Give the test notification action clear loading, error, and success feedback.

### Non-goals

- Change Android or iOS notification delivery.
- Override operating-system notification or focus settings.
- Claim that the operating system displayed a notification after Electron accepted it.

### QA

- Contributor manually verified the original sound fix on Windows 11 in the desktop development app.
- Maintainer ran the macOS Electron development app and verified the dedicated Notifications section, notification permission state, default-on Play sound toggle, test notification delivery, loading state, persistent success feedback, and success styling.
- Verified the Permissions section now contains only microphone permission.
- `npx vitest run packages/desktop/src/settings/desktop-settings.test.ts packages/desktop/src/settings/desktop-settings-commands.test.ts packages/app/src/utils/host-routes.test.ts packages/app/src/i18n/resources.test.ts packages/app/src/hooks/use-settings/storage.test.ts packages/app/src/desktop/permissions/desktop-permissions.test.ts --bail=1` — 6 files and 133 tests passed.
- `npm run typecheck` passed.
- `npm run lint` passed with 0 warnings and 0 errors.
- `npm run format` and `git diff --check` passed.

Risk surface: Electron notification creation, persisted desktop settings, the desktop settings navigation, translated settings copy, and the shared success Alert colors/border.

### Checklist

- [x] One focused change
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence
- [x] Tests added or updated where it made sense

Issue #1841, opened 2026-07-01
bug: notification sound has never taken effect.


winezer0:
### What's broken

Under both Windows and Android systems, no sound alerts were triggered after the tasks were completed.

Under the Android system, no notification was triggered even after the task was completed.

The current version is v0.1.102

Hope to prioritize the resolution of this issue

### Steps to reproduce

1. Start the ai task
2. Wait for the background task to complete
3. No notification was made by sound.

### Where did this happen

Android app

### Paseo version

0.1.102

### OS version

Andorid 16 and windows 11

### Agent provider

OpenCode

### Provider configuration

opencode  1.17.12
codex-cli 0.142.4

download:
https://github.com/getpaseo/paseo/releases/download/v0.1.102/Paseo-Setup-0.1.102-x64.exe
https://github.com/getpaseo/paseo/releases/download/v0.1.102/paseo-v0.1.102-android.apk

### Logs

```text
[2026-07-01 09:53:52.301] [warn]  [login-shell-env] failed; keeping inherited env {
  reason: 'win32',
  durationMs: 0,
  timeoutMs: 10000,
  error: 'login shell env is not resolved on Windows',
  errorCode: null,
[...]

Issue #1465, opened 2026-06-11
macOS notifications for agent events have no sound


shouwangzhe:
## Description

Agent-related notifications (agent finished, permission request) arrive in the macOS notification center but **without any sound**, even when "Play sound for notifications" is enabled in System Settings → Notifications → Paseo.

## Steps to reproduce

1. Enable notification sound in macOS System Settings → Notifications → Paseo → "播放通知的声音" (Play sound for notifications)
2. Switch away from the Paseo app
3. Trigger an agent completion or permission request
4. Notification banner appears — **no sound plays**

## Expected behavior

Notifications should play the system notification sound, matching what macOS users expect.

## Notes

- The "Test Audio" button in Paseo settings does produce sound, so audio output itself works fine.
- The notification badge and banner content display correctly.
- macOS Sequoia (Darwin 25.5.0)
- Paseo desktop app (latest)

PR #3872, opened 2026-08-26
feat(app,desktop): let users pick a custom notification sound

JuanRdBO:
### Linked issue

Refs #1841, #1465 (closed). Related: #3860 takes notification audio in a complementary direction (decoupled synthesized chime + floating/in-app notifications); this PR is only about letting the user pick their own sound file for the existing desktop notification path.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

The desktop "Play sound" toggle (from #2582) can only play the OS default notification sound. When you run several agents all day, that sound is the same one every other app makes — you can't tell "agent finished" from "calendar reminder", and you can't make it something you actually want to hear. This PR adds one setting under the existing Play sound switch: pick an audio file, and finished-agent notifications play that instead of the OS default.

### Goals

- `notifications.customSoundPath: string | null` in desktop settings (default `null`, old settings files unaffected).
- A "Notification sound" row in Settings → Notifications, visible only while Play sound is on: shows the chosen file name or "System default", Choose opens the native file dialog (mp3/wav/m4a/aac/ogg/flac/aiff), Reset clears back to the OS default.
- When a custom sound is set, the Electron notification is created `silent: true` and the renderer plays the file (Chromium decodes all listed formats), so the OS default and the custom sound never double up.
- The existing "Send test notification" button exercises the same path, so it previews the chosen sound.
- Files are read in the main process, capped at 5 MB, and any failure (missing file, oversized, unreadable) degrades to a silent no-op.

### Non-goals

- No change to mobile/web notification paths (native and plain-browser code paths are untouched).
- No bundled sound library or synthesized chime (see #3860 for that direction).
- No per-event sounds (different sound per notification type).
- Trade-off: if the configured file becomes unreadable, the notification shows silently rather than falling back to the OS sound — kept simple on purpose; happy to add the fallback if you'd rather.
- Trade-off: the renderer caches the resolved sound; the cache is invalidated by the settings handlers that can change it, not reactively.

### QA

Automated (all run locally on macOS, paste-trimmed):

```
$ npm run typecheck   → all 8 workspaces clean
$ npm run lint -- <21 changed files>   → 0 warnings, 0 errors
$ npx vitest run packages/desktop/src/settings/desktop-settings.test.ts --bail=1   → 16 passed
$ npx vitest run packages/desktop/src/features/notifications.test.ts --bail=1      → 15 passed
$ npx vitest run packages/app/src/utils/os-notifications.test.ts --bail=1          → 13 passed
$ npx vitest run packages/app/src/i18n/resources.test.ts --bail=1                  → 35 passed
$ npx vitest run packages/app/src/hooks/use-settings/storage.test.ts --bail=1      → 71 passed
```

Test coverage added: settings coercion (path survives restart byte-for-byte incl. whitespace, `null`/`""` clear, an invalid patch cannot overwrite a persisted valid path, pre-existing settings files load with `null`); `resolveCustomSound` and `planNotificationSound` (full mime table, the silent/playsCustomSound matrix, no file read when sound is off or unset, missing file, size-cap boundary both sides); `readBoundedFile` against real temp files (short-read-safe, over-cap files rejected end-to-end with an injected tiny cap); renderer playback (fetch only when the send reports a custom sound, cache never retains a failed/null resolution, `Audio.play()` rejection is a silent no-op that retries next send, legacy boolean send-result tolerance).

The diff was additionally put through two independent model review passes (architecture/defect lens); findings fixed in place: bounded file reads (a single `fs.read` can return short — reads now loop to EOF/cap), the send IPC now returns `{ shown, playsCustomSound }` so the renderer plays only when the main process actually silenced the OS sound (no double-sound race), no caching of failed sound resolutions, a pending state on the Choose button, and byte-for-byte path preservation in settings coercion.

Manual QA (macOS, `npm run dev:desktop`): exercised the full flow in the running Electron dev app — Settings → Notifications → Choose (native dialog, audio-filtered) → picked an mp3 → Send test notification plays the custom sound with the OS notification silent (no double-play) → Reset returns to "System default" and the test notification plays the OS sound again. Video of the whole interaction:

https://github.com/user-attachments/assets/04b6d997-8c78-45e2-a188-8ad7c094e83c

**Windows and Linux untested** — the playback path is Chromium `Audio` + a plain `fs` read, so it should be platform-neutral, but I could only verify macOS.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence (automated + manual macOS run with video; Windows/Linux untested)


boudra:

JuanRdBO:
> Closing this feature PR. Please discuss the workflow and shared need in [Discussions](https://github.com/getpaseo/paseo/discussions), following [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

That is one way to thank contributors 👍

PR #3886, opened 2026-08-26
feat(app): give each sidebar status group its own icon

cleiter:
### Type of change

- [x] Enhancement

### Reasoning

Two of the five sidebar status group headers do not say much on their own. Working uses a circle-with-a-dot, the same shape the workspace rows use for "there is a state here", so it reads as a generic status pip rather than "busy". Ready to review and Done share the same check icon and are told apart only by colour. Each state now gets its own glyph, so the header stack is readable by shape.

### Goals

- Each of the five status group headers renders a distinct glyph, so the state is readable without relying on colour.
- Every header glyph keeps lucide's standard ring, so the five marks read as one icon set at 14pt.
- Header colours are unchanged and still resolve through `getStatusDotColor`, so each header keeps the exact hue its workspace rows use.

The mapping after this change:

| Bucket | Label | Glyph |
| --- | --- | --- |
| `needs_input` | Needs input | `CircleAlert` (unchanged) |
| `failed` | Failed | `CircleX` (unchanged) |
| `running` | Working | `Clock` |
| `attention` | Ready to review | `CircleArrowRight` |
| `done` | Done | `CircleCheck` (unchanged) |

image

### Non-goals

- **The animated `StatusRing` on workspace rows, tabs, and project tiles.** That is activity motion rather than a glyph, and changing it would desync the two halves of the running indicator.
- **`CircleDot` anywhere else.** It keeps its own meaning in the composer, the git check presentation, the task list row, and the message view.
- **A test.** `packages/app/vitest.config.ts` aliases `lucide-react-native` to a stub where every export is the same object, so no unit test can tell one glyph from another. The browser tier could assert the rendered SVGs differ, but the headers only exist for buckets that have workspaces, so it needs a five-bucket fixture and a new testID. Deferred as disproportionate to a glyph swap.
- **Reworking `getStatusDotColor`.** Colours are out of scope; this change is shape only.

Some candidates were rejected for colliding with an existing meaning in the app: `Eye` (shown vs hidden), `CircleDot` (in progress), `Archive` (a real Paseo lifecycle concept), `CircleStar` (`Star` is the default agent-profile glyph). `Clock` also appears in the display-preferences menu for "timestamp"; that ambiguity is accepted, since a menu option icon and a status header glyph are far apart and a clock for "busy" is near-universal.

### QA

image

Tested: browser web, Electron desktop.
Not tested: iOS, Android.

To reproduce the full header stack you need one workspace in each of the five buckets — `buildStatusGroups` omits empty buckets, so a plain dev run only shows the states you happen to have.

```
npm run typecheck   # 0 errors
npm run lint        # 0 warnings, 0 errors
npm run format:check # all matched files use the correct format
```

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:

PR #3897, opened 2026-08-26
fix(claude): add a selectable Opus 5 1M model entry

0-Captain:
## Problem

#3000: since v0.2.4, Opus 5 at 1M context cannot be selected from the model
picker, and the remaining `claude-opus-5` entry advertises a window that
third-party endpoints do not deliver.

#2497 (`f91a984`) removed the `claude-opus-5[1m]` manifest entry and flipped the
plain entry from `200_000` to `1_000_000`, on the premise that Opus 5 is always 1M.

That premise holds on a first-party endpoint, where Claude Code resolves Opus 5's
native 1M window without the suffix. It does not hold when `ANTHROPIC_BASE_URL`
points at a gateway or proxy: the client only sends the `context-1m-2025-08-07`
beta for the suffixed id, so a picker-started session runs at 200k while the
catalog says 1M.

Every other family already ships both spellings — `claude-sonnet-5` beside
`claude-sonnet-5[1m]`, same for opus-4-8, 4-7 and 4-6. Opus 5 is the only one
that lost its 1M entry.

## Fix

Add `claude-opus-5[1m]` back as its own manifest entry, same shape as the other
paired families. `claude-opus-5` keeps `defaultPriority: 2`, so the default
selection is unchanged.

No alias is added: a real entry already makes a persisted `claude-opus-5[1m]`
preference resolve to itself, which is the only thing an alias would have bought.

## Relationship to #3727

#3727 targets the same bug but adds `aliases: ["claude-opus-5[1m]"]` on the plain
entry *and* a separate `claude-opus-5[1m]` manifest entry. Those collide:
`getClaudeManifestModels()` emits one definition per alias *and* one per manifest
entry, so the catalog ends up with a duplicate id. Running that branch:

```
entries whose id startsWith 'claude-opus-5': 3
   id="claude-opus-5"     label="Opus 5"    isSelectable=undefined isDefault=true
   id="claude-opus-5[1m]" label="Opus 5"    isSelectable=false
   id="claude-opus-5[1m]" label="Opus 5 1M" isSelectable=undefined
DUPLICATE ids in catalog: [["claude-opus-5[1m]",2]]
```

Its own updated assertions expect two Opus 5 entries, so those tests do not pass
as written. No checks have run on that branch.

## QA evidence

Platform: macOS (Darwin 25.6.0), Node 24.19.0. Server + CLI only; no app/UI code
is touched by this change, so no screenshots.

```
packages/server $ CLAUDE_CONFIG_DIR=<empty> npx vitest run \
    src/server/agent/providers/claude/models.test.ts
Test Files  1 passed (1)      Tests  37 passed (37)

packages/server $ CLAUDE_CONFIG_DIR=<empty> npx vitest run \
    src/server/agent/providers/claude/agent.test.ts
Test Files  1 passed (1)      Tests  74 passed (74)

packages/cli $ npx tsx tests/15-provider.test.ts
✓ Test 6: provider models claude lists canonical model aliases

packages/server $ npm run typecheck      # exit 0
$ npm run lint -- <the four changed files>   # 0 warnings, 0 errors
```

Note on `models.test.ts`: `createCatalogClient` does not pass `configDir`, so
`getClaudeModelsWithSettings` reads the *real* `~/.claude/settings.json` of
whoever runs the suite. On a machine that sets `ANTHROPIC_DEFAULT_OPUS_MODEL`,
`addSettingsModel` injects that id and the version-gating test fails — on
unmodified `main` too. Not touched here; flagging it as a separate pre-existing
issue.

### DIFF
--- a/packages/cli/tests/15-provider.test.ts
+++ b/packages/cli/tests/15-provider.test.ts
@@ -58,6 +58,11 @@ const EXPECTED_CLAUDE_MODELS = [
     descriptionFragment: "Latest release",
   },
   {
+    id: "claude-opus-5[1m]",
+    model: "Opus 5 1M",
+    descriptionFragment: "1M context window",
+  },
+  {
     id: "claude-fable-5",
     model: "Fable 5",
     descriptionFragment: "Most powerful",
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -425,6 +425,7 @@ describe("ClaudeAgentClient.fetchCatalog", () => {
 
       expect(models.map((m) => m.id)).toEqual([
         "claude-opus-5",
+        "claude-opus-5[1m]",
         "claude-fable-5",
         "claude-fable-5[1m]",
         "claude-opus-4-8[1m]",
--- a/packages/server/src/server/agent/providers/claude/model-manifest.ts
+++ b/packages/server/src/server/agent/providers/claude/model-manifest.ts
@@ -46,6 +46,16 @@ export const CLAUDE_MODEL_MANIFEST = [
     supportsFastMode: true,
   },
   {
+    id: "claude-opus-5[1m]",
+    label: "Opus 5 1M",
+    description: "Opus 5 with 1M context window",
+    minimumClaudeCodeVersion: "2.1.219",
+    contextWindowMaxTokens: 1_000_000,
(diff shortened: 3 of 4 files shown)


0-Captain:
@boudra This PR fixes #3000 (adds a selectable Opus 5 1M model entry) and has been open for a couple of days without CI running or review. Could you take a look when you get a chance? Note there's also #3727 targeting the same issue.

PR #3899, opened 2026-08-26
feat(client): expose paseo.projects.list() through the public SDK

### LINKED ISSUE #3834: Expose registered projects through the public plugin SDK
## Problem

The public Paseo plugin API exposes `paseo.workspaces.list()` and `paseo.workspaces.create()`, but it does not provide a supported way to list registered Projects.

A plugin can infer Projects from existing Workspace descriptors, but this omits registered Projects that currently have no active Workspaces.

The daemon and protocol already support `project.list`, and workspace responses may include `emptyProjects`, but neither capability is exposed as a typed public API through `PaseoApi`.

This prevents plugins from presenting a complete Project picker before creating a Workspace.

## Expected behavior

Expose a supported, typed public API such as:

```ts
paseo.projects.list()
```

The result should include all registered Projects, including Projects with zero active Workspaces, with at least:

- `projectId`
- display name
- root path
- project kind

An equivalent typed API that provides the complete Project catalog would also solve the problem.

## Current workaround

Plugins must either infer Projects from existing Workspaces, which produces an incomplete list, or depend on internal/undeclared interfaces. Neither is a reliable public SDK contract.

omercnet:
## Problem

Plugins have to infer projects from active workspaces, so they miss registered projects that have no active workspaces.

## Change

Expose `paseo.projects.list()` through the public SDK. This only surfaces the existing `project.list` RPC; it adds no protocol messages or daemon behavior.

Fixes #3834

## QA evidence

```text
$ npm run typecheck
> paseo@0.5.2 typecheck
> npm run typecheck --workspaces --if-present
...
> @getpaseo/cli@0.5.2 typecheck
> tsgo --noEmit

$ npm run lint
> paseo@0.5.2 lint
> oxlint

Finished in 664ms on 3796 files with 177 rules using 16 threads.

$ npx vitest run packages/client/src/index.test.ts --bail=1
Test Files  1 passed (1)
Tests  12 passed (12)
Duration  3.59s
```

## Platforms tested

SDK-only and platform-independent. Verified through the client SDK unit test and repository-wide typecheck and lint.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -139,8 +139,8 @@ RPC contracts validate inputs and outputs in both the app and plugin subprocess.
 typed async function. Use the host-provided `@tanstack/react-query` for request state and caching;
 Paseo gives each plugin installation its own query client.
 
-`usePaseo()` and the handler's `{ paseo }` context expose the same `PaseoApi`: workspaces, agents,
-providers, and daemon config. They do not expose connection lifecycle. A surface borrows the
+`usePaseo()` and the handler's `{ paseo }` context expose the same `PaseoApi`: projects,
+workspaces, agents, providers, and daemon config. They do not expose connection lifecycle. A surface borrows the
 selected host's existing connection; switching the screen's host changes both `usePaseo()` and
 `useRpc()` to that host. An offline selected host fails there and never falls through to another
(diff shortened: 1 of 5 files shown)


boudra:
@paseo-bot /deliver

boudra:
@paseo-bot deliver

PR #3902, opened 2026-08-26
fix(server): plugin daemon sessions see all providers by pinning the daemon appVersion

### LINKED ISSUE #3763: bug: plugin daemon sessions only see claude/codex/opencode agents — plugin client sends no appVersion, so provider visibility falls back to the legacy allow-list

omercnet:
## What's broken

Plugin-owned daemon sessions omit `appVersion`, so `packages/server/src/server/session.ts` treats them as legacy clients and filters every provider except Claude, Codex, and OpenCode. Calls such as `paseo.agents.list()` therefore silently omit Pi, OMP, Copilot, Cursor, ACP, and other agents.

## Fix

The daemon now passes its resolved version from `packages/server/src/server/bootstrap.ts` through `PluginService`, `PluginRuntime`, and the internal `PluginProcessRequest` initialize message. `packages/server/src/server/plugins/plugin-process.ts` supplies that value as the plugin `DaemonClient`'s `appVersion`.

This keeps the compatibility filter unchanged for genuinely old external clients while stating the actual invariant directly: a plugin subprocess is pinned to the daemon version that spawned it. No public protocol or wire schema changed.

The plugin Paseo API E2E coverage now creates a Pi agent and verifies that the plugin's own `agents.list()` call returns it. Removing the `appVersion` assignment makes this test fail with `agentIds: []`.

Fixes #3763

### Regression proof before the fix

```text
$ npx vitest run packages/server/src/server/plugins/plugin-paseo-api.e2e.test.ts --bail=1

FAIL  packages/server/src/server/plugins/plugin-paseo-api.e2e.test.ts > plugin handlers create workspaces and agents through their Paseo API
AssertionError: expected { agentIds: [] } to deeply equal { agentIds: ArrayContaining{…} }

Test Files  1 failed (1)
Tests       1 failed (2)
EXIT=1
```

### Regression test after the fix

```text
$ npx vitest run packages/server/src/server/plugins/plugin-paseo-api.e2e.test.ts --bail=1

✓ packages/server/src/server/plugins/plugin-paseo-api.e2e.test.ts (2 tests) 1360ms
  ✓ plugin handlers create workspaces and agents through their Paseo API 903ms
  ✓ daemon config reload enables and disables configured plugins without restarting 457ms

Test Files  1 passed (1)
Tests       2 passed (2)
Duration    6.86s
EXIT=0
```

### Typecheck

```text
$ npm run typecheck

@getpaseo/expo-two-way-audio: tsgo --noEmit
@getpaseo/highlight: tsgo --noEmit
@getpaseo/plugin: tsgo --noEmit && tsgo -p tsconfig.examples.json --noEmit
@getpaseo/protocol: tsgo --noEmit
@getpaseo/client: tsgo --noEmit
@getpaseo/server: tsgo -p tsconfig.server.typecheck.json --noEmit
@getpaseo/app: tsgo --noEmit
@getpaseo/relay: tsgo --noEmit
@getpaseo/website: tsgo --noEmit
@getpaseo/desktop: tsgo --noEmit -p tsconfig.json
@getpaseo/cli: tsgo --noEmit
EXIT=0
```

### Lint

```text
$ npm run lint

> paseo@0.6.1 lint
> oxlint

Finished in 673ms on 3796 files with 177 rules using 16 threads.
EXIT=0
```

### Format

```text
$ npm run format

> paseo@0.6.1 format
> oxfmt .

Finished in 953ms on 4086 files using 16 threads.
EXIT=0
```

## Platforms tested

- Linux: daemon and plugin subprocess E2E regression test.
- macOS: not tested; daemon-only cross-platform TypeScript change.
- Windows: not tested; daemon-only cross-platform TypeScript change.
- iOS, Android, web, Electron UI: not affected.

## Supersedes

- [PR #3765 — Earlier plugin provider-visibility fix](https://github.com/getpaseo/paseo/pull/3765) — This PR takes over the same plugin agent-listing workflow while passing the daemon’s already-resolved version into the subprocess.


boudra:
@paseo-bot /deliver

boudra:
@paseo-bot deliver

boudra:
@paseo-bot deliver

PR #3913, opened 2026-08-26
fix(app): align the scrolling diff order with the Changes tree

cleiter:
### Type of change

- [x] Bug fix

### Reasoning

In the Changes view the tree rail and the scrolling diff list the same files in a different order, so scrolling the diff does not walk the tree from top to bottom. The tree puts directories before files at each level of the path, while the diff compares the whole path byte by byte, so a root file such as `.gitlab-ci.yml` opens the diff but sits below every directory in the tree. Reviewing a change means reading one order and clicking in the other. This PR makes the tree the single ordering authority and derives the flat order from it.

### Goals

- The scrolling diff lists files in the same order as the Changes tree, on every platform.
- One ordering rule exists in one place: `sortTree` in `git/diff-tree.ts`. The flat list derives from the tree instead of reimplementing the rule.
- Clicking a tree row still scrolls to the matching file, and the folder collapse state still works.

### Non-goals

- Commit diffs are unchanged. `CommitDiffPanel` renders no tree rail, so there is no second order to disagree with, and sorting it would only replace git's own order with ours.
- The daemon keeps its whole-path sort in `checkout-git.ts`. Fixing this in the client means a daemon that is not updated also gets the fix, per `docs/protocol-compatibility.md`. The CLI and other clients keep the old order.

### QA

**Repro.** A checkout with changes in both a root-level file and a directory. The tree rail showed `bin/…`, `samera-one/…`, then `.gitlab-ci.yml` and `process-compose.yaml`. The scrolling diff opened with `.gitlab-ci.yml`.

**Before / after.** Desktop app (Electron, macOS), the same worktree with an untracked root file `foo.txt` next to changed directories. Before, `foo.txt` is first in the diff and last in the tree. After, it is last in both, and the diff reads top to bottom in tree order.

Before:

image

After:

image

I also clicked tree rows after the change and confirmed each one still scrolls to its file, and opened a commit from the Commits list and confirmed its file order is unchanged.

**Tested on:** desktop (Electron, macOS). **Not tested by hand on:** iOS, Android, browser web. The browser end-to-end test below runs the Changes view in a real browser against a real daemon.

**The test that proves it.** `packages/app/e2e/browser/changes-pane.spec.ts` compares what the tree renders against what the diff renders. To show the test is not vacuous I reverted `diff-order.ts` to the old comparator and ran it again. The tree assertion still passed and the diff assertion failed with the reported order:

```
Expected: ["changed.ts", "root.ts", "use-mounted-tab-set.ts", "a-root-note.txt"]
Received: ["a-root-note.txt", "use-mounted-tab-set.ts", "changed.ts", "root.ts"]
```

The old fixture could not have caught this. It only wrote `src/…` files plus a root `zz-untracked.txt`, and `"src" < "zz"`, so that file sorts last under both rules. The fixture now writes `a-root-note.txt`, which sorts first as a whole string and last in the tree. The test also asserts equal row counts in both surfaces, because `flattenDiffTree` drops the children of a collapsed folder, which would make the tree a subsequence of the diff and hide a real difference.

**Tests:**

- `packages/app/src/git/diff-order.test.ts` — rewritten around `orderCheckoutDiffFiles`: sibling files, a directory before a file beside it, loose root files below root directories (the reported case), a nested directory above a shallower file with the same prefix, a directory ahead of a file sharing its name, lists too short to reorder, and stability on equal paths. 7 tests, green.
- `packages/app/e2e/browser/changes-pane.spec.ts` — the tree-versus-diff order agreement test above. All 33 tests in the file green.

**Commands:**

- `npm run typecheck` — pass
- `npm run lint` — pass
- `npm run format:check` — pass
- `npx vitest run packages/app/src/git/diff-order.test.ts packages/app/src/git/diff-tree.test.ts` — 22 pass
- `npx playwright test changes-pane.spec.ts --project=browser` — 33 pass

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/changes-pane.spec.ts
+++ b/packages/app/e2e/browser/changes-pane.spec.ts
@@ -19,6 +19,10 @@ interface WorkspaceFixtureOptions {
   includeDeletedFile?: boolean;
   includeNestedFolders?: boolean;
   includeRenamedFile?: boolean;
+  // A root-level file whose name sorts BEFORE the root-level "src" directory.
+  // zz-untracked.txt cannot stand in for it: "src" < "zz", so that file lands
+  // last whether paths are compared whole or segment by segment.
+  includeRootFileSortingFirst?: boolean;
   includeUntrackedFile?: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


boudra:
@paseo-bot deliver

PR #3917, opened 2026-08-26
Close agent runtime on terminal error instead of leaking its process tree

frontend-london:
## Summary

Incident: on the `agents-worker` host, 5 Devin sessions sat in lifecycle `error` for ~3.5h, each leaking ~7 child processes (`devin acp` + duplicated `mcp-server-slack`) — 35 processes total, causing memory pressure and event-loop stalls in `paseo.service`. Processes were manually reaped and cgroup limits (`MemoryHigh=12G`, `MemoryMax=16G`, `MemorySwapMax=2G`) were set as a stopgap.

**Root cause**: `AgentManager.finalizeForegroundTurn` and `AgentManager.onStreamTurnFailed` (`packages/server/src/server/agent/agent-manager.ts`) set `agent.lifecycle = "error"` on a terminal turn failure but never called anything to tear down the underlying provider session. `docs/agent-lifecycle.md` already documents `error → closed` as the intended state diagram, and `AgentManager.closeAgent` → `closeAgentRuntime` already implements exactly this teardown (`agent.session.close()`, which tree-kills the whole process group — that's why fixing this also cleans up the duplicated MCP children, since they're descendants of the same process tree) for archive/reload. Nothing routed the `error` transition through it.

## Changes

- Both places lifecycle flips to `"error"` now call a new `closeAgentAfterTerminalError(agentId)`, which fires `closeAgent(agentId)` as a tracked background task (same pattern as other post-`emitState` cleanup in this file) and swallows a "runtime already gone" rejection instead of letting it crash the daemon or leak an unhandled rejection.
- `lastError` and the `error` attention reason are preserved onto the closed record (verified this is what the UI's state-bucket derivation already keys off, independent of raw `status`), so the failure stays visible even though `lastStatus` becomes `"closed"`.
- Retrying an errored agent now goes through the normal `ensureAgentLoaded()` resume path and gets a fresh provider process, instead of the old behavior of reusing an in-memory session object with no live process behind it.
- `docs/agent-lifecycle.md`: noted terminal error as a runtime-closing trigger alongside archive/reload/etc.

## Why not just kill the process and leave lifecycle at "error"?

Traced through `ensureAgentLoaded()` (`agent-loading.ts`) and the `deriveAgentStateBucket` UI logic before implementing:
- If the agent stayed live in `this.agents` with a dead session, the next prompt would hit `agent.session.startTurn()` on a closed connection and throw confusingly, instead of resuming cleanly.
- The UI already renders `"failed"` off `attentionReason === "error"` independent of `status`, so collapsing to `"closed"` doesn't hide the error from users.
- This matches the docs' `error → closed` diagram and reuses proven teardown code instead of adding a second, parallel process-killing path.

## Test plan

- [x] New regression test: `terminal error closes the agent runtime (kills the provider session) instead of leaking it` — asserts the session's `close()` (the real tree-kill teardown in ACP) is called exactly once, the agent is no longer live, the closed record keeps `attentionReason: "error"`, and retry via `ensureAgentLoaded` spins up a distinct fresh session.
- [x] New regression test: `a successful turn does not close the agent runtime` — no regression for a normal completing turn.
- [x] Updated 4 pre-existing tests that assumed an errored agent stays live and reusable in-place (`clearAgentAttention on errored agent...`, `streamAgent clears pending run when startTurn fails...`, the two `turn_failed` system-error-message tests) to route retries through `ensureAgentLoaded`, matching production behavior.
- [x] `npx vitest run src/server/agent/agent-manager.test.ts` — 145/145 passing
- [x] Related suites (`agent-prompt`, `chat-mentions`, `agent-loading`, `agent-archive`, `lifecycle-command`, `create-agent-lifecycle-dispatch`, `session*`, `workspace-archive-service`, `provider-snapshot-manager`, `mcp-server`, `agent-projections`) — all passing
- [x] `npm run typecheck` (full monorepo, via pre-commit hook) — clean
- [x] `npm run lint` / `npm run format` — clean
- [x] `git diff --check` — clean

Not deployed. This PR is for review only.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -22,7 +22,11 @@ same Paseo agent ID. Provider history is not appended again when the canonical t
 primed.
 
 Idle agents remain resident indefinitely. Runtime closure happens only through an explicit lifecycle
-action such as archive, replacement, reload, workspace teardown, or daemon shutdown.
+action such as archive, replacement, reload, workspace teardown, or daemon shutdown -- or the terminal
+`error` transition itself, which closes the runtime the same way archive does (killing the provider
+process tree) instead of leaving a dead session's process resident forever. `lastError` and the
+`error` attention reason survive onto the closed record so the failure stays visible; retrying goes
+through `ensureAgentLoaded()`'s normal resume path and gets a fresh provider process.
 
 ### Cancellation
 
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -5757,55 +5757,64 @@ test("onAgentAttention is not called for delegated child agents", async () => {
   expect(attentionCalls).toEqual([]);
 });
 
-test("clearAgentAttention on errored agent stays cleared until a new error transition", async () => {
+test("a turn failure raises attention as error before the runtime closes", async () => {
   const workdir = mkdtempSync(join(tmpdir(), "agent-manager-attention-error-"));
   const storagePath = join(workdir, "agents");
   const storage = new AgentStorage(storagePath, logger);
 
   class FailingSession extends TestAgentSession {
-    private attempt = 0;
+    closeCalls = 0;
 
     override async startTurn(): Promise<{ turnId: string }> {
-      this.attempt += 1;
-      const attempt = this.attempt;
-      const turnId = `fail-turn-${attempt}`;
+      const turnId = "fail-turn-1";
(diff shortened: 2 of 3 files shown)


frontend-london:
Closing — this was opened against upstream by mistake. We're going to maintain and deploy this fix locally on our own fork/branch (frontend-london/paseo, fix/close-agent-runtime-on-terminal-error) rather than pursue an upstream merge. Not intended for further review here; no action needed from maintainers.

PR #3926, opened 2026-08-27
feat(app): show host battery next to the hostname

Tommypop2:
The host badge now reads `fedora - 37%`.

In the mobile workspace header, under the workspace name:

> **battery demo**
> batt-demo · ▤ fedora - 77%

## Why?

I tend to leave my laptop unplugged sometimes when using Paseo remotely from my phone. I thought seeing my laptop's battery percentage would be a nice-to-have so I know when I need to return to my laptop to plug it in.

Totally understand if this feature is a little too niche and happy to close if that is the case :)

## Agent-written how it works

The daemon samples its own charge and broadcasts `host_battery` when the whole-number percent moves. `server_info` carries the last reading so a connecting client paints immediately instead of waiting out a tick. The sampler skips the probe entirely while nobody is connected, and stops for good once a host reports no battery.

**Linux** reads sysfs directly — no subprocess. It combines multiple packs by total charge rather than averaging their percentages, and filters to `BAT*` because the power-supply class also carries peripherals: a wireless mouse registers as `hidpp_battery_2`, and folding that into a laptop's charge reports a number belonging to neither device. **Other platforms** go through `systeminformation` (zero runtime deps, pure JS, so Docker/F-Droid/bundling are unaffected), which shells out to pmset/WMI with parsers already debugged across OS versions. Windows treats an empty WMI result as a *retryable failure* rather than settled absence, since elevation and "no battery" look identical there.

## Protocol

Three states stay distinct on the wire and must remain so: absent means unmeasured, `null` means no battery, a value means it has one. Collapse the first two and every laptop looks like a desktop for the first seconds after connect.

`percent` is an unconstrained `z.number()` on purpose — a range or integer check would reject the whole enclosing message if a daemon ever reported a fraction. Clients clamp and round at render, through `Intl.NumberFormat`, so French gets `37 %` and Arabic its own numerals.

Both additions are COMPAT-tagged. Wire-compat tests cover an old client parsing new `server_info`, and an old client tolerating a `host_battery` status it doesn't know.

## Client

Battery lives beside `serverInfo` in the session store rather than inside it: it changes on its own clock and would otherwise rewrite that object every few minutes and wake every reader. `HostBadgeModel` stays a pure function of the host registry and appearance settings — the badge reads live charge itself.

## Known limitation

With default settings this is invisible on a **local-only** setup: `resolveHostBadgeDisplay` defaults a local host to `"hidden"`, so no badge renders and the battery has nothing to attach to. Mobile-to-remote works by default (remote hosts default to `"name"`). Decoupling battery from `badgeDisplay` is a deliberate follow-up rather than something snuck into this PR.

## Verification

Verified on a real Fedora laptop, not just in tests:

- **Live daemon socket** — `features.hostBattery: true`, `server_info` seeded with a reading, `host_battery` push delivered on change.
- **Probe arithmetic** — energy-counter ratio matched the kernel's own `capacity` exactly (67.35% vs 67%).
- **Peripheral exclusion** — the machine has a `hidpp_battery_2` mouse alongside `BAT0`; correctly ignored, and now a regression test.
- **Real app UI** — Expo web at 390x844 against the dev daemon, opened into a workspace, header rendered `batt-demo · fedora - 77%`.

336 tests pass across the affected suites. Lint, format, and a full workspace typecheck are clean.


boudra:

Issue #3256, opened 2026-08-12
GJC/ACP agent remains stuck running after cancel timeout and foreground-turn restart race


probepark:
## Foreground-turn investigation cases

This tracker coordinates cancellation and foreground-turn failures. Cases remain separately assessable; matching error wording does not establish a common cause. All observations below are reporter evidence, not independently reproduced here.

| Case | Environment and trigger | Observed failure and evidence | Completion criterion |
| --- | --- | --- | --- |
| A — GJC cancellation | macOS Desktop/daemon 0.3.1, GJC via ACP; cancel then retry | Original report below: timeout at 2026-08-12 14:05:13 +09:00, ACP active-turn rejection, persisted running with no matching process | Cancel/retry reaches a truthful terminal state and accepts a later prompt without stranded ownership; check persistence/restart |
| B — GJC concurrent send/fan-out | 0.3.1, gjc-codex-pro via ACP; send_agent_message_request or parent/child notification while running | [Detailed follow-up](https://github.com/getpaseo/paseo/issues/3256#issuecomment-5267064128): cancellation timeout, child durable error, caller notification failures; another run stayed running | Concurrent sends and notifications do not strand or falsely fail agents; capture terminal events and durable state |
| C — Rewind/abort | Provider, platform and versions unreported; execute /rewind, abort, then send | [Original #349](https://github.com/getpaseo/paseo/issues/349): foreground-turn-active error and all subsequent messages fail; reported April 14, no regression version established | Reproduce exact sequence with versions/provider and post-failure diagnostics; subsequent prompt works after cancellation |
| D — Native Codex scheduling case | GPT 5.6 Sol; coordinate one subagent, then schedule a message; platform/versions unreported | [August 9 report](https://github.com/getpaseo/paseo/issues/349#issuecomment-5230456501): same error with GPT but not Claude; native Codex classification carried from prior evaluation, exact provider configuration still to confirm | Capture scheduling and foreground events with exact provider/version; scheduled message and later prompts remain usable |
| E — Antigravity long command | Antigravity via ACP; shell command longer than 10 seconds; platform/versions unreported | [September 4 report](https://github.com/getpaseo/paseo/issues/349#issuecomment-5538283499): foreground-turn-active error | Reproduce >10s command with cancellation/terminal logs; completion and next prompt work without stranded turn |
[...]

PR #3937, opened 2026-08-27
fix(server): recover unresponsive agent stops

### LINKED ISSUE #3540: Cancel gives up after a 2s RPC interrupt and never escalates, leaving the agent permanently stuck
### Summary

When an agent's CLI subprocess stops responding to RPC, cancelling the turn is impossible from the UI. The daemon sends an interrupt over RPC, waits **2 seconds**, gives up, and never escalates to a signal. The agent stays `running` with `withActiveForegroundTurn: 1` indefinitely, and the user has no way to recover it short of `docker exec … kill` or restarting the container.

I hit this with an agent wedged on a hung tool call. The turn was stuck for ~60 minutes and every cancel click was a no-op.

### Environment

- Paseo daemon **0.3.1**, container built on `ghcr.io/getpaseo/paseo`
- Agent provider: `omp` (`@oh-my-pi/pi-coding-agent` 17.2.15), spawned as `--mode rpc-ui`
- The subprocess was alive but blocked (~0.6% CPU, no I/O) with an unresolved tool call, so its RPC loop never serviced the interrupt

### Reproduction

1. Have an agent provider process block inside a turn such that it stops servicing RPC (in my case a tool call that never settled — any uninterruptible hang works).
2. Press cancel/stop in the UI.

### Actual

Five cancel attempts produced only:

```
{"level":40,"component":"agent-manager","agentId":"044b7acc-…",
 "timeoutMs":2000,"msg":"Timed out interrupting session during cancel"}
{"level":40,"component":"agent-manager","agentId":"044b7acc-…",
 "msg":"Session interrupt failed after timeout during cancel"}
```

After that the daemon takes no further action. The agent remained `lifecycle: running` / `withActiveForegroundTurn: 1` for an hour. Nothing in the UI could clear it.

Sending `SIGINT` to the subprocess from outside resolved it **instantly** — the process exited cleanly with code 130, wrote a complete session-exit record preserving its pending state, and the daemon then handled it correctly:

```
{"level":40,"component":"agent-manager","turnId":"4dd2dce3-…","lifecycle":"running",
 "error":"OMP RPC process exited with code 130 and signal null","msg":"handleStreamEvent: turn_failed"}
```

So the recovery path works fine once the process is signalled — the daemon just never signals it.

### Expected

When the RPC interrupt times out, escalate rather than stop:

1. `SIGINT` the child process (graceful — the provider CLI handles it cleanly and flushes session state).
2. If still alive after a grace period, `SIGTERM`, then `SIGKILL`.

### Suggestions

- Escalate to signals automatically after the RPC interrupt times out. A cancel that silently does nothing is worse than a forceful one, since the alternative for the user is losing the whole container.
- Expose a **force stop** in the UI for an agent whose interrupt has already failed, so recovery doesn't require shell access to the host.
- Consider making the 2000ms interrupt timeout configurable, and surfacing "interrupt failed" in the UI instead of only in daemon logs — from the user's side the button simply appeared to do nothing, with no indication that anything had failed.

### Impact

The agent is unrecoverable through the product. Anyone without shell access to the container has to restart Paseo, which disrupts every other agent in it.

edihasaj:
Fixes #3540.

Related: #3256 and #2679.

## Summary

- Escalate an unresponsive ACP provider process through `SIGINT → SIGTERM → SIGKILL` after its normal interrupt fails.
- Settle the stale manager run, close the compromised provider session, and reload the durable session before accepting the next prompt.
- Surface recovery failures as attention errors instead of leaving an agent permanently `running`.

## Testing

- `npx vitest run packages/server/src/server/agent/lifecycle-command.test.ts packages/server/src/server/agent/agent-manager.test.ts packages/server/src/server/agent/providers/acp-agent.test.ts --bail=1` — 275 passed.
- `npm run build --workspace=@getpaseo/server` — passed.
- `npm run lint` — passed.

## Risk

Force recovery deliberately terminates an unresponsive provider process only after the normal provider interrupt rejects or times out. It retains the provider persistence handle and resumes the same Paseo agent session; if resumption fails, the agent visibly enters an error state.

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -42,9 +42,10 @@ into that contract; lifecycle callers do not interpret provider-specific errors.
 
 After an acknowledged interrupt, the manager settles the captured run even when no terminal event
 arrives or the run was still waiting for its provider turn id. The captured run token prevents an
-older cancellation from settling a newer turn. If interruption is rejected or times out, the agent
-keeps its active foreground turn and replacement, reload, rewind, and Stop report the failure.
-Accepting new work after an ambiguous interruption would create a split-brain session.
+older cancellation from settling a newer turn. If the normal interrupt rejects or times out, Paseo
+escalates the provider runtime when it supports direct process interruption, then closes and reloads
+the session. A successful recovery leaves the agent idle and ready for the next prompt. If the
+session cannot be restored, the agent becomes an attention error instead of remaining `running`.
 
 ## Relationships
 
--- a/packages/server/src/server/agent/agent-manager.test.ts
+++ b/packages/server/src/server/agent/agent-manager.test.ts
@@ -1194,6 +1194,8 @@ class McpCapableTestAgentClient extends TestAgentClient {
 
 class ControlledInterruptSession extends TestAgentSession {
   interruptCalled = false;
+  forceInterruptCalled = false;
+  closeCalled = false;
 
   constructor(
     config: AgentSessionConfig,
@@ -1214,6 +1216,14 @@ class ControlledInterruptSession extends TestAgentSession {
     this.interruptCalled = true;
     await this.interruptBehavior(this);
   }
+
+  override async forceInterrupt(): Promise<void> {
+    this.forceInterruptCalled = true;
+  }
+
+  override async close(): Promise<void> {
+    this.closeCalled = true;
+  }
 }
 
 interface ControlledInterruptFixture {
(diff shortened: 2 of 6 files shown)


boudra:
Closing for now: this needs a real ACP before/after stop-recovery run showing the old process exits, the durable session survives, and a follow-up prompt completes. Please reopen with that evidence.

boudra:

PR #3959, opened 2026-08-28
docs(cli): fix remote host option placement

### LINKED ISSUE #3949: README's --host usage is wrong: it is a per-subcommand flag, not a global option (0.4.0 and 0.6.1)
The README documents remote-daemon usage as a global option placed *before* the subcommand:

```
paseo --host workstation.local:6767 run "run the full test suite"
```

That form does not work — `--host` is registered per-subcommand, so it must follow the verb:

```
$ paseo --host 127.0.0.1:6767 ls
error: unknown option '--host'

$ paseo ls --host 127.0.0.1:6767          # works
$ paseo send <id> --host <h>:6767 "..."   # works
$ paseo run -d --host <h>:6767 --cwd /path "task"   # works
```

Reproduced on **0.4.0** and again on **0.6.1** (macOS, Node 26 and Node 22), against both a local daemon and a remote daemon over a private network — so this isn't a transient regression.

Two smaller things noticed alongside it, both probably documentation rather than code:

1. **`paseo status` doesn't accept `--host` at all** (it suggests `--home`). It reports only the local daemon, so there's no obvious way to health-check a remote one — `paseo ls --host …` works as a substitute. If that's intended, it's worth stating, since `status` is the natural first thing to reach for.

2. **`paseo run --host …` requires `--cwd`.** Without it the remote daemon appears to inherit the *caller's* working directory and fails with `Directory not found: /Users/<me>/...` — a path that exists on the calling machine, not the remote one. Sensible once understood, but the error points at the wrong machine; a note in the remote-usage docs would save others the same few minutes.

Happy to send a README patch if useful — the fix is just moving the flag after the verb in the example.

ZacharyZcR:
## Summary

- place `--host` after each CLI subcommand in README and configuration examples
- clarify that remote `run` paths are resolved on the daemon host and include `--cwd` in the example
- document that these are per-command options and that `status` does not support `--host`
- keep the translated README examples in sync

## Validation

- `npm run format:check:files -- README.md README.zh-CN.md README.ja.md README.ko.md public-docs/configuration.md public-docs/cli.md`
- `npm run typecheck`
- `git diff --check`

Closes #3949

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -101,8 +101,8 @@ paseo ls                           # 実行中のエージェントを一覧表�
 paseo attach abc123                # ライブ出力をストリーミング
 paseo send abc123 "also add tests" # 追加タスクを送信
 
-# リモートデーモンで実行
-paseo --host workstation.local:6767 run "run the full test suite"
+# リモートデーモンで実行（パスはリモートマシン上で解決されます）
+paseo run --host workstation.local:6767 --cwd /path/to/project "run the full test suite"
 ```
 
 詳細は[完全な CLI リファレンス](https://paseo.sh/docs/cli)を参照してください。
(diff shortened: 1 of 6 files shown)


boudra:
Closing in favor of the implementation merged in [#4238](https://github.com/getpaseo/paseo/pull/4238).

PR #3972, opened 2026-08-28
fix(app): stop the model browser from using sheet scrollables outside a sheet

colonelpanic8:
## Problem

Opening the model picker on a non-compact native screen — a tablet, an unfolded foldable, or a phone in landscape — crashes the app to the error boundary with:

```
'useBottomSheetInternal' cannot be used out of the BottomSheet!
```

`CombinedModelSelector` picks the browser's scroll mode from the platform (`isWeb ? "independent" : "sheet"`), but `Combobox` only presents as a bottom sheet when the breakpoint is compact. Anywhere else it is a popover, so `ModelBrowser`'s `BottomSheetScrollView` renders with no sheet context to read and gorhom throws. (`BottomSheetScrollView` reaches `useBottomSheetInternal()` before its own friendlier `'Scrollable' cannot be used out of the BottomSheet!` check, which is why the message names the hook.)

Fixing only that leaves the picker rendering but inert on Android: the popover lives inside a plain RN `Modal`, whose children mount into a separate native view hierarchy outside the app's root `GestureHandlerRootView`. Gesture handlers there never receive touches, and `ModelBrowser`'s independent-scrolling rows are `GestureDetector` taps with no `Pressable` fallback — so every press in the picker silently does nothing.

## Fix

Two commits:

1. Resolve the scrolling mode from the render scope, the way #3942 made text inputs do it: a `"sheet"` request holds only where a sheet actually is, and degrades to `"independent"` — exactly what web already uses for the popover — anywhere else. The mapping is one-directional, so a caller asking for `"independent"` inside a sheet is left alone. `resolveModelBrowserScrolling` is pure and unit tested; `ModelBrowser` reads the existing `useIsInsideBottomSheet()` scope.
2. Wrap the `Combobox` popover modal's content in its own `GestureHandlerRootView`, the same repair `AdaptiveModalSheet` already carries for its own modal path, so those rows are pressable.

## Test plan

- `npx vitest run src/components/model-browser-view.test.ts` (17 passed)
- `npm run typecheck` (all workspaces)
- `oxlint` / `oxfmt --check` on the touched files
- Android phone: model picker opens as a popover, rows scroll and respond to taps

### DIFF
--- a/packages/app/src/components/model-browser-view.test.ts
+++ b/packages/app/src/components/model-browser-view.test.ts
@@ -4,10 +4,11 @@ import type {
   ProviderSelectorProvider,
 } from "@/provider-selection/provider-selection";
 import {
+  clampModelBrowserScrolling,
   resolveInitialModelBrowserView,
   resolveModelBrowserAllView,
-  groupProfilesByProviderModel,
   resolveModelBrowserScrolling,
+  groupProfilesByProviderModel,
 } from "./model-browser-view";
 
 function provider(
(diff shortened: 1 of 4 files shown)


boudra:
Closing in favor of the model-picker scrolling fix merged in [#4392](https://github.com/getpaseo/paseo/pull/4392).

PR #4004, opened 2026-08-28
fix(server): batch initial GitHub PR status polls

dezchai:
## Problem

The shared GitHub PR-status scheduler has three startup paths that can leave every workspace without visible PR status:

1. The first poll is due immediately while repository host and slug discovery runs asynchronously. Sending unresolved targets through the legacy path recreates the account-wide GraphQL exhaustion that #3825 was intended to prevent.
2. When the account is already below the polling reserve, the discovery batch succeeds but its response was discarded before PR identity was published.
3. Waiting for repository identity across the entire due set lets one non-settling workspace block every ready workspace indefinitely, so the daemon makes no GitHub request at all.
4. Workspace Git snapshots warm through a bounded FIFO queue after restart. A selected workspace can remain behind the entire background backlog, leaving its branch and PR metadata absent even though the PR poller itself is healthy.
5. Foreground priority cannot make every sidebar badge immediate: reconstructing all 157 local checkout identities still took roughly three to four minutes in the packaged app.

## Change

- Defer only targets whose repository identity is still resolving; retry each target when its own reads settle.
- Let ready targets form a shared batch immediately instead of falling back or waiting for unrelated workspaces.
- Preserve the legacy fallback for repositories that settle without a usable origin slug.
- Ignore targets unsubscribed while identity discovery is in flight.
- When a successful discovery response crosses the reserve, publish PR identity and merge state while skipping the follow-up checks request until reset.
- Raise the GraphQL reserve from 40% to 80%, leaving four-fifths of the advertised budget for users and agents.
- Add regression coverage for delayed identity resolution, one permanently unresolved target beside a ready target, low-budget startup, and the 80% reserve boundary.
- Make background warming local-Git-only; remote forge status remains owned by the batched poller instead of running one forge/auth read per workspace during snapshot startup.
- Replace FIFO-only workspace refresh admission with a bounded foreground/background queue. Background warming may occupy three of the four slots; the fourth remains available for direct snapshot requests. Foreground work promotes at both layers: the workspace refresh queue and the lower-level Git process scheduler. Neither concurrency ceiling increases.
- Upgrade an already-running background refresh when its workspace becomes foreground, so later Git commands in the same pipeline enter the high-priority lane instead of retaining their startup priority.
- Give foreground restoration a five-second grace period before background warming begins, matching the measured Electron cold-restore window. Defer each workspace's observation setup until its initial local snapshot settles so watcher setup cannot preempt restoration.
- Keep initial checkout and PR-status RPCs local-only; the retained poller publishes remote PR state asynchronously instead of blocking workspace readiness on a direct forge query.
- Collect independently resolved repository identities on the next shared five-second grid boundary so startup produces real multi-target batches instead of concurrent one-target GraphQL calls.
- Persist a versioned, 24-hour stale-while-revalidate sidebar cache under `$PASEO_HOME/runtime`. It contains only exact workspace path, branch/remote identity, forge id, and the last PR/check summary—not dirty state, diffs, refs, ahead/behind counts, or complete Git snapshots.
- Render cached PR badges in the initial workspace-list response, then validate them through local warming and the retained forge poller. A branch/remote mismatch removes the entry immediately; an authoritative null or updated status removes or replaces it.
- Coalesce a startup batch into one atomic cache write instead of writing once per workspace.

## Quota impact

Measured against real pull requests using the exact pre-#3825 and batched query shapes. GitHub reported these primary-rate-limit costs through `rateLimit.cost`:

| Targets | Before #3825 | Batched | Savings |
| ---: | ---: | ---: | ---: |
| 1 | 2 points | 2 points | 0% |
| 5 | 10 points | 2 points | 80% |
| 10 | 20 points | 2 points | 90% |
| 25 | 50 points | 2 points | 96% |

The old matched-PR path cost one point for `gh pr view` with checks plus one point for merge/auto-merge facts: two points per target per cycle. The batched path cost one point for discovery plus one point for checks for as many as 25 targets: two points per full batch.

For `N` ready targets:

- Old matched-PR cost: approximately `2N` points per cycle.
- New open-PR cost: approximately `2 × ceil(N / 25)` points per cycle.
- New merged/closed cost after checks are cached: approximately `ceil(N / 25)` points per cycle.

At a full 25-target batch, the open-PR path saves 48 of 50 points per cycle (96%, or 25× less quota). Cached terminal PRs save 49 of 50 points (98%, or 50× less quota).

The reserve is a separate hard bound. With GitHub's usual 5,000-point user allowance, the 80% reserve permits background polling to consume approximately 1,000 points and preserves approximately 4,000 for users and agents. Query cost is controlled by GitHub and can change; see [GitHub's GraphQL rate-limit documentation](https://docs.github.com/en/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api).

## Evidence

- Computer Use reproduced neutral workspace dots with no PR badges in the packaged build.
- The live daemon received `checkout_pr_status_request` but made zero GitHub calls. A known workspace was confirmed to exactly match its merged PR by branch and head SHA, isolating the global identity wait.
- Direct query-cost measurement returned one point for the old status query, one point for the old facts query, one point for batched discovery, and one point for batched checks at batch sizes 1, 5, 10, and 25.
- `npx vitest run packages/server/src/services/github-service.test.ts --bail=1` — 108 tests passed.
- `npx vitest run packages/server/src/server/workspace-git-service.test.ts --bail=1` — 36 tests passed, including foreground admission plus cache identity reuse, invalidation, and retained-poller persistence.
- `npx vitest run packages/server/src/server/workspace-sidebar-status-cache.test.ts --bail=1` — 4 tests passed, covering version/expiry validation, coalesced writes, deletion, and real atomic persistence/restore.
- `npx vitest run packages/server/src/server/session.workspaces.test.ts --bail=1` — 107 tests passed (4 skipped), including cached PR status in the first cold workspace-list response.
- `npx vitest run packages/server/src/server/workspace-git-service.primitive.test.ts --bail=1` — 55 tests passed.
- Background registration coverage proves initial snapshots do not issue per-workspace forge reads; generic non-GitHub adapters start their own self-heal poll immediately.
- `npx vitest run packages/server/src/utils/run-git-command.test.ts --bail=1 -t "carries high-priority admission"` — passed, covering async propagation into the Git process scheduler.
- `npx vitest run packages/server/src/server/session/checkout/checkout-session.test.ts --bail=1` — 38 tests passed, covering local-only initial status projection.
- `npm run lint` — passed with zero warnings or errors.
- The rebuilt arm64 artifact passed the packaged desktop smoke: renderer, preload, isolated daemon, bundled CLI, and terminal smoke all succeeded.
- In the packaged app, the first cache-enabled run populated the versioned file through real retained GitHub polls. On the next cold restart, cached PR numbers and check states were visible with the restored workspace list in under nine seconds, versus roughly three to four minutes without the disk cache; background Git and GitHub validation continued afterward.

Follow-up to #3825.


boudra:
Thanks for following up. I reproduced the cold-start batching escape and merged the focused fix in #4025. I split persisted forge metadata into #4031 so it can be evaluated separately from polling correctness. Closing this PR in favor of those two focused tracks.

PR #4016, opened 2026-08-29
Hide Pi subagent control reports from chat

evaera:
Pi subagent control reports need to remain in the native session and model context, but Paseo currently displays them as ordinary assistant messages. This clutters the conversation and makes internal orchestration look like an agent response.

This keeps those reports in Pi’s context while omitting them from the visible transcript. Ordinary extension messages and subagent lifecycle updates are unchanged.


boudra:
I prefer this Pi extension-specific projection to live in a provider plugin. Provider plugins can own the mapping from native session data to the visible timeline. That API is on main for the upcoming 0.8 release: [provider plugin guide](https://paseo.sh/docs/plugins/v0.8/providers). Closing the core PR in favor of that approach.

PR #4021, opened 2026-08-29
fix(server): support downloads of files with non-ASCII names

lgjx123:
# fix(server): support downloads of files with non-ASCII names

## Problem

Downloading a file whose **filename contains non-ASCII characters** (e.g. Chinese names like `要求.txt`) through the file explorer always fails:

- The daemon writes the raw UTF-8 basename into the HTTP header:
  ```js
  res.setHeader("Content-Disposition", `attachment; filename="要求.txt"`);
  ```
- Node rejects non-Latin-1 characters in header values with `ERR_INVALID_CHAR: Invalid character in header content ["Content-Disposition"]`.
- The error is swallowed by the handler's catch block, which then returns a misleading `404 {"error": "File not found"}` for an existing file.
- Desktop clients (e.g. the Windows app) crash/black-screen while handling this failed download and must be restarted.

ASCII filenames (e.g. `1.PNG`) are unaffected, which is why only some files fail. Note the directory path is irrelevant — only the basename appears in the header.

## Fix

Emit an ASCII-safe `filename=` fallback plus the RFC 5987 `filename*=UTF-8''...` form when the name contains non-ASCII characters, so the header stays a valid ASCII value and clients (browsers, Electron, WebView2) still save the file under its original name:

```
attachment; filename="__.txt"; filename*=UTF-8''%E8%A6%81%E6%B1%82.txt
```

ASCII filenames keep the exact same header as before (no behavior change).

## QA evidence

- `cd packages/server && npm run typecheck` → passes (exit 0)
- e2e: `npx vitest run src/server/daemon-e2e/file-download.e2e.test.ts --maxWorkers=1` → **5/5 passed**:

```
✓ ... > file download tokens > issues token over WS and downloads via HTTP 373ms
✓ ... > file download tokens > downloads a file whose name contains non-ASCII characters 127ms
✓ ... > file download tokens > rejects invalid token 76ms
✓ ... > file download tokens > rejects expired token 395ms
✓ ... > file download tokens > rejects paths outside the workspace cwd 106ms
Test Files  1 passed (1)
Tests       5 passed (5)
```

- Regression proof: with the fix reverted, the new test fails exactly as the reported bug does:

```
× ... > downloads a file whose name contains non-ASCII characters 382ms
  → expected 404 to be 200 // Object.is equality
AssertionError: expected 404 to be 200 // Object.is equality
```

- Platforms tested: Linux (server side, e2e daemon). Not tested: actual Windows/macOS desktop client UI (the server-side header fix is platform-independent).

## Test added

`packages/server/src/server/daemon-e2e/file-download.e2e.test.ts` — new case "downloads a file whose name contains non-ASCII characters" that requests a token for `要求.txt`, downloads it, asserts HTTP 200, the `filename*=UTF-8''...` disposition, and the file body.

### DIFF
--- a/packages/server/src/server/bootstrap.ts
+++ b/packages/server/src/server/bootstrap.ts
@@ -811,8 +811,18 @@ export async function createPaseoDaemon(
       }
 
       const safeFileName = entry.fileName.replace(/["\r\n]/g, "_");
+      // Node rejects non-Latin-1 characters in header values (ERR_INVALID_CHAR),
+      // so a raw UTF-8 basename in Content-Disposition makes every download of a
+      // file whose name contains non-ASCII characters fail. Emit an ASCII-safe
+      // filename= fallback plus the RFC 5987 filename*=UTF-8'' form for clients
+      // that support it (browsers, Electron, WebView2).
+      const hasNonAscii = /[^\x20-\x7E]/.test(safeFileName);
+      const fallbackFileName = safeFileName.replace(/[^\x20-\x7E]/g, "_");
+      const disposition = hasNonAscii
+        ? `attachment; filename="${fallbackFileName}"; filename*=UTF-8''${encodeURIComponent(safeFileName)}`
+        : `attachment; filename="${safeFileName}"`;
       res.setHeader("Content-Type", entry.mimeType);
-      res.setHeader("Content-Disposition", `attachment; filename="${safeFileName}"`);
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the Content-Disposition package implementation in [#3409](https://github.com/getpaseo/paseo/pull/3409), which is still open.

PR #4043, opened 2026-08-29
fix(app): commit default model for slow-probing providers

### LINKED ISSUE #3961: bug: Kiro New Workspace crashes ("모델을 선택하세요" in buildDraftAgentSnapshot) unless model is re-selected
### What's broken

On the iOS app (v0.5.2), starting a **New Workspace with the Kiro provider** shows a model already selected in the composer (the persisted default, `claude-opus-4.8`). But sending the first message throws a fatal client-side error and drops to the "Paseo에 문제가 발생했습니다 / Something went wrong" screen:

```
Name: Error
Message: 모델을 선택하세요        // "Select a model"

Error: 모델을 선택하세요
    at buildDraftAgentSnapshot (.../main.jsbundle)
    at t20 (.../main.jsbundle)
    at useDraftAgentCreateFlow (.../main.jsbundle)
    at WorkspaceDraftAgentTab (.../main.jsbundle)
    at renderWithHooks (.../main.jsbundle)
    at updateFunctionComponent (.../main.jsbundle)
    at beginWork (.../main.jsbundle)
    at performUnitOfWork (.../main.jsbundle:21150:25)
    at workLoopSync (.../main.jsbundle)
    at renderRootSync (.../main.jsbundle)
    at performWorkOnRoot (.../main.jsbundle)
    at performSyncWorkOnRoot (.../main.jsbundle)
    at flushSyncWorkAcrossRoots_impl (.../main.jsbundle)
    at processRootScheduleInMicrotask (.../main.jsbundle)
```

A model is clearly displayed in the composer, so `buildDraftAgentSnapshot` should be able to build a valid snapshot. Instead it throws "select a model" — the displayed default model is **not committed into the draft-agent state** on the Kiro path.

**This only happens with the Kiro provider.** Other providers auto-populate the draft model, so the first send works. Workaround: after the crash, tap "다시 시도" (Try again), manually open the model selector and pick the *same* model, then send again — it works. So the model value only lands in draft state after an explicit manual selection.

Expected: when the composer already shows a default/remembered model, the New Workspace draft should carry that model into `buildDraftAgentSnapshot` and create the agent without requiring a manual re-selection.

### Steps to reproduce

1. iOS app (v0.5.2) connected to a daemon (v0.2.5).
2. Start a **New Workspace**; provider = **Kiro**. The composer already shows a default model (`claude-opus-4.8`), mode = Chat.
3. Type the first prompt and send it **without** manually opening the model picker.
4. The app crashes to the error screen with `Error: 모델을 선택하세요` (stack above).
5. Tap "다시 시도" (Try again), open the model selector, pick the same model, and send again → it works.

Reproduces only with the Kiro provider. This looks like the same New-Workspace draft-state family as:

- #3316 (closed) — New Workspace forwards an unsupported `modeId` for Pi; draft state not reconciled against the resolved provider.
- #3584 — remembered provider dropped in new workspaces after a snapshot error.
- #623 — respect model/mode defaults in workspace creation.
- #3217 — same `WorkspaceDraftAgentTab` / draft-create flow.

Here the unreconciled field is the **model** for the Kiro provider.

### Where did this happen

iOS app

### Paseo version

0.5.2 (iOS app); daemon 0.2.5

noahg9:
### Linked issue

Fixes #3961

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Starting a New Workspace with a slow-probing custom/ACP provider (e.g. Kiro) showed a default model but crashed on send. `buildDraftAgentSnapshot` threw for a null `selectedProvider`. The provider sits in `status:"loading"` during its ACP probe, so the ready-only guards dropped an early pick, and the one-shot `COMPLETE_RESOLUTION` never reconciled once the provider became ready.

The fix applies to any enabled provider still probing, with no provider-id branching. Providers are now committable while `ready` or `loading`. Once the committed provider becomes ready, its default model is backfilled from the catalog; this is keyed on an empty model slot, so a user's own choice is never overridden. A missing provider is gated at submit rather than thrown during render.

### Goals

- Send with the shown default on a still-probing provider — no crash, no manual re-selection.
- Keep committed draft state and the model-selector display in agreement.

### Non-goals

- Change server-side Kiro/ACP probe timing or the snapshot `status`/`enabled` model.
- Touch composer flows outside the New Workspace draft-create path.

### QA

Reproduced at the hook level with a deterministic loading→ready transition. Pre-fix: `expected null to be 'kiro'` (provider dropped) and `expected '' to be 'claude-opus-4.8'` (model never backfilled). After the fix:

```text
npx vitest run packages/app/src/hooks/use-agent-form-state.test.tsx packages/app/src/provider-selection/resolve-agent-form.test.ts packages/app/src/provider-selection/provider-selection.test.ts --bail=1   # 91 passed
npm run typecheck   # passed
npm run lint        # 0 warnings, 0 errors (changed files)
npm run format      # passed
```

Not exercised on-device against a real Kiro session — verified by unit and regression tests only.

### Risk surface

Draft provider/model resolution in the New Workspace path. Ready-only providers behave as before; disabled/unavailable picks are still dropped; the reconcile is idempotent and never overrides a user-chosen model.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/composer/draft/create-flow.ts
+++ b/packages/app/src/composer/draft/create-flow.ts
@@ -93,7 +93,7 @@ interface UseDraftAgentCreateFlowOptions<TDraftAgent, TCreateResult> {
   onBeforeSubmit?: (ctx: CreateRequestContext) => Promise<void> | void;
   onCreateStart?: () => void;
   createRequest: (ctx: CreateRequestContext) => Promise<CreateRequestResult<TCreateResult>>;
-  buildDraftAgent: (attempt: CreateAttempt) => TDraftAgent;
+  buildDraftAgent: (attempt: CreateAttempt) => TDraftAgent | null;
   onCreateSuccess: (ctx: { result: TCreateResult; attempt: CreateAttempt }) => Promise<void> | void;
   onCreateError?: (error: Error) => void;
 }
--- a/packages/app/src/composer/draft/workspace-tab.tsx
+++ b/packages/app/src/composer/draft/workspace-tab.tsx
@@ -156,7 +156,7 @@ async function submitDraftCreateRequest(input: {
     featureValues: Record<string, unknown> | undefined;
   };
   hostDisconnectedMessage: string;
-  selectModelMessage: string;
+  providerNotSelectedMessage: string;
 }): Promise<{ agentId: string | null; result: AgentSnapshotPayload }> {
   const {
     attempt,
@@ -179,7 +179,7 @@ async function submitDraftCreateRequest(input: {
 
   const provider = autoSubmitConfig?.provider ?? composerState.selectedProvider;
   if (!provider) {
-    throw new Error(input.selectModelMessage);
+    throw new Error(input.providerNotSelectedMessage);
   }
   const modeIdOverride = resolveDraftModeIdOverride({
     autoSubmitConfig,
@@ -227,8 +227,7 @@ function buildDraftAgentSnapshot(input: {
     selectedProvider: string | null;
     agentControls: { features?: Agent["features"] };
   };
-  selectModelMessage: string;
-}): Agent {
+}): Agent | null {
   const { attempt, serverId, tabId, workspaceDirectory, autoSubmitConfig, composerState } = input;
   invariant(workspaceDirectory, "Workspace directory is required");
   const now = attempt.timestamp;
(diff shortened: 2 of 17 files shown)


boudra:
Closing in favor of the implementations merged in [#4332](https://github.com/getpaseo/paseo/pull/4332) and [#4401](https://github.com/getpaseo/paseo/pull/4401).

PR #4083, opened 2026-08-30
fix(claude): use finalized usage for empty iterations

### LINKED ISSUE #4019: bug(claude): contextWindowUsedTokens only reports output_tokens for backends returning empty iterations
## What's broken

The context-window meter's **used** value (`agent.lastUsage.contextWindowUsedTokens`) severely understates actual usage for the Claude provider when the upstream backend returns a result `usage` with an **empty `iterations` array** — which non-Anthropic-native backends (e.g. GLM via a translating proxy) always do on normal (non-tool-use) turns.

In this case the meter reports only the turn's `output_tokens` instead of `input_tokens + cache_read_input_tokens + output_tokens`, so it looks like ~0.5% used when the real figure is ~33%.

The model itself is NOT affected — the full context (328K tokens below) is sent to the backend by the Claude Code SDK; only the number Paseo displays is wrong. But the meter that's meant to tell you how close you are to the limit is off by ~2 orders of magnitude.

### Real captured data

Final GLM result message `usage` (from `~/.claude/projects/.../<sessionId>.jsonl`):

```json
{
  "input_tokens": 61489,
  "cache_creation_input_tokens": 0,
  "cache_read_input_tokens": 266816,
  "output_tokens": 4790,
  "iterations": []
}
```

Real context in use = `61489 + 266816 = 328305` tokens (~33% of 1M).

Paseo `agent.lastUsage` for the same turn:

```json
{
  "inputTokens": 61489,
  "cachedInputTokens": 266816,
  "outputTokens": 4790,
  "contextWindowMaxTokens": 1000000,
  "contextWindowUsedTokens": 4790   // ← should be 328305 (or 333095 incl. output)
}
```

`contextWindowUsedTokens` equals `outputTokens` exactly. The input + cache_read figures are present on the very same `lastUsage` object but are not summed into the "used" field.

## Steps to reproduce

1. Configure the Claude provider against a non-Anthropic-native backend that returns `iterations: []` on regular turns, e.g. via a translating proxy:
   ```
   ANTHROPIC_BASE_URL=https://<proxy>/v1/messages
   ANTHROPIC_AUTH_TOKEN=<token>
   ANTHROPIC_MODEL=glm-5.3[1m]   # resolves upstream to glm-5.2, context_length=1000000
   ```
2. Send at least 2 turns in the same session (so `completedResultTurns > 0`).
3. After turn ≥2 completes, read `agent.lastUsage.contextWindowUsedTokens` via the MCP `get_agent_status`, or watch the composer context-meter.
4. Observe `contextWindowUsedTokens` ≈ `outputTokens` of the last turn, instead of `input + cache_read + output`.

Single-turn sessions are NOT affected (see root cause — the legacy fallback only runs when `completedResultTurns === 0`).

## Root cause (traced in source)

`packages/server/src/server/agent/providers/claude/agent.ts`, class `ClaudeContextUsageState.buildResultUsage`:

```ts
const activeResultUsageTokens =
    readActiveUsageTokens(message.usage) ??
    (this.completedResultTurns === 0 ? readLegacyResultUsageTokens(message.usage) : undefined);
const usedTokens =
    this.streamUsedTokens() ?? activeResultUsageTokens ?? this.compactedContextWindowUsedTokens;
if (usedTokens !== undefined) {
    usage.contextWindowUsedTokens = usedTokens;
}
```

Two fallback paths both fail for `iterations: []` on a multi-turn session:

1. `readActiveUsageTokens` → `readLastUsageIteration` returns `undefined` for an **empty** `iterations` array (`Array.isArray([])` is true, but the loop body never runs, so it falls through to `return undefined`). The correct `readUsageTokenTotal` (which sums input + cache_creation + cache_read + output = 328305) is therefore never called. (For Anthropic-native backends, `iterations` is populated on multi-iteration tool-use turns, so this path works there.)

2. `readLegacyResultUsageTokens` — the flat-fields fallback that *would* compute the right total via `readUsageTokenTotal` — is gated on `this.completedResultTurns === 0`, so it only runs on the **first** completed turn. Any session with ≥2 turns skips it entirely.

3. That leaves `streamUsedTokens()` (priority highest in the `??` chain), which uses `streamRequestInputTokens` captured from the streaming `message_start` event. At `message_start` time the input-token count is not yet final (the backend hasn't finished counting), so it reports `0`/incomplete, and `readStreamRequestInputTokens` returns `0` (a valid number ≥0, not `undefined`) → `streamUsedTokens() = 0 + outputTokens = outputTokens`. That's why the meter shows exactly the output-token count.

Net effect: the only code path that can compute the correct total (`readUsageTokenTotal`) is unreachable for `iterations: []` on multi-turn sessions, and the displayed value collapses to the streaming-time output count.

I did not find any test covering the empty-`iterations` case, and searching issues I didn't find a report of this specific symptom (closest are #1392 wrong *max*, #2493/#2494 meter *disappearing*, #1390/#1391 ACP meter — all different layers).

## Suggested fix (either/or)

- Drop the `this.completedResultTurns === 0` gate on `readLegacyResultUsageTokens`, so the flat-fields fallback works on every turn; or
- Make `readActiveUsageTokens` (or `readLastUsageIteration`) fall back to `readUsageTokenTotal(usage)` when `iterations` is present but **empty** (`Array.isArray(iterations) && iterations.length === 0`).

Either lets `readUsageTokenTotal` run and sum the real total. The first option is the smaller change.

## Where did this happen

Daemon

## Paseo version

0.6.1

## OS version

Ubuntu 24.04 (Linux 6.8.0-138-generic)

## Agent provider

Claude Code

## Provider configuration

Claude provider pointed at a translating proxy (Anthropic Messages API → upstream GLM). Upstream model `glm-5.2`, `context_length = 1000000`. Reproduced with Paseo daemon 0.6.1; the bundled `@getpaseo/server` dist matches `main` at the relevant functions (`readUsageTokenTotal`, `readLastUsageIteration`, `ClaudeContextUsageState`).

Any non-Anthropic-native backend that returns `iterations: []` on plain turns triggers this; it is not specific to GLM.

githubbzxs:
### Linked issue

Closes #4019

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Some Anthropic-compatible backends return finalized per-turn usage as flat fields together with an explicit empty `iterations` array. On later turns, the existing logic ignored those finalized flat fields and preferred the incomplete streaming counters, so a result with 61,489 input tokens, 266,816 cache-read tokens, and 4,790 output tokens reported only 4,790 tokens used.

An explicit empty `iterations` array is different from a missing `iterations` field: it tells us there is no populated iteration to read while the flat fields are the finalized per-turn values. This change uses those flat values first only for that explicit-empty case. Results with populated iterations keep their existing active-iteration behavior, and later results with no `iterations` field retain the existing guard against treating legacy aggregate billing totals as active context.

### Goals

- Report the finalized flat usage total when a later Claude result explicitly has `iterations: []`.
- Include input, cache-creation, cache-read, and output tokens in that context total.
- Preserve the existing populated-iteration and stream fallbacks.
- Preserve the multi-turn protection for legacy results where `iterations` is missing.
- Add a two-turn regression using the real token values from #4019.

### Non-goals

- Change token billing fields or `totalCostUsd`.
- Infer finalized flat usage when the backend omits `iterations` entirely.
- Change context-window maximum resolution or any app UI.

### QA

The regression test failed against the unpatched implementation for the reported reason:

```text
expected contextWindowUsedTokens: 333095
received contextWindowUsedTokens: 4790
```

The focused test drives two turns through `ClaudeAgentSession.run`. The second result uses the issue payload (`input_tokens: 61489`, `cache_read_input_tokens: 266816`, `output_tokens: 4790`, `iterations: []`) and now returns a `contextWindowUsedTokens` value of `333095`.

```text
$ npx vitest run packages/server/src/server/agent/providers/claude/agent.test.ts --maxWorkers=1 --reporter=verbose
Test Files  1 passed (1)
Tests       75 passed (75)
```

The adjacent existing regression `does not use aggregate result totals after the first result turn` also passes, covering the missing-`iterations` safety boundary.

Repository checks:

```text
$ npm run format
Finished in 1029ms on 4179 files using 10 threads.

$ npm run lint

$ npm run build:server
Server dependencies, server, and CLI builds exited 0.

$ npm run typecheck
All workspace typecheck scripts exited 0.

$ git diff --check
(no output)
```

No full local test suite was run; only the modified test file was run locally, leaving the full matrix to CI as requested by the repository testing guidance.

Platform coverage (macOS 27.0 arm64, Node 26.7.0):

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | No | No app/UI code changed. |
| Android | No | No app/UI code changed. |
| Web | No | No app/UI code changed. |
| Desktop macOS | Partial | Source provider logic and regression test run locally; no Electron UI change. |
| Desktop Windows | No | Not available locally. |
| Desktop Linux | No | Not available locally. |

The external GLM translating proxy from the report was not available locally, so this was not claimed as a live-provider run. The automated regression uses the captured SDK result and stream sequence from the issue. Screenshots/video are not included because no UI code changed and the behavior is asserted at the provider result boundary.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


githubbzxs:
Hi! For visibility, #4083, #4084, and #4086 are all currently waiting for approval to run workflows (`action_required`) on their latest heads, so no CI jobs have started yet. When convenient, could a maintainer approve the fork workflow runs? No rush—thank you!

githubbzxs:
Quick follow-up on the workflow gate: #4083, #4084, and #4086 are still waiting for workflow approval, with no CI jobs started. When available, could a maintainer approve the fork workflow runs? Thanks!

PR #4092, opened 2026-08-30
feat(server): per-profile plan usage for pinned Claude accounts

eddiemachaalani:
### Linked issue

Closes #

### Type of change

- [x] New feature

### Reasoning

Paseo documents running multiple profiles of the same provider (`docs/custom-providers.md`, "Multiple profiles for the same provider") — e.g. several Anthropic accounts pinned via `claude setup-token` + `CLAUDE_CODE_OAUTH_TOKEN`, so account switching tools can't log agents out mid-run. But plan usage is blind to profiles today: the single `claude` usage entry reads the machine's shared credential stores, i.e. whichever account the CLI happens to be on. For anyone spreading work across accounts, the context-meter tooltip and the host Usage page can't answer the question that matters when a session stalls: *which account is this agent on, and how much of its session/weekly/per-model limit is left?* Worse, a pinned agent's tooltip shows either nothing or another account's numbers.

### Goals

- Every provider profile that extends `claude` and pins a `CLAUDE_CODE_OAUTH_TOKEN` gets its own usage entry, fetched with that token, keyed by the profile id and labeled with the profile's `label`.
- The context-meter tooltip picks it up with zero app changes — it already matches usage entries by the agent's provider id, which for profile agents *is* the profile id. Same for the host Usage page, which lists all entries.
- `paseo daemon reload` applies profile changes to usage with no daemon restart: profile fetchers are re-derived from live config on every usage refresh (`dynamicFetchers` on `ProviderUsageService`).
- Per-entry validation: one malformed provider entry in config can't take down other profiles' usage.
- The fetcher stays strictly read-only on credentials; a dead/revoked token yields an honest `unavailable` entry (no refresh attempts).

### Non-goals

- Usage for profiles **without** a pinned token (e.g. `CLAUDE_CONFIG_DIR`-based profiles): their agents run on a shared login, which the base `claude` entry already describes. Emitting a duplicate per profile would double-count.
- Codex/Copilot/etc. profile usage — their usage endpoints aren't derivable from a pinned env token the same way; the builder is structured so more bases can be added later.
- `planLabel` for profile entries: subscription type/tier live in the credential stores, not the token, so profile entries show windows without a plan badge.

### QA

**Automated (macOS, local):**

- `packages/server` `src/services/quota-fetcher/profile-fetchers.test.ts` — new: one fetcher per pinned Claude profile keyed by profile id + label; skips disabled profiles, non-claude bases, missing/blank tokens; empty config → no fetchers.
- `packages/server` `src/services/quota-fetcher/service.test.ts` — new: "re-evaluates dynamic fetchers on every refresh so config reloads apply" and "fetches profile usage with the profile's pinned token, ignoring shared stores" (asserts `Authorization: Bearer <profile token>` while the shared store holds a different account, `providerId`/`displayName` from the profile, `planLabel: null`). Full file: 76 passed.
- **Mutation check:** reverting `claude.ts` + `service.ts` to the base branch with the new tests in place → exactly the 3 new tests fail; restoring → green. The tests bind to the changed code.
- `npm run typecheck` ✅ · `npm run lint` ✅ · `npm run format` ✅ (pre-commit hook re-runs all three).

**Live (dev daemon, real WebSocket RPC):**

- Added a `claude-demo-pinned` profile (synthetic token) to the dev home's `config.json`, restarted the dev daemon, and queried `provider.usage.list` over the real WS client:

```json
{
  "fetchedAt": "2026-08-30T21:18:21.319Z",
  "rows": [
    { "providerId": "claude", "displayName": "Claude", "status": "available", "planLabel": "Max 20x",
      "windows": "Session=99% | Weekly=48% | Weekly · Fable=92%" },
    { "providerId": "codex", "displayName": "Codex", "status": "available", "planLabel": "team",
      "windows": "Session=0% | Weekly=0%" },
    { "providerId": "claude-demo-pinned", "displayName": "Claude (Demo Pinned Account)",
      "status": "unavailable", "planLabel": null, "windows": "" }
  ]
}
```
*(six other built-in providers omitted — all `unavailable`, no credentials on this machine)*

- The base `claude` entry keeps real account data (via the shared stores), and the profile entry appears under its own id/label with `unavailable` status for the synthetic token — the exact honest failure a revoked token should produce. A real pinned token was exercised by the unit tests' Bearer-header assertions; I avoided copying a live account token into QA artifacts.

**Platforms:**

| iOS             | ❌     | server-only change; app consumes existing data shape |
| Android         | ❌     | 〃 |
| Web             | ✅     | RPC-level probe; tooltip/Usage-page rendering of these entries is the existing code path |
| Desktop macOS   | ❌     | 〃 |
| Desktop Windows | ❌     | 〃 |
| Desktop Linux   | ❌     | 〃 |

### Notes for review

Stacked on #4091 (`fix: keep the context-meter tooltip available and its provider usage fresh`) — both touch the Claude quota fetcher, so this branch includes that commit to avoid guaranteed conflicts. Review the last commit only; if #4091 lands first this rebases clean.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -290,6 +290,8 @@ Example: two different Anthropic accounts as separate profiles:
 
 Each profile appears as a separate provider in the Paseo app. You can select which one to use when launching an agent.
 
+Claude profiles that pin an account with `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) also get their own plan-usage entry: the context-meter tooltip and the host Usage page show that account's session/weekly limits under the profile's label, fetched with the pinned token. Profiles without a pinned token share the base `claude` entry, which reads the machine's active Claude login.
+
 You can also combine profiles with model overrides to pin specific models per profile:
 
 ```json
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -290,7 +290,7 @@ Example: two different Anthropic accounts as separate profiles:
 
 Each profile appears as a separate provider in the Paseo app. You can select which one to use when launching an agent.
 
-Claude profiles that pin an account with `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) also get their own plan-usage entry: the context-meter tooltip and the host Usage page show that account's session/weekly limits under the profile's label, fetched with the pinned token. Profiles without a pinned token share the base `claude` entry, which reads the machine's active Claude login.
+Claude profiles that pin an account with `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) also get their own plan-usage entry in the context-meter tooltip and the host Usage page, under the profile's label. Anthropic's usage API requires the `user:profile` scope and setup-tokens carry only inference scope, so a setup-token profile's entry reports unavailable rather than another account's numbers. A profile whose token has `user:profile` shows live limits. Profiles without a pinned token share the base `claude` entry, which reads the machine's active Claude login.
 
 You can also combine profiles with model overrides to pin specific models per profile:
 
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -290,7 +290,7 @@ Example: two different Anthropic accounts as separate profiles:
 
 Each profile appears as a separate provider in the Paseo app. You can select which one to use when launching an agent.
 
-Claude profiles that pin an account with `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) also get their own plan-usage entry in the context-meter tooltip and the host Usage page, under the profile's label. Anthropic's usage API requires the `user:profile` scope and setup-tokens carry only inference scope, so a setup-token profile's entry reports unavailable rather than another account's numbers. A profile whose token has `user:profile` shows live limits. Profiles without a pinned token share the base `claude` entry, which reads the machine's active Claude login.
+Claude profiles that pin an account with `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) also get their own plan-usage entry in the context-meter tooltip and the host Usage page, under the profile's label. Anthropic's usage API requires the `user:profile` scope, which setup-tokens do not carry, so those profiles read their session and weekly windows from the rate-limit headers of a minimal one-token API call instead — spending a few tokens of the pinned account's own quota per refresh. The plan label and per-model weekly windows exist only on the usage API, so they stay absent for setup-token profiles. Profiles without a pinned token share the base `claude` entry, which reads the machine's active Claude login.
 
 You can also combine profiles with model overrides to pin specific models per profile:
 
--- a/packages/app/src/components/context-window-meter.tsx
+++ b/packages/app/src/components/context-window-meter.tsx
@@ -129,32 +129,55 @@ export function ContextWindowMeter({
 
   // No usage yet: reserve the footprint with a track-only ring while a session is
   // active so the real ring fades in without shifting siblings. Render nothing when
-  // no usage is expected.
+  // no usage is expected. The pending ring keeps the tooltip: a running or stalled
+  // agent with no usage report is exactly when someone hovers to check plan limits.
   if (percentage === null || maxTokens === null || usedTokens === null) {
     if (!pending) {
       return null;
     }
     return (
-      <View style={geometry.containerStyle}>
-        <Svg
-          width={geometry.svgSize}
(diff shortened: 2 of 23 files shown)


eddiemachaalani:
**Live-testing update — a scope limitation worth knowing before review.**

The PR's live probe used a synthetic token (as noted in the body). I've since run the same path with a **real** `claude setup-token` value in a pinned profile, and Anthropic's usage endpoint refuses it:

```json
{"type":"error","error":{"type":"permission_error","message":"OAuth token does not meet scope requirement user:profile","details":{"required_scopes":["user:profile"],"match":"any","error_code":"oauth_scope_insufficient"}}}
```

`claude setup-token` mints inference-only tokens, and `GET /api/oauth/usage` requires `user:profile` — so a setup-token profile's entry reports `unavailable` (the fetcher already maps 401/403 → needs-auth). A profile whose token carries `user:profile` gets live data; setup-token pins — the common case from `docs/custom-providers.md` — do not, on any build.

What the feature still delivers for setup-token pins: the per-profile row correctly attributes the agent to its own account and reports unavailable, instead of silently showing nothing or another account's numbers. But I didn't want the PR reviewed on an overstated premise, so:

- Pushed 0c37dff1d amending the doc section to state the limitation plainly.
- If you'd rather not carry the feature until it can show real numbers for setup-token pins, options I see: (a) merge as-is with the documented limitation, (b) extend profiles with an optional full-OAuth credential source (access+refresh) plus refresh handling in the fetcher, or (c) wait on upstream — a `user:profile`-scoped setup-token from Anthropic's CLI would make this work as originally described.

Happy to amend in whichever direction.

eddiemachaalani:
One more live-testing observation on the needs-auth path: a profile whose token persistently 403s on scope eventually gets **429-throttled per-token** by Anthropic. After a morning of low-volume periodic fetches (the 15s force-refresh floor + 5-min TTL), the endpoint switched from `403 oauth_scope_insufficient` to `429` for that token — while a sibling setup-token from the same IP still received the plain 403 in the same second, so the throttle keys on the token, not the caller.

Implication: the service currently refetches a needs-auth profile on the same cadence as healthy ones, so a scope-less pin produces a permanent drip of failing calls whose only effect is surfacing `Claude usage API returned 429` in the tooltip. A longer negative cache for needs-auth results (or not retrying until the profile's config changes) would avoid that. Happy to implement here if wanted — flagging it so the behavior has an explanation on record either way.

eddiemachaalani:
**Implemented the fallback (d23abae74) — setup-token profiles now show live usage.**

Inference responses carry the same session and weekly windows in `anthropic-ratelimit-unified-5h/7d-utilization` (+ `-reset`) headers, and those come back for inference-scoped tokens. So when the usage endpoint refuses a pinned token (401/403) — or throws (429/5xx) — the profile fetcher sends a one-output-token `/v1/messages` ping (cheapest listed model, resolved once via `/v1/models`) and parses the headers.

- **Calibration:** headers are 0..1 fractions. Verified with a full-scope token at the same instant: usage API `utilization: 4.0 / 2.0` ↔ headers `0.04 / 0.02`, identical reset timestamps.
- **What's absent:** plan label and per-model weekly windows exist only on the usage API, so setup-token entries show Session + Weekly only. Documented in `docs/custom-providers.md`.
- **Throttle behavior:** after a scope refusal the endpoint is skipped on later refreshes (a refusal is permanent for a token, and re-knocking is what attracted the 429s); after a 429/5xx it stays first choice since that's transient. Profile fetcher instances now persist across config refreshes (keyed by id, token, label) so that memo and the resolved ping model survive; token/label changes or profile removal evict.
- **Cost:** one output token of the pinned account's own quota per cache refresh (default TTL 5 min, 15 s force-refresh floor).
- Tests cover the happy path, the skip-after-refusal memo, the throttled path, the ping-failure path (→ `unavailable`), and instance caching; mutation-checked (3 red on revert).

Live on my dev daemon: a pinned profile whose usage endpoint is currently 429ing reports `available` with Session 0% / Weekly 46% under its own label.

eddiemachaalani:
Follow-up: #4093 (stacked on this branch) closes the per-model gap noted above — the Claude agent now reads Claude Code's own `rate_limit_event` (which carries every window, including the per-model weekly bucket such as `Weekly · Fable`) and the tooltip prefers those windows for the running agent. The header-ping fallback here stays useful for the host Usage page and agents that have not run yet.

boudra:

PR #4095, opened 2026-08-30
Voice: drain TTS audio at synthesis time so replies stop losing segments

eddiemachaalani:
## Problem

Spoken replies were losing whole segments — the tail or middle sentences of an answer silently never played. Field trace from a live daemon (OpenAI TTS, five-segment reply):

```
08:50:47.831 TTS segment 1 synthesized (synth=664ms, 179 chars)
08:50:48.117 TTS segment 0 synthesized (synth=951ms, 42 chars)
08:50:48.962 TTS segment 2 synthesized (synth=845ms, 119 chars)
08:50:52.381 TTS segment 0 playback confirmed (emit+play 4264ms)   ← played
08:50:52.382 TTS segment 1 playback confirmed (emit+play 1ms)      ← vanished
08:50:52.385 TTS segment 2 playback confirmed (emit+play 3ms)      ← vanished
08:50:59.052 TTS segment 3 playback confirmed (emit+play 5739ms)   ← played
08:50:59.053 TTS segment 4 playback confirmed (emit+play 0ms)      ← vanished
```

Root cause: prefetched segments held their provider **HTTP response streams open** until their turn to play. A response body left unread for a few seconds while other segments streamed and played was observed (OpenAI + undici) to drain **empty** — zero bytes, no error, no abort. `emitPreparedSegment` then took the "nothing to emit" path and resolved instantly (the 0–3ms "confirms" above), so the audio was never sent to the client. An instrumented run caught it directly: `TTS segment collected (0 bytes, aborted=false)`. Segments drained within ~1s of synthesis always played; segments that sat 3–6s behind a playing one drained empty.

## Change

- Each segment is drained into a `Buffer` as part of synthesis, so prefetch holds audio, not live sockets. `PreparedTtsSegment` carries `audio: Buffer` instead of `stream`.
- A zero-byte synthesis now **throws** (surfacing as an error instead of a silent skip), and the per-segment log records the collected byte count for field diagnosis.

Independent of #4010 — applies cleanly with or without it.

## QA evidence

After the fix, on the same live daemon driven from the iOS app: full multi-sentence replies play to the last word (previously reproducible loss within a few exchanges; user-confirmed fixed).

Tests on this branch (abort test updated: an aborted utterance stops emitting and resolves promptly):

```
npx vitest run src/server/agent/tts-manager.test.ts
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

## Platforms

Server-only. Tested on a macOS daemon end-to-end from the iOS app with OpenAI TTS; local (sherpa/Kokoro) TTS streams are in-memory generators and are unaffected by the ordering change. Not separately tested: Android app, Windows/Linux daemons.

### DIFF
--- a/packages/server/src/server/agent/tts-manager.test.ts
+++ b/packages/server/src/server/agent/tts-manager.test.ts
@@ -165,26 +165,11 @@ describe("TTSManager", () => {
     await task;
   });
 
-  it("destroys prefetched streams after abort", async () => {
-    const destroyed: string[] = [];
+  it("stops emitting and resolves promptly when aborted mid-playback", async () => {
     const tts: TextToSpeechProvider = {
       async synthesizeSpeech(text: string): Promise<{ stream: Readable; format: string }> {
-        const stream = new Readable({
-          read() {
-            this.push(Buffer.from(text));
-            this.push(null);
-          },
-        });
-        const destroySpy = vi.spyOn(stream, "destroy").mockImplementation(function (
-          this: Readable,
-          error?: Error,
-        ) {
-          destroyed.push(text);
-          return Readable.prototype.destroy.call(this, error);
-        });
-        void destroySpy;
         return {
-          stream,
+          stream: Readable.from([Buffer.from(text)]),
           format: "pcm;rate=24000",
         };
       },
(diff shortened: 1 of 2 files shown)


boudra:
Closing for now: this needs a failing-first regression reproducing the lost prefetched audio segments and verifying that every segment is emitted. Please reopen with that evidence.

boudra:

Issue #3335, opened 2026-08-13
bug: background fetch can massively inflate partial-clone repositories


aivanov93:
### What's broken

Paseo's daemon automatically runs `git fetch origin --prune` immediately after observing a repository and every 180 seconds afterward. In a large partial-clone repository whose `remote.origin.fetch` maps all remote branches, opening the repository in Paseo caused `.git/objects` to grow extremely quickly.

One occurrence grew into the hundreds of GiB before I stopped Paseo and cleaned up the checkout. In the retained reproduction, `.git` reached about 18 GiB, including a single 11 GiB promisor pack. This was a `blob:none` partial clone, but the remote had thousands of branches, so repeatedly applying the broad refspec still materialized a very large commit/tree object graph.

I expected opening a local checkout for agent use not to initiate an unconfigurable, repository-wide background fetch. At minimum, I expected a way to disable automatic fetches for repositories whose fetch strategy is managed externally.

The current latest stable release still appears to have this behavior:

- `packages/server/src/server/workspace-git-service.ts` sets `BACKGROUND_GIT_FETCH_INTERVAL_MS = 180_000`, starts the interval, and performs an immediate fetch.
- `packages/server/src/server/workspace-git-fetch.ts` runs `git fetch origin --prune`.

There does not appear to be a documented global or per-project setting to disable or constrain this fetch.

Sanitized repository evidence:

```text
$ git config --get remote.origin.promisor
true

$ git config --get remote.origin.partialclonefilter
blob:none

$ git config --get-all remote.origin.fetch
+refs/heads/*:refs/remotes/origin/*

$ git for-each-ref --format='%(refname)' refs/remotes/origin | wc -l
approximately 9,000

$ git count-objects -vH
in-pack: approximately 38 million
packs: 10
size-pack: 16.32 GiB
```

The reflog showed repeated entries of this form during the growth window:

```text
fetch origin --prune: fast-forward
```

No private repository name, path, remote URL, or daemon log is included because the affected repository is private.

Possible safe fixes:

1. Add a global and per-project option to disable background Git fetches.
2. Avoid fetching every ref selected by a repository's broad default refspec merely to refresh Paseo's base/default branch metadata.
3. Detect partial/promisor clones and preserve their intended narrow-fetch behavior.
4. Surface that automatic fetch is enabled and which command/refspec it will use.

### Steps to reproduce

1. Create or use a large repository with many remote branches.
2. Clone it as a partial clone with `--filter=blob:none`.
[...]

PR #4116, opened 2026-08-31
Start new worktrees from the latest origin

ismatBabirli:
Discussion: #4115

## Problem

Nothing in the `branch-off` worktree creation path fetches. `resolveBaseBranchForWorktree` only `rev-parse`s local refs, and `resolveRepositoryDefaultBranch` deliberately prefers the local branch name over the remote-tracking ref — so an unqualified base resolves to whatever your local branch last pulled. The freshest that can be is the daemon's background `git fetch origin --prune`. **Correction (see [discussion](https://github.com/getpaseo/paseo/discussions/4115#discussioncomment-18240129)):** I originally described that as a 180 s ceiling. It is not a ceiling at all — `ensureRepoTarget`, which installs the interval and fires the immediate fetch, is only reached from the observed-workspace setup path, so a repo with no observed workspace is never fetched and its remote-tracking refs can be arbitrarily old. @odoo-mvds reproduced this on 0.7.1 with a **qualified** `origin/main` base branching off a commit two days stale.

The failure is silent. `createWorktree` stores the resolved ref as the worktree's comparison base (`baseRef: "refs/heads/main"`), and `getCheckoutStatus` compares against exactly that ref, so a worktree cut from a stale `main` reports `aheadBehind {ahead: 0, behind: 0}`. The agent works on out-of-date history and nothing says so.

`public-docs/worktrees.md` documented the manual workaround ("Use `origin/main` rather than `main`"), which doesn't help the New Workspace default path or `resolveDefaultBranch`.

## Change

Refresh the base from its remote before `git worktree add`, then start from the remote tip when the local branch is *purely behind* it.

- **Narrow fetch**: `git fetch <remote> <branch>` — one branch, never `--prune`, never the repository-wide refspec. Deliberate, given #3335.
- **Your checkout is never touched**: no branch moves, no working tree changes, no `git pull`.
- **Strictly-behind only**: if local holds commits the remote doesn't, local wins and unpushed work is never dropped.
- **Non-fatal**: unreachable remote falls through to local resolution instead of failing creation.
- **Fork-aware**: the remote is derived from the ref, so `refs/remotes/upstream/main` fetches `upstream`.
- **Opt-out**: `fetchBase: false` on the create request, surfaced as a "Fetch latest from remote" row in the ref picker footer, remembered like the `isolation` preference and gated on the new `workspaceCreateFetchBase` host feature.

Absent means fetch, so existing clients, the CLI and MCP get the behaviour without changes.

## Before / after

Repro: clone, teammate pushes two commits, daemon's background fetch runs, then create a worktree off `main`.

```
After the daemon's background fetch:
  origin/main : three
  local  main : one
  local main is behind origin by 2 commits

BEFORE (fetchBase: false — the old behaviour)
  worktree HEAD subject : one
  ahead/behind vs base  : ahead=0 behind=0     <-- staleness invisible

AFTER  (default — fetch the base first)
  worktree HEAD subject : three
  ahead/behind vs base  : ahead=0 behind=0

User's own checkout after both creations:
  branch          : main
  local main moved: no
  working tree    : clean
```

## Cost

Two extra local git commands when there is no remote (one `config --get remote.<name>.url`, one `rev-list`), plus one `fetch` when there is. The creation census in `workspace-create-git-fanout.local.e2e.test.ts` is updated by exactly that delta: `config` 22→23, `rev-list` 3→4, total 103→105 (that fixture has no origin remote, so no `fetch`).

> Note: those three census assertions, the metadata ledger assertion, and the 100-sibling latency assertion in that file already fail on `main` at `63923fdd0` on my machine — the counts are git-version sensitive (I see an extra `check-ref-format` and lower counts throughout). I verified my delta against a pristine worktree of `63923fdd0` rather than against the committed numbers, and applied the delta to the committed numbers so CI's baseline stays authoritative. Worth a look independently.

## Tests

New:
- `packages/server/src/utils/fetch-base-ref.test.ts` — 19 tests: remote derivation table, narrowness (asserts no `--prune`/`--all`), skip when no remote, non-throwing failure, and the strictly-behind rule.
- `packages/server/src/utils/worktree-fetch-base.test.ts` — 7 tests against **real** git repos: starts at the freshest commit, checkout left untouched, opt-out preserves old behaviour, diverged local keeps unpushed commits, offline still creates, qualified remote base works.
- `packages/server/src/server/fetch-base-wire.local.e2e.test.ts` — 3 tests through a real daemon over the WebSocket: capability advertised, default create branches off the fresh commit, wire-level opt-out honoured.
- `preferences.test.ts` — 5 tests: unset means fetch, round-trips, non-boolean falls back to fetching.

```
$ npx vitest run packages/server/src/utils/fetch-base-ref.test.ts \
    packages/server/src/utils/worktree-fetch-base.test.ts \
    packages/app/src/create-agent-preferences/preferences.test.ts \
    packages/app/src/i18n/resources.test.ts \
    packages/server/src/server/resolve-worktree-creation-intent.test.ts \
    packages/server/src/server/worktree-session.test.ts --bail=1
 Test Files  6 passed (6)
      Tests  122 passed (122)

$ npx vitest run packages/server/src/server/fetch-base-wire.local.e2e.test.ts \
    packages/server/src/server/workspace-git-noop-fetch.local.e2e.test.ts --maxWorkers=1
 Test Files  2 passed (2)
      Tests  10 passed (10)
```

`workspace-git-noop-fetch.local.e2e.test.ts` is included deliberately: the create-time fetch is a new caller outside the `FETCH_METADATA_ECHO_TTL_MS` echo-suppression path, so I checked it does not trigger a broad sibling refresh. It does not.

```
$ npm run typecheck && npm run lint && npm run format:check
typecheck=0   lint=0 (Found 0 warnings and 0 errors)   format:check=0
```

## Platforms

| iOS             | no     | UI is a plain `ComboboxItem` in the existing ref picker footer |
| Android         | no     | as above |
| Web             | no     | server behaviour covered by the daemon e2e test |
| Desktop macOS   | no     | |
| Desktop Windows | no     | |
| Desktop Linux   | partial | daemon + git paths exercised via the test suites above |

I have not run the app UI on any platform — the toggle row is untested visually and I'd rather say so than imply otherwise. The server behaviour is covered end to end through the daemon. Happy to add screenshots if you want the UI verified before merge.

## Notes for review

- **Local isolation is untouched by design.** There the workspace *is* your checkout, and pulling it automatically would mutate your working tree. If you'd rather this also covered that case, say so — it's a different and riskier change.
- `resolveWorktreeSourcePlan` was one point over the `complexity` lint ceiling with the new branch, so the base resolution moved into `resolveBranchOffBaseRef` next to `resolveBaseBranchForWorktree`.
- Two open questions from the discussion: default on vs opt-in, and whether the control belongs in the ref picker footer or Settings.


boudra:
Closing in favor of the narrower remote-ref fix in [#4287](https://github.com/getpaseo/paseo/pull/4287), which is open and unmerged. The unqualified local-main behavior remains a separate decision in [#4115](https://github.com/getpaseo/paseo/discussions/4115).

PR #4120, opened 2026-08-31
fix(server): rewind Codex conversations with thread/revert

### LINKED ISSUE #4104: Codex rewind fails on paginated threads: `paginated threads do not support thread/rollback`
## Description

Using Paseo with a recent Codex CLI, attempting to rewind/undo a Codex conversation fails with:

```text
paginated threads do not support thread/rollback
```

It looks like Paseo is attempting to use `thread/rollback` on a Codex paginated thread, where that operation is not supported.

## Environment

- Paseo: `0.7.0-beta.2`
- Codex CLI: `0.151.0`
- OS: Linux

## Steps to reproduce

1. Start/use a Codex conversation through Paseo.
2. Use a conversation that is represented as a paginated thread.
3. Attempt to rewind/undo the conversation in Paseo.
4. Paseo reports:

```text
paginated threads do not support thread/rollback
```

## Expected behavior

Rewind/undo should work with Codex paginated threads, or Paseo should use the appropriate newer Codex app-server mechanism for reverting/forking paginated conversation history.

## Actual behavior

The rewind/undo operation fails because `thread/rollback` is invoked for a paginated thread.

## Additional context

I searched the existing open issues for `paginated threads` / rollback / rewind / undo and did not find an issue matching this specific error.

SameTrouble:
### Linked issue

Closes #4104

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Rewinding the first message of a Codex conversation never worked for me — it
always failed with `paginated threads do not support thread/rollback`, while
rewinding any later message was fine. If you want to throw away a whole
conversation and start over without leaving the agent, that is the one rewind
you actually reach for.

The split turned out not to be "first message" but "message without a turn id".
Rewind resolves the target message to a Codex turn id and calls `thread/revert`.
Messages carry a turn id while their turn is live, but a message that reached us
through persisted history has none, so the old code fell back to `thread/fork` +
`thread/rollback`. Every thread a modern Codex creates is
`historyMode: "paginated"`, and rollback is rejected there — with `-32600`
rather than "method not found", which is why the fallback guard never caught it
and the raw Codex error reached the UI.

So the fix is in two parts: stop using the dead RPC, and make sure a turn id is
available for messages that came from history.

### Goals

- Rewinding the first message of a Codex conversation works and produces an
  empty conversation.
- The empty conversation stays usable: sending a new message starts a fresh turn
  on the same thread.
- Rewinding any message works on a thread that has already been rewound.
- An unsupported `codex` binary gets a clear reason instead of a raw Codex error.

### Non-goals

- Keeping rewind reversible. `thread/revert` rewrites the thread in place; the
  old rollout file stays on disk but the thread no longer points at it. This
  matches Claude Code's rewind semantics. The old fork path was recoverable, but
  it is the path that stopped working.
- Supporting `codex` older than 0.148.0. `thread/revert` landed there
  (openai/codex#38440, first shipped in 0.148.0-alpha.13). Older binaries now
  report `supportsRewindConversation: false` and lose the entry point rather
  than failing on click — their previous path was already broken.
- Reworking history loading. It still goes through `thread/read`, which returns
  no turns for paginated threads. This PR only fills in turn ids for rewind.
- Subscribing to `thread/reverted` notifications, so a second device does not
  reload its timeline after a rewind on the first one.

### QA

Rebuilt the Paseo desktop app with this branch and drove it against the local
Codex 0.151.0 install on macOS (daemon on 6767). Three turns, then two rewinds:

```
14:57:11  Created agent 90b55efd (codex)   thread 01a05853-0795-77c3-b08e-7afb45950b0e
14:57:11  Starting Codex app-server turn   codex-turn-0
14:57:16  Starting Codex app-server turn   codex-turn-1
14:57:20  Starting Codex app-server turn   codex-turn-2
14:57:25  agent.rewind.complete            msg_1788188240404_1or3sa3ej
14:57:28  agent.rewind.complete            01a05853-08a5-76c1-86e2-b7a431bb811d
14:57:32  Starting Codex app-server turn   codex-turn-3   thread 01a05853-0795-77c3-b08e-7afb45950b0e
```

The 14:57:28 rewind targets the first message — a native Codex item id, not a
Paseo `msg_*` id. After it the timeline in the app was empty; the follow-up turn
at 14:57:32 ran on the same thread and answered normally.

Before this change the same operation failed every time, on the same machine and
Codex install (`~/.paseo/daemon.log`):

```
08:47:30  agent.rewind.failed  01a056e7-b20c-70a2-8dda-1345fa1e3019  paginated threads do not support thread/rollback
08:48:10  agent.rewind.failed  01a05700-bd62-7330-99fc-3f4869f80d8b  paginated threads do not support thread/rollback
08:48:57  agent.rewind.failed  01a05700-bd62-7330-99fc-3f4869f80d8b  paginated threads do not support thread/rollback
08:49:17  agent.rewind.failed  01a05700-bd62-7330-99fc-3f4869f80d8b  paginated threads do not support thread/rollback
```

Note the 08:20:26 entry in the same log: a fresh session whose very first rewind
targeted the *second* message and failed the same way. That is what ruled out
"first message" as the cause and pointed at the turn id.

I also probed the Codex 0.151.0 app-server directly to check the assumptions:

- `thread/start` reports `historyMode: "paginated"` for a brand new thread, and
  `thread/rollback` answers `paginated threads do not support thread/rollback`
  (-32600) on it.
- `thread/read` returns `turns: []` for paginated threads — which is why
  persisted history has no turn ids at all.
- `thread/turns/list` with `cursor: null` returns every turn with its id.
- `thread/revert` with the first turn's id succeeds, keeps the thread id, empties
  the thread, and a new turn afterwards works.

Tests:

```
npx vitest run packages/server/src/server/agent/providers/codex/rewind.test.ts \
  packages/server/src/server/agent/providers/codex/app-server-transport.test.ts \
  packages/server/src/server/agent/providers/codex-app-server-agent.test.ts
# 162 passed
```

New coverage: revert to the first message, revert to a later message, pagination
cursor hand-off, turn ids backfilled from the thread, rewind refused when the
turn id is unknown, rewind refused when the binary is too old, and — in
`codex-rewind.real.e2e.test.ts` — a real Codex conversation rewound to the first
message that then accepts a follow-up turn. That e2e now asserts the thread id is
*unchanged*; it previously asserted the opposite for the fork path.

Pre-commit hooks (format, lint, typecheck across all workspaces) pass. I did not
run the `*.real.e2e.test.ts` suites locally — they need provider credentials.
CI runs them with `OPENROUTER_API_KEY`.

No UI change, so no screenshots: the app-side effect of this is that the rewind
entry point disappears on a `codex` binary older than 0.148.0.

| iOS             | N/A    | Daemon-side change; no app or protocol change. |
| Android         | N/A    | Same. |
| Web             | N/A    | Same. |
| Desktop macOS   | Yes    | Manual verification above, unit tests, and the app-server probe. |
| Desktop Windows | No     | No platform-specific code; the version probe uses the same path. |
| Desktop Linux   | No     | Same. |


SameTrouble:
#4119 has now been merged and fixes the same Codex rewind issue.

Its implementation preserves Paseo's existing recoverable rewind semantics,
while this PR uses `thread/revert`, which rewrites the source thread in place
and changes that behavior.

Since this PR is now superseded by #4119, I’m closing it to avoid maintaining
two competing implementations. Any remaining gaps can be addressed in a
follow-up PR based on the updated main branch.

PR #4129, opened 2026-08-31
feat: go to definition from the file viewer

half144:
### Linked issue

No issue. Direction discussed in the Discord feature-ideas channel: https://discord.com/channels/1481169421832814616/1544080014541066300

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Reading code in Paseo is currently a dead end. You open a file, see `SessionsScreen`, and to find out what it is you have to leave for your editor or guess the path and open it by hand. That breaks the thing Paseo is for, which is staying in the loop on your agents from wherever you are, including a phone or a browser far from your machine.

This adds the affordance every editor already trained people to expect: hold Cmd (or Ctrl), the symbol under the pointer underlines, click it and the declaration opens at its line.

The daemon answers by talking to a real language server over LSP rather than guessing from a symbol index, so resolution is whatever the user's own toolchain would say. It ships no binaries.

### Goals

- Cmd/Ctrl+click a symbol in a file tab opens its declaration at the right line.
- Holding the modifier underlines the range the language server reports, so the target is visible before committing to the click.
- Resolution comes from a real language server, so aliases (`@/screens/...`), re-exports and shadowing behave like they do in the user's editor.
- An absent language server is a reported state with the command to install it, not an error or a dead click.
- A `config.json` override points at a binary that is not on the daemon's PATH.
- Older daemons render no affordance at all, gated once on `server_info.features.codeNavigation`.

### Non-goals

- No language server binaries are bundled. Per-platform binary distribution is where this feature would die.
- Native is unchanged. There is no pointer to hold a modifier over, and a tap-to-navigate design is a separate question.
- Only `textDocument/definition`. Hover, references and document symbols reuse the same session and are cheap to add later, but they are not in this PR.
- No incremental document sync. Full text replace on change, which is enough for read-only navigation.
- No navigation from a buffer with unsaved edits. The daemon resolves against the file on disk, so those positions would not match; the links come back on save.

### QA

Tested on web (Chromium, dev daemon on `127.0.0.1:6768`). Not tested on iOS, Android, or Electron desktop; the macOS window-chrome path in particular is unexercised.

**Automated**

```
$ npx vitest run packages/server/src/server/lsp/ packages/app/src/file-pane/
 Test Files  13 passed (13)
      Tests  79 passed (79)
   Duration  7.40s
```

```
$ npm run typecheck   # exit 0
$ npm run lint        # Found 0 warnings and 0 errors
$ npm run format      # clean
```

The LSP session tests run against a real `typescript-language-server` and resolve `packages/app/src/app/sessions.tsx:2` to `packages/app/src/screens/sessions-screen.tsx:62`, which is `export function SessionsScreen()`.

**The readiness gate, and how I know it is load-bearing**

tsserver answers `textDocument/definition` from a partially loaded project with a well formed *wrong* result: the target is the import statement you clicked, not the declaration. No error, no warning. Instrumenting the round trip showed the boundary exactly:

```
+2128ms  window/workDoneProgress/create
+2203ms  definition #1  -> sessions.tsx        (wrong)
   ...   9 more wrong answers
+5820ms  publishDiagnostics sessions.tsx (0)
+5833ms  definition #10 -> sessions-screen.tsx (correct)
```

So requests gate on the document's first `publishDiagnostics`. To confirm the test actually covers this, I disabled the gate and re-ran:

```
AssertionError: expected '…/screens/sessions-screen.tsx' to be '…/app/sessions.tsx'
      Tests  1 failed | 5 passed (6)
```

**Failure paths**

A review pass found the daemon-crash class first: the spawned server had no `'error'` listener, so a spawn failure (EAGAIN under load, the binary removed while its resolution was still cached) was rethrown by Node and took the daemon down with every agent it was running. Every other spawn site in the repo already listens. The regression test for it then surfaced a second route to the same crash — `sendNotification` returns a promise whose write rejects once the process is gone:

```
Unhandled Rejection: Error: Cannot call write after a stream was destroyed
  at StreamMessageWriter.doWrite (vscode-jsonrpc/lib/common/messageWriter.js:141)
```

Both are fixed and the suite now runs with zero unhandled errors. Also closed: two concurrent hovers on one file each sent `didOpen` and orphaned a document in a `ready` state nothing resolved; a server that never publishes diagnostics paid the full 20s gate on every request rather than the first; a failed handshake poisoned the session for the whole idle window; a rejected request stranded the progress cursor.

The command override was being read on every request — a synchronous `config.json` read and Zod parse on the daemon's event loop — rather than per cold session as both the code comment and the doc claimed. It now sits behind the resolution cache.

**Behavior, against the running daemon**

```
SessionsScreen                9264ms  packages/app/src/screens/sessions-screen.tsx
HostRouteBootstrapBoundary       6ms  packages/app/src/components/host-route-bootstrap-boundary.tsx
export (keyword)                 3ms  (no target)
return (keyword)                 3ms  (no target)
```

First call pays the project load; the rest are warm. Keywords resolve to nothing because a target enclosing the clicked position is dropped: tsserver answers a click on `return` with the function containing it, which would put a link on every keyword and jump nowhere.

**Perceived latency**

Measuring in the browser, the underline appears in 2-4ms because it is drawn on the word under the pointer immediately and replaced by the server's own range when the answer lands. An earlier version waited for the round trip and measured 132ms, of which 120ms was a debounce; that felt sluggish warm and broken cold.

A word that resolves to nothing loses its underline when the answer arrives. Verified live on `return`: underlined at 2ms, gone once the daemon replied.

**Video**

https://github.com/user-attachments/assets/7435ad5e-f6bc-4e22-881a-fe0d35b1f065

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### Notes for review

- New dotted RPC pair `code.definition.request` / `code.definition.response`, per `docs/rpc-namespacing.md`. Gate tagged `COMPAT(codeNavigation)`.
- Targets outside the workspace root are dropped rather than sent: the client cannot open them and the path would leak a location from outside the workspace.
- Containment uses `getRealpathAwareRelativePath`. Plain `path.relative` reads a symlinked workspace root (macOS `/tmp`, any symlinked checkout) as outside itself and silently drops every target.
- The pool holds one server per (workspace root, language), reaped after ten idle minutes or when a fifth is needed, since Paseo hands out a worktree per agent.
- `docs/code-navigation.md` covers installing a server, the config override, and why the first click is slow.


boudra:

half144:
@boudra https://discord.com/channels/1481169421832814616/1544080014541066300 i did and it's on PR's body, i've sent to contributing channel in discord and got 5 fires from the community, they liked it

PR #4146, opened 2026-09-01
feat(workspaces): workspace pin groups

aliozkanozdurmus:
## Problem

Workspace pinning is daemon-global: one `pinnedAt` per workspace, broadcast to every connected client. If you share access to your daemon (e.g. over the relay), everyone sees — and changes — the same pinned set. There is no way to keep your own pinned view, let alone separate pinned sets for different contexts.

## What changes

Pinning becomes **named pin groups**, stored on the daemon, with a per-device active group:

- **Daemon**: a workspace registry gains a pin-group catalog (create / rename / delete groups) and per-workspace group membership. New dotted RPCs: `workspace.pinGroups.*` request/response pairs. Catalog changes are broadcast to all connected clients; membership changes ride the existing workspace-changed pipeline. Mutations require `workspace.manage`; reads require `workspace.read`.
- **App**: the sidebar Pinned section gets a group switcher (create / rename / delete, compact bottom-sheet on small screens). The active group is a per-device choice persisted as `{serverId, groupId}`, so two daemons never mix memberships and two devices on one daemon can each view a different group.
- **Compatibility** (per `docs/protocol-compatibility.md`): the feature is capability-gated on `server_info.features.workspacePinGroups`. New app + old daemon gets an update-host state; there is no silent fallback. Old app + new daemon keeps working: the legacy `workspace.pin.set` RPC maps to the built-in `default` group and memberships in `default` are projected back out as legacy `pinnedAt`, so old clients see exactly the default group. All compatibility shims carry dated `COMPAT(...)` tags.

## Persistence and downgrade safety

`workspaces.json` keeps its legacy array shape — a downgraded daemon reads and writes it unchanged. Group catalog and membership live in a sidecar (`workspace-pin-groups.json`) with a mirrored backup, an expected-state marker, a single-entry transaction journal, and a SHA-256 content hash of the primary:

- Crash mid-transaction: known before/after states roll back or forward automatically; a genuinely unknown state blocks startup with an actionable error naming every file and the exact recovery steps, and is routed as terminal (no supervisor respawn loop). Recovery guarantee is scoped to process crashes in `docs/data-model.md`.
- Downgrade interleaving: if a legacy daemon writes `workspaces.json` while pin groups exist, the hash detects it on next start and legacy `pinnedAt` changes are reconciled into the `default` group instead of being overwritten by stale sidecar state.
- Corruption: malformed primary/sidecar/backup/journal blocks startup without touching a single byte. Future `formatVersion` values block with "keep everything, run a newer daemon" guidance. Only `EAGAIN`/`EBUSY`/`EMFILE`/`ENFILE` stay on the transient retry path.

The recovery machinery is deliberately more than a bare JSON file: every layer here is the fix for a specific data-loss path found by adversarial downgrade/crash probing (an older daemon's writes silently discarded, a crash journal overwriting newer work, sidecar loss silently deleting every group). The state machine converges ambiguous states to "block with actionable steps" rather than guessing.

## Tests

All suites pass at HEAD (`899cdf5be`):

| Suite | Result |
|---|---|
| `workspace-registry.test.ts` | 47 passed |
| `workspace-registry-bootstrap.test.ts` | 10 passed |
| `session.workspaces.test.ts` | 114 passed, 4 skipped |
| `supervisor.logging` / `lifecycle-intents` / `workspace-registry-integrity` | 15 passed |
| Protocol pin-group messages | 5 passed |
| `sidebar-view-store` / `catalog` / `delete-action` / `menu-model` / `use-sidebar-pins` | 54 passed |
| E2E `workspace-pin-groups.spec.ts` | 6 passed |
| E2E `sidebar-workspace-pin-shortcut.spec.ts` | 5 passed |

Coverage highlights: downgrade round-trips execute the **actual** pre-feature registry implementation (materialized from git history) against a temp home; crash-recovery windows are probe-tested (prepared/committed journal × missing/corrupt/third-state artifacts); the supervisor test launches the real entrypoint + worker + PID lock in a temp `PASEO_HOME`; E2E uses real daemon-backed workspaces, including a two-daemon isolation test and a 390×844 compact bottom-sheet flow (rename, failed delete + retry without duplicate requests).

`npm run typecheck`, `npm run lint`, and `npm run format:check` all pass.

## QA evidence

Captured from real Chromium E2E runs of the final code (isolated daemon + Metro harness):

| | |
|---|---|
| Default group holds Alpha only | Default pin group |
| "Focus" group active after reload, Beta only | Active group after reload |
| Switcher menu: groups + create/rename/delete | Switcher menu open |

Videos: [group switch persists across reload](https://raw.githubusercontent.com/aliozkanozdurmus/paseo/qa-assets/pin-groups/workspace-pin-groups-reload.webm) · [two-daemon membership isolation](https://raw.githubusercontent.com/aliozkanozdurmus/paseo/qa-assets/pin-groups/workspace-pin-groups-two-daemon-isolation.webm)

One pre-existing note: the switcher menu rows use the shared menu engine's standard foreground tokens; in light theme the submenu text is pale — identical to existing menus elsewhere in the app, and not changed by this PR.

## Tested / not tested

| Platform | Status |
|---|---|
| Web (Chromium, desktop viewport) | ✅ E2E — 11 tests |
| Web (Chromium, 390×844 compact sheet) | ✅ E2E — rename, delete, retry |
| Daemon + protocol (Linux) | ✅ Unit + adversarial persistence probes |
| Windows daemon | ⚠️ Not run locally; baseline-execution test is portable and covered by `windows-latest` CI |
| iOS / Android native | ❌ Not tested (shared code paths, compact-sheet E2E covers the small-screen flow) |
| Desktop (Electron) | ❌ Not tested (same web build as browser) |


boudra:

PR #4149, opened 2026-09-01
fix(voice): share the native audio engine between voice mode and dictation

### LINKED ISSUE #4147: bug: Microphone capture could not start because Android audio focus is unavailable - but I am using an iphone
### What's broken

I tried to use the voice mode in the iphone app, it used to work, but now I suddenly get this error.

`Microphone capture could not start because Android audio focus is unavailable`

### Steps to reproduce

1. Open the iphone app
2. Open project
3. Click on the voice mode icon

### Where did this happen

iOS app

### Paseo version

0.5.2

### OS version

_No response_

### Agent provider

None

### Provider configuration

_No response_

### Logs

```text

```

### Screenshots or video

Image

1-Bart-1:
Fixes #4147

## What was wrong

`expo-two-way-audio` keeps **one** engine per app process. The JS side wrapped it **twice**, each wrapper caching its own `initialized` flag:

- `VoiceProvider` — mounted at the app root (`packages/app/src/app/_layout.tsx:665`), lives for the whole process.
- Dictation — created per composer mount (`use-dictation-audio-source.native.ts:24`).

`destroy()` called `native.tearDown()`, which nils the shared engine, and the dictation wrapper destroys on unmount. So:

1. Voice mode runs once → the provider's wrapper sets `refs.initialized = true`.
2. Dictation runs → `initialize()` returns `true` early (engine already exists), so it marks itself initialized too.
3. Composer unmounts (switch agent, leave screen) → dictation `destroy()` → `tearDown()` → **engine is nil**.
4. Voice mode → `ensureInitialized()` short-circuits on the provider's now-stale flag → `toggleRecording` returns `false` → error, **every time, for the rest of the process**.

On iOS that return value can only mean one thing: `ExpoTwoWayAudioModule.swift:62-74` returns `false` solely from `guard let audioEngine = self.audioEngine else { print("AudioEngine not initialized"); return false }` — `AudioEngine.swift:309-323` returns `val` unconditionally otherwise. Android reports genuine audio-focus denial through the same `false`, which is how an iPhone ended up showing an Android audio-focus message.

## The fix

Ownership of the native engine is now reference counted in `native-audio-session.ts`: created for the first holder, torn down only when the last one releases. `startRecording` rebuilds the engine once if `toggleRecording` reports failure, so a stale engine is recoverable instead of permanent. The message no longer names a platform.

`native-audio-module.ts` is a typed lazy loader for the native module — the `require` stays inside a function because the package calls `requireNativeModule` at import time, which throws on a binary without the module linked.

## Evidence

The regression test run against the pre-fix engine (HEAD version, one line changed to use the loader seam):

```
FAIL  |unit| src/voice/audio-engine.native.test.ts > native audio engine > keeps capturing after another engine is destroyed
AssertionError: promise rejected "Error: Microphone capture could not start…" instead of resolving
Caused by: Error: Microphone capture could not start because Android audio focus is unavailable.
 ❯ Object.startCapture src/voice/audio-engine.native.ts:284:17
```

That is the exact error from #4147, reproduced in CI. With the fix:

```
$ npx vitest run src/voice --project unit
 Test Files  3 passed (3)
      Tests  27 passed (27)

$ npm --workspace packages/app run typecheck   # after npm run build:server
exit 0, no errors

$ npm run lint -- packages/app/src/voice

$ npm run format:check
```

Coverage: `native-audio-session.test.ts` covers refcounting, single init under concurrent callers, engine recovery, focus-denied, and init failure. `audio-engine.native.test.ts` drives two real `createAudioEngine` wrappers over one fake native module — the regression itself. Both share the fake in `src/voice/test-utils/fake-native-audio-module.ts`.

## Platforms

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | Not yet on device | The reported platform. Repro: voice mode once → dictation → leave the composer → voice mode. Broken before, works after. |
| Android | Not yet on device | Same code path. Genuine audio-focus denial still throws, with the reworded message. |
| Web | N/A | Separate implementation in `audio-engine.web.ts`, untouched. |
| Desktop | N/A | Web implementation. |

JS-only change, so a dev-client reload is enough to verify on device — no native rebuild.

### DIFF
--- a/packages/app/src/voice/audio-engine.native.test.ts
+++ b/packages/app/src/voice/audio-engine.native.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it, vi } from "vitest";
+
+import type { FakeNativeAudioModule } from "@/voice/test-utils/fake-native-audio-module";
+
+const fakeRef = vi.hoisted(() => ({ current: null as FakeNativeAudioModule | null }));
+
+vi.mock("@/voice/native-audio-module", async () => {
+  const { createFakeNativeAudioModule } =
+    await import("@/voice/test-utils/fake-native-audio-module");
+  fakeRef.current = createFakeNativeAudioModule();
+  return { loadNativeAudioModule: () => fakeRef.current!.native };
+});
+
+import { createAudioEngine } from "@/voice/audio-engine.native";
+
+function createEngine() {
+  return createAudioEngine({
+    onCaptureData: () => undefined,
+    onVolumeLevel: () => undefined,
+  });
+}
+
+describe("native audio engine", () => {
+  /**
+   * Voice mode and dictation wrap the same native engine. Destroying the dictation wrapper used
+   * to tear that engine down under voice mode, which then failed every later capture.
+   */
+  it("keeps capturing after another engine is destroyed", async () => {
+    const state = fakeRef.current!.state;
+    const voice = createEngine();
+    await voice.initialize();
+    await voice.startCapture();
+    await voice.stopCapture();
+
+    const dictation = createEngine();
+    await dictation.initialize();
+    await dictation.startCapture();
+    await dictation.stopCapture();
+    await dictation.destroy();
+
+    await expect(voice.startCapture()).resolves.toBeUndefined();
+    expect(state.recording).toBe(true);
+    expect(state.tearDownCalls).toBe(0);
+
+    await voice.destroy();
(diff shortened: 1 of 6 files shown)


boudra:
Closing for now: this needs native-device evidence for voice → dictation → leaving the composer → voice on iOS, plus Android audio-focus denial and recovery. Please reopen with those results.

boudra:

Issue #4061, opened 2026-08-30
bug: send-now on a queued message interrupts the active turn instead of steering it


lxsolutions:
### What's broken

With **Send behavior = Queue**, a message sent while the agent is busy correctly goes to the queue. Clicking **send now** on that queued message then **interrupts and replaces the active turn** rather than steering it, discarding the in-flight reasoning and tool results.

Active-turn steering is available on this path — the daemon and the Claude provider both implement `steerActiveTurn` (shipped in v0.5.0 via #3394). The steer intent is discarded in the app layer before it reaches the daemon, so a supported capability is never attempted.

As in #3999, the destroyed turn surfaces as an ordinary turn end, so an aborted turn is hard to distinguish from a completed one without reading the transcript.

### Root cause

Line references against `main` @ `86ef4680031d`.

`sendQueuedComposerMessageNow` (`packages/app/src/composer/actions.ts:290`) removes the item from the queue and calls `submitMessage({ text, attachments })` — with no behavior argument. `submitMessage` derives the behavior solely from the persisted default (`packages/app/src/composer/index.tsx:1426`):

```ts
appSettings.sendBehavior === "steer" ? "steer" : "interrupt",
```

In Queue mode `appSettings.sendBehavior` is `"queue"`, which is not `"steer"`, so the ternary resolves to `"interrupt"`. There is no path by which an explicit send-now from the queue can produce `"steer"`.

Because `activeTurnId` is attached only when the behavior is `"steer"` (`index.tsx:1457`), the daemon also receives no turn id, so it cannot recover the intent.

The same collapse appears twice more:

- `resolveActiveSendBehavior` (`packages/app/src/composer/input/state.ts:12`) — Queue plus a pending permission resolves to `"interrupt"`, never `"steer"`.
- `runAlternateSendAction` (`packages/app/src/composer/input/state.ts:128`) — the keyboard "send anyway" override in Queue mode routes into the same `submitMessage` path.

### Steps to reproduce

1. Set **Send behavior** to **Queue**. Use a provider with native steering (Claude Code).
2. Send a prompt that runs for a while (long tool chain or large read).
3. While the turn is running, type a follow-up and send it — it goes to the queue.
4. Click **send now** on the queued message.
5. The running turn aborts and a replacement turn starts; the in-flight work is lost.

Expected: the turn stays alive and absorbs the message, as it does with **Send behavior = Steer**.

### Expected

[...]

Issue #3999, opened 2026-08-28
bug: sending a message during an active turn replaces and kills it, losing in-flight work


pmbemax:
### What's broken

Sending a message to an agent that is mid-turn **replaces and kills the running turn** instead of queueing behind it or steering it. The in-flight reasoning is lost and the turn is reported as aborted.

This is distinct from an explicit interrupt: no stop was requested. A second message arriving during an active turn is enough.

### Impact, measured

Over a single long-running orchestration session on a self-hosted deployment:

- **299 turn aborts** attributable to a message arriving during an active turn
- **59 of those lost work outright** — the turn had produced reasoning/tool results that were discarded rather than persisted
- **17 turn kills in one week** across the fleet from the same cause

The pattern is worst for supervisory setups: a scheduled prompt, a hourly pulse, or a second operator message lands while the agent is genuinely working, and the work in flight is destroyed. Because the abort surfaces as a normal turn end, it is easy to mistake for the agent finishing.

### Steps to reproduce

1. Start an agent and send a prompt that will run for a while (long tool chain, large read).
2. While the turn is still running, send a second message.
3. The first turn terminates; its in-flight work is not preserved.

### Expected

One of:
- **Queue** the incoming message and deliver it at the next turn boundary, or
- **Steer** the active turn with it (see the sibling issue on `steerActiveTurn` declaration), or
- at minimum, **refuse the send with a clear error** rather than silently destroying the running turn.

Any of the three is recoverable. Silent replacement is not, because the operator cannot tell an aborted turn from a completed one without reading the transcript.

### Environment

- `@getpaseo/cli` 0.6.1, self-hosted, Linux
- Provider: built-in `omp` (declared `"omp": { "enabled": true }`, not a custom ACP provider)

## Separate OMP propagation case from #2033

Preserve this independently of the mid-turn replacement reports above and the steering-capability allegation in [#4000](https://github.com/getpaseo/paseo/issues/4000). These are reporter observations, not a reproduced run or proof of a common cause.

| Case | Environment and trigger | Evidence | Next action and completion criterion |
| --- | --- | --- | --- |
[...]

Issue #3964, opened 2026-08-28
bug: ACP provider: no steerActiveTurn implementation, so mid-turn input fails with the agent's busy error


deokcheoncityboy:
### What's broken

Sending a message while an ACP-provider agent is busy fails with the agent's raw
busy error instead of steering or falling back.

I expected the message to either steer the active turn, or to interrupt and
replace it. Instead the turn fails and the error is surfaced to me as-is.

acp-agent.ts has no steerActiveTurn implementation, while
codex-app-server-agent.ts and opencode-agent.ts both do.

### Steps to reproduce

1. Configure a custom ACP provider in ~/.paseo/config.json (extends: "acp").
2. Start an agent with it and send a prompt that runs for a while.
3. While the turn is still running, type another message and send it
   (or hit stop and immediately send a new one).
4. The turn fails with the busy error below.

Reproduces every time.

### Where did this happen

Desktop (Electron)

### Paseo version

0.6.1

### OS version

macOS 26.5.2 (arm64)

### Agent provider

Custom provider

### Provider configuration

gajae-code (gjc) v0.15.3, via openai-codex / gpt-5.6-luna

~/.paseo/config.json:
"gjc": {
  "extends": "acp",
  "label": "Gajae Code",
  "command": ["/Users/gimdongha/.local/bin/gjc", "acp"],
  "env": {
    "GJC_ACP_PERMISSION_MODE": "auto",
    "GJC_ACP_ABORT_SCOPE": "owned",
    "GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS": "600000",
    "GJC_OPENAI_CODE_WEBSOCKET_IDLE_TIMEOUT_MS": "600000"
  },
  "params": { "supportsMcpServers": false },
  "enabled": true
}

initialize response advertises:
  loadSession: true
  sessionCapabilities: { list, fork, resume, close, delete }

### Logs

```text
[...]

PR #4154, opened 2026-09-01
fix(acp): steer Hermes without replacing active turns

atomlink-ye:
### Linked issue

Addresses #3964 and #3999. Related to #4061 and #4112.

This PR carries the exact patch from #4113 as a dedicated commit with Burak Cabadan's original authorship, because derived ACP providers otherwise lose `steerActiveTurn` at the wrapper boundary.

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Refactor
- [x] Docs

### Reasoning

Paseo currently treats an unavailable ACP steer as interrupt-and-replace. That is particularly destructive with Hermes: a message sent while Hermes is running a tool cancels the tool, starts another foreground turn, and can strand Hermes' own queued prompt behind its interrupted ACP turn.

Hermes already has the transport needed for the intended behavior. A second plain-text `session/prompt` received while its turn is active is redirected into the same `AIAgent` turn. This change lets a generic ACP provider explicitly opt into that transport. A small internal `HermesACPAgentClient` owns the Hermes default, following the existing provider-specific ACP client pattern instead of adding command sniffing to the generic registry. Paseo no longer sends `session/cancel` for an ordinary Hermes steer.

This follows the optional `AgentSession.steerActiveTurn` contract used by the Pi implementation in #3752 while preserving provider-specific acknowledgment semantics. Pi can queue first because its native `steer` RPC acknowledges admission; ACP `session/prompt` resolves only when the turn ends, so concurrent ACP steering clears a blocking permission before dispatch and sends nothing if that clear fails.

### Goals

- Steer an active Hermes ACP turn through a concurrent `session/prompt`, without canceling or replacing its foreground turn.
- Keep the initial prompt, correction, tool output, and final answer on one Paseo turn.
- Preserve explicit Stop/Interrupt: Stop can preempt a steer whose auxiliary ACP response never settles.
- Keep unsupported ACP providers on their existing behavior unless they explicitly opt in.
- Enable existing canonical `hermes acp` provider configs while preserving `activeTurnSteering: "none"` as an explicit opt-out.
- Preserve exact user-message identity and suppress only matching provider echoes.
- Handle both original-first and steer-first ACP response ordering without losing late updates.
- Prevent a late steer completion from an interrupted turn from mutating or settling the next turn.

### Non-goals

- No Hermes source changes.
- No generic `_session/steering` extension support; #3921's vendor-extension approach was reverted and is separate from Hermes' existing redirect transport.
- No change to the global composer send policy or explicit Interrupt behavior.
- No `/reload-mcp` or broader Hermes CLI/ACP command parity work.
- Concurrent steering is plain non-command text only. Attachments, slash commands, and control input fail before dispatch.

#### Before

On Paseo 0.7.0 with the installed Hermes ACP provider, I started a foreground `sleep 60` tool call and sent a message while it was active. Paseo canceled the original turn and sent a replacement prompt. The observed result was:

```text
[Command interrupted]
Internal error: 'NoneType' object has no attribute 'startswith'
Queued for the next turn. (1 queued)
```

The replacement produced a second Paseo foreground turn and the queued Hermes message did not run.

A deterministic ACP subprocess fixture also recorded one `session/cancel` followed by the replacement `session/prompt` on the unmodified path.

#### After

Tested on macOS 26.2, Apple Silicon, Paseo 0.7.0, and:

```text
Hermes Agent v0.20.5 (2026.8.19)
```

The final exact-SHA test bundle used an in-process daemon on an OS-assigned loopback port with a temporary `PASEO_HOME` and temporary workspace. It did not restart or send test traffic through the normal daemon on port 6767.

Focused deterministic daemon E2E:

```text
$ npx vitest run packages/server/src/server/daemon-e2e/acp-active-turn-dispatch.e2e.test.ts --bail=1
Test Files  1 passed (1)
Tests       3 passed (3)
```

Those cases cover:

- steer response before original response;
- original response before steer response, including a permission release and late update;
- an unresolved steer response followed by explicit Stop, which must deliver `session/cancel` promptly, emit one canceled turn, and never resume from the late steer response.

Real Hermes E2E on exact candidate `0b00d2c59f60e65a958faeffeb3f4d1aa6c14bb8`:

```text
$ npm run test:e2e:hermes-steering --workspace=@getpaseo/server
Test Files  1 passed (1)
Tests       1 passed (1)
```

Observed redacted summary:

```json
{
  "candidateHead": "0b00d2c59f60e65a958faeffeb3f4d1aa6c14bb8",
  "actual": {
    "finalStatus": "idle",
    "turnStartedCount": 1,
    "turnCanceledCount": 0,
    "assistantText": "Redirected the active turn with your correction. HERMES_ACP_STEER_ACK"
  },
  "nextPrompt": {
    "status": "idle",
    "acknowledged": true
  },
  "cleanup": {
    "workspaceRemoved": true,
    "paseoHomeRemoved": true
  }
}
```

The completed shell tool contained both `HERMES_ACP_STEER_START` and `HERMES_ACP_STEER_END`. The initial prompt and correction share `<turn-1>`; the health-check prompt is `<turn-2>`. The evidence artifacts contain no raw UUIDs or local home paths, and both temporary roots were removed.

Focused adapter, registry, catalog, deterministic daemon, and evidence tests on the final candidate:

```text
Test Files  7 passed (7)
Tests       184 passed (184)
```

An additional lifecycle regression holds a steer from turn A, interrupts A, admits and defers a steer on turn B, then proves the stale A completion cannot decrement or flush B's settlement state.

Repository gates:

```text
npm run typecheck     # passed
npm run lint          # 0 warnings, 0 errors
npm run format:check  # all matched files correctly formatted
```

Not tested manually on Windows, Linux, iOS, Android, or Electron UI. This change has no UI surface.

### Known follow-ups from isolated local deployment

Before the history cleanup, I built the equivalent code tree at `9d334208a26990211824c91bd9f0b891ee3ec747` and ran it as a second daemon with an isolated `PASEO_HOME` on `127.0.0.1:6769`, while the normal daemon remained on `6767`. The installed Hermes provider completed a 60-second tool, accepted a correction through `activeTurnBehavior: "steer"` on the same turn without cancellation or replacement, and then completed a normal next turn. The rewritten candidate above was revalidated only through local unit, deterministic daemon, and ephemeral real-Hermes tests; it was not redeployed.

That deployment exposed two independent follow-ups which are intentionally outside this PR:

1. The `paseo send` CLI currently calls ordinary `sendAgentMessage` without setting `activeTurnBehavior: "steer"`. A CLI message sent during an active turn is therefore queued for the next turn, while the UI/composer and direct client API can exercise same-turn steering. A future CLI change could expose this explicitly, for example as `paseo send --steer`.
2. When Hermes deliberately runs a tool in the background and then waits through the process tool, the `process wait` timeline row can remain `running` after the process and agent have completed. The same stale status reproduces without any steering message, so it is an existing Hermes ACP tool-status mapping issue rather than a regression from concurrent steering.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


atomlink-ye:
cc @boudra

This is the Hermes ACP counterpart to #3752 and is ready for independent review/QA at `0b00d2c59f60e65a958faeffeb3f4d1aa6c14bb8`.

Highlights:

- keeps a running Hermes tool and correction on one Paseo turn through a concurrent ACP `session/prompt`, without interrupt-and-replace;
- preserves explicit Stop preemption, both prompt-response orderings, exact echo identity, permission release, and stale completion isolation;
- documents the provider-specific distinction from Pi: native Pi steering can acknowledge queueing before permission cleanup, while ACP clears blocking permissions before dispatch because `session/prompt` only resolves at turn completion;
- carries Burak's #4113 wrapper-forwarding patch as its own author-preserving commit;
- history is cleaned to three reviewable commits: feature, wrapper fix, and E2E/evidence coverage.

Local evidence on the rewritten exact SHA:

- focused adapter/registry/catalog/deterministic E2E: 7 files, 184 tests passed;
- `npm run typecheck`, `npm run lint`, and `npm run format:check` passed;
- real Hermes E2E passed with one started turn, zero canceled turns, the tool reaching END, the correction acknowledged on the same turn, and a healthy next turn;
- independent final history/code review: APPROVE.

The PR body also records two non-blocking follow-ups found during isolated deployment: the CLI does not yet expose an explicit `send --steer`, and Hermes background `process wait` rows can remain visually running even without steering.

Could you take a look and run the normal independent QA/delivery flow when appropriate? Fork Actions are currently waiting for maintainer approval rather than reporting code failures.

PR #4155, opened 2026-09-01
feat(app): show a workspace's agents in the sidebar

kiawin:
### Linked issue

Closes #

### Type of change

- [x] New feature

### Reasoning

A workspace row shows one status dot no matter how many agents are running inside it. `utils/workspace-agent-activity.ts` walks every agent, derives a status bucket for each, then keeps only the most recently active one and discards the rest.

So a workspace running three agents looks exactly like one running one. If the second agent needs input, nothing says so, and there is no way to reach it without opening the workspace and hunting through its tabs. That is the case this change is for: work you are *not* currently looking at.

Everything needed already existed and was simply never joined up — `useAggregatedAgents` has the agents, `deriveSidebarStateBucket` has the status, `navigateToAgent` already resolves an agent to its workspace and reveals its tab. What was missing was the collapse, and somewhere to put the result.

### Goals

- A workspace row shows how many agents are working in it, and can expand into them.
- Each agent row carries its own status and provider, and clicking one goes straight to that agent.
- The workspace row's own dot and status are unchanged, so a collapsed sidebar looks exactly as it does today.
- Both sidebar list modes (Group by Project and Group by Status) behave identically.
- The preference is discoverable next to its siblings, with an off switch.

### Non-goals

- **No sub-list for a single-agent workspace.** The row already *is* that agent's row — its dot, its status, and its click target all belong to that agent, so a lone child restating them is noise. The list appears at two or more.
- **No stale-decay to `idle`.** Orca repaints a quiet agent idle after 30 minutes because its status is inferred from hook callbacks that can be missed. Paseo's status is authoritative, so an agent quiet for an hour is genuinely `running` and must not be relabelled.
- **No lineage tree, per-row dismiss, or send-to-agent.** Subagents already have the subagents track; duplicating it here would put one fact in two places.
- **No protocol change.** App-only — no new RPC, no capability gate, no daemon change.

### Design notes

Two things live on the row, and they are deliberately different kinds of thing:

- **The count is a fact**, the same class as host, pull request, and checks, so it sits on the meta line and leads it — the only item there about live work rather than about the workspace itself.
- **The disclosure is a control**, so it sits in the leading slot, where a project row already swaps its icon for a chevron on hover. A workspace row now says it expands in the same place a project row does.

They started as one chip on the title line, which is the one line a kebab overlays. Every geometry problem this change had came from that; splitting them dissolved it.

The workspace chevron needs its own press target, which a project row never did: a project header's press already toggles collapse, so its chevron can be decoration, while a workspace row's press navigates and a chevron inheriting it would open the workspace instead.

The count is also a press target. On a phone there is no hover to reveal the chevron, so without it the only way in would be the 6px status dot.

### QA

Two platforms. Each screenshot below names the file to drop in its place — they are staged locally in `qa-evidence/` (not committed).

#### Desktop web — Chromium, 1280x720

Generated from this branch by a committed spec, not hand-taken:

```
PASEO_SIDEBAR_EVIDENCE=1 npx playwright test --project=browser \
  e2e/browser/sidebar-agent-rows-evidence.spec.ts
```

Each capture seeds **two** workspaces — one with three agents and one with a single agent — so every screenshot also shows what the change does *not* do to an ordinary row.

**1. Resting state.** The busy workspace carries `3` on its meta line; the single-agent row is untouched.

web-01-collapsed

**2. Hovered.** The status indicator swaps for the expand chevron, the way a project row's icon does. The count stays on the meta line, clear of the kebab.

web-02-hover-chevron

**3. Expanded.** One row per agent — status dot, provider mark, title. Dots sit on the count's left rail.

web-03-expanded

**4. Where the preference lives.** Display preferences → Show → **Agents**. Not the Settings screen: every sidebar display preference lives in this menu, so this one joins Checks rather than starting a second home for the same kind of decision.

web-04-display-menu

**5. The three answers**, with the shipped default marked.

web-05-agents-options

**6. Group by Status.** The other list mode, same behaviour.

web-06-status-grouping

#### Android — emulator, API 36

Debug build from this branch, driven over `adb`, against a scratch daemon seeded with one three-agent workspace and two single-agent ones.

This is the half the first round of review could not speak to. The touch path has no hover, so the chevron never appears and the count is what makes the sub-list reachable at all.

**1. Resting state.** The count on the busy row; the two single-agent rows draw nothing.

android-01-collapsed

**2. Tapping the count expands it.**

android-02-expanded-via-count

**3. Tapping an agent row opens that agent** and dismisses the sidebar.

android-03-agent-opens

**4. Reopened after navigating away.** The expansion survived, and the three agents no longer agree: one is grey while two stay green. The single-winner model this replaces could not show that.

android-04-persisted-and-per-agent-status

**5. The status dot collapses it again.**

android-05-collapsed-via-dot

**6. Where the preference lives on mobile.** `Agents` sits under `Checks`, rendered as a sheet rather than a popover.

android-06-show-menu

**7. The three modes**, `Collapsed` checked.

android-07-agents-options

**8. The off switch.** Rows identical to pre-change.

android-08-hidden

**9. Group by Status.**

android-09-status-grouping

**10. Expanded in that mode**, where the leading slot is a project badge rather than a plain indicator.

android-10-status-grouping-expanded

#### Automated

- `packages/app/e2e/browser/sidebar-agent-rows.spec.ts` — count appears and expands; a single-agent workspace draws nothing; and two geometry assertions that compare **bounding boxes** rather than visibility.

  That distinction is not academic. The disclosure originally shipped inside the trailing slot, where the kebab overlay painted straight over it — and `toBeVisible()` passed the whole time, because a fully occluded element is still "visible". The measured overlap was 3px; the visual loss was the entire digit, because the 48px scrim behind the kebab is wider than the icon. Both assertions exist because that shipped once.

- Unit tests across the activity index's identity and clock invariants, the preference parser, the expansion overrides and their persistence round-trip, and the new meta-row item.

- Full app suite and all four Playwright shards green on CI.

- Rebased onto current `main` and re-verified there: `npm run typecheck` and `npm run lint` clean, and the unit tests for the activity index, the preference parser, the expansion store, the meta-row item, and the type scale pass (51 tests).

#### Reviewed

Two external passes over the first commit. One found nothing; the other found a real regression I had introduced — the reuse pass required title equality before carrying `enteredAt` forward, so a provider renaming a session restarted its status clock and pushed the row's "Last activity" stamp to now. Fixed in `fix(app): hold the agent status clock through a rename`, with a test verified to fail without the fix.

#### Not covered

- **No iOS, no Electron.** Desktop Chromium and Android only.
- **Auto-expand on a blocked agent is untested on device.** It needs an agent in `needs_input`, which the mock provider does not produce.
- **The sub-list rail is a constant pinned to desktop geometry.** The e2e asserts it within 1px, but only on the desktop project.
- **Provider icons in the screenshots are the generic fallback**, because the harness uses the `mock` provider. Real Claude/Codex/Copilot marks resolve through the same `getProviderIcon` path used by History rows.
- **Row spacing on a real phone is unjudged.** Adjacent agent rows are below the 24dp comfortable target; an emulator cannot answer whether that matters in the hand.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/sidebar-agent-rows-evidence.spec.ts
+++ b/packages/app/e2e/browser/sidebar-agent-rows-evidence.spec.ts
@@ -1,0 +1,160 @@
+import { test, expect } from "../support/fixtures";
+import { gotoAppShell } from "../support/helpers/app";
+import { seedWorkspace } from "../support/helpers/seed-client";
+import { getServerId } from "../support/helpers/server-id";
+
+/**
+ * Capture-only. Produces the PR evidence for the sidebar agent sub-list into
+ * `test-results/evidence/`, at 3x so the crops stay readable when scaled down in a PR body.
+ *
+ * Opt-in because it exists to write files, not to assert: the behaviour it photographs is
+ * covered by `sidebar-agent-rows.spec.ts`, which runs on every push. Same shape as the
+ * `PASEO_DIFF_PERF_E2E` capture spec.
+ *
+ *   PASEO_SIDEBAR_EVIDENCE=1 npx playwright test --project=browser \
+ *     e2e/browser/sidebar-agent-rows-evidence.spec.ts
+ */
+
+const MULTI_AGENT_TITLES = [
+  "Rebase onto upstream main",
+  "Port sidebar agent rows",
+  "Audit COMPAT tags",
+];
+
+/** The sidebar plus a little of the pane behind it, so the row's right edge is in frame. */
+const SIDEBAR_CLIP = { x: 0, y: 124, width: 340, height: 292 };
+
+/**
+ * Wide enough for the sidebar plus the preference menu and its submenu, which open to the right
+ * of the trigger as separate popovers. Framing them together is the point: the shot has to show
+ * where the control lives, not just what it says.
+ */
+const MENU_CLIP = { x: 0, y: 96, width: 720, height: 480 };
+
+/**
+ * Popovers fade in, and `toBeVisible` is satisfied before the animation lands — a shot taken then
+ * catches the menu half-transparent over whatever is behind it. Capture-only, so waiting a beat
+ * beats plumbing an animation signal through the menu engine.
+ */
+async function settlePopover(page: import("@playwright/test").Page): Promise<void> {
+  await page.waitForTimeout(400);
+}
+
+test.describe("sidebar agent rows evidence", () => {
+  test.skip(!process.env.PASEO_SIDEBAR_EVIDENCE, "Set PASEO_SIDEBAR_EVIDENCE=1 to capture.");
+  test.use({ viewport: { width: 1280, height: 720 }, deviceScaleFactor: 3 });
+
+  test("captures the sub-list states", async ({ page }) => {
+    // Two workspaces on purpose: one with several agents and one with a single agent, so every
+    // shot also shows what the change does NOT do to an ordinary row.
+    const busy = await seedWorkspace({ repoPrefix: "paseo-evidence-busy-" });
+    const quiet = await seedWorkspace({ repoPrefix: "paseo-evidence-quiet-" });
+    try {
+      for (const title of MULTI_AGENT_TITLES) {
+        await busy.client.createAgent({
+          provider: "mock",
+          cwd: busy.repoPath,
+          workspaceId: busy.workspaceId,
+          title,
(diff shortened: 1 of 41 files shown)


boudra:

PR #4173, opened 2026-09-01
fix(app): create new workspaces on the host you are already working on

### LINKED ISSUE #4172: Sidebar project + creates the workspace on the local host instead of the host you are working on
> **Edited 2026-09-14.** The original report blamed the New workspace composer and proposed persisting the host picker's selection. Both were wrong — see [the correction comment](https://github.com/getpaseo/paseo/issues/4172#issuecomment-5510991444). The composer already follows the active workspace's host. The actual bug is the sidebar project `+`, and it is reproducible with two live daemons. Rewritten to describe that.

## Problem

On a multi-host setup, the `+` on a sidebar project row creates the new workspace on the **local** host even when I am currently working in a workspace on a **remote** host, and the project exists on both.

Paseo otherwise has one consistent rule for which host a new workspace lands on — the host you are already working on:

| Entry point | Host it picks | Follows the rule? |
| --- | --- | --- |
| Cmd+N / global **New workspace** | active workspace's `serverId`, passed as a route param | yes |
| Composer at bare `/new` | `lastActiveProject` → the active workspace's host | yes |
| **Sidebar per-project `+`** | **first qualifying host in registry order** | **no** |

The sidebar `+` is the odd one out, and it is the entry point I actually use, because it is right there on the project I am working in.

Getting this wrong is expensive: the agent starts against the wrong checkout on the wrong machine, and I only notice after it has begun working.

## Repro

1. Register two hosts, both online.
2. Check out the same repository on both, so the two placements merge into a single sidebar project row (shared git remote → same `projectId`).
3. Open a workspace belonging to that project **on the remote host**, so it is the active workspace.
4. Click the `+` on that project's sidebar row.

**Expected:** the new workspace is created on the remote host — the one I am working on, matching Cmd+N and the composer.
**Actual:** the composer opens on the local host and the workspace is created there.

Version: `0.7.2`, web (Chromium). Not platform-specific — the resolution is shared app code.

## Evidence

This reproduces in the existing two-live-daemon e2e harness, so it does not need my particular machines. `e2e/browser/sidebar-project-grouping.spec.ts` already stands up two real daemons via `startIsolatedHostDaemon` with one repository grouped across both.

Adding a test that opens the **secondary** host's workspace and then clicks the project `+`, against unmodified `main`:

```
Expected substring: "Secondary Host"
Received string:    "Primary Host"
```

The workspace is then created on Primary too, so it is not only a preselection problem.

Worth pairing it with the mirror-image case — open the **primary** host's workspace and assert the `+` follows back to Primary. That one passes on `main` as well, because `main` always picks Primary. The asymmetry is what makes the pair meaningful: one pins the bug, the other stops a "fix" from simply hard-coding the opposite host.

## Where it comes from

`resolveNewWorkspaceTarget` in `packages/app/src/utils/sidebar-project-row-model.ts` walks `project.hosts` and returns the first host that can create a workspace:

```ts
for (const host of project.hosts) {
  if (host.worktreeSupport === "unsupported" && !supportsMultiplicityByServerId.get(host.serverId)) {
    continue;
  }
  const target = hostTarget(host);
  if (target) return target;
}
```

Nothing in that walk considers where the user is working. For a merged project, `hosts[]` is in host-registry iteration order, which puts the local daemon first in the common case since it registers first. The resulting `serverId` is baked into the route as `?serverId=`, and the route param wins outright in `resolveNewWorkspaceInitialServerId` — so the composer cannot correct it downstream.

That last part is why an online check matters for any fix here in a way it does not in the composer: routing to a host that is registered but offline would stick.

The parallel to follow is already in the component — `activeWorkspaceSelection` is a prop on `sidebar-workspace-list.tsx`, at the same level as the `buildSidebarProjectRowModel` call.

## Status

I had a patch for this in #4173, closed under the new PR policy. Not resubmitting, and no follow-up needed on my end — filing the corrected diagnosis and the repro so this is actionable if it is ever worth picking up. The branch is still on my fork if the e2e test is useful on its own.

mwhitman-steward:
### Linked issue

Closes #4172

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

If you work across two machines and a project is registered on both, the sidebar's per-project "+" always creates the workspace on the local daemon — even when you are currently working on the remote host. You notice after the agent starts, against the wrong checkout, and have to throw the workspace away and start again.

Paseo already has one rule for this: new work happens on the host you are already working on. Cmd+N follows it, and so does the composer on a bare `/new`. The sidebar "+" is the only entry point that ignores it, which is unfortunate because it is the one people reach for most.

| Entry point | Host it picks | Follows the rule? |
| --- | --- | --- |
| Cmd+N / global "New workspace" | active workspace's `serverId`, passed on the route (`use-global-new-workspace-action.ts`) | yes |
| Composer on a bare `/new` | `lastActiveProject`, synthesised from the active workspace by `hostProjectFromWorkspace`, so it carries exactly one host | yes |
| Sidebar per-project "+" | first qualifying entry in the project's merged `hosts[]` | **no** |

The sidebar row is the odd one out because its `SidebarProjectEntry.hosts` is the *merged* multi-host array from `buildWorkspaceStructureProjects`, ordered by host registry, which puts the local daemon first in the common case. `resolveNewWorkspaceTarget` takes the first qualifying entry, so a project on both machines always resolves local. That decision is final rather than a default: the row navigates with an explicit `?serverId=`, and `resolveNewWorkspaceInitialServerId` honours a route param outright, so the composer never gets a say.

### Goals

- The sidebar "+" targets the active workspace's host when the project is on that host and a workspace can be created there.
- Fall back to today's behaviour, unchanged, whenever it cannot: no active workspace, the project is not on that host, the host cannot host a workspace, or the host is not online.
- Require the host to be online. This matters more here than elsewhere — because the route param wins outright, an offline host would stick with nothing downstream to correct it.
- No new persisted state.

### Non-goals

- **The global "New workspace" / Cmd+N action is untouched.** It already passes the active workspace's `serverId`, which is correct context propagation rather than the bug.
- **`orderHostsLocalFirst` is untouched.** Local-first ordering is correct for *display*; this changes *selection* only.
- **No persisted host preference.** Earlier commits on this branch took that approach — a stored "last host you picked by hand". Reviewing it against the code showed the composer already follows the active workspace, so the preference was a second competing convention rather than an extension of the existing one, and it did not fix the sidebar at all. The final commit removes it. Trade-off accepted: with no stored state, a cold start with no active workspace still falls back to the local-first walk.
- Intermediate commits are left in place rather than force-pushed over the maintainer's merge. The net diff is three files.

### QA

**Reproduced before, confirmed after, on two live daemons.** `sidebar-project-grouping.spec.ts` boots a real second host via `startIsolatedHostDaemon` and groups one repository across both through a shared Git remote, which is the exact precondition for this bug.

Open the secondary host's workspace so it is the active one, then click the project row's "+":

| | Composer opens on | Workspace created on |
| --- | --- | --- |
| `main` | **Primary Host** (the bug) | Primary Host |
| this branch | **Secondary Host** | Secondary Host |

On `main` the assertion fails with `Expected substring: "Secondary Host"` / `Received string: "Primary Host"`. The sibling test opens the *primary* workspace and asserts the "+" follows back to Primary; it passes on `main` too, because `main` always picks primary. That asymmetry is what makes the pair conclusive rather than merely green. Both tests assert the workspace is actually created on that host, not just preselected.

Commands run:

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npx vitest run src/utils/sidebar-project-row-model.test.ts` — 17 passed
- `npx playwright test --project=browser e2e/browser/sidebar-project-grouping.spec.ts` — 9 passed
- `npx playwright test --project=browser e2e/browser/sidebar-workspace.spec.ts e2e/browser/new-workspace-entry.spec.ts` — 1 pre-existing failure, `new-workspace-entry.spec.ts` › "Ctrl+P opens the project picker with search focused", reproduced identically on unmodified `main` with my changes reverted

**No screenshots.** There is no visual change — the sidebar row layout, host ordering and every label are untouched. The only difference is which `serverId` the "+" puts on the route. **I have not tested this on iOS or Android.**

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence — with the two limitations called out above


boudra:
Closing for now: this needs a live two-host check showing the sidebar + creates the workspace on the active remote host when the project exists on both. Please reopen with that evidence.

mwhitman-steward:
Evidence added — and you were right to ask, because it caught something the unit tests could not.

I had said the e2e harness could only fake a second host as offline. That was wrong: `startIsolatedHostDaemon` boots a genuinely running second daemon, and `sidebar-project-grouping.spec.ts` already stands up two live hosts with one repository grouped across both via a shared Git remote. That is exactly the scenario you asked for, so I added the check there rather than inventing a new fixture.

**Live two-host result, both daemons online, project registered on both:**

Open the *secondary* host's workspace so it is the active one, then click the project row's "+":

| | Composer opens on | Workspace created on |
| --- | --- | --- |
| `main` | **Primary Host** ❌ | Primary Host |
| this branch | **Secondary Host** ✅ | Secondary Host |

On `main` the assertion fails with:

```
Expected substring: "Secondary Host"
Received string:    "Primary Host"
```

The sibling test opens the *primary* workspace and asserts the "+" follows back to Primary. It passes on `main` too — because `main` always picks primary. That asymmetry is what makes the pair conclusive rather than merely green: one test pins the bug, the other proves the fix follows the active host instead of just hard-coding the other direction.

Both tests assert the workspace is actually **created** on that host, not merely that the composer preselects it.

Full spec on this branch: 9/9 passing.

Two things I want to be straight about:

- My earlier QA section claimed this was untestable and offered a manual check. That was a limitation of my search, not of the harness. It is now automated and permanent regression coverage, which is better than the screenshots I offered.
- I cannot reopen this PR myself — I only have read access here, and both `gh pr reopen` and the REST equivalent return a validation error. The branch is pushed and up to date with `main`. Could you reopen it? If you would rather have a fresh PR, say the word and I will open one.

boudra:

PR #4230, opened 2026-09-02
Let loopback clients skip the daemon password

panrafal:
### Linked issue

None. There's no existing issue or discussion for this — it came out of running a password-protected daemon day to day. Happy to open a product discussion first if you'd rather that came before the code.

### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Setting a daemon password is the right move the moment you bind past localhost, but today it locks out the machine the daemon runs on. `paseo` on that box, agents that call back into the daemon, and a browser at `localhost` all have to present the password too. In practice that means exporting `PASEO_PASSWORD` into your shell profile and into the environment of every local tool — which spreads the plaintext password across exactly the places you set a password to avoid.

The flow I actually want is "authenticate my phone, don't authenticate me." This adds a config-file option for that: remote devices still need the password, connections from the daemon's own machine don't.

### Goals

- A daemon with a password set can still be driven from its own machine with no password: `paseo ls`, agents, the MCP endpoint, and a browser at `localhost`.
- Remote clients are unaffected — the password is still required, and nothing about the relay or Hub paths changes.
- The exemption cannot be tricked by a header. It's decided from the kernel-reported peer address, never `req.ip`, because `req.ip` honours `trust proxy` and a daemon with `trustedProxies: true` would otherwise accept `X-Forwarded-For: 127.0.0.1` from anyone.
- A reverse proxy on the same host can't launder remote clients through it. Any request carrying `Forwarded`, `X-Forwarded-For`, or `X-Real-IP` counts as remote even on a loopback socket, because such a proxy connects from `127.0.0.1` for every client it relays.
- Off by default; existing daemons behave exactly as before.
- The security trade-off is written down where someone deciding whether to enable it will read it.

### Non-goals

- **No environment variable and no CLI flag.** Config file only. This punches a hole in the daemon's only authentication, so it should live next to the password it exempts, not somewhere a container image or a shell profile can set it by accident. Deliberate ergonomic cost.
- **Not a defence against local processes.** Enabling it hands the daemon to anything that can open a socket on that machine. The stored bcrypt hash is not a usable credential, so the password genuinely was keeping local callers out and this gives that up on purpose. The docs say so plainly.
- **Doesn't help behind a header-stripping reverse proxy.** If a proxy removes the forwarding headers, the check can't see it. Documented as "leave this off when a proxy fronts the daemon" rather than papered over.
- **No UI.** No settings toggle, no app surface.
- **No protocol change.** `packages/protocol` is untouched; this is server-local persisted config.

### QA

Config used throughout, in a throwaway `PASEO_HOME`, daemon on port 6799. The main daemon on 6767 was not touched.

```json
{
  "daemon": {
    "listen": "127.0.0.1:6799",
    "auth": {
      "password": "$2b$12$...",           // "correct-password"
      "allowLoopbackWithoutPassword": true
    }
  }
}
```

**Startup confirms the field is read:**

```
{"host":"127.0.0.1","port":6799,"authRequired":true,"msg":"Server listening on http://127.0.0.1:6799"}
{"allowLoopbackWithoutPassword":true,"msg":"Daemon password authentication enabled"}
```

**HTTP, exemption on:**

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:6799/api/status
200
$ curl -s -o /dev/null -w "%{http_code}\n" -H "X-Forwarded-For: 203.0.113.7" http://127.0.0.1:6799/api/status
401
$ curl -s -o /dev/null -w "%{http_code}\n" -H "X-Forwarded-For: 203.0.113.7" \
       -H "Authorization: Bearer correct-password" http://127.0.0.1:6799/api/status
200
```

**CLI with no `PASEO_PASSWORD` in the environment, exemption on** — connects and exits 0:

```
$ env -u PASEO_PASSWORD PASEO_HOST=127.0.0.1:6799 paseo ls -a
$ echo $?
0
```

Daemon log for that session, no bearer presented:

```
{"connectionId":"conn_193416...","transport":"direct","peer":"loopback","remoteAddress":"127.0.0.1",
 "clientId":"cid_fb89...","appVersion":"0.7.0","msg":"Client connected via hello"}
```

**Then I flipped the flag to `false`, restarted, and re-ran the same commands** to confirm nothing was left open:

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:6799/api/status
401
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer correct-password" http://127.0.0.1:6799/api/status
200
$ env -u PASEO_PASSWORD PASEO_HOST=127.0.0.1:6799 paseo ls -a
Error: Cannot reach the daemon at 127.0.0.1:6799: Password required
$ env PASEO_PASSWORD=correct-password PASEO_HOST=127.0.0.1:6799 paseo ls -a   # connects, exit 0
```

**Tests — 12 added,** going through the real interfaces (`loadConfig` against a real `config.json`, and a real daemon over HTTP and WebSocket via `createTestPaseoDaemon`), no mocks:

- `auth.test.ts` (+7) — loopback recognition incl. `::1`, `0:0:0:0:0:0:0:1`, `::ffff:127.0.0.1`; Unix-socket peer; non-loopback peers rejected; each of the three forwarding headers rejected on a loopback socket; opt-in required; `/mcp/agents` opens to exempt callers.
- `config-auth.test.ts` (+4) — parsed from `config.json`; absent by default; survives `PASEO_PASSWORD` supplying the password; ignored when no password is set.
- `bootstrap-auth.test.ts` (+1) — against a running daemon: 200 without credentials, 401 with `X-Forwarded-For`, WebSocket connects with no subprotocol, stale bearer from loopback ignored rather than rejected.

Run over the changed files plus the adjacent auth, config, and connection suites:

```
$ npx vitest run \
    packages/server/src/server/auth.test.ts \
    packages/server/src/server/config-auth.test.ts \
    packages/server/src/server/bootstrap-auth.test.ts \
    packages/server/src/server/config.test.ts \
    packages/server/src/server/persisted-config.test.ts \
    packages/server/src/server/bootstrap.smoke.test.ts \
    packages/server/src/server/websocket-server.origin.test.ts \
    packages/server/src/server/daemon-config-store.test.ts

 Test Files  8 passed (8)
      Tests  141 passed (141)
```

```
$ npm run typecheck    → exit 0
$ npm run lint         → Found 0 warnings and 0 errors.
$ npm run format:check → All matched files use the correct format.
```

**Pre-existing failure, not from this PR.** `websocket-server.liveness.e2e.test.ts > a resumed stale socket is bounded and removed without disrupting its replacement` times out at 30s. I checked out the base versions of the five source files I touched and ran it again — it fails identically on unmodified `main`. Left alone.

**What I did not test.** A genuinely non-loopback peer end to end: that needs binding to `0.0.0.0`, and this machine's only interface is a public IP, so I wasn't going to expose a daemon with a known test password to the internet. It's covered by unit tests over `isLoopbackConnection`, and the middleware path itself is proven by the `X-Forwarded-For` → 401 above. Also not run on macOS or Windows — it's plain Node `net`/`http`, and the "no peer address" branch covers Windows named pipes as well as Unix sockets, but I haven't executed it there.

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | n/a | No app change |
| Android | n/a | No app change |
| Web | n/a | No app change |
| Desktop macOS | ❌ | Daemon-side change, not run there |
| Desktop Windows | ❌ | Daemon-side change, not run there |
| Desktop Linux | ✅ | All evidence above |

No UI change, so no screenshots or video.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -205,7 +205,7 @@ snapshot so a mixed edit can apply its live subset and still name the paths that
     agentProfiles: AgentProfile[],        // named agent launch bundles; omitted means none
     cors: { allowedOrigins: string[] },
     relay: { enabled: boolean, endpoint: string, publicEndpoint: string, useTls: boolean, publicUseTls: boolean }, // new homes materialize enabled: false
-    auth: { password: string }    // bcrypt hash, optional
+    auth: { password: string, allowLoopbackWithoutPassword: boolean }  // bcrypt hash, optional; exemption is config-file only
   },
   app: {
     baseUrl: string
--- a/packages/server/src/server/auth.test.ts
+++ b/packages/server/src/server/auth.test.ts
@@ -8,6 +8,8 @@ import {
   isAgentMcpRequestAuthorized,
   isBearerTokenValidAsync,
   isBearerTokenValid,
+  isLoopbackConnection,
+  isLoopbackPasswordExempt,
   shouldBypassBearerAuth,
 } from "./auth.js";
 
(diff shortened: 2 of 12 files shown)


boudra:
Closing this feature PR. Please start a [workflow discussion](https://github.com/getpaseo/paseo/discussions) describing what you’re trying to do, where Paseo gets in the way, and whether others share the need, as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #4243, opened 2026-09-02
docs(plugins): hold public entry docs until launch

boudra:
### Linked issue

Refs #4206

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [x] Docs

### Reasoning

The plugin client-entry implementation is on `main`, but its public documentation should appear only when the feature is available to users. This restores the public plugin pages to their pre-launch state and removes references and the drift test that depend on the withheld migration page.

### Goals

- Restore the public plugin index and reference to their pre-#4206 content.
- Withhold the public migration guide until launch.
- Avoid dead migration links and a test that reads the withheld file.

### Non-goals

- Revert any plugin runtime, compiler, SDK, scaffold, example, or internal documentation behavior.
- Rewrite or replace the launch documentation.

### QA

- Confirmed `public-docs/plugins/index.md` and `reference.md` exactly match their pre-#4206 versions.
- `npm run format` passed.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `git diff --check` passed.

Risk surface: documentation visibility only. The companion add-back PR restores these exact files, links, and drift test for launch.

### Checklist

- [x] One focused change
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence
- [x] Tests added or updated where it made sense

PR #4244, opened 2026-09-02
docs(plugins): publish client entry guides

boudra:
## Summary

Publishes the client-entry plugin documentation when the feature goes live.

This is intentionally stacked on #4243. Keep it draft until that revert lands, then retarget it to `main` and merge it at launch.

Refs #4206 and #4243.

## Goals

- Restore the public plugin overview and reference for the directory-only client/server entry layout.
- Restore the standalone mechanical migration guide.
- Restore links from the internal plugin docs and plugin-authoring skill.
- Restore the migration-table drift test.

## Non-goals

- No runtime, compiler, scaffold, or SDK changes.
- Do not merge before the feature is publicly available.

## Verification

- `npx vitest run packages/plugin/src/migration-doc.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`
- `git diff --check`
- Restored public pages and drift test match merge `53d824a320ccacecb08583be3cffc61ed2974576`.

## Risk

Documentation-only launch switch. The risk is publishing it before the corresponding release is available.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -417,6 +417,8 @@ stable rather than arrival-ordered. The app resolves that id
 against the installed catalog on every change; an id nothing contributes falls back to the default
 preference instead of painting the reserved slot's placeholder colors.
 
+Existing plugin authors should follow the standalone [runtime-entry migration guide](../public-docs/plugins/migration.md).
+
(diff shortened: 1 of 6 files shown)


boudra:
Superseded by #4245, which publishes the v0.8 material as a versioned preview alongside the current v0.7 docs.

PR #4246, opened 2026-09-02
Group projects in the sidebar under names you choose

totally-tim:
### Linked issue

No linked issue.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

A sidebar with twenty projects is one flat list, so the four that belong to the same client sit wherever they were added and you scan the whole column to find them. This change lets you put a name on a project and nests every project carrying that name under one collapsible header, so the sidebar reads the way you think about the work.

A group is a name on each project rather than a record of its own, stored per host next to `customName`. Nothing to create, nothing left over: a group exists while a project names it and disappears with its last member. Group order and project order are the client's own, like the existing drag order.

Cross-group drags land the new order before the host answers. Rolling that write back on a refusal turned out to be impossible to do consistently, because with two drags in flight the result depended on which host answered first. The client now replays instead: it keeps the order the first pending write found plus a log of every write since, and a refusal drops one write and nothing you did after it.

### Goals

- Group projects under a name, from a project's own menu, the display-preferences menu, or by dragging a project onto "New group".
- Drag a project onto a group header or between another group's rows to move it, and onto "Remove from group" to take it out.
- Drag a group header to reorder whole groups, with "Move up" and "Move down" in the group menu for native and keyboard users.
- Keep a refused write from disturbing anything else the user did while it was pending.
- Gate the feature on host support and say so when a host is too old to take it.

### Non-goals

- A group catalog. Groups have no existence apart from their members, so an empty group cannot be created or left behind.
- Native cross-list dragging. React Native's drag lists cannot drag between each other; native gets the menus.
- Sharing group names between hosts. The name is written per host, the same as `customName`.
- Reaching members on a host the host filter excludes. Those projects are never fetched, so a whole-group rename cannot name them.

### QA

- Created "Client Alpha" from a project's Group menu, then dragged a second project onto its header. The row landed first in the group and the header count followed.
- Dragged a grouped project onto "Remove from group" and it returned to the ungrouped rows.
- Dragged a project onto "New group" and the create sheet opened with that project ticked.
- Dragged the "Client Bravo" header above "Client Alpha". The sections swapped, and the order held across a reload.
- Confirmed a plain click on a header still collapses the group, and that the header's kebab menu still opens, since the header is also the drag handle.
- Used "Move down" in the group menu and watched the order change; "Move up" was disabled for the group already at the top.
- `cd packages/app && npx vitest run src/components/sidebar/ src/utils/sidebar-reorder.test.ts src/stores/sidebar-order-store.test.ts src/project-groups/ src/i18n/resources.test.ts`: 218 tests passed.
- `npm run test:e2e --workspace=@getpaseo/app -- e2e/browser/sidebar-project-group-drag.spec.ts e2e/browser/sidebar-project-groups.spec.ts e2e/browser/sidebar-reorder.spec.ts e2e/browser/sidebar-project-grouping.spec.ts`: 11 tests passed in 1.5 minutes.
- `npm run typecheck`: passed for all workspaces.
- `npm run lint`: 0 warnings and 0 errors.
- `npm run format`: passed.
- Every ordering fix was checked by reverting it and watching the test fail before restoring it.
- Video: attach recording here.

| Platform | Tested | Notes |
| --- | --- | --- |
| Web | Yes | Chromium end-to-end tests, plus the running app by hand |
| iOS | No | Shared app code; the menu path is untested |
| Android | No | Shared app code; the menu path is untested |
| Desktop macOS | No | Not tested |
| Desktop Windows | No | Not tested |
| Desktop Linux | No | Not tested |

### Known gaps

- A rename or ungroup issued in the moment between a host accepting a move and pushing the updated project leaves that one member under the old name. It is visible, and a second rename picks it up. Inferring those members from the client's own record of pending moves was worse: the record goes stale when another surface writes the group, and the rename then reaches a project that had already moved elsewhere.
- Setting a group from a project's own menu, or dragging one project into two groups in a row, can leave a later header drop one position lower inside that group for a moment. Both are written down in `project-group-move.ts`.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

https://github.com/user-attachments/assets/d4d1e7ed-f106-4170-adb4-a5d60359dde7


boudra:
Closing this feature PR. Please start a [workflow discussion](https://github.com/getpaseo/paseo/discussions) describing what you’re trying to do, where Paseo gets in the way, and whether others share the need, as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

Issue #4144, opened 2026-09-01
bug: a single oversized timeline item crashes the renderer, and route restore makes it an unrecoverable loop


huangguang1999:
## Summary

A single oversized timeline item crashes the desktop renderer process (`EXC_BREAKPOINT` / `SIGTRAP` on `CrRendererMain`, inside V8 GC). Because the app restores the last workspace route on launch with no crash-loop guard, every restart navigates straight back into the same view and crashes again within seconds — the app becomes permanently unusable until the offending agent is archived or `Local Storage` is cleared.

The main process, GPU/network helpers, and the daemon all stay alive, so the window just sits there frozen. Nothing in the UI indicates a crash.

## Environment

- Paseo desktop `0.7.0-beta.3` (Electron 41.2.0, Chrome 146.0.7680.179)
- macOS 15 (Darwin 25.6.0), Apple Silicon
- Daemon `0.7.0-beta.3`, `desktopManaged: true`
- The workspace lives on a remote host connected over direct TCP; the agent runs `claude-super-relay` with a 1M-context Claude model

## What triggers it

An agent whose provider produced runaway output. On the agent side:

- 7 assistant turns with `output_tokens: 64000` (the max_tokens cap)
- The content of those turns is a single short word repeated tens of thousands of times
- The upstream cause was provider-side (`API Error: 503 No available accounts` from the relay, plus prompt-cache misses forcing ~875k tokens to be re-uploaded per turn) — **that part is not a Paseo bug**, but Paseo has no defense against the resulting timeline

How big those individual items are, measured through the CLI:

```
$ paseo logs <agent-id> --tail 1 | wc -l
1278
$ paseo logs <agent-id> --tail 2 | wc -l
33229
$ paseo logs <agent-id> --tail 8 | wc -l
97080
```

So this is **a handful of items that are each enormous**, not a large number of items. `packages/app/src/agent-stream/` virtualizes the item *list* (`web-virtualization.ts`, `use-stream-history-window.ts`), but I could not find any cap on the rendered size of an individual item — and virtualization does not help here, because the oversized item is mounted.

## Steps to reproduce

1. Have an agent produce one or more assistant messages of a few hundred KB of text (7 × 64k output tokens in my case).
2. Open the workspace containing that agent in the desktop app.
3. The renderer process dies within seconds to a few minutes.
4. Relaunch the app — it restores the same route and crashes again.

## Evidence

Renderer crashed 5 times, all with an identical signature:

```
[...]

PR #4249, opened 2026-09-03
fix(cli): honor --json in the logs command

huangguang1999:
### Linked issue

No dedicated issue — I raised this as a separate nit at the end of #4144 and offered to file it on its own. Happy to open one if you'd rather track it that way.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

`paseo logs --json` parses without error but always prints the human transcript, so a script cannot consume it. `ls --json` and `inspect --json` both emit valid JSON, so the inconsistency is easy to hit — I hit it while scripting timeline inspection and assumed my parser was wrong before I checked the source.

Root cause is two-part:

- `logs` is wired with `addDaemonHostOption` (`cli.ts`, `commands/agent/index.ts`) rather than `addJsonAndDaemonHostOptions`, so `--json` is never advertised in `--help`. The global `--json` still parses, which is why it looks supported.
- `runLogsCommand` never consults the flag; before this change `commands/agent/logs.ts` contained no reference to JSON at all.

### Goals

- Make `--json` work on both `paseo logs` and `paseo agent logs`, and advertise it in help.
- Keep `--tail` and `--filter` meaning the same thing with and without `--json`.
- Give follow mode a stream a consumer can parse incrementally.

### Non-goals

- Change the text transcript output in any way.
- Add a schema/envelope around the payload. This emits the `AgentTimelineItem[]` the command already fetched; say the word if you'd prefer a wrapper like the other commands' result shapes and I'll rework it.

### Implementation notes

`--tail` parity is the one judgement call. The text path passes `tailCount` to `curateAgentActivity`, which does `items.slice(-maxItems)` on the raw timeline (`activity-curator.ts:144`), so the JSON path applies the same slice via `selectTimelineItemsForJson`. `--tail n` therefore selects the same entries in both modes.

Follow mode emits newline-delimited JSON — one item per line — and omits the `--- Following logs ---` banner, which would not parse.

### QA

**Before / after.** Released 0.7.2 CLI vs. this build:

```
$ /Applications/Paseo.app/Contents/Resources/bin/paseo logs --help | grep -E '^\s+--(json|tail|host)'
  --tail <n>       Show last n entries
  --host <host>    Daemon host target: host:port, tcp://host:port, or

$ npx paseo logs --help | grep -E '^\s+--(json|tail|host)'
  --tail <n>       Show last n entries
  --json           Output in JSON format
  --host <host>    Daemon host target: host:port, tcp://host:port, or
```

**New unit tests** (`packages/cli/src/commands/agent/logs.test.ts`) drive the real `runLogsCommand` with the daemon client and timeline fetch mocked, covering: JSON output parses and equals the timeline; `--tail` applied to the JSON payload; `--filter` applied to the JSON payload; and a control that the text transcript is unchanged when `--json` is absent (that one asserts the output does *not* parse as JSON and still contains the message text).

```
$ npx vitest run src/commands/agent/logs.test.ts src/cli-surface.test.ts
 Test Files  2 passed (2)
      Tests  18 passed (18)
```

**Real CLI end-to-end** (`packages/cli/tests/08-agent-logs.test.ts`, two cases added — help advertises `--json`, and `--json` is accepted without an option-parsing error):

```
$ npx tsx tests/08-agent-logs.test.ts
...
Test 10: logs --help advertises --json
✓ logs --help advertises --json

Test 11: logs --json flag is accepted
✓ logs --json flag is accepted

=== All logs tests passed ===
```

**Full CLI unit suite:**

```
$ npx vitest run src
 Test Files  1 failed | 37 passed (38)
      Tests  2 failed | 284 passed (286)
```

The 2 failures are `src/commands/plugin/scaffold.test.ts` and are pre-existing — I verified they fail identically with my changes stashed on a clean tree (they need `@getpaseo/plugin` typecheck artifacts). Everything else passes.

**Hooks:** the pre-commit hook ran clean on this commit — `format ✔`, `lint ✔`, repo-wide `typecheck ✔`. No `--no-verify`.

### What I did not test

- The JSON payload against a **real provider-backed agent timeline**. The mocked tests exercise the command function, and the E2E case covers option acceptance against the real binary, but I did not stand up a provider to produce genuine timeline items. If you want that wired through the E2E daemon helper, tell me and I'll add it.
- Follow-mode NDJSON against a live stream — the follow path is covered by reading, not by a test.
- macOS only (Apple Silicon, Node 22). Not tested on Linux or Windows.

### DIFF
--- a/packages/cli/src/cli-surface.test.ts
+++ b/packages/cli/src/cli-surface.test.ts
@@ -27,6 +27,16 @@ describe("canonical CLI surface", () => {
     expect(nestedReload?.helpInformation()).toContain("--json");
   });
 
+  it("offers JSON output on both top-level and agent logs", () => {
+    const cli = createCli();
+    const logs = cli.commands.find((command) => command.name() === "logs");
+    const agent = cli.commands.find((command) => command.name() === "agent");
+    const nestedLogs = agent?.commands.find((command) => command.name() === "logs");
+
+    expect(logs?.helpInformation()).toContain("--json");
+    expect(nestedLogs?.helpInformation()).toContain("--json");
+  });
+
   it("names explicit workspace creation without exposing older syntax", () => {
     const run = createCli().commands.find((command) => command.name() === "run");
     const help = run?.helpInformation();
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -92,7 +92,7 @@ export function createCli(): Command {
     withGlobalOptions(runAttachCommand),
   );
 
-  addDaemonHostOption(addLogsOptions(program.command("logs"))).action(
+  addJsonAndDaemonHostOptions(addLogsOptions(program.command("logs"))).action(
     withGlobalOptions(runLogsCommand),
   );
 
--- a/packages/cli/src/commands/agent/index.ts
+++ b/packages/cli/src/commands/agent/index.ts
@@ -41,7 +41,7 @@ export function createAgentCommand(): Command {
     withGlobalOptions(runAttachCommand),
   );
 
-  addDaemonHostOption(addLogsOptions(agent.command("logs"))).action(
+  addJsonAndDaemonHostOptions(addLogsOptions(agent.command("logs"))).action(
     withGlobalOptions(runLogsCommand),
   );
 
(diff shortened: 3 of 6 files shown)


boudra:
Closing for now: this needs a real CLI run against a daemon timeline proving parseable JSON and follow-mode NDJSON, including tail/filter behavior. Please reopen with that evidence.

boudra:

Issue #3509, opened 2026-08-18
bug: packaged desktop app can't start outside the repo — peer-only `@codemirror/lang-*` never make it into app.asar


cleiter:
### What's broken

Since 1917fdd13 (`fix(highlight): highlight Svelte components`, #3487), a packaged desktop build crashes on launch with `ERR_MODULE_NOT_FOUND` as soon as the bundle is run from anywhere other than inside the repo it was built in — e.g. after being installed to `/Applications`, or by anyone who downloads a release.

```
A JavaScript error occurred in the main process

Uncaught Exception:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@codemirror/lang-css' imported from
/Applications/Paseo.app/Contents/Resources/app.asar/node_modules/@replit/codemirror-lang-svelte/dist/index.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
    at packageResolve (node:internal/modules/esm/resolve:780:81)
    at moduleResolve (node:internal/modules/esm/resolve:871:18)
    at defaultResolve (node:internal/modules/esm/resolve:1003:11)
    ...
    at ModuleJobSync.syncLink (node:internal/modules/esm/module_job:160:33)
```

### Steps to reproduce

1. Check out `main` at or after 1917fdd13, `npm ci`.
2. `npm run build:desktop -- --publish never --dir` (any platform).
3. Copy the produced bundle **out of the repo** — e.g. `cp -R packages/desktop/release/mac-arm64/Paseo.app /Applications/`.
4. Launch it. It dies before showing a window with the `ERR_MODULE_NOT_FOUND` above.

Launching it in place under `packages/desktop/release/` works, which is why this is easy to miss locally as well as in CI.

### Manual bisect

Three packaged builds, each from a clean checkout with no local patches, each packaged with electron-builder for macOS arm64 and then installed to `/Applications` — i.e. outside the checkout — and launched:

| base commit | result |
| --- | --- |
| current `main` | crashes with the error above |
| `f3202ac41` (parent of the change) | starts normally |
| `1917fdd13` (#3487) | crashes with the error above |

## Root cause

[...]

PR #4267, opened 2026-09-03
fix(server): declare the plugin SDK's react peer so it reaches app.asar

### LINKED ISSUE #4266: bug: no plugin loads in a packaged desktop build — peer-only `react` never reaches app.asar
### What's broken

Since 53d824a32 (`feat(plugins): split client and server plugin entries`, #4206), **no plugin can load in a packaged desktop build**. Every plugin goes straight to `failed`:

```
$ paseo plugin ls
PLUGIN          STATUS   ENABLED  ERROR
guide-manager   failed   yes      Plugin guide-manager exited during initialization

$ paseo plugin logs guide-manager
[paseo] Loading plugin
Error: Cannot find module 'react'
Require stack:
- /…/Paseo.app/Contents/Resources/app.asar/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1463:15)
    …
[paseo] Plugin failed to load: Plugin guide-manager exited during initialization
```

Running the daemon from a source checkout works, because the monorepo's hoisted `node_modules/react` is resolvable there. Only the packaged app is broken, so this does not show up in dev.

This is the same failure class as #3509 — a peer-only package that never reaches `app.asar` — on a different import chain.

### Steps to reproduce

1. Check out `main` at or after 53d824a32, `npm ci`.
2. `npm run build:desktop`.
3. Launch `packages/desktop/release/mac-arm64/Paseo.app` (installing it elsewhere is not required — unlike #3509, this one fails in place too).
4. Configure any directory plugin and load it: `paseo plugin ls` → `failed`, `paseo plugin logs <id>` → `Cannot find module 'react'`.

Reproduced with two local directory plugins; the import chain is in the host, so it is plugin-independent.

### Root cause

`@getpaseo/plugin` declares `react` as a **peerDependency**, which is right for a UI SDK — the host supplies React. That assumption holds in the renderer. It does not hold in the **daemon**, which is where the plugin host process runs, and nothing there supplies React.

Before #4206 that did not matter, because the host imported only React-free values:

```ts
// packages/server/src/server/plugins/plugin-process.ts — before #4206
import { defineAttachmentSource, defineRpc, type PluginHandlerContext, type PluginRpcContract }
  from "@getpaseo/plugin/server";
```

#4206 reduced `packages/plugin/src/server.ts` to `export type { PluginHandlerContext }` and moved the host to the package root:

```ts
// after #4206
import { defineAttachmentSource, defineRpc, type PluginRpcContract } from "@getpaseo/plugin";
import type { PluginHandlerContext } from "@getpaseo/plugin/server";
```

The root entry re-exports `usePaseo`, `useRpc`, `useAgent`, `useWorkspace`, so requiring it pulls React in eagerly:

```
packages/server/dist/.../plugins/plugin-process.js
  → @getpaseo/plugin                       (root, external + resolved from disk by the host)
  → dist/client-state.js                   import … from "use-sync-external-store/shim/with-selector"
  → use-sync-external-store/cjs/…/with-selector.development.js   require("react")   ← not in app.asar
```

electron-builder's collector walks declared `dependencies`/`optionalDependencies` only, so:

- `use-sync-external-store` **is** packaged — it is a real dependency of `@getpaseo/plugin`
- `react` is **not** — the only edge to it is `@getpaseo/plugin`'s peer edge

Nothing in the desktop production graph declares `react` as a real dependency: `@getpaseo/app` does, but `@getpaseo/desktop` does not depend on it (the renderer ships as a prebuilt export via `extraResources: ../app/dist → app-dist`, so it needs no `node_modules/react`).

Confirmed in the built archives — `react` was never packaged, in this version or the previous one:

```
$ npx asar list .../Paseo-0.7.2/Paseo.app/Contents/Resources/app.asar | grep '^/node_modules/react/package.json'
(no output)

$ npx asar list .../Paseo-0.6.1/Paseo.app/Contents/Resources/app.asar | grep '^/node_modules/react/package.json'
(no output)
$ … | grep '^/node_modules/use-sync-external-store/package.json'
/node_modules/use-sync-external-store/package.json
```

So packaging did not change — the requirement did. #4206 introduced a runtime need for React in the daemon that the archive has never satisfied.

### Suggested fixes

Either alone fixes it.

**1. Declare the peer where the runtime graph needs it** — the #3509 idiom. `@getpaseo/server` is what imports `@getpaseo/plugin`, so it owns the peer:

```json
// packages/server/package.json
"react": "19.1.0",
```

Verified locally: `react` then lands in `app.asar`, and both plugins reach `running` in the packaged app. A PR with this change and a regression test follows.

**2. Keep the daemon off the React-laden entry** — removes the dependency instead of shipping it. The host only needs `defineRpc` and `defineAttachmentSource`, both React-free (`rpc.ts`, and a plain identity function). Restoring those two value exports on a host-facing entry would let `plugin-process.ts` avoid the root again. This contradicts the freshly published statement that `@getpaseo/plugin/server` "now exports only handler-side types", so it is the maintainers' call — hence the PR takes option 1.

### The blind spot

#3509 closed with a suggestion to smoke the packaged app from outside the checkout. That would not have caught this one: this failure happens in place too, but only once a plugin is loaded, and the smoke never loads a plugin. Loading a trivial plugin in the packaged smoke would turn this class of break into a red build.

### Where did this happen

Desktop (Electron)

### Paseo version

main (0.7.2, at eb753813a)

### OS version

macOS 26.3.1 (arm64)

### Agent provider

Not relevant

leeyc09:
Fixes #4266.

## The problem

Since #4206, no plugin loads in a packaged desktop build. The plugin host requires `@getpaseo/plugin` from disk inside the daemon; that entry re-exports `usePaseo`/`useRpc`/`useAgent`/`useWorkspace`, so requiring it pulls React in eagerly. `react` is a `peerDependency` of `@getpaseo/plugin` — right for a UI SDK, since the renderer supplies React — but nothing supplies it in the daemon, and electron-builder's collector walks declared `dependencies` only. `use-sync-external-store` is packaged because it is a real dependency of the SDK; `react` is not, because the only edge to it is the peer edge.

Dev is unaffected: a source checkout resolves `react` from the hoisted root `node_modules`. Only the archive is short.

## The change

Declare `react` on `@getpaseo/server`, the package whose runtime graph reaches it — the same shape as the fix for #3509 on `@getpaseo/highlight`.

I considered the alternative of keeping the daemon off the React-laden entry (the host only needs `defineRpc` and `defineAttachmentSource`, both React-free). That means putting value exports back on a host-facing entry, which contradicts the just-published statement that `@getpaseo/plugin/server` "now exports only handler-side types", so I left that call to you and took the packaging fix.

The test derives the requirement from the SDK source instead of a hand-kept list: every bare specifier `@getpaseo/plugin` imports for its value and declares only as a peer must be a real dependency of `@getpaseo/server`. It also asserts the SDK declares every bare specifier it imports.

## QA evidence

**The test fails without the fix and passes with it.**

```
$ npx vitest run packages/server/src/server/plugins/plugin-sdk-peers.test.ts   # react removed from packages/server
AssertionError: @getpaseo/server must declare react: @getpaseo/plugin imports it
for its value and only declares it as a peer: expected undefined to be defined
 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)

$ npx vitest run packages/server/src/server/plugins/plugin-sdk-peers.test.ts   # with the fix
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

**Before — packaged app on `main` (eb753813a), plugins in a local directory:**

```
$ paseo plugin ls
PLUGIN          STATUS   ENABLED  ERROR
guide-manager   failed   yes      Plugin guide-manager exited during initialization
tour-poster     failed   yes      Plugin tour-poster exited during initialization

$ paseo plugin logs guide-manager
[paseo] Loading plugin
Error: Cannot find module 'react'
Require stack:
- /…/Paseo.app/Contents/Resources/app.asar/node_modules/use-sync-external-store/cjs/use-sync-external-store-shim/with-selector.development.js
[paseo] Plugin failed to load: Plugin guide-manager exited during initialization

$ npx asar list …/app.asar | grep '^/node_modules/react/package.json'
(no output)
```

The 0.6.1 archive is missing `react` too — packaging did not change, the requirement did:

```
$ npx asar list …/Paseo-0.6.1/…/app.asar | grep '^/node_modules/react/package.json'
(no output)
$ npx asar list …/Paseo-0.6.1/…/app.asar | grep '^/node_modules/use-sync-external-store/package.json'
/node_modules/use-sync-external-store/package.json
```

**After — same tree with this commit, rebuilt with `npm run build --workspace=@getpaseo/desktop`:**

```
$ npx asar list …/app.asar | grep -E '^/node_modules/react/(package.json|index.js)$'
/node_modules/react/index.js
/node_modules/react/package.json

$ npm ls react --workspace=@getpaseo/server
paseo@0.7.2
└─┬ @getpaseo/server@0.7.2 -> ./packages/server
  ├─┬ @getpaseo/plugin@0.7.2 -> ./packages/plugin
  │ ├── react@19.1.0 deduped
  │ └─┬ use-sync-external-store@1.6.0
  │   └── react@19.1.0 deduped
  └── react@19.1.0

$ paseo plugin ls
PLUGIN          STATUS    ENABLED  ERROR
guide-manager   running   yes
tour-poster     running   yes

$ paseo plugin logs guide-manager
[paseo] Loading plugin
[paseo] Plugin ready
```

The panels both render and their RPCs answer in the relaunched packaged app.

`npm run lint`, `npm run format:check` and `npm run typecheck` all pass (they run on the pre-commit hook).

## Platforms

Tested on **macOS 26.3.1, arm64**, packaged via `electron-builder` from this checkout. I did not test Linux or Windows — the missing dependency is platform-independent (the collector behaves the same), but I cannot claim I ran it there.

## Note on the blind spot

#3509 closed with a suggestion to smoke the packaged app from outside the checkout. That would not have caught this one: this failure happens in place too, but only once a plugin is loaded, and `packages/desktop/e2e/packaged-app-smoke.js` never loads one. Loading a trivial plugin in that smoke would turn this whole class of break into a red build. I left it out of this PR to keep the change focused — happy to follow up if you want it.


boudra:
Thanks for the reproduction and packaging analysis. Closing in favor of https://github.com/getpaseo/paseo/pull/4347, which removes React from the plugin server import path. That is the direction we want to take for this failure.

PR #4270, opened 2026-09-03
feat(hub): let Hub send a message to an execution's live agent

as-cto:
## The workflow this unblocks

We self-host Hub and drive it from Linear agent sessions. A session panel is one conversation: the user writes, the agent answers, the user writes again in the same panel.

Hub can only create an agent, interrupt it, or archive it. There is no way to reach the agent it already started, so every message in that one panel starts a **new** agent and replays the thread as text. Everything the previous agent had built — the files it read, what it had already tried, why it answered as it did — is thrown away between two sentences of the same exchange. It also costs a full cold start per message: measured on our deployment, ~20 s per turn against ~6 s once the agent can simply be spoken to.

## The change

One message, `hub.execution.agent.prompt.request`, carrying `{executionId, prompt, activeTurnBehavior}`. It reuses `sendPromptToAgent`, so it inherits the dispatch semantics chat and schedules already use, and `activeTurnBehavior` decides what happens when a turn is in flight — `steer` folds the message into it, `interrupt` replaces it.

Two deliberate limits keep the new reach narrow:

- `unarchive: false` — a Hub prompt continues a conversation, it never revives an agent whose workspace the daemon already tore down;
- an execution with no live agent answers `delivered: false` rather than raising, because "start a fresh one" is the caller's ordinary fallback, not an error.

Prompts are ordered behind the same per-execution tail as control actions: a prompt racing an archive, or two prompts racing each other, would otherwise reach the agent in whichever order the event loop picked.

## QA

- Unit tests added for the controller (delivery, no live agent, blank prompt, daemon failure, post-cleanup silence) and for `DaemonExecutions.prompt` (delivered, archived agent, unknown execution, revoked authority). `vitest` green on the hub suites; typecheck, lint and format clean on the changed files.
- Running in production on three of our daemons (0.7.2 + this commit) since today, driven by a Hub build that uses it: three consecutive messages in one Linear session served by the same agent, which then restated the earlier requests from memory rather than from a replayed transcript.
- Platforms tested: Linux (Ubuntu 24.04, Node 24). Not tested: macOS, Windows.

## Compatibility

Additive. A Hub that sends it to a daemon predating the message gets `rpc_error` and falls back to creating an agent, which is today's behaviour.

Draft because the direction is worth confirming before I polish: happy to reshape the message, the naming, or the ordering guarantees. Maintainer edits enabled.


boudra:
Closing in favor of the implementation merged in [#4354](https://github.com/getpaseo/paseo/pull/4354), which lets Hub continue agents through the standard daemon API.

PR #4271, opened 2026-09-03
feat(worktree): let a caller reuse the worktree of a slug instead of suffixing beside it

as-cto:
## The workflow this unblocks

We self-host Hub and drive it from Linear. A trigger names its worktree after the **issue** it serves rather than after the run — `eng-42` — so that everything about that issue keeps iterating on the same branch: the first agent writes code, a later session reviews it, a third applies feedback.

`createWorktree` suffixes on collision, so each run lands in `eng-42`, `eng-42-1`, `eng-42-2` — a fresh copy of the base branch every time, blind to what the previous ones changed, and the work ends up split across several worktrees. Suffixing is right when the slug names a run; it is wrong when the slug names the work.

## The change

`reuseExisting`, opt-in and narrow:

- reads `git worktree list` rather than the filesystem, so a leftover directory git no longer tracks is not mistaken for a worktree and handed back as a checkout no branch points at;
- matches only the canonical path, so a slug suffixed by an earlier run is left alone;
- absent from every existing call site, which keeps today's behaviour unchanged.

Hub executions pass it through `toCreateAgentWorktree`, since their worktree name comes from the work, not from the execution.

## QA

- New test cutting a real git repository twice with the same slug and asserting one worktree, one branch. `vitest` green; lint and format clean on the changed files.
- Running in production on our daemons since today: a second session on the same issue lands in the existing worktree with the previous work in it, where it previously created a suffixed copy of `main`.
- Platforms tested: Linux (Ubuntu 24.04, Node 24). Not tested: macOS, Windows.

Draft because the direction is worth confirming before I polish. Maintainer edits enabled.


as-cto:
Note: this branch also carries `fix(server): accept an MCP protocol version newer than the bundled SDK`, which is unrelated to worktree reuse — I have opened it separately as #4387 so it can be reviewed on its own. It stays here only because our daemon build tracks this branch.

boudra:
Closing this feature PR. Please start a [workflow discussion](https://github.com/getpaseo/paseo/discussions) about what you’re trying to do, where Paseo gets in the way, and whether others share the need, as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #4283, opened 2026-09-03
Reject pending Codex app-server requests on dispose

### LINKED ISSUE #4282: Codex model rows stay "Loading" forever in the new workspace model picker
## Summary

Opening the model picker in a new workspace sometimes shows every Codex provider (the built-in one and each Codex profile) stuck at "Loading" indefinitely. Claude rows load normally. Nothing recovers the rows short of restarting the daemon, and the daemon log never shows a "Timed out refreshing Codex" entry.

## Cause

The per-workspace provider snapshot probes Codex by spawning `codex app-server` and awaiting `initialize`. The snapshot manager wraps the probe in a two-minute refresh deadline, but the deadline aborts cooperatively: it disposes the app-server client and waits for the probe to settle.

`CodexAppServerClient.dispose()` killed the child process without rejecting in-flight requests. Pending requests only rejected on *unexpected* termination, and that handler returns early once `disposed` is set. The request fallback timeout is 14 days. So when an app-server was slow or unresponsive, the `initialize` promise never settled, `fetchCatalog` never returned, the deadline never surfaced, and the snapshot entry stayed `loading` forever. The picker only force-refetches when the *selected* provider is loading, so the Codex rows never recover.

Each Codex profile spawns its own app-server, which is why every Codex row freezes together.

## Related

#4281 reduces how often a new workspace has to wait on a fresh app-server start. This issue is about the deadline not firing when it still does.

colonelpanic8:
## Problem

Opening the model picker in a new workspace could leave every Codex provider row at "Loading" forever, with no timeout in the daemon log. Details in #4282.

The provider refresh deadline aborts a Codex catalog probe by disposing the app-server client and waiting for the probe to settle. `CodexAppServerClient.dispose()` killed the child without rejecting in-flight requests, and the request fallback timeout is 14 days. An unanswered `initialize` therefore left `fetchCatalog` pending past the deadline, so the workspace snapshot entry never left `loading`.

## Fix

`dispose()` now rejects every pending request with "Codex app-server client is closed", using the same helper as the unexpected-termination path. The deadline surfaces as an error entry with the timeout message, which the picker renders with a retry.

#4281 addresses the other half (one app-server start queued per new workspace). The two changes do not overlap.

## Verification

- New transport test: a pending request rejects when the client is disposed.
- New client test: a deadline-aborted `fetchCatalog` with an unanswered `initialize` settles with the timeout error and kills the child. Both fail on `main` and pass here.
- `npx vitest run` on `codex/app-server-transport.test.ts`, `codex-app-server-agent.spawn-error.test.ts`, and `codex-app-server-agent.test.ts` (148 tests) pass.
- `npm run typecheck`, `npm run lint`, `npm run format:check` clean.

Fixes #4282

### DIFF
--- a/packages/server/src/server/agent/providers/codex-app-server-agent.spawn-error.test.ts
+++ b/packages/server/src/server/agent/providers/codex-app-server-agent.spawn-error.test.ts
@@ -1,7 +1,9 @@
-import { describe, expect, test } from "vitest";
+import { describe, expect, test, vi } from "vitest";
 
 import { CodexAppServerAgentClient } from "./codex-app-server-agent.js";
+import { runProviderRefreshWithDeadline } from "../provider-refresh-deadline.js";
 import { createTestLogger } from "../../../test-utils/test-logger.js";
+import { createCodexAppServerChildProcess } from "./codex/test-utils/fake-app-server.js";
 
 describe("CodexAppServerAgentClient spawn error handling", () => {
   const logger = createTestLogger();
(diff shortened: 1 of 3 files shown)


boudra:
@paseo-bot QA and deliver

PR #3832, opened 2026-08-25
fix(server): give the web UI connection hint an explicit port

kiawin:
### Linked issue

Closes #1842

Related implementations: #2618 and #3754.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

When a browser reaches the daemon-served web UI through a proxy on the scheme's default port, the `Host` header has no port. The daemon injects that value as the initial connection hint, but the app requires an explicit `host:port`. A new browser profile therefore rejects the hint and falls back to its own `localhost:6767` instead of connecting to the daemon.

Add the scheme default (`443` for HTTPS and `80` for HTTP) when the header omits a port. Keep explicit ports unchanged and handle bracketed IPv6 hosts. This stays on the producer side, where the daemon knows the request scheme, instead of weakening the shared client parser.

This is a narrower server-only alternative to #2618 and #3754.

### Goals

- Make first-run browser web bootstrap work behind a TLS-terminating proxy on port 443.
- Emit port 80 or 443 only when the `Host` header has no explicit port.
- Preserve explicit ports and bracketed IPv6 hosts.
- Cover the real daemon bootstrap path with a regression test.

### Non-goals

- Change trusted-proxy or `X-Forwarded-Host` handling.
- Add a client-side fallback for portless connection strings.
- Change the connection UI or native app behavior.

### QA

[Raw Linux evidence for the exact PR head](https://github.com/getpaseo/paseo/pull/3832#issuecomment-5406658240)

Manual:

- Browser web with the daemon running on Linux: tested the affected reverse-proxy flow and confirmed that it connects with this change.
- Not manually tested on macOS or Windows. Native apps are not affected by this server-rendered bootstrap hint.

Automated evidence from the fork PR:

- `bootstrap-web-ui.test.ts`: 3 passed.
- Negative control: reverting only `web-ui.ts` fails the new assertion (`expected 'paseo.example.test' to be 'paseo.example.test:443'`) while the two existing tests still pass.
- `web-ui.test.ts` and `config-web-ui.test.ts`: 27 passed.
- The pre-commit hook passed lint, format, and typecheck across all workspaces.

### Checklist

- [x] One focused change
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence
- [x] Tests added or updated where it made sense

PR #4288, opened 2026-09-04
fix(server): append default port to connection hint when Host header omits it

rophy:
## Summary

- When the daemon is behind a reverse proxy on a standard port (443/80), the `Host` header omits the port per HTTP spec
- `parseHostPort()` requires `host:port` format, so the connection hint was silently dropped and the SPA fell back to a localhost probe — showing the welcome/setup screen instead of auto-connecting
- Append `:443` (HTTPS) or `:80` (HTTP) when the `Host` header has no port suffix

## Reasoning

The web UI hint injection constructs a `listen` value from `req.headers.host`. RFC 7230 §5.4 says clients should omit `:443`/`:80` for default ports, and reverse proxies (nginx, Caddy, Traefik) follow this. `parseHostPort()` in `packages/protocol` requires the `host:port` format, so a bare hostname like `paseo.example.com` throws, the hint is silently skipped, and the SPA shows the welcome screen.

## Goals

- Standard-port reverse proxy deployments auto-connect without manual configuration
- No behavior change for explicit-port or localhost deployments

## Non-goals

- Changing `parseHostPort()` itself (it's used broadly; better to normalize at the source)
- Adding `X-Forwarded-Port` support (Express `trust proxy` + `req.protocol` already handles this)

## QA Evidence

- **Unit tests:** 2 new tests added covering port-omitted and explicit-port cases (38/38 pass)
- **Manual test:** Verified on a self-hosted instance behind a reverse proxy on port 443 — SPA auto-connects without showing the welcome screen
- **Regression:** All existing web-ui tests pass unchanged

## Checklist

- [x] Linked issue (standalone fix — no existing issue)
- [x] Tests added
- [x] Typecheck passes
- [x] Lint passes
- [x] Format passes


boudra:
Closing in favor of the same connection-hint fix in [#3832](https://github.com/getpaseo/paseo/pull/3832), which is still open.

PR #4310, opened 2026-09-04
fix(pi): settle autonomous turns Paseo never started

### LINKED ISSUE #4309: bug(pi): agent stuck in running when a pi-subagents background notification starts a turn
### What's broken

A Pi agent hosted by Paseo can get stuck showing **running** and never return to idle. The stuck turn is not one Paseo started — it is a turn Pi runs on its own when the `pi-subagents` extension delivers a background-completion notification with `triggerTurn: true` (`pi-subagents/src/runs/background/notify.ts`, which calls `pi.sendMessage({ customType: "subagent-notify" }, { triggerTurn: true })`).

Paseo emits `turn_started` for that provider-initiated run, which flips the agent lifecycle to `running`, but never emits `turn_completed`. There is no watchdog on the autonomous run, so the status stays `running` until the daemon restarts or an unrelated foreground turn happens to reset it.

Evidence from a stuck agent on Paseo 0.7.2:

| Check | Result |
| --- | --- |
| Pi session `.jsonl` | Last entry is an assistant message with `stopReason: stop`; 96 lines, no unclosed tool call. Pi finished. |
| Pi process (child of the Paseo daemon) | Alive but idle: no open socket, CPU flat, blocked on stdin. Not waiting on the model. |
| Paseo agent record | `lastStatus: "running"`, `requiresAttention: false`, `PendingPermissions: []`. Not a permission/approval prompt. |
| Daemon stream log | `thread_started` + `turn_started` at 15:17:24; **no `turn_completed` after 15:17:41**. `byLifecycle.running` went 2 → 3 and never moved. |
| Cross-check | Of 4 agents marked `running`, the three Claude/Codex ones refresh `updatedAt` continuously; only the Pi agent is frozen, and it is the only session containing a `custom_message` (the subagent notification). |

### Root cause

The asymmetry is in `packages/server/src/server/agent/providers/pi/agent.ts`, in `PiRpcAgentSession`.

- The **start** side has no gate. `handleSessionEvent` emits `thread_started` / `turn_started` for every `agent_start` / `turn_start`, with `turnId = activeTurnId ?? undefined`. `activeTurnId` is only set by Paseo's own `startTurn()`, so a provider-initiated run reports `turnId: undefined` but still emits `turn_started`.
- The **end** side is gated on `activeTurnId`. In `handleTurnBoundaryEvent`, `agent_end` (Pi always carries `willRetry`) stashes `pendingSettledMessages` under `if (this.activeTurnId)`, and the real completion on the `agent_settled` branch is also wrapped in `if (this.activeTurnId)`. When `activeTurnId` is `null`, no `turn_completed` is ever emitted.

On the daemon side, `onStreamTurnStarted` tracks the non-foreground run and sets `lifecycle = "running"`; the only branch that returns it to `idle` for an autonomous run lives in `onStreamTurnCompleted` (`agent-manager.ts`). No terminal event arrives, so it never runs.

### Steps to reproduce

1. Install the `pi-subagents` extension for a Pi provider Paseo manages.
2. Start a background subagent run and let it complete while Paseo holds the Pi session. The completion notification triggers a Pi-side turn.
3. Watch the Paseo agent status — it stays `running` after the turn's output has already streamed in.

### Fix

Treat a turn as open when Paseo owns it (`activeTurnId`) **or** Pi already published `turn_started` for it (`activeTurnStartedEmitted`), so provider-initiated runs reach a terminal event too. The legacy `willRetry === undefined` path is unchanged. A regression test in `agent.test.ts` reproduces the stuck state on main (turn settles only after the fix). PR to follow.

### Where did this happen

Daemon

### Paseo version

0.7.2

### OS version

macOS (daemon host)

### Agent provider

Pi

### Provider configuration

Default Pi provider with the `pi-subagents` extension installed under `~/.pi/agent/npm` (background subagent completion notifications, `triggerTurn` defaults to `true` and has no off switch).

JrDw0:
Closes #4309

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

A Pi agent stays showing `running` forever after the `pi-subagents` extension delivers a background-completion notification. That notification runs a turn inside Pi itself (`triggerTurn: true`), and Paseo never called `startTurn()` for it. The start side of `PiRpcAgentSession` emits `turn_started` for any `turn_start` the provider publishes, which sets the daemon's autonomous lifecycle to `running`. The end side — `handleTurnBoundaryEvent` — gated both the `agent_end` staging and the `agent_settled` completion on `activeTurnId`, which is `null` for a run Paseo didn't start. So `turn_completed` is never emitted and there is no watchdog on the autonomous run: the agent never returns to `idle` until the daemon restarts or an unrelated foreground turn resets it.

The fix keys turn completion off "did we already announce this turn" instead of only "did Paseo start it". Foreground turns are unaffected.

### Goals

- An autonomous Pi run reaches a terminal event so the agent lifecycle returns to `idle`.
- Foreground (Paseo-started) turn behavior is unchanged, including the retry path and the legacy `willRetry === undefined` floor.

### Non-goals

- No watchdog / timeout on autonomous runs. The missing terminal event is the bug; papering over it with a timer isn't.
- No change to the daemon-side `agent-manager.ts` autonomous tracking.

### QA

Behavior fix, no UI. Verified with the added regression test against the real `PiRpcAgentSession`, both directions.

**Before** — new test against `upstream/main` `agent.ts` (bug reproduced, stuck after `turn_started`):

```text
$ npx vitest run src/server/agent/providers/pi/agent.test.ts -t "settles a Pi-run autonomous turn"
 ❯ src/server/agent/providers/pi/agent.test.ts (79 tests | 1 failed | 78 skipped)
     × settles a Pi-run autonomous turn that Paseo never started
AssertionError: expected [ { type: 'turn_started', …(1) } ] to deeply equal [ Array(2) ]
```

**After** — with the fix, full file:

```text
$ npx vitest run src/server/agent/providers/pi/agent.test.ts
 Test Files  1 passed (1)
      Tests  79 passed (79)
```

The test drives the exact failure: `agent_start` → `turn_start` → `agent_end(willRetry: true)` → `agent_settled`, with no `startTurn()`, and asserts `turn_completed` arrives.

Field evidence from the stuck agent is in #4309.

**Platforms tested:** daemon, Linux and macOS hosts. **Not tested:** mobile/web UI (no UI change).

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:
The autonomous Pi settlement fix is covered by [#3849](https://github.com/getpaseo/paseo/pull/3849), which is merged. Closing this PR.

PR #4324, opened 2026-09-04
fix(terminal): preserve CJK wide characters in snapshots and capture

### LINKED ISSUE #2193: bug: terminal snapshots and capture insert spaces between CJK wide characters

heshunme:
### Linked issue

Closes #2193

### Type of change

- [x] Bug fix

### Reasoning

A PTY that emits `中文` was captured by `paseo terminal capture` as `中 文`,with a real U+0020 between the characters, and the same corruption showed up when the desktop app repaints a restored terminal snapshot on session switch. xterm stores a wide glyph as a width-2 content cell followed by a width-0 placeholder cell; state extraction serialized the placeholder's blank char as a real space, and every state→text/ANSI conversion then emitted it.

The fix records `cell.getWidth()` during state extraction and skips width-0 placeholder cells in every conversion: plain-text capture (CLI/MCP), snapshot ANSI rendering (client restore and daemon restart restore), the native row model, and terminal selection/copy. The new `TerminalCell.width` field is optional and gated on a `terminal_cell_width` client capability, so old strict-schema clients keep parsing and old daemons keep working.

### Goals

- `paseo terminal capture` (with and without `--ansi`) returns `中文` exactly
- Snapshot replay no longer inserts a column after wide glyphs, so restored terminals render CJK/Hangul rows without drifting
- Protocol stays backward compatible in both directions (capability-gated, COMPAT-tagged)
- Regression coverage on the real PTY path

### Non-goals

- Snapshots persisted before this fix carry no width data and still replay with the inserted spaces
- Mixed-version setups keep the old behavior: width is only sent to clients advertising `terminal_cell_width`, and old daemons never send it
- No string-level whitespace rewriting (would erase genuine spaces)

### QA

Reproduced the issue's scenario end-to-end with a real PTY (node-pty + `/bin/sh`, `echo A界B`), added as `wide characters` tests in `packages/server/src/terminal/terminal.posix.test.ts`:

- capture path: `captureTerminalLines` returns `A界B`, not `A界 B`
- restore path: snapshot → `renderTerminalSnapshotToAnsi` → written into a fresh `@xterm/headless` terminal reads back `A界B` with no phantom column

Regression tests verified to fail on the pre-fix code for the reported reason (implementation files checked out at HEAD~1, tests kept):

- terminal.posix.test.ts (real PTY): fails with
  `expected '$ echo A界 B' to match /A界B$/` and
  `expected [ '$ echo A界 B', 'A界 B', ... ] to include 'A界B'`
- terminal-snapshot.test.ts (protocol): replay renders `A界 B`
- app row-model/selection: 4 tests fail on placeholder handling

All green again with the fix restored.

Test runs (on this branch, after the fix):

- `packages/protocol`: terminal-snapshot.test.ts — 4 passed
- `packages/app`: terminal-row-model + terminal-selection — 24 passed
- `packages/client`: daemon-client.test.ts — 116 passed
- `packages/server`: terminal-session-controller + terminal.posix — 50 passed, 4 skipped
- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean

| iOS             | no     | native renderer + selection changes covered by unit tests only |
| Android         | no     | same |
| Web             | no     | snapshot rendering covered by protocol unit tests |
| Desktop macOS   | no     | issue reporter's platform; fixed path (xterm headless width) is platform-independent |
| Desktop Windows | no     | Hangul report on 0.3.1 shares the same root cause and fix path |
| Desktop Linux   | yes    | real-PTY end-to-end tests + full unit coverage above |

Extraction runs per snapshot request, not per frame; the added work is one numeric field read and one filter pass per cell, no allocation change on the render path.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


boudra:
Closing in favor of the open implementation in [#4101](https://github.com/getpaseo/paseo/pull/4101).

Issue #826, opened 2026-05-08
feat: support VSCode Remote SSH / Cursor Remote URI schemes in the Open button


alexkreidler:
## Summary

When the Paseo daemon runs on a remote machine (e.g. a devbox, cloud VM, or GPU server), the "Open" button should be able to open the workspace in **VSCode Remote SSH** or **Cursor Remote SSH** on the user's local machine — not just local editors or Finder.

## Motivation

Paseo already supports managing agents across multiple remote hosts via the relay + offer URL pairing. But when you want to jump into the code yourself, the "Open" button only offers local options (Cursor, Finder per #261). On a remote daemon, those options don't work — Finder can't browse a remote filesystem, and launching a local Cursor instance opens the wrong machine's path.

This is especially relevant for **devbox / cloud development** workflows where:
- Developers provision ephemeral or persistent remote VMs
- The Paseo daemon runs on the VM alongside the agent
- The developer wants to open the agent's workspace in their local IDE via SSH

## Proposed behavior

When a workspace is on a remote host, the "Open" button should offer options like:

- **VSCode Remote SSH** → opens `vscode://vscode-remote/ssh-remote+<host>/<workspace-path>`
- **Cursor Remote SSH** → opens `cursor://cursor-remote/ssh-remote+<host>/<workspace-path>`

The `<host>` could be:
1. The daemon's hostname/IP (from the connection config)
2. A user-configured SSH alias (e.g. matching an entry in `~/.ssh/config`)
3. Derived from the offer URL metadata

### Configuration

Since the SSH host string may differ from the relay/daemon address, users should be able to configure the SSH host per daemon, e.g.:

```json
{
  "sshHost": "my-devbox",
  "sshUser": "dev"
}
```

Or via a setting in the host config UI.

## Prior art

- #261 — requests more editors in the Open button (local only)
- #246 / #680 — Cursor as a provider (agent integration, not IDE opening)
- VSCode URI scheme docs: `vscode://vscode-remote/ssh-remote+{host}/{path}`

## Scope

1. Detect whether the workspace is on a remote host vs. local
2. For remote hosts, offer VSCode Remote SSH and Cursor Remote SSH as "Open" options
3. Allow configuring the SSH host/user per daemon connection
4. Optionally auto-detect if VSCode/Cursor are installed locally (same as #261 proposes for local editors)

josecanciani:
[...]

PR #2516, opened 2026-07-28
fix(app): hide open in file manager action for remote host workspaces

PTK030:
### Linked issue

Closes #2509

### Type of change

- [x] Bug fix
- [ ] New feature (with prior issue + design alignment)
- [ ] Refactor / code improvement
- [ ] Docs

### What does this PR do

Hides the "Open in file manager" sidebar context menu action for projects/workspaces that belong to a remote host daemon (`isLocalDaemon === false`). 

Previously, on a desktop client connected to a remote host (e.g. Windows client to Ubuntu host), selecting this action attempted to open remote POSIX paths like `/home/user/project` in the local file manager and failed with the generic `Couldn't open folder` toast error.

### How did you verify it

- Added unit tests in `packages/app/src/workspace/open-in-file-manager/menu-item.test.tsx` verifying that `OpenInFileManagerMenuItem` renders `null` when `serverId` belongs to a remote daemon (`isLocalDaemon === false`).
- Verified all 15 unit tests in `menu-item.test.tsx` pass.
- Verified `SidebarWorkspaceMenu` forwards `serverId` correctly.
- Ran `npm run typecheck`, `npx oxlint`, and `npx oxfmt` across modified files.

### Checklist

- [x] One focused change. Unrelated cleanups split out.
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` ran (Biome)
- [x] Tests added or updated where it made sense

PR #4330, opened 2026-09-04
fix(app): hide open in file manager for remote-host workspaces

### LINKED ISSUE #2509: bug: Open in file manager fails for a remote-host workspace
### What's broken

In Paseo Desktop, the sidebar **Open in file manager** action is offered for a project or workspace that belongs to a paired remote host. On a Windows Electron client connected to an Ubuntu host, selecting the action shows the generic toast:

> Couldn't open folder

The workspace path belongs to the remote host (for example, `/home/user/project/paseo`). The native Windows file manager cannot open that path.

I expected this local-only action to be hidden or disabled when the selected workspace is owned by a remote host, with an explanation if appropriate. The action should not be presented as available and then fail with a generic error.

### Steps to reproduce

1. Run a Paseo daemon on an Ubuntu host with a workspace at a POSIX path such as `/home/user/project`.
2. Connect Paseo Desktop on Windows to that host via relay/pairing.
3. In the sidebar, open the project or workspace actions menu.
4. Select **Open in file manager**.
5. Observe the `Couldn't open folder` toast.

### Where did this happen

Desktop (Electron)

### Paseo version

0.2.3 (built from current main at `89c2fac3c`, which includes #2491)

### OS version

Windows 11 client; Ubuntu 24.04 remote daemon/workspace host

### Agent provider

Not relevant

### Provider configuration

Not relevant

### Logs

No daemon error is expected: the failed operation is the Electron client's attempt to open a remote-host filesystem path. I did not capture the Windows Electron main log for this click.

### Screenshots or video

Not captured. The visible result is the `Couldn't open folder` toast after selecting the new sidebar action.

---

Related but distinct from #826: that issue proposes opening remote workspaces through Remote SSH URIs. This report is about the existing native file-manager action being offered for a remote-host path even though it can only operate on the desktop client's local filesystem.

KiraPC:
### Linked issue

Fixes #2509

### Type of change

- [x] Bug fix

### Reasoning

The sidebar's "Open in file manager" entry decided whether it could act by asking "am I running in Electron, and is the path non-empty". It never asked which daemon owns the workspace. Paseo Desktop can be paired with remote hosts, so the entry was offered for workspaces that live on another machine, and choosing it handed a remote path to the local file manager and failed with the generic "Couldn't open folder" toast. Someone on a Windows client connected to an Ubuntu host sees a working-looking action, clicks it, and gets an error with nothing to do about it.

The fix applies the rule the other desktop open targets already use: this action is for the local daemon's workspaces, so it is only offered for them.

### Goals

- Gate the entry on workspace ownership (`useIsLocalDaemon`) instead of on "is this Electron".
- The entry is absent for a remote-host workspace or project, in the row context menu and in the `...` menu.
- Nothing changes for a local-daemon workspace.
- Automated coverage that fails if the gate is removed again.

### Non-goals

- Opening remote workspaces. That is #826 (Remote SSH URIs), a separate feature.
- Any change to the desktop bridge or to `desktop-open-targets`.
- Rendering the entry disabled with an explanation, see below.

### What changed

`OpenInFileManagerMenuItem` now takes the `serverId` of the host that owns the path and derives `isLocalExecution` from `useIsLocalDaemon(serverId)`. The prop is required rather than optional (nullable, but you have to pass it) so a future call site has to answer the ownership question instead of quietly defaulting to offering the action.

Both call sites already had the value, so nothing new is threaded through the component tree. The sidebar workspace menus already carry `serverId` for the label submenu, and every `SidebarWorkspaceMenu` and `SidebarWorkspaceContextMenu` call site passes the workspace's own `serverId`. The project menu passes the local daemon id, because its `projectPath` comes from `resolveSidebarProjectLocalPath`, which already resolves either to the local daemon's checkout or to nothing at all; that makes the local daemon the host the path belongs to. `settingsTarget` would have been the wrong source there, since it is just `project.hosts[0]` and can be a remote host.

The entry is hidden, not disabled. `docs/menus.md` documents no pattern for disabled menu items, hiding is what the other desktop targets do when execution is not local, and it is what the reporter asked for.

One thing worth knowing for review: the project menu was already correct, by accident, because `resolveSidebarProjectLocalPath` returns an empty path for a project that only exists on a remote host. The bug was reachable only through the workspace menus, which pass `workspace.workspaceDirectory` unconditionally for every row regardless of host.

### QA

New test file `packages/app/src/workspace/open-in-file-manager/menu-item.test.tsx`, three cases: offered for a local-daemon workspace and actually calls `openTarget` with that path, hidden for a workspace owned by a paired remote host, hidden when the owner is unknown.

The test reproduces the bug rather than just describing it. Reverting only the gate in `menu-item.tsx` (`isLocalDaemon` back to `isElectron`) and leaving everything else in place makes the two "hidden" cases fail:

```
 ❯ src/workspace/open-in-file-manager/menu-item.test.tsx (3 tests | 2 failed)
   × hides the action for a workspace owned by a paired remote host
   × hides the action when the owning host is unknown

AssertionError: expected <button type="button" …(1)></button> to be null

- Expected:
null

+ Received:
<button
  data-testid="open-folder"
  type="button"
>
  sidebar.project.actions.openFolder
</button>
```

With the gate in place:

```
$ npx vitest run --project unit src/workspace/open-in-file-manager/menu-item.test.tsx

 Test Files  1 passed (1)
      Tests  3 passed (3)
```

`src/components/sidebar-workspace-list.test.tsx` also still passes (3 tests), since this PR touches that file.

Manual verification in the desktop app, against a workspace on a paired remote host and a local one:

- Remote-host workspace: the entry is gone from the row context menu and from the `...` menu.
- Remote-host project: same, gone from both menus.
- Local-daemon workspace: the entry is still there and still opens the file manager at the right directory.

No video. The visible effect is a menu entry not being offered, and the two states are described above.

| iOS             | n/a    | Entry is behind `getIsElectron()`, never shown |
| Android         | n/a    | Same |
| Web             | n/a    | Same |
| Desktop macOS   | yes    | Manual pass described above |
| Desktop Windows | no     | The gate is a `serverId` comparison in JS with no OS-specific path handling, so behaviour is the same. The original report is Windows client + Ubuntu host |
| Desktop Linux   | no     | Same |

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/src/components/sidebar-workspace-list.tsx
+++ b/packages/app/src/components/sidebar-workspace-list.tsx
@@ -541,6 +541,10 @@ function ProjectMenuItems({
 }) {
   const { t } = useTranslation();
   const toast = useToast();
+  // `projectPath` comes from `resolveSidebarProjectLocalPath`, so it is the local daemon's
+  // checkout of the project or nothing at all. `settingsTarget` is the first host, which can
+  // be a remote one, so it is not the owner of this path.
+  const localDaemonServerId = useLocalDaemonServerId();
   const handleOpenProjectSettings = useCallback(() => {
     if (!settingsTarget) return;
     router.navigate(buildProjectSettingsRoute(settingsTarget.serverId, settingsTarget.projectId));
@@ -581,6 +585,7 @@ function ProjectMenuItems({
       ) : null}
       <OpenInFileManagerMenuItem
         surface={surface}
+        serverId={localDaemonServerId}
         path={projectPath}
         testID={`sidebar-project-menu-open-folder-${projectViewKey}`}
       />
--- a/packages/app/src/components/sidebar/sidebar-workspace-menu.tsx
+++ b/packages/app/src/components/sidebar/sidebar-workspace-menu.tsx
@@ -214,6 +214,7 @@ function SidebarWorkspaceMenuItems({
       ) : null}
       <OpenInFileManagerMenuItem
         surface={surface}
+        serverId={serverId ?? null}
         path={openInFileManagerPath}
         testID={`sidebar-workspace-menu-open-folder-${workspaceKey}`}
       />
--- a/packages/app/src/workspace/open-in-file-manager/menu-item.test.tsx
+++ b/packages/app/src/workspace/open-in-file-manager/menu-item.test.tsx
@@ -1,0 +1,138 @@
+/**
+ * @vitest-environment jsdom
+ */
+import React from "react";
+import { act } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, expect, it, vi } from "vitest";
+
+const { LOCAL_SERVER_ID, REMOTE_SERVER_ID, fileManagerTarget, openTargetMock } = vi.hoisted(() => ({
+  LOCAL_SERVER_ID: "local-daemon",
+  REMOTE_SERVER_ID: "paired-remote-host",
+  fileManagerTarget: {
+    id: "file-manager",
+    label: "Finder",
+    kind: "file-manager" as const,
+    icon: { kind: "symbol" as const, name: "folder" as const },
+  },
+  openTargetMock: vi.fn(async () => {}),
+}));
+
+vi.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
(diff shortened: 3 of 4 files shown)


KiraPC:
Closing this: #2516 covers the same bug, came first, and handles two cases mine doesn't (local placement for a project on several daemons, and the stale target cache). I hadn't seen it when I opened this.

PR #4333, opened 2026-09-04
Collapse the fork onto main and stamp versions from the release tag

totally-tim:
## Why

The fork had two trunks. `main` already carried fork features, and `team-main` was `main` plus fork identity, so every feature needed two merges and the nightly drift check never ran because GitHub only schedules workflows from the default branch. The committed 1.0.0 version bump also conflicted with upstream's release commit on 13 files every one to three days.

## What changes

- `main` is the only trunk. This PR merges `team-main` in; `team-main` gets deleted after merge.
- Committed versions go back to upstream's. `scripts/fork/stamp-version.mjs` writes the tag version into every workspace at build time, and the release workflow runs it before each desktop build. `scripts/fork/build-desktop.sh` does the same locally and restores the files afterwards.
- The drift workflow targets `main`.
- `docs/fork.md` is rewritten as the ledger of fork changes plus the sync, version, and release procedure. The README gets a two-line fork notice.
- `.npmrc` enforces a two-day minimum release age on installs.

## Verification

- `node scripts/fork/stamp-version.mjs v1.0.1` stamps all 12 package.json files; `1.0.1-rc.1` is rejected by the shared tag parser; stamping `0.7.2` back leaves zero diff against `upstream/main`.
- Both edited workflows parse; the stamp step appears once per desktop build job (mac, linux, windows).
- `npm run typecheck` and `npm run lint` pass after `npm ci` and `npm run build:server`.
- Not covered: an actual tagged CI release. The first `v1.0.0` tag on `main` after merge is the runtime proof.


totally-tim:
Opened against the wrong repository by mistake, sorry for the noise. This belongs on the totally-tim fork.

PR #4362, opened 2026-09-05
Add terminal close confirmation with remember option

panrafal:
### Linked issue

None

### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Closing a terminal tab always asks for confirmation, every single time.

This change makes the confirmation optional, and lets you turn it off from the dialog itself with "Remember this choice", the same affordance Paseo already uses when it asks where to open a service link. The default is unchanged, so nobody loses the safety net without asking to.

### Goals

- New General setting, **Confirm closing terminals**, on by default — existing behaviour is untouched for everyone who doesn't change it.
- **Remember this choice** in the close dialog. Confirming with it checked flips the setting off, so the dialog stops appearing.
- Reversible from Settings → General at any time.
- Remember is honoured wherever the platform can express it: a real checkbox on desktop and VS Code, an extra button on iOS/Android.

### Non-goals

- **No process detection.** An earlier draft added a `terminal.inspect` RPC that read the pty's foreground process group so the dialog could be skipped only for idle shells, and name the process otherwise. It worked, but it cost a protocol message, a feature flag, worker plumbing and per-platform pty semantics for a preference toggle. Dropped deliberately in favour of a plain on/off.
- **Bulk close paths unchanged.** Close others / left / right / pane keep their own confirmation. Those close several things at once and are worth a prompt.
- **No remember on plain browser web.** `window.confirm` has no checkbox and no third button. The dialog behaves as it does today there; the setting is still reachable from Settings.
- **No new rendered modal.** Confirmations keep using the OS dialog, consistent with every other confirmation in the app.

### QA

8 new tests in `confirm-dialog.test.ts` cover the new dialog surface:

- reports the checkbox from the desktop dialog
- never remembers a cancelled desktop dialog
- falls back to the plain desktop dialog when the checkbox bridge is missing
- falls back to browser confirm with no way to remember
- maps the native Cancel / Close / "Close and don't ask again" buttons (3 cases)
- declines when the native alert is dismissed

4 new tests in `storage.test.ts` cover persistence: defaults to true, loads a persisted false, falls back to true on an invalid stored value, round-trips a save.

#### Manual QA

| Platform | Tested | Notes |
| iOS             | ✅     | Three-button `Alert` |
| Android         | No     |  |
| Web             | No     |  |
| Desktop macOS   |  ✅    | `askWithCheckbox` |
| Desktop Windows | No     | |
| Desktop Linux   | No     | |

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [ ] QA evidence — automated only, no screenshots or video

### DIFF
--- a/packages/app/src/hooks/use-settings/storage.test.ts
+++ b/packages/app/src/hooks/use-settings/storage.test.ts
@@ -955,3 +955,45 @@ describe("parseClampedFontSize", () => {
     expect(parseClampedFontSize("abc", { min: 11, max: 24 })).toBeNull();
   });
 });
+
+describe("confirmTerminalClose", () => {
+  it("defaults to true", async () => {
+    expect(DEFAULT_CLIENT_SETTINGS.confirmTerminalClose).toBe(true);
+    expect((await loadAppSettingsFromStorage(makeDeps())).confirmTerminalClose).toBe(true);
+  });
+
+  it("loads a persisted false back", async () => {
+    const deps = makeDeps({
+      storage: createInMemoryKeyValueStorage({
+        [APP_SETTINGS_KEY]: JSON.stringify({ confirmTerminalClose: false }),
+      }),
+    });
+
+    expect((await loadAppSettingsFromStorage(deps)).confirmTerminalClose).toBe(false);
+  });
+
+  it("falls back to true when the persisted value is invalid", async () => {
+    const deps = makeDeps({
+      storage: createInMemoryKeyValueStorage({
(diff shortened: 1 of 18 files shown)


boudra:
Closing this feature PR. Please start a [workflow discussion](https://github.com/getpaseo/paseo/discussions) about what you’re trying to do, where Paseo gets in the way, and whether others share the need, as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #2241, opened 2026-07-20
Generalize the question form model for typed answers

supermomonga:
## Summary

- replace the component-specific question form core with a reusable plain TypeScript form model
- separate stable question keys from display headers and option values from display labels
- return single-select and text answers as strings and multi-select answers as ordered string arrays
- preserve the existing question permission wire shape through a dedicated header/label adapter
- document the reusable question form contract and lifecycle expectations

## Why

The existing Question Form coupled display text to submitted values and kept its state management inside the renderer. That made it unsuitable as a foundation for other structured form consumers.

This change introduces a normalized internal contract while keeping the existing provider question flows unchanged. The reusable model exposes typed machine-keyed answers, while the existing permission path still submits header-keyed strings with display labels. The adapter also preserves legacy selection order, raw Other text, and optional empty-answer behavior.

No protocol schema or server provider changes are included.

## Validation

- `npx vitest run packages/app/src/components/question-form-model.test.ts packages/app/src/components/question-form-permission-adapter.test.ts --bail=1`
- `npm run typecheck --workspace packages/app`
- targeted `npm run lint -- ...` for the changed app files
- targeted `npm run format:files -- --check ...` for the changed app files and documentation
- `git diff --check`

The focused unit tests cover machine keys and option values, single/multi/text answer types, Other-value collisions, legacy selection ordering, optional empty answers, and prototype-named keys.

## Out of scope

- ACP SDK updates or elicitation/create lifecycle wiring
- client or runtime capability advertisement
- renderer capability wiring
- visual styling changes

PR #4372, opened 2026-09-05
Keep checked options and the Other answer together on multi-select questions

### LINKED ISSUE #4370: bug: multi-select question card treats Other and checked options as exclusive, and since #3517 silently drops the typed text
## What's broken

On a multi-select question card ("select all that apply"), Other and the checked options cannot coexist. Which one survives depends on the order you fill them in, and one of the two orders loses data without telling you.

**Order A, check first, then type.**
Typing anything into the Other field unchecks every option that was selected. The boxes visibly clear, so at least you can see it.

**Order B, type first, then check.**
Tapping an option deletes the Other text from the card's state, but the field keeps showing it. The screen now shows typed text plus checked boxes. Submit sends the labels alone. The typed text never reaches the agent.

I confirmed order B end to end on 0.7.2: typed a sentence into Other, checked four options, submitted. The tool result the agent received was the four labels joined with commas and nothing else.

Expected: on a multi-select question, checked options and a typed Other answer are both kept and both submitted, the way Claude Code's own UI does it. Single-select is fine as is, since Claude Code also replaces the option with the custom text there.

Order B used to be visible too. #3517 (merged 2026-08-18, in every 0.7.x release) switched the Other field from a controlled `value` to `initialValue`, so state no longer replays into the field. The card was not updated to call `replaceText` when it clears the text, so the field and the state disagree. Order A is unchanged since the card was introduced.

## Steps to reproduce

1. Run Paseo daemon 0.7.2 with Claude Code 2.1.258 on PATH.
2. Create a Claude agent and send a prompt that makes it call `AskUserQuestion` with `multiSelect: true` and a few options. Asking it to "use AskUserQuestion to ask me which of A, B, C, D I want, select all that apply" is enough.

Order A:

3. Check two options. The checkboxes fill.
4. Type one character into the Other field.
5. Both checkboxes clear. Only the typed text remains.

Order B:

6. Reload or get a fresh question. Type a sentence into Other first.
7. Check two options. The text stays visible and the boxes fill.
8. Submit.
9. Ask the agent what it received. The answer is the two labels joined with `", "`. The typed sentence is gone.

Same result in the bundled web UI (desktop and mobile browser) and in the Android app. It is client state, so the daemon never sees the dropped half.

## Where did this happen

Web (browser) and Android app. The card is shared, so I expect the same on iOS and Electron but did not test those.

## Paseo version

0.7.2, the latest on npm as of 2026-09-05 (`dist-tags.latest` is `0.7.2`, published `2026-09-02T00:10:54Z` per https://registry.npmjs.org/@getpaseo/server). The relevant code is unchanged on `main` at `78b2850` (2026-09-05).

## OS version

Daemon: Debian 12 (bookworm), x86_64, Docker, image built `FROM ghcr.io/getpaseo/paseo:0.7.2`.
Not relevant to the bug; it is client-side state.

## Agent provider

Claude Code. Also affects Codex and OpenCode, see below.

## Provider configuration

Claude Code 2.1.258, Anthropic first-party API, subscription login via `claude` OAuth.

## Logs

None that help. Both inputs are reconciled in React state before submit, so `daemon.log` only ever sees the final `answers` map: the Other text alone in order A, the labels alone in order B.

## Screenshots or video

Recording of steps 3 to 9 attached below.

https://github.com/user-attachments/assets/de3fd95a-ce8b-4e50-9444-592d4657b35f

---

## Where the selection gets dropped

`packages/app/src/components/question-form-card.tsx`, the Other text handler (https://github.com/getpaseo/paseo/blob/78b2850/packages/app/src/components/question-form-card.tsx#L363-L372):

```ts
const setOtherText = useCallback((qIndex: number, text: string) => {
  setOtherTexts((prev) => ({ ...prev, [qIndex]: text }));
  if (text.length > 0) {
    setSelections((prev) => {
      if (!prev[qIndex] || prev[qIndex].size === 0) return prev;
      return { ...prev, [qIndex]: new Set<number>() };
    });
  }
}, []);
```

Any non-empty text empties the selection set. There is no check on `multiSelect`. This is order A.

The option toggle does the mirror image (https://github.com/getpaseo/paseo/blob/78b2850/packages/app/src/components/question-form-card.tsx#L349-L354):

```ts
setOtherTexts((prev) => {
  if (!prev[qIndex]) return prev;
  const nextTexts = { ...prev };
  delete nextTexts[qIndex];
  return nextTexts;
});
```

This is order B. The deletion happens in state only. The field is rendered with `initialValue={value}` (https://github.com/getpaseo/paseo/blob/78b2850/packages/app/src/components/question-form-card.tsx#L308), and `EditingTextInput` reads `initialValue` once into a ref and strips the `value` prop from its type (`packages/app/src/components/ui/text-input/types.ts`, `text-input.web.tsx`, `text-input.native.tsx`). The only way to change what the field shows is the `replaceText` handle, and the card never takes a ref to call it. So the text stays on screen while the state that feeds submit has already forgotten it.

The answer builder in `question-form-card-core.ts` then does what the state tells it (https://github.com/getpaseo/paseo/blob/78b2850/packages/app/src/components/question-form-card-core.ts#L115-L129):

```ts
if (questionShowsTextInput(q)) {
  if (otherText && otherText.length > 0) {
    answers[q.header] = otherText;
    continue;
  }
  ...
}
if (selected && selected.size > 0) {
  const labels = Array.from(selected).map((idx) => q.options[idx].label);
  answers[q.header] = labels.join(", ");
}
```

Even if the state kept both, this would still send one or the other.

`git log -L` on the handlers shows the exclusivity has been this way since the card was introduced (`b80f4e0e9`, `061c08426`, February 2026). #1422 and #1643 touched the server flag and the checkbox rendering respectively and did not change it. #3517 turned order B from visible into silent.

## Reference behavior in Claude Code

Claude Code's own terminal UI renders Other as one more checkbox row in the multi-select list. Typing into it adds it to the selection, clearing the text removes only that row, and submit concatenates the checked labels with the typed text.

An Anthropic collaborator reproduced exactly this in https://github.com/anthropics/claude-code/issues/73136 on 2.1.233: a multi-select with Apple, Banana, Cherry, checked Apple plus Other, typed `durian`, and the tool result contained `"Which fruits do you like?"="Apple, durian"`. The 2.1.181 changelog entry reads "Fixed AskUserQuestion multi-select questions silently dropping a typed 'Other' free-text answer when submitting", so the combination is supported, not accidental.

The wire format needs no change. The Agent SDK docs (https://code.claude.com/docs/en/agent-sdk/user-input) say multi-select answers are the labels joined with `", "` or passed as an array, and that per-question free text goes in `answers[question]` as the text itself, not the word "Other". Paseo's Claude provider already passes the string through unchanged (`normalizeClaudeAskUserQuestionUpdatedInput` in `packages/server/src/server/agent/providers/claude/agent.ts`), so `"Apple, durian"` is what it would forward today if the card produced it.

## Other providers

The card is shared, so Codex (`request_user_input`, `multiSelect` + `isOther`) and OpenCode (`question.asked` with `multiple`, `allowOther` always on) have the same two symptoms. Both providers already split the comma-joined answer into an array on the server, so a combined `"App, Desktop, my custom text"` arrives in the shape their APIs expect. OMP and Pi force their `ask_user` bridge to single-select and are unaffected.

## Out of scope: comma splitting in Codex and OpenCode

The Codex and OpenCode providers split multi-select answers on every comma with no quoting (https://github.com/getpaseo/paseo/blob/78b2850/packages/server/src/server/agent/providers/codex-app-server-agent.ts#L1240-L1246, https://github.com/getpaseo/paseo/blob/78b2850/packages/server/src/server/agent/providers/opencode-agent.ts#L4846-L4849). A custom answer containing a comma is broken into several answers. That is already true today when Other text is sent alone, so it is not introduced by this fix, but a combined answer makes it easier to hit. Claude Code's client JSON-quotes any segment containing `", "` or `"` before joining, and the server side could adopt the same scheme. It should be a separate patch. Noting it here so nobody testing this fix with Codex mistakes a split custom answer for a regression.

## Related

- #1421 / #1422 added the Other input for Claude questions (server flag only, no UI change).
- #1642 / #1643 made multi-select checkboxes visible in the card.
- #3517 made the Other field own its text. Correct on its own; the card was not updated to match.
- #2241 (open) generalizes the question form model. Its `setTextAnswer` resets the selection whenever text is non-empty, and its test "uses typed other text instead of selected option values" asserts that, so it carries order A into the new model. If it lands first, the fix and the tests need to move there.

## Proposed fix

App only, no protocol or daemon change:

1. In the Other text handler, only clear the selection when the question is single-select.
2. In the option toggle, only clear the Other text when the question is single-select, and when it does clear it, call `replaceText("")` on the field so the screen matches the state. Anything that drops text from state without touching the field recreates order B.
3. In `buildQuestionFormAnswers`, for multi-select, emit the selected labels followed by the trimmed Other text, joined with `", "`. Single-select keeps the replace behavior.
4. Tests in `question-form-card-core.test.ts` for both modes and both orders.

I have a fix ready and will open a PR referencing this issue.

ThePharmer:
### Linked issue

Closes #4370

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

On a multi-select question card, the Other field and the checked options cannot coexist. Check first and then type, and the boxes clear. Type first and then check, and the card silently drops the typed text on submit: since #3517 the field owns its text and never replays state, so the screen keeps showing a sentence the state has already forgotten. The agent receives the labels alone. Claude Code's own AskUserQuestion UI keeps both and submits `"Apple, durian"`, and the Agent SDK docs describe exactly that shape, so a Paseo user answering from their phone gets a worse answer than they would at the terminal.

This PR makes multi-select keep both inputs and submit the checked labels followed by the trimmed Other text, joined with `", "`. Single-select keeps the replace behavior, which also matches Claude Code, but when it clears the Other text it now calls `replaceText("")` on the field so the screen and the state agree.

### Goals

- Multi-select: checked options survive typing into Other, and Other text survives checking options, in either order.
- Multi-select: the submitted answer is the checked labels followed by the Other text, comma-joined, so Codex and OpenCode split it into the array their APIs expect and Claude receives the same string its own UI would send.
- Single-select: picking an option after typing clears the Other field on screen, not just in state.
- No protocol or daemon change. `answers` stays `Record<string, string>`.

### Non-goals

- Comma quoting for Codex and OpenCode. Both providers split multi-select answers on every comma with no quoting, so a custom answer containing a comma is broken into pieces. That predates this change and is called out as a separate patch in #4370.
- Any change to the question form model in #2241. If that lands first, the same two rules and these tests need to move into `setTextAnswer` and `toggleOption` there.
- Ordering of the checked labels. They are emitted in selection order, as before.

### QA

**Reproduced before the fix** (Paseo 0.7.2, Claude Code 2.1.258, web UI and Android app): on a multi-select AskUserQuestion, typed a sentence into Other, checked four options, submitted. The agent received the four labels joined with commas and nothing else. Checking first and typing afterwards visibly cleared the boxes. Recording is attached to #4370.

**Tests.** `packages/app/src/components/question-form-card-core.test.ts` gains two cases for the answer builder (multi-select combines, single-select replaces). New `packages/app/src/components/question-form-card.browser.test.tsx` mounts the real card in headless Chromium through the Vitest browser project, with the real web `EditingTextInput` and no mocks, and drives both orders through clicks and input events, asserting the answers the card submits and what the field shows. On `main` without the source change, four of the ten tests fail:

```
core:    × keeps checked options and appends the other answer for multi-select
           expected { Fruits: 'durian' } to deeply equal { Fruits: 'Apple, Cherry, durian' }
browser: × keeps checked options when the other answer is typed afterwards (multi-select)
         × keeps the typed other answer when options are checked afterwards (multi-select)
           expected { Fruits: 'Apple, Banana' } to deeply equal { Fruits: 'Apple, Banana, durian' }
         × clears the typed other answer on screen when an option is picked afterwards (single-select)
           expected 'OpenCode' to be ''
```

With the change, all ten pass. `oxlint` and `oxfmt` are clean on the four touched files. Typecheck, lint, and format results for the whole repo are from CI.

Platforms tested for the bug: web UI in a browser and the Android app. The fix is in the shared card and has no platform branches. Not re-tested on device after the fix; the browser tests cover the same interaction sequence against the real web input.

### Checklist

- [x] One focused change
- [x] `npm run typecheck` passes (CI)
- [x] `npm run lint` passes (CI)
- [x] `npm run format` passes (CI)
- [x] QA evidence

### DIFF
--- a/packages/app/src/components/question-form-card-core.test.ts
+++ b/packages/app/src/components/question-form-card-core.test.ts
@@ -54,6 +54,54 @@ describe("question form card core", () => {
     });
   });
 
+  test("keeps checked options and appends the other answer for multi-select", () => {
+    const questions = parseQuestionFormQuestions({
+      questions: [
+        {
+          question: "Which fruits do you like?",
+          header: "Fruits",
+          options: [{ label: "Apple" }, { label: "Banana" }, { label: "Cherry" }],
+          multiSelect: true,
+          allowOther: true,
+        },
+      ],
+    });
+
+    if (!questions) throw new Error("questions did not parse");
+    expect(buildQuestionFormAnswers(questions, { 0: new Set([0, 2]) }, { 0: " durian " })).toEqual({
+      Fruits: "Apple, Cherry, durian",
+    });
+    expect(buildQuestionFormAnswers(questions, { 0: new Set([0, 2]) }, {})).toEqual({
+      Fruits: "Apple, Cherry",
+    });
+    expect(buildQuestionFormAnswers(questions, { 0: new Set() }, { 0: "durian" })).toEqual({
+      Fruits: "durian",
+    });
+    expect(buildQuestionFormAnswers(questions, {}, { 0: "durian" })).toEqual({ Fruits: "durian" });
+  });
+
+  test("replaces the selected option with the other answer for single-select", () => {
+    const questions = parseQuestionFormQuestions({
+      questions: [
+        {
+          question: "Which provider?",
+          header: "Provider",
+          options: [{ label: "Claude Code" }, { label: "Codex" }],
+          multiSelect: false,
+          allowOther: true,
+        },
+      ],
+    });
+
+    if (!questions) throw new Error("questions did not parse");
+    expect(buildQuestionFormAnswers(questions, { 0: new Set([1]) }, { 0: "OpenCode" })).toEqual({
+      Provider: "OpenCode",
+    });
+    expect(buildQuestionFormAnswers(questions, { 0: new Set([1]) }, {})).toEqual({
+      Provider: "Codex",
+    });
+  });
+
   test("shows text input for explicit other questions", () => {
     const questions = parseQuestionFormQuestions({
       questions: [
--- a/packages/app/src/components/question-form-card-core.ts
+++ b/packages/app/src/components/question-form-card-core.ts
@@ -111,10 +111,13 @@ export function buildQuestionFormAnswers(
     const q = questions[i];
     const selected = selections[i];
     const otherText = otherTexts[i]?.trim();
+    const labels = selected ? Array.from(selected).map((idx) => q.options[idx].label) : [];
 
     if (questionShowsTextInput(q)) {
       if (otherText && otherText.length > 0) {
-        answers[q.header] = otherText;
+        // Multi-select keeps the checked options and appends the custom answer, the way
+        // Claude Code's own AskUserQuestion UI does. Single-select replaces the option.
+        answers[q.header] = q.multiSelect ? [...labels, otherText].join(", ") : otherText;
         continue;
       }
       if (q.allowEmpty && q.options.length === 0) {
@@ -123,8 +126,7 @@ export function buildQuestionFormAnswers(
       }
     }
 
-    if (selected && selected.size > 0) {
-      const labels = Array.from(selected).map((idx) => q.options[idx].label);
+    if (labels.length > 0) {
       answers[q.header] = labels.join(", ");
     }
   }
--- a/packages/app/src/components/question-form-card.browser.test.tsx
+++ b/packages/app/src/components/question-form-card.browser.test.tsx
@@ -1,0 +1,154 @@
+import React, { act } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import type { AgentPermissionResponse } from "@getpaseo/protocol/agent-types";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { PendingPermission } from "@/types/shared";
+import { QuestionFormCard } from "./question-form-card";
+
+// App sources compile against the classic JSX runtime, which expects React on the global.
+beforeEach(() => vi.stubGlobal("React", React));
+
+/**
+ * A real browser with the real web `EditingTextInput`, because the bug under test lives in the
+ * gap between that input and React state: the input owns its text and never replays state, so a
+ * card that drops the Other text from state alone keeps showing it while submit ignores it.
+ */
+
+interface Mounted {
+  root: Root;
+  container: HTMLDivElement;
+}
+
+const mounted: Mounted[] = [];
+
+afterEach(() => {
+  for (const entry of mounted.splice(0)) {
+    act(() => entry.root.unmount());
+    entry.container.remove();
+  }
+});
+
+function buildPermission(question: Record<string, unknown>): PendingPermission {
+  return {
+    key: "perm-1",
+    agentId: "agent-1",
+    request: {
+      id: "perm-1",
+      provider: "claude",
+      name: "AskUserQuestion",
+      kind: "question",
+      input: { questions: [question] },
+    },
+  };
+}
+
+function mountCard(question: Record<string, unknown>) {
+  const onRespond = vi.fn<(response: AgentPermissionResponse) => void>();
+  const container = document.createElement("div");
+  document.body.appendChild(container);
+  const root = createRoot(container);
+  act(() =>
+    root.render(
+      <QuestionFormCard
+        permission={buildPermission(question)}
+        onRespond={onRespond}
+        isResponding={false}
+      />,
+    ),
+  );
+  mounted.push({ root, container });
+
+  const option = (label: string): HTMLElement => {
+    const element = container.querySelector(`[aria-label="${label}"]`);
+    if (!(element instanceof HTMLElement)) throw new Error(`option ${label} did not render`);
+    return element;
+  };
+  const otherInput = (): HTMLInputElement => {
+    const element = container.querySelector("input");
+    if (!(element instanceof HTMLInputElement)) throw new Error("other input did not render");
+    return element;
+  };
+  const check = (label: string) => act(() => option(label).click());
+  const type = (text: string) => {
+    const input = otherInput();
+    const valueSetter = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, "value")?.set;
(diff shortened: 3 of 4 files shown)


ThePharmer:
Two fixes pushed in response to the review:

- The branch was cut from my fork's main, which carries a fork-only release workflow, so `.github/workflows/native-headers-release.yml` leaked into this PR. Rebased onto upstream main; the PR is now the single fix commit and touches only the four question-form files. The two workflow comments do not apply to this PR.
- The component test was a mocked JSDOM test, which docs/testing.md rules out. Replaced with `question-form-card.browser.test.tsx` in the Vitest browser project: real Chromium, the real web `EditingTextInput`, no mocks. Three of its four cases fail on main without the source change, including the silent-drop case (`Apple, Banana` received instead of `Apple, Banana, durian`).

ThePharmer:
I was able to conduct a live device test with a build on my fork.

I installed the fork build of upstream v0.7.2 with this PR's commit (1f33b1a02) cherry-picked on top, on Android: https://github.com/ThePharmer/paseo/releases/tag/v0.7.2-native-headers

Sequence: multi-select AskUserQuestion with Apple, Banana, Cherry as options. Typed into Other first, then checked all three, submitted. The agent received:

`Cherry, Banana, Apple, also durian, which is not on this list and now survives`

Three labels plus the typed text in one string. Before the fix the same sequence delivered only the labels.

The build also carries the native-headers cherry-pick from #4290 and the fork's Firebase push setup; nothing in it touches the question card besides this PR.

Issue #3009, opened 2026-08-08
bug: Codex app-server receives mcpServers.paseo but browser/Paseo MCP tools never appear on the model tool surface (0.2.5 / Codex 0.146.1)


zengyuzhi:
## Investigation cases: missing injected tools and protocol rejection

These cases coordinate the agent MCP investigation; they do not establish one root cause. All runtime observations are attributed reporter evidence, not independently reproduced during this review. The original report and its evidence are preserved verbatim below.

| Case | Environment, source and evidence | Independent completion criterion |
| --- | --- | --- |
| A — original fresh stock Codex missing tools | @zengyuzhi, original #3009 below: macOS arm64, Paseo 0.2.5, stock Codex 0.146.1/OpenAI path, fresh session. Metadata contains URL/bearer and hasCodexConfig is true; direct initialize and tools/list succeed (60 tools, 22 browser tools), and browser_list_tabs returns live tabs, while the model exposes no Paseo/browser tools. Grok works on the same daemon; Claude also reported working. | Trace real fresh-session app-server MCP startup and tool discovery; verify the model can call injected browser/workspace tools while the direct endpoint control still succeeds. Static codex mcp list and injected metadata alone do not prove runtime tool discovery. |
| B — explicit rejection | @wilgon456 in [#3599](https://github.com/getpaseo/paseo/issues/3599): macOS 26.5.2 arm64, Desktop/daemon 0.5.0-beta.2, stock Codex 0.148.0; three fresh agents log Unsupported protocol version: 2026-07-28. The later Linux matrix below locates its observed rejection after initialize; that stage was not independently captured for every beta.2 start. | Capture initialize negotiation and subsequent request headers with real Codex, then verify tools/list and an injected tool call complete without rejection. |
| C — missing tools without logged 400 | @wilgon456: #3599 beta.3 isolated direct-loopback repro on Codex 0.148.0 returns LIST_WORKSPACES_UNAVAILABLE despite hasCodexConfig:true and no explicit protocol error. [0.6.1 retest](https://github.com/getpaseo/paseo/issues/3599#issuecomment-5432888569), macOS arm64/Codex 0.149.1: two fresh agents return MCP_UNAVAILABLE and complete normally; neither Unsupported protocol version nor Agent MCP transport error occurs in those log windows. | Reproduce fresh-session missing tools with full startup/tool-list tracing even when no 400 is logged; verify an actual list_workspaces call and normal turn completion. Do not infer that fixing B fixes C. |
[...]

PR #4387, opened 2026-09-06
fix(server): accept an MCP protocol version newer than the bundled SDK

as-cto:
## What breaks today

A client that ships ahead of the SDK sends an `MCP-Protocol-Version` we do not know, and the Streamable HTTP transport answers `400 Unsupported protocol version` on **every** request.

For an agent this is fatal in a way nobody sees. It keeps running, but every hub tool is gone — so it can neither reply nor call `finish_execution`. The run ends `succeeded` with nothing said, and the surface shows "finished this workflow without posting a reply".

Measured on 2026-09-06 with **Claude Code 2.1.259**, which sends `2026-07-28`:

```
Agent MCP transport error
  Bad Request: Unsupported protocol version: 2026-07-28
  (supported: 2025-11-25, 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)
```

Three Linear issues ran to completion and answered nothing. Upgrading the SDK does not help: the newest published `@modelcontextprotocol/sdk` (1.30.0) still tops out at `2025-11-25`, so any daemon is one client release away from this.

## The change

One guard before `transport.handleRequest`: a protocol date **newer** than ours is downgraded to ours. That is what version negotiation is for — MCP is backward compatible, and a client that asked for more can be served less.

Left untouched, so the SDK keeps refusing what it genuinely cannot speak:
- a version the SDK already supports,
- a date older than ours,
- anything that is not a date, or a missing header.

Four tests cover those cases.

### DIFF
--- a/packages/server/src/server/bootstrap.ts
+++ b/packages/server/src/server/bootstrap.ts
@@ -517,14 +517,27 @@ function createBootstrapManagedProcessRegistry(
  * ours, or one that is not a date at all, is left untouched: the SDK must keep rejecting what it
  * genuinely cannot speak.
  */
-export function clampAgentMcpProtocolVersion(headers: Record<string, unknown>): void {
-  const raw = headers["mcp-protocol-version"];
+export function clampAgentMcpProtocolVersion(request: {
+  headers: Record<string, unknown>;
+  rawHeaders?: string[];
+}): void {
+  const raw = request.headers["mcp-protocol-version"];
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the open implementation in [#4375](https://github.com/getpaseo/paseo/pull/4375). The unresolved MCP cases remain tracked in [#3009](https://github.com/getpaseo/paseo/issues/3009).

PR #3776, opened 2026-08-24
Terminalize abandoned compactions

Ctrl-Creeper:
## Summary
- Terminalize unmatched compaction markers at the shared `AgentManager` boundary when the active turn ends.
- Carry a backward-compatible optional compaction error to the app, so usage-limit failures stop the spinner and render the real error.
- Add a turn-scoped app reducer fallback for clients connected to older daemons.

## Root cause
Providers may emit a compaction start followed by a terminal turn event without a compaction completion. The marker remained `loading` indefinitely because neither the canonical timeline nor the app reducer closed it.

## Validation
- `npx vitest run packages/server/src/server/agent/agent-manager.test.ts --bail=1`
- `npx vitest run packages/app/src/types/stream.test.ts --bail=1`
- `npx vitest run packages/protocol/src/messages.wire-compat.test.ts --bail=1`
- `npx vitest run packages/app/src/components/message-compaction-label.test.ts --bail=1`
- `npm run format`
- `npm run typecheck`
- `npm run lint`

PR #4391, opened 2026-09-06
fix(claude): open a single compaction marker per compaction

### LINKED ISSUE #4390: bug: one Claude compaction leaves several stuck "Compacting..." rows
### What's broken

A single Claude compaction leaves several `Compacting...` rows spinning in the timeline forever. One compaction produced one `Context manually compacted` row plus **three** stuck `Compacting...` rows below it.

Expected: one compaction produces one timeline row, which reaches a terminal state.

### Raw evidence

Claude Code emits `{type:"sdk_status", status:"compacting"}` once when compaction starts, and then **re-emits it every 30 seconds until the compaction finishes**. From the shipped Claude Code binary (`~/.local/share/claude/versions/2.1.263`, `strings`):

```
var Lvo=30000;
function g7(e){let t=setInterval(Fvo,Lvo,e);return()=>clearInterval(t)}
function Fvo(e){Bnr(),e?.({type:"sdk_status",status:"compacting"})}
```

and the call site that arms it for the duration of the compaction:

```
if(N)t.onCompactEvent?.({type:"compact_progress",event:{type:"compact_start"}}),
     t.onCompactEvent?.({type:"sdk_status",status:"compacting"});
let F=null,U=!1,V=N?g7(t.onCompactEvent):void 0;try{ ... }finally{ V?.() }
```

Identical in 2.1.261 (`var Ivo=30000; function d7(...)`), which is the version that produced the screenshot.

`packages/server/src/server/agent/providers/claude/agent.ts` appends a new timeline item for **each** of those messages:

```ts
if (status === "compacting") {
  this.compacting = true;
  events.push({
    type: "timeline",
    item: { type: "compaction", status: "loading" },
    provider: "claude",
  });
}
```

`packages/app/src/types/stream.ts` (`reduceTimelineCompaction`) resolves only the first one when the completed item arrives:

```ts
const loadingIdx = state.findIndex((s) => s.kind === "compaction" && s.status === "loading");
```

The compaction from the screenshot, from that agent's own transcript (`~/.claude/projects/<project>/<session>.jsonl`, `compact_boundary` record):

```json
{"trigger": "manual", "preTokens": 800233, "durationMs": 102942}
```

102.9 seconds = 1 initial status + heartbeats at t=30s, 60s, 90s = 4 loading items. One was resolved, three were left spinning. Three earlier compactions in the same session ran 120s, 130s and 140s, so each of those stranded four.

### Steps to reproduce

1. Run a Claude agent until its context is large enough that compacting takes longer than 30 seconds (a few hundred thousand tokens; the one above was 800k).
2. Send `/compact`.
3. Watch the timeline: one row resolves to `Context manually compacted`, and one extra `Compacting...` row per 30 seconds of compaction time stays spinning permanently.

Compactions shorter than 30 seconds never fire a heartbeat and look correct, which is why this appears intermittent.

### Where did this happen

Desktop (Electron)

### Paseo version

0.7.2

### OS version

macOS 26.6.2

### Agent provider

Claude Code

### Provider configuration

Claude Code 2.1.261 and 2.1.263, Anthropic subscription (OAuth), model `claude-opus-5` with a 1M context window.

### Screenshots or video

One compaction, one resolved row and three that never resolve:

three stuck Compacting rows

tomgrin10:
### Linked issue

Closes #4390

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Claude Code doesn't send `status: "compacting"` once. It sends it when compaction starts and then repeats it every 30 seconds until the compaction finishes, as a heartbeat. The Claude provider treated every one of those as a new compaction and pushed a new `loading` timeline item, while the app only ever resolves one of them when the boundary arrives. So a compaction that takes 100 seconds leaves three `Compacting...` rows spinning in the user's timeline forever, and the longer the session, the longer the compaction and the more stuck rows it leaves — worst exactly when the user is already deep in a long piece of work.

This makes the heartbeat a heartbeat: the first `compacting` status opens the marker, the repeats are ignored, and the boundary closes it.

### Goals

- One compaction produces exactly one timeline marker, whatever its duration.
- A second compaction in the same session still gets its own marker.
- A compaction abandoned mid-turn does not silence the marker for the next one.

### Non-goals

- Terminalizing markers that never receive a boundary. That is a different root cause — a compaction that starts and never completes — and is tracked separately in #3776. This fix does not depend on it.
- Any app-side change. With one marker emitted per compaction the existing reducer resolves it correctly.

### QA

Full raw QA transcript, every command with its output: [qa-transcript.md](https://github.com/tomgrin10/paseo/blob/qa-evidence/qa-transcript.md)

**Before** — 112.9s compaction, 1 resolved row + 3 stuck:

before

**After** — 98.4s compaction, all three heartbeats fired, exactly 1 row:

after

**Reproduced before, confirmed fixed after — same corpus, same viewport, both runs long enough for all three heartbeats to fire.**

|  | BEFORE (unfixed) | AFTER (fixed) |
|---|---|---|
| `preTokens` | 304,441 | 312,666 |
| `durationMs` | 112,921 | 98,371 |
| heartbeats fired (t=30/60/90s) | 3 | 3 |
| compaction markers opened | **4** | **1** |
| rows once complete | 1 resolved + **3 stuck** `Compacting...` | 1 resolved, **0 stuck** |

BEFORE, observed live in the web app:

```
t = 12s    1 x "Compacting..."
t = 30s    2 x "Compacting..."     <- heartbeat #1
t = 43s    2 x "Compacting..."
t = 1m01s  3 x "Compacting..."     <- heartbeat #2
t = 1m32s  4 x "Compacting..."     <- heartbeat #3
done       1 x "Context manually compacted" + 3 x "Compacting..." STUCK
```

AFTER, same scenario:

```
t = 7s     1 x "Compacting..."
t = 38s    1 x "Compacting..."     <- heartbeat #1 fired, no new marker
t = 1m05s  1 x "Compacting..."     <- heartbeat #2 fired, no new marker
t = 1m35s  1 x "Compacting..."     <- heartbeat #3 fired, no new marker
done       1 x "Context manually compacted", zero leftover spinners
```

`compact_boundary` records from the agents' own Claude transcripts:

```json
BEFORE  { "trigger": "manual", "preTokens": 304441, "durationMs": 112921, "postTokens": 19592 }
AFTER   { "trigger": "manual", "preTokens": 312666, "durationMs": 98371,  "postTokens": 9446 }
```

#### Automated tests

New regression test `repeated compacting statuses open a single compaction marker`. Confirmed it fails on the unfixed code for the reported reason — the second `compacting` status opens a second marker:

```
$ npx vitest run packages/server/src/server/agent/providers/claude/agent.test.ts \
    -t "repeated compacting statuses"        # fix reverted

 FAIL  ... > repeated compacting statuses open a single compaction marker
AssertionError: expected 1 to be +0 // Object.is equality
- Expected  0
+ Received  1

 ❯ agent.test.ts:2886
    2884|       expect(markersOpenedBy(createCompactingStatus())).toBe(1);
    2885|       // Claude Code repeats the status every 30 seconds until the compaction ends.
    2886|       expect(markersOpenedBy(createCompactingStatus())).toBe(0);
```

With the fix:

```
$ npx vitest run packages/server/src/server/agent/providers/claude/agent.test.ts
 Test Files  1 passed (1)
      Tests  76 passed (76)

$ npx oxfmt --check packages/server/src/server/agent/providers/claude/agent.ts \
                    packages/server/src/server/agent/providers/claude/agent.test.ts

$ npm run lint

$ npm run typecheck:server
exit 0
```

The test also asserts that a `compact_boundary` re-arms the marker, so a second compaction in the same session still gets one.

#### One caveat worth recording

If Claude Code has a precomputed compaction cached, `/compact` takes a different branch that returns before the heartbeat interval is armed. Those compactions never showed the bug, which is why it looks intermittent. Both runs above took the heartbeat branch.

#### Platforms

| iOS             | ❌     | Not tested |
| Android         | ❌     | Not tested |
| Web             | ✅     | Before and after captured here, dev app against a dev daemon |
| Desktop macOS   | ⚠️     | Where the bug was originally observed; the fix was not re-verified on Electron |
| Desktop Windows | ❌     | Not tested |
| Desktop Linux   | ❌     | Not tested |

The change is daemon-side and provider-side only, so the surface behaves the same on every client; the timeline items it emits are what differ.

### DIFF
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -24,6 +24,12 @@ interface TestClaudeSession {
   close(): Promise<void>;
 }
 
+function isLoadingCompactionEvent(event: AgentStreamEvent): boolean {
+  return (
+    event.type === "timeline" && event.item.type === "compaction" && event.item.status === "loading"
+  );
+}
+
 function isPermissionResolvedEvent(
   event: AgentStreamEvent,
 ): event is Extract<AgentStreamEvent, { type: "permission_resolved" }> {
@@ -1981,6 +1987,15 @@ describe("ClaudeAgentSession context window usage", () => {
     };
   }
 
+  function createCompactingStatus(): Record<string, unknown> {
+    return {
+      type: "system",
+      subtype: "status",
+      status: "compacting",
+      session_id: "session-1",
+    };
+  }
+
   test("emits turn_started before the submitted user message", async () => {
     const session = await createSessionForTurns([[]]);
     const events: AgentStreamEvent[] = [];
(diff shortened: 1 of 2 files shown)


tomgrin10:
@boudra sorry for the ping, this is a very small fix for an issue proven to affect multiple people, would like you to take a look

stevecastaneda:
Seeing this a lot here as well.  Bump.

CleanShot 2026-09-09 at 8 55 07 AM@2x

PR #4405, opened 2026-09-06
Let hosts declare a default color so devices agree

colonelpanic8:
## Summary

Host badge color used to be purely per-device and defaulted to no color, so the same host looked different on every client. Color now resolves in three steps:

1. The device's own choice on `HostProfile.appearance.color`, if any.
2. The color the daemon declares for itself (`daemon.appearance.color` in its config, advertised in `server_info`).
3. A color derived from the server id, the same way project icons get theirs.

Steps 2 and 3 are identical on every client, so hosts line up by default. Changing the host-level setting moves every device that has not picked its own color.

## Changes

- **protocol**: `IDENTITY_COLOR_NAMES` lives here now so the daemon and clients share one list. `server_info` gains an optional `appearance.color` (a string on the wire so a future color does not break older clients) and a `hostAppearance` feature flag. The mutable daemon config gains `appearance.color`; the patch schema accepts `null` to clear it. Both tagged `COMPAT(hostAppearance)`.
- **server**: `daemon.appearance.color` in the persisted config, reloadable and patchable through `set_daemon_config`. `server_info` is re-sent to every connection when the color changes.
- **app**: `HostProfile.declaredColor` caches what the daemon advertises, refreshed on every `server_info`. The host Appearance section shows the resolved default behind the existing "Default" option, and, against a daemon that advertises the feature, a new "Default color" row that edits the host's own setting. The badge no longer has an uncolored state.
- **docs**: glossary and design entries updated for the new resolution order.

## Testing

- Focused suites: protocol wire-compat, server `daemon-config-store` and `persisted-config`, app `appearance`, `host-connection`, `host-runtime`, i18n resources, and `browser-tools-config`.
- Not run locally: the browser e2e host-appearance spec (heavy two-daemon fixture); CI covers it.


boudra:

PR #4425, opened 2026-09-07
Add generic ACP active-turn steering

IEatCodeDaily:
## Summary

- add opt-in active-turn steering for generic ACP providers through `params.activeTurnSteerCommand`
- forward accepted steering as a concurrent ACP prompt without canceling the foreground turn
- preserve interrupt-and-replace fallback for providers that do not opt in, stale turns, and structured prompts
- document the custom-provider setting

## Root cause

Paseo's provider-neutral manager supports `steerActiveTurn`, but the generic ACP adapter did not implement it because standard ACP has no steering method. Hermes exposes in-place steering through `/steer`, so Paseo unnecessarily canceled and replaced Hermes turns.

## Configuration

```json
"params": { "activeTurnSteerCommand": "/steer" }
```

## Verification

- focused ACP suites: 105 passed
- clean `npm run build:server`
- full workspace typecheck through pre-commit: passed
- lint and format checks: passed
- isolated daemon on port `16767`: a Hermes turn running `sleep 20` accepted a mid-turn correction, preserved the tool call, emitted the correction once, and finished with `STEERED DONE`

Production Paseo on port `6767` was not modified or restarted.

### DIFF
--- a/docs/custom-providers.md
+++ b/docs/custom-providers.md
@@ -503,6 +503,17 @@ Paseo tools such as subagent creation come from the shared internal tool catalog
 }
 ```
 
+ACP does not define native active-turn steering. Agents that expose steering through a slash
+command can opt in without changing the fallback for other ACP providers:
+
+```json
+"params": { "activeTurnSteerCommand": "/steer" }
+```
+
+Paseo sends that command as a concurrent ACP prompt while preserving the foreground turn. Use this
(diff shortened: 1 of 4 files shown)


boudra:
Closing in favor of the Hermes steering implementation in [#4154](https://github.com/getpaseo/paseo/pull/4154), which remains open.

PR #4434, opened 2026-09-07
fix(server): decouple native Paseo tools from MCP injection

thomasvan:
After #4277 landed per-provider control of the Paseo tool catalog, disabling MCP injection also disables the native catalog, even where the provider policy is enabled:

```ts
// bootstrap.ts — startup and both live field-change handlers
agentManager.setPaseoToolsEnabled(mcpEnabled && mcp.injectIntoAgents !== false);
```

**Why it matters.** Deployments that serve MCP caller-scoped — each seat gets its own caller-scoped server, so daemon-wide injection is off — still need the native catalog for coordinator seats (and, via the per-provider `paseoTools` policy, none for worker seats). Gating the catalog on `injectIntoAgents` removes the whole channel for exactly those deployments; the per-provider policy from #4277 is a second gate on top of a switch that is already off.

**Fix.** Gate native availability on `mcp.enabled` at startup and on both `onFieldChange` handlers; injection keeps controlling only the injected MCP server URL (`setMcpBaseUrl`). The per-provider `paseoTools` policy remains the seat-level gate. Adds `native-tools-gate.ts` owning the decision, with its truth table (`native-tools-gate.test.ts`).

Default behaviour is unchanged where MCP is enabled and injection is on; the change is only for injection-off deployments, whose native tools previously disappeared as a side effect of a flag about something else. Server typecheck and the new tests pass.

### DIFF
--- a/packages/server/src/server/bootstrap.ts
+++ b/packages/server/src/server/bootstrap.ts
@@ -148,6 +148,7 @@ import {
 import { CheckoutDiffManager } from "./checkout-diff-manager.js";
 import { ScheduleService } from "./schedule/service.js";
 import { DaemonConfigStore, type MutableDaemonConfig } from "./daemon-config-store.js";
+import { isNativePaseoToolsEnabled } from "./native-tools-gate.js";
 import { createOrchestrationSkills } from "./orchestration-skills/index.js";
 import { resolveConfigFromPersisted, type CliConfigOverrides } from "./config.js";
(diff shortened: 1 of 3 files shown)


boudra:
Closing for now: this needs a daemon-level regression and real-provider evidence that native tools remain usable with MCP injection off at startup and after live config changes, while disabled provider policies still hide them. Please reopen with that evidence.

boudra:

PR #4441, opened 2026-09-07
feat(app): persist dismissed provider subagents across app restarts

SezginKahraman:
### Linked issue

Discussion: #4440 (this is a feature request, not a bug — no issue to close). Opening as a **draft** per the discussion until there's agreement this is the right shape; see CONTRIBUTING.md guidance on unsolicited PRs.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

"Archive finished" hides completed/failed/canceled provider-owned (Task-tool) subagent rows only in an in-memory `hiddenFromTrack` Set (`packages/app/src/subagents/provider-store.ts`), documented as intentional in `docs/agent-lifecycle.md`. Every reload or app restart wipes it, since `refreshProviderSubagents` refetches the full descriptor list from the daemon on mount and the daemon has no concept of a "dismissed" provider subagent to filter back out. On a long-running parent that fans out a lot of subagents, this means re-dismissing the same finished rows over and over across reconnects. See #4440 for the full framing.

This PR persists just the `hiddenFromTrack` set (not the much larger, non-serializable-safe `descriptors`/`timelines` maps) to the device's local storage, using the same `persist` + `createValidatedPersistStorage` + `AsyncStorage` pattern already used by `sidebar-collapsed-sections-store` and `workspace-service-route-preferences`. The store construction moves into an exported `createProviderSubagentStore(storage)` factory (matching `workspace-service-routes/store.ts`) so tests can inject an in-memory storage and exercise a real write → restart → read round trip.

### Goals

- A dismissal made via "Archive finished" for a provider-owned row survives a reload/app restart on the same device.
- No change to managed Paseo subagent archiving (already daemon-side and correct).
- No change to `descriptors`/`timelines` persistence — those stay in-memory only, refetched/rebuilt from the daemon as before.

### Non-goals

- Not adding a daemon-side "archived" concept for provider subagents — this stays a client-local presentation dismissal, exactly as documented today, just a durable one. It does not propagate to other clients, and does not survive the provider reporting the same subagent id through a different daemon.
- Not pruning `hiddenFromTrack` against the live provider list — an existing test (`keeps hidden state when a child temporarily disappears from the provider list`) already asserts a dismissal must survive a subagent temporarily vanishing from a `replaceList` response, so the set isn't safe to prune on absence; it only grows. Flagged explicitly in the doc update. A size-bounded eviction policy would need a design decision I didn't want to bundle into this PR.

### QA

Commands run:
```
npm run build:client && npm run build:server   # workspace declarations must be current, see CLAUDE.md
npx vitest run packages/app/src/subagents/provider-store.test.ts packages/app/src/subagents/select.test.ts packages/app/src/subagents/hidden-track-persistence.test.ts
# 28 passed, 0 unhandled rejections
npm run typecheck   # clean across all workspaces
npm run lint -- packages/app/src/subagents/provider-store.ts packages/app/src/subagents/provider-store.test.ts packages/app/src/subagents/hidden-track-persistence.ts packages/app/src/subagents/hidden-track-persistence.test.ts
# 0 warnings/errors
```

New test `survives an app restart instead of resetting Archive finished dismissals` (`provider-store.test.ts`) creates two separate store instances sharing one in-memory storage backend — the second simulates the app restarting — and asserts the dismissal from the first instance is visible in the second after `persist.rehydrate()`. Confirmed it fails (`createProviderSubagentStore is not a function`) against the pre-change code and passes after.

One thing I had to work around: `@react-native-async-storage/async-storage`'s web build reads `window.localStorage` unconditionally, which crashed as an unhandled rejection when the *existing* `provider-store.test.ts`/`select.test.ts` tests (which mutate the real exported singleton, not an injected instance) ran in this package's Node-environment unit test project. `resolveHiddenTrackStorage` swaps in a no-op storage specifically for that gap (web-targeted evaluation with no `window` — native and real-browser evaluation are unaffected); covered by `hidden-track-persistence.test.ts`.

I did not test this manually against a running app (web only, per my role in the linked discussion) — the QA here is the automated round-trip test above, which exercises the real persistence mechanism rather than a mock.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -184,7 +184,9 @@ Claude Code announces subagent lifecycle on the SDK stream (`task_started` / `ta
 
 Archived Paseo subagents disappear from the track, by design. To remove one from the track without closing its tab, use the **archive button** on the row — it opens a confirm dialog and archives the subagent on confirm. Provider-owned rows have no individual Paseo lifecycle controls.
 
-The **Archive finished** row at the foot of the panel covers every finished row. It archives idle or errored managed Paseo subagents one at a time, and hides completed, failed, or canceled provider-owned rows in the current app session. Native sessions and timelines are untouched. Running and initializing children remain in the track. If a hidden provider child starts running again, the app brings it back to the track.
+The **Archive finished** row at the foot of the panel covers every finished row. It archives idle or errored managed Paseo subagents one at a time, and hides completed, failed, or canceled provider-owned rows. Native sessions and timelines are untouched. Running and initializing children remain in the track. If a hidden provider child starts running again, the app brings it back to the track.
+
+Hiding a provider-owned row is client-local, not a daemon lifecycle action: it has no `archivedAt`, does not propagate to other clients, and is keyed by `serverId`/`parentAgentId`/`subagentId` in `useProviderSubagentStore`'s `hiddenFromTrack` set (`packages/app/src/subagents/provider-store.ts`). That set persists to the device's local storage so a dismissal survives an app restart, but it stays scoped to the device that dismissed it — another client, or the same client against another daemon, still shows the row.
 
 To keep the agent alive but remove it from the parent's track, use **detach**. The daemon clears the relationship lifecycle labels, emits the normal agent update, and every client reclassifies the agent from subagent to root/sibling from that updated snapshot.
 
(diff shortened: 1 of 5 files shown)


boudra:
Closing this feature PR. Please describe the workflow and shared need in [Discussions](https://github.com/getpaseo/paseo/discussions), as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #4445, opened 2026-09-07
Add a Usage history page: daily token and cost usage from provider transcripts

colonelpanic8:
## What this adds

A **Usage history** page under each host's settings: a daily chart of tokens or API-equivalent cost, per provider and per model, over the last 7, 30, or 90 days.

It is deliberately **separate from the existing "Usage" page**. That one reports live subscription quota windows from `services/quota-fetcher`. This one answers a different question ("what have I actually burned, and on what?") from a different source, and the two share no code and no RPC.

It gathers across **every connected host**, not just the one you are looking at, and keeps working when a host is down, too old, or reading the same transcripts as another host.

Ported from the equivalent feature in T3 Code, including its visual treatment, minus the parts that don't apply here (see *Scope* below).

## How it works

The daemon scans the provider CLIs' **own on-disk session transcripts** rather than Paseo's agent records, so usage is complete even for turns that never went through Paseo. This is the approach `ccusage` takes.

| Provider | Directory | Usage carrier |
| --- | --- | --- |
| Claude Code | `$CLAUDE_CONFIG_DIR/projects/**/*.jsonl` (default `~/.claude`) | `type: "assistant"` records, `message.usage` |
| Codex | `$CODEX_HOME/sessions/**/*.jsonl` (default `~/.codex`) | `token_count` events; model carried from the preceding `turn_context` |

New RPC `provider.usage_history.read.request` / `.response`, gated on a new `server_info.features.providerUsageHistory` flag and tagged `COMPAT(providerUsageHistory)`. Buckets are `(day, provider, providerId, model)` in the client's IANA time zone, so a turn lands on the day the user experienced it.

Costs are priced from LiteLLM's rate table at the base tier and are labeled **"API estimate"** in the UI. They are not a subscription bill. Models the table doesn't know are counted in tokens and reported as `unpriced`. Partially priced totals show a lower bound (`≥`); wholly unknown costs show `—`. The warning applies to the chart and tables, and incomplete cost shares are withheld.

`docs/usage-history.md` documents the sources, the counting rules that are easy to get wrong, and the cache.

### Multiple providers of one kind (#2785)

> **Correction.** The first version of this PR scanned only `process.env` and the two default homes. Every total and screenshot it reported was therefore incomplete on any machine running a second account, and the page gave no sign of it. The numbers below replace them.

A user runs a second account by configuring a provider that extends `claude` or `codex` with its own `CLAUDE_CONFIG_DIR` or `CODEX_HOME`. Each of those homes holds its own transcripts. The scan now enumerates one home per **configured provider**:

- `extends` is followed transitively to a transcript-keeping built-in; anything that lands elsewhere keeps no token transcripts and is skipped, and a cycle resolves to nothing.
- Homes are deduplicated by `realpath`, so two ids pointing at one directory cannot double count.
- Built-in defaults are always scanned, including when the provider is disabled, because past usage is still real.
- Homes resolve per scan, so adding or removing a provider takes effect without a daemon restart.

A bucket carries both `provider`, the base kind that groups the chart series, and a new optional `providerId`, the configured provider owning the home. `sources[]` reports one entry per home with that provider's configured `label`, so a home whose scan failed can be named instead of quietly dropping its tokens.

The page keeps its per-kind rows and adds indented sub-rows under a kind used by **more than one** configured provider, plus a **Provider** breakdown mode. A kind with a single provider is unchanged.

On this machine that is not a rounding difference. Four homes, 30-day window:

| Provider | Cost | Tokens |
| --- | --- | --- |
| Claude Code | ≥$1,649.54 | 1.51B |
| Claude (Dean) | ≥$294.87 | 223M |
| Codex | ≥$477.24 | 1.10B |
| Codex (Colonel Panic) | ≥$132.82 | 330M |
| **Total** | **≥$2,554.47** | **3.17B** |

The two non-default homes are **$427.69 and 553M tokens — about 17%** of all activity, and the previous version reported none of it.

### Every host, and hosts that are down

The page lives under **Settings → App**, not under a single host: one read is issued per connected host and results render as soon as **any** host answers, so one slow or unreachable machine never blocks the rest. Each host resolves to `ready`, `pending`, `offline`, `unsupported`, or `error`, and everything short of `ready` is named in one muted line under the summary rather than silently dropped:

> `Scanning jimi-hendnix…` · `jay-lenovo is offline` · `mac-demarco-mini runs an older Paseo and is excluded`

**Two hosts can read the same transcripts.** Two daemons on one machine, or a shared home, would otherwise count every token twice. Each source now carries a fingerprint — `hostId` (the daemon's hostname), the resolved path, and `volumeId` (the directory's `device:inode`) — and the first host in a stable order claims it; a later host matching it has that source's buckets dropped and is named:

> `Counted once: ryzen-shine (localhost:31878) reads the same transcripts as ryzen-shine (localhost:31877)`

Hostname alone is not enough, which is why `volumeId` exists: every Mac in a fleet resolves `/Users/<user>/.claude`, so two machines that happen to share a hostname would collapse into one source and have real usage dropped. Both fields are optional, and a source lacking either can never be *proven* a duplicate, so it is always counted.

Hosts whose names collide are told apart by endpoint everywhere they are named. A host that contributed nothing after de-duplication is not a row in the Host breakdown; it is already named in the coverage line.

### Presentation

The page follows T3 Code's usage page: a layered smooth area chart (one line per provider, in that provider's own color, filled at low opacity), a summary column beside it on wide layouts, and flat sections with divider-separated tables. The chart's scale tops out at the largest single provider-day rather than the stacked sum, so a layered series is never drawn into the top of the plot.

This is a deliberate departure from [design.md](docs/design.md) §5's card-for-row-groups rule, at the maintainer's request; everything else stays on the documented rails. Provider brand color lives in one place (`providers.ts`) with `identity-colors.ts` as the precedent for theme-independent hexes.

### The counting rules that matter

These are the parts that silently produce wrong numbers, and each has a test:

- **Claude repeats the full `usage` object on every content-block record** of one assistant message. Dedupe is on `message.id:requestId`, globally across the scan rather than per file, because resumed and forked sessions copy history forward into other files. Summing naively overcounts by more than 2x.
- **Codex re-emits an unchanged `token_count`** on some stream boundaries; consecutive identical payloads are dropped, which is what makes the deltas reconcile with the session total.
- **Codex forked/subagent rollouts** open with the parent's history re-stamped to the fork instant. That leading burst is suppressed until the first event more than a second later, or it would double-count the parent.
- **`reasoningTokens` is a subset of `outputTokens`** and is never added on top. Total processed tokens is uncached input + cached input + cache creation + output.

A hand tally that skips two of these reads high enough to look like a scanner bug, so `docs/usage-history.md` now spells them out: filter *records* by day rather than admitting whole files by mtime, and dedupe Claude across files rather than within one. Checked by running the parser and an independent raw tally over the same file list at the same instant across all four homes on this machine — 3,157,544,080 tokens, 3,188 Codex events, **delta zero**, with no record dropped by the fork-copy or model-context gates.

### Performance

Transcripts are append-only, so parsed records are cached per file by `(size, mtime, provider)` along with the byte offset where parsing stopped. A file that merely grew resumes from that offset after a 64-byte FNV-1a guard hash confirms its tail is unchanged. The cache is per file rather than per day, so changing the reporting time zone invalidates nothing. A corrupt or missing cache costs one cold scan, never an error.

Measured on a real corpus (243 MB under `~/.claude/projects`, 1.6 GB under `~/.codex/sessions`), 30-day window:

| | |
| --- | --- |
| Cold scan | 2.5 s |
| Warm scan, same process | 114 ms |
| Warm after daemon restart | 212 ms |
| Scan cache on disk | 2.8 MB |

Windows are capped at 90 inclusive calendar days, and unsupported time zones are rejected before scanning. Scans and cache persistence run serially, with at most eight distinct pending requests; excess requests fail for retry. Concurrent identical requests share one scan. The rate table is fetched at most daily, with a 60 s floor on the explicit refresh, and falls back to an on-disk snapshot so the page works offline.

## Screenshots

All captured against real daemons reading my own transcripts, at 2x device scale. Two dev daemons were run side by side, one pointed at a different set of provider homes, to exercise the cross-host paths.

### Cost, 30 days, two hosts merged
Usage history summary and layered area chart

### Per-host provider rows
Sub-rows under a provider carry the host once more than one host contributes. Same-named hosts are told apart by endpoint, and the label wraps rather than clipping the part that distinguishes them.
Provider sub-rows labelled by host

### Host breakdown
Breakdown table in Host mode

### A host that is down
The other host's numbers still render; the missing one is named.
Coverage line naming an offline host

### Day readout on hover
The readout row is reserved at a fixed height and shows the window range when nothing is selected, so selecting a day never shifts the plot under the pointer.
Chart readout showing one day's per-provider values

### Dark
Usage history in the dark theme

### Compact (420 px)
Usage history at a compact width

### Tokens metric + Day breakdown
Tokens metric with the day breakdown table

### Provider breakdown
Breakdown table in Provider mode

### 90-day window
Usage history over a 90-day window

### Electron desktop (Linux)
From the earlier pass, and therefore **pre-restyle** — it shows the desktop shell mounting the page, not the current visual treatment.
Usage history in the Electron desktop app on Linux

### Does it work well

- Layout shift: the loading state reserves the loaded report's height; the chart's day readout occupies a fixed-height row whether or not a day is selected. Verified by toggling metric, window, and selection with the page warm and on first load.
- Cost-axis labels stay compact (`$600`) but retain fractional steps (`$0.5`, `$1.5`) so small totals do not produce duplicate tick labels.
- Empty, error, unsupported-host, and unavailable-pricing states all render as their own quiet card or `<Alert>` rather than blanking the page.

### Does it regress anything else

The neighboring quota Usage page is untouched. The additions are a new RPC, a new service, and a new settings page; the edits to shared files add feature negotiation, permission entries, dispatch, settings navigation, and access to current provider configuration.

### Platforms

| Platform | Tested | Notes |
| --- | --- | --- |
| iOS | ❌ | No Apple hardware available to me. |
| Android | ❌ | The only device attached to this machine is a Wear OS watch running the production build, with no Expo dev client. |
| Web | ✅ | Chromium via Playwright against real isolated daemons; light and dark, 1600 px / 420 px, all three windows, both metrics, all four breakdown tables, hover and press. Cross-host paths exercised with two daemons: merged, de-duplicated, and one stopped mid-session. |
| Desktop macOS | ❌ | No hardware. |
| Desktop Windows | ❌ | No hardware. |
| Desktop Linux | ⚠️ | Electron under Xvfb over CDP, from the original pass — **not** re-run for the cross-host change or the restyle. It renders the same component the browser run covers, and the change adds no Electron-specific code. |

The page contains **no platform-gated code**: no `.web.`/`.native.`/`.electron.` variants, no `isWeb`/`isNative`/`getIsElectron()` branches. Hover is additive via `onHoverIn`/`onHoverOut` on the column pressable, and press selection works on every platform, so a device with no hover loses nothing. `useUnistyles()` is not used anywhere in the change.

**Someone should still put this in front of a phone before it ships.** The updated Playwright test covers 90 days at 420 px, but this does not replace native-device testing.

### Automated coverage

The changed service, aggregation, derivation, formatting, view-state, and locale tests were run individually after review. The unchanged parser/cache/handler results below are from the original implementation pass. The full suite was not run locally.

```
packages/server/src/services/usage-history/   7 files, 79 passed
  transcripts.test.ts, transcript-reader.test.ts, aggregation.test.ts,
  pricing.test.ts, scan-cache.test.ts, service.test.ts,
  provider-homes.test.ts        <- new: extends chain, cycles, realpath dedupe
packages/app/src/provider-usage-history/      6 files, 51 passed
  merge.test.ts    <- new: cross-host merge, fingerprint dedupe, contributed hosts
  view.test.ts     <- new: multi-host page state
  curve.test.ts    <- new: monotone tangents, no overshoot
packages/app/src/i18n/resources.test.ts            36 passed
packages/app/e2e/browser/provider-usage-history.spec.ts   2 passed
```

The server tests cover the counting rules above against real fixture transcripts written to temp directories, plus cache corruption, resume-after-growth, guard-hash mismatch, and pruning. The handler tests cover success, service failure, forced refresh, and the invalid-window rejection.

The JSDOM component test was replaced by pure view-state tests and `packages/app/e2e/browser/provider-usage-history.spec.ts`. The browser test uses a real isolated daemon, real transcript files, and an on-disk pricing snapshot, with no mocked transport or app modules. It verifies partial/unknown costs, token totals, 90-day selection, compact rendering, and refresh after transcript removal. Errors use localized retry guidance; raw errors are logged separately.

The Arabic “Uncached input” label was checked in the original reviewed commit and already reads `الإدخال غير المخزّن`; the reported replacement characters are not present.

Review screenshots from the deterministic browser fixture:

Partial and unknown costs, with fractional axis ticks
Partial and unknown costs in the compact 90-day view

```
$ npm run typecheck      # all workspaces
exit 0
$ npm run lint -- <every touched file>

$ npm run test:e2e --workspace=@getpaseo/app -- <local NixOS browser config> provider-usage-history.spec.ts
2 passed
```

The browser run used the system Chrome executable and cleared the inherited `PASEO_PASSWORD` for the isolated test daemon. No production daemon was restarted.

End to end against a live dev daemon, through the real client:

```
features.providerUsageHistory = true
scanDurationMs 932  pricing fresh
source claude              ok  scanned 78  skipped 3  sessions 69   Claude Code
source codex               ok  scanned 79  skipped 2  sessions 79   Codex
source codex-colonelpanic  ok  scanned 46  skipped 0  sessions 46   Codex (Colonel Panic)
source claude-dean         ok  scanned 69  skipped 1  sessions 69   Claude (Dean)
total 3,157,544,080 tokens  (independent tally: identical, delta 0)
invalid window -> Invalid usage history window: sinceDay '2026-09-08' is after
  untilDay '2026-09-07'  code=provider_usage_history_invalid_window
```

## Protocol compatibility

- Request and response are new message types; nothing existing was narrowed, removed, or made required. All new `server_info.features` entries are optional booleans.
- `providerId` on a bucket and `providerId`/`label` on a source are **optional additions**. `provider` still carries the base kind and still drives the chart series, so an old client renders a new daemon unchanged, and a new client falls back to the base kind when an old daemon omits them.
- Wire schemas stay pure: no `.transform()`, `.catch()`, or `.preprocess()`.
- An old app against this daemon ignores a flag it doesn't read. A new app against an old daemon sees `providerUsageHistory` absent and renders the "update the host" state instead of calling the RPC. There is no fallback path.
- The shim is tagged `COMPAT(providerUsageHistory): added in v0.7.3, remove gate after 2027-03-07`.

## Scope

Deliberately left out:

- **Providers beyond Claude Code and Codex.** Copilot, OpenCode, and Pi keep no on-disk transcript with token counts, so there is nothing to scan. Adding one later is a parser plus a directory entry.
- **Hourly resolution.** T3 Code has a "past 24h" mode; this is daily only, which keeps the bucket key and the cache simpler.
- **Custom per-model price overrides**, and the multi-environment merge with transcript-directory fingerprinting. The latter is meaningless here because the page is scoped to one host.
- **A public SDK method.** `providers.listUsage()` landed on `main` in #4062 while I was working, and the symmetric `readUsageHistory` would be a natural follow-up. I left it out to keep this one focused change; happy to add it here if you'd rather it land together.

One defect I know about and did not fix: an offline host has no live endpoint, so if its display name collides with another host it cannot be disambiguated in the coverage line. Fixing it means reading the profile's configured endpoint rather than the active connection.

Maintainer edits are enabled.


boudra:
Closing this feature PR. Please describe the workflow and shared need in [Discussions](https://github.com/getpaseo/paseo/discussions), as described in [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #4466, opened 2026-09-08
Fix context usage for Anthropic-compatible providers

ele-yufo:
Follow-up to #2903, reopened as a new PR because author-side reopen is not permitted after the close. The evidence @boudra asked for is below and in the new commit.

## Summary

- Treat zero-token `message_start` usage as incomplete so output deltas cannot replace the context meter with an output-only value.
- Reconcile each parent assistant frame's per-request input, cache-read, cache-creation, and output usage into live and completed context usage.
- Keep subagent usage excluded through the existing parent-frame routing.

## Root cause

Some Anthropic-compatible gateways emit a zeroed `message_start` usage object. Paseo accepted the zero input as authoritative, added the later output delta, and displayed only that output count (for example, `799 / 1m`; on a captured real session the meter read **22 / 1M** while the turn actually consumed 38,917 input tokens).

## Evidence from an affected gateway

GLM Coding Plan via `open.bigmodel.cn/api/anthropic`, reached through a local CLIProxyAPI that only maps `glm-5.3-flash[1m]` → `glm-5.3-flash`. Synthetic prompt, response bodies only, credentials stripped.

`message_start` (line 2):

```json
data: {"type": "message_start", "message": {..., "usage": {"input_tokens": 0, "output_tokens": 0}}}
```

Terminal `message_delta` (line 89 direct, line 98 via the proxy):

```json
data: {"type": "message_delta", "delta": {"stop_reason": "end_turn", ...}, "usage": {"input_tokens": 818, "output_tokens": 25, "cache_read_input_tokens": 0, ...}}
```

Direct and proxied captures are identical in shape: the zeroing originates at the vendor endpoint; the proxy passes it through unchanged.

### Before/after context-meter readings (real sessions through the production wrapper)

- **main (`1f5b6143d`)** — turn consumed input 38,917 + output 22; meter showed **22 / 1M**, `turn_completed` also carrying `contextWindowUsedTokens: 22`.
- **this branch (`196681a65`)** — meter showed the correct total after the turn (38,935 / 1M that run), but no live `usage_updated` events.

## What the traces added in `a11e45513`

Instrumenting the SDK stream showed the completed **assistant frames also carry zeroed usage** on this gateway (`{"input_tokens":0,"output_tokens":0}`); accurate per-request input/cache/output is published only on the terminal `message_delta`. The assistant-frame reconcile alone therefore never fired live. The follow-up commit:

- adopts `message_delta` input tokens (incl. cache fields) when they improve on the current per-request value, gated on the current request's `message_start` so a straggler terminal delta from a canceled request cannot seed usage for the next turn;
- applies the same non-regressing merge to `buildAssistantUsageEvent` so a partially-populated assistant frame cannot undo the reconciled value;
- adds four regression tests mirroring the captured trace and these orderings, each mutation-verified.

Reading with the commit on the same chain: live `usage_updated: used=38957` (input 31,621 + cache_read 7,296 + output 40), meter **38,957 / 1M** mid-turn, `turn_completed` matching.

## Validation

- `npx vitest run packages/server/src/server/agent/providers/claude/agent.test.ts --bail=1` (66 passed)
- `npm run lint` and format check clean on changed files
- repo-wide typecheck: no new errors versus the branch baseline (this checkout predates current main and lacks `@/partial/watcher` native builds, so a fixed set of pre-existing errors remains in untouched files)

### DIFF
--- a/packages/server/src/server/agent/providers/claude/agent.test.ts
+++ b/packages/server/src/server/agent/providers/claude/agent.test.ts
@@ -2182,6 +2182,84 @@ describe("ClaudeAgentSession context window usage", () => {
     }
   });
 
+  test("assistant usage corrects incomplete Anthropic-compatible stream usage", async () => {
+    const session = await createSessionForTurns(
+      [
+        [
+          createInitMessage(),
+          createMessageStartEvent({
+            input_tokens: 0,
+            cache_creation_input_tokens: 0,
+            cache_read_input_tokens: 0,
+          }),
+          createMessageDeltaEvent(799),
+          {
+            type: "assistant",
+            parent_tool_use_id: null,
+            message: {
+              id: "assistant-compatible-provider-1",
+              role: "assistant",
+              content: [{ type: "text", text: "Continuing with a tool call." }],
+              usage: {
+                input_tokens: 1_514,
+                cache_creation_input_tokens: 0,
+                cache_read_input_tokens: 178_752,
+                output_tokens: 799,
+              },
+            },
+            uuid: "assistant-compatible-provider-event-1",
+            session_id: "session-1",
+          },
+          createSuccessResult({
+            usage: {
+              input_tokens: 1_514,
+              cache_creation_input_tokens: 0,
+              cache_read_input_tokens: 178_752,
+              output_tokens: 799,
+              iterations: [],
+            },
+            modelUsage: {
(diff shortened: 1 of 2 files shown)


boudra:
Closing in favor of the streaming-usage implementation in [#4295](https://github.com/getpaseo/paseo/pull/4295), which is still open.

PR #4467, opened 2026-09-08
fix(app): collapse commit diff files from their header

CN-liuzhiyang:
### Linked issue

Closes #

### Type of change

- [x] Bug fix

### Reasoning

Clicking a file header in the working diff collapses that file. The same header in a commit diff does nothing — it is not even pressable. Nothing about collapsing is specific to the working tree; a commit that touches twenty files is exactly where you want to fold away the ones you have already read.

The reason it was missing is structural: `collapseState` was declared on the working branch of the `DiffDocumentProps` union, so the commit panel had no way to pass one, and `DocumentFileHeader` returned early on any non-working mode.

### Goals

- A commit diff file header collapses and expands on press, like the working diff.
- Collapsed paths persist per tab, like the Changes panel.
- Right click on a commit diff header does not regress.

### Non-goals

- No file-actions menu (Open file, Revert, Add to chat) on commit diff headers. Those are working-tree actions and the commit panel owns none of them.

### Implementation notes

`collapseState` moves onto the document's base props so both modes toggle.

`FileHeader` separates the two things `interactive` used to mean. Press and hover now follow `onActivate`; the drag source and the file-actions menu stay behind `interactive`, which the commit diff does not set. Without that split a commit header would mount a `ContextMenu` whose content renders `null`, and right click would be swallowed for a menu that never appears.

That is also why `ContextMenuTrigger` changes: a trigger with `enabled={false}` still called `preventDefault`/`stopPropagation`, so it ate the right click and showed nothing. It now lets the event through, and the diff surface's own Copy / Copy line / Select all menu still opens over a commit diff header. No existing caller passes `enabled={false}`, so nothing else changes.

### QA

Automated, on this branch:

```
$ npm run typecheck
(all workspaces, no errors)

$ npm run lint

$ npx vitest run src/components/ui/context-menu.test.tsx src/git/use-diff-files.test.ts --bail=1
 Test Files  2 passed (2)
      Tests  3 passed (3)
```

New browser coverage in `packages/app/e2e/browser/commit-diff-panel.spec.ts`:

```ts
test("a commit diff file collapses and expands from its header", ...)
  await panel.getByTestId("diff-file-0").click();
  await expect(panel.getByTestId("diff-file-0-body")).not.toBeVisible();
  await panel.getByTestId("diff-file-0").click();
  await expect(panel.getByTestId("diff-file-0-body")).toBeVisible();
```

Manual, Windows desktop, against a real daemon in a real browser, through the same steps the spec runs (Changes → Commits → commit row):

Expanded:

expanded

After pressing the `feature.txt` header:

collapsed

Pressing again restores the body. Right clicking the header still opens the diff Copy / Copy line / Select all menu.

| iOS             |        | Shares the `DocumentFileHeader` path with web; not run on device |
| Android         |        | Same |
| Web             | yes    | Playwright browser project |
| Desktop macOS   |        |       |
| Desktop Windows | yes    | Electron dev build |
| Desktop Linux   |        |       |

`commit-diff-panel.spec.ts` could not be run locally: on this machine the `withWorkspace` fixture never gets past "No projects yet", and it fails the same way on a clean checkout of `main`, so it is an environment problem rather than something this branch introduced. The collapse behaviour above was verified through a `seedWorkspace`-based run of the identical steps. CI covers the committed spec.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence

### DIFF
--- a/packages/app/e2e/browser/commit-diff-panel.spec.ts
+++ b/packages/app/e2e/browser/commit-diff-panel.spec.ts
@@ -1,7 +1,7 @@
 import { execFileSync } from "node:child_process";
 import { writeFile } from "node:fs/promises";
 import path from "node:path";
-import type { Locator } from "@playwright/test";
+import type { Locator, Page } from "@playwright/test";
 import { test, expect } from "../support/fixtures";
 import { openChangesTreePanel } from "../support/helpers/workspace-tabs";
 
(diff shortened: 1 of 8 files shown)


boudra:
Closing this commit-diff folding proposal. Please discuss the workflow and shared need in [Discussions](https://github.com/getpaseo/paseo/discussions), following [CONTRIBUTING.md](https://github.com/getpaseo/paseo/blob/main/CONTRIBUTING.md).

PR #4492, opened 2026-09-08
fix(app): redraw terminal after switching away from a hidden tab

### LINKED ISSUE #3376: bug: desktop viewport becomes visually corrupted after switching away and back
## What's broken

On the macOS desktop app, returning to a terminal/agent screen after navigating to another screen can leave the entire content area severely visually corrupted.

Instead of repainting cleanly, stale portions of the previous terminal/agent frame remain composited at incorrect vertical offsets. Large horizontal white bands and black rectangles cut across the content, and multiple chunks of terminal output overlap. The sidebar and top navigation remain mostly intact while the main terminal/agent viewport is corrupted.

Expected: returning to the screen performs a complete repaint and shows the terminal/agent content normally.

Actual: the viewport contains stale, duplicated, offset content and broad blank bands, making it unreadable.

This appears related to the focus/visibility/rendering family of issues in #39, #135, and #288, but the symptom is more severe than an incorrect terminal size and still occurs on the latest v0.4.0 release.

## Steps to reproduce

1. Open Paseo Desktop and enter a workspace with a terminal/agent screen containing enough output to scroll.
2. Navigate to another Paseo screen, tab, or workspace.
3. Return to the original terminal/agent screen.
4. Intermittently, the main viewport is rendered with stale overlapping frame fragments and large horizontal white/black bands.

The issue is intermittent rather than reproduced on every switch.

## Where did this happen

Desktop (Electron)

## Paseo version

0.4.0

## OS version

macOS 26.6.1 (25G76), Apple Silicon arm64, Apple M1, Metal 4, 1920×1080 display

## Agent provider

Not relevant

## Logs

No relevant renderer/GPU error was found in `~/Library/Logs/Paseo/main.log` around the available sessions.

## Screenshots or video

A screenshot was captured at the time of the issue. It clearly shows broad horizontal white bands, black rectangles, and stale terminal/agent content duplicated and vertically offset across the viewport. The original screenshot contains local workspace names and filesystem paths, so it is not attached publicly here; a redacted copy can be added if needed.

omorsi45:
## Summary
- Hidden retained tabs measure 0x0 under `display: none`. The next same-size fit skipped xterm `refresh()`, which left a blank WebGL canvas with only a thin strip of glyphs.
- Remember that a fit was skipped while hidden, then force a visible-row refresh on the next real fit.
- Tab reflow now uses `forceRefresh` (plus a follow-up rAF), matching the existing window `visibilitychange` path.

Fixes #3376

## Test plan
- [x] `npx vitest run src/terminal/runtime/terminal-emulator-runtime.test.ts` (20/20)
- [ ] Desktop: open a terminal with enough scrollback, switch to another tab/workspace/agent, switch back. Viewport should repaint fully instead of staying blank.
- [ ] Repeat a few times (issue is intermittent). Sidebar/chrome should stay intact.
- [ ] Window minimize/restore still refreshes as before.

Platforms tested: Windows unit tests only. Not smoke-tested in Desktop yet.


omorsi45:
Addressed the Greptile finding: dropped the fake-DOM unit regression and added a Chromium browser test that hides the terminal with `display: none` (same as RetainedPanel), restores it, then asserts xterm `refresh` on the next same-size fit.

Verified: browser test fails without the runtime fix and passes with it. `npx vitest run --project browser ... -t retained tab` and unit suite both green.

boudra:
Closing in favor of [#4101](https://github.com/getpaseo/paseo/pull/4101), which is still open and includes the retained-tab surface restoration path.

PR #4493, opened 2026-09-08
fix(agent): preserve Cursor Auto model id default

frontend-london:
## Summary
- Paseo `normalizeConfig` stripped `model: "default"` as a legacy “unspecified” sentinel, then filled `isDefault` / first catalog model (often `claude-opus-5`).
- Cursor Auto’s real provider-native id is `default`, so explicit Auto was rewritten to Opus and hit plan limits (`Upgrade your plan to continue`).
- Fix: keep `"default"` when the provider catalog lists that id; otherwise keep legacy strip→resolve behavior (Codex tests unchanged).

## Test plan
- [x] unit: preserve provider-native `default` (Cursor Auto)
- [x] unit: legacy Codex `model=default` still resolves to provider default
- [x] live: `paseo run --provider cursor --model default` → `config.model=default`, agent replies without plan upgrade
- [x] ACP E2E: `start_role` runtime=cursor model=auto → execution_model=default, `config.model=default`, `git status` runs

### DIFF
--- a/docs/agent-lifecycle.md
+++ b/docs/agent-lifecycle.md
@@ -22,7 +22,11 @@ same Paseo agent ID. Provider history is not appended again when the canonical t
 primed.
 
 Idle agents remain resident indefinitely. Runtime closure happens only through an explicit lifecycle
-action such as archive, replacement, reload, workspace teardown, or daemon shutdown.
+action such as archive, replacement, reload, workspace teardown, or daemon shutdown -- or the terminal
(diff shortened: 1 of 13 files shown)


boudra:
Closing this bundled change to model selection, daemon restart, and provider lifecycle behavior. Please submit the Cursor Auto model-ID fix separately with its regression coverage and reported live-provider evidence.

PR #4426, opened 2026-09-07
chore(arch): drop dead re-exports, extract listen-target, align --host wording

nillikechatchat:
## 架构清理 PR（3 个独立 commit，每个都可单独 revert）

源自 `REVIEW.md` 的三项最小改动。每个 commit 单一目的、零行为变化、易回退。

### Commit 1 — `refactor(protocol): drop dead chat/loop type re-exports from messages.ts`

- 删 `packages/protocol/src/messages.ts` 末尾 24 行 `export type ChatXxxRequest/Response` 和 `LoopXxxRequest/Response`
- 经 ripgrep 确认 monorepo 内 0 消费者
- 模式本身在 P1-5 review 笔记里点名过；这是它的一半（另一半是 schedule，留作独立清理）
- Schema 本身保留（仍被 `SessionInboundMessageSchema` union 内部使用）
- 破坏性？仅 TS 导出层；运行时、wire format 都不变。需要这些类型的未来消费者请走 `@getpaseo/protocol/chat/rpc-schemas` / `@getpaseo/protocol/loop/rpc-schemas`（subpath 已在 `package.json` `exports` 中声明）

### Commit 2 — `refactor(server): extract listen-target utilities from bootstrap.ts`

- 把 80 行的纯函数（`ListenTarget`、`parseListenString`、`resolveBoundListenTarget`、`formatListenTarget`、`WINDOWS_DRIVE_RE`）从 `packages/server/src/server/bootstrap.ts` 抽出到 `packages/server/src/utils/listen-target.ts`
- 行为零变化；`bootstrap.ts` 和 `bootstrap.smoke.test.ts` 只改 import 路径
- `service-proxy.ts` 的私有 `resolveBoundListenTarget` 故意不去重——它是个 sibling-but-isolated 副本，统一是 P2-3 的另一片
- 价值：纯工具不再依赖 daemon 生命周期/MCP/WS；可被未来 relay 控制通道等复用

### Commit 3 — `refactor(cli): align --host option wording with glossary (daemon, not host)`

- `docs/glossary.md` 把 "Daemon host target" 列为已知术语混淆：Host 是客户端连接 profile，daemon 才是服务端进程
- 改 `command-options.ts` 的 option 描述、对应 JSDoc、内部常量名 `DAEMON_HOST_OPTION_DESCRIPTION` → `DAEMON_ADDRESS_OPTION_DESCRIPTION`
[...]

PR #4505, opened 2026-09-08
refactor(cli): clarify --host option as daemon address

nillikechatchat:
## Why

`docs/glossary.md` separates two concepts that the current `--host` help text conflates:

- **daemon** — the service the CLI targets
- **Host** — a client-side connection profile that points *at* a daemon

Today the help string says `Daemon host target: ...`, which reads as if a daemon is a host. The value the user actually supplies to `--host` is the daemon's own network address, so the wording should reflect that.

This is a focused, follow-up to #4426 (closed per @boudra — "we're not taking this refactor. A separate PR limited to the --host help wording can be considered"). It carries only the wording change.

## What

- `packages/cli/src/utils/command-options.ts`
  - Constant `DAEMON_HOST_OPTION_DESCRIPTION` → `DAEMON_ADDRESS_OPTION_DESCRIPTION`
  - Description: `Daemon host target: ...` → `Daemon address to target: ...`
- `packages/cli/src/output/with-output.ts`
  - `CommandOptions.host` JSDoc: `Daemon host target from --host option` → `Daemon address from --host option`

The user-facing `--host` flag and the `host` field name are intentionally untouched. The flag is a public CLI contract, and the field is referenced widely enough that renaming it would balloon the diff past the scope of a wording fix.

## Scope

- 2 files, +4 / −4
- No protocol, server, or build-system changes
- No new dependencies, no formatting churn
- `docs/glossary.md` is not modified — it does not currently reference this CLI string, and editing it in the other direction would just invite another round of back-and-forth

## Verification

Static checks already done locally:

- `git diff --stat` → 2 files, +4 / −4
- `rg "DAEMON_HOST_OPTION_DESCRIPTION"` → 0 hits (old constant name removed)
- `rg "Daemon host target"` → 0 hits (old wording removed)
- `rg "Daemon address to target"` → 1 hit (`command-options.ts`)

I deliberately did not run `npm install` / `tsc --noEmit` / `biome format` in this environment. The change is a string + private constant rename inside a leaf CLI utility, so the PR's required `format / lint / typecheck` checks on CI should be the source of truth here.

## Test Plan

- After merge, run `paseo --help` (or any subcommand) and confirm the `--host` line reads:
  `Daemon address to target: host:port, tcp://host:port, or ssh://user@host (default: local socket/pipe, then localhost:6767)`
- Existing CLI tests that exercise `--host` parsing should remain green; the field name and flag shape are unchanged.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -54,6 +54,4 @@ Authoritative terminology. UI label wins. Don't invent synonyms; use what's here
 - **Product discussion** — Where non-bug product talk goes: [GitHub Discussions](https://github.com/getpaseo/paseo/discussions) and `#product` in Discord. Covers workflows, features, and product direction. The container term; the ask inside it is always a **workflow** — what you're doing, how you do it today, where Paseo gets in the way. Issues are bugs only. Forbidden: "Feature request" as a Paseo surface, label, or invitation; the phrase is only used when answering someone who arrived with it. Don't invent "feature discussion", "idea", or "workflow thread" as synonyms. See [CONTRIBUTING.md](../CONTRIBUTING.md).
 
 ## Inconsistencies (documented, not papered over)
-
-- CLI `--host <host>` description `"Daemon host target"` (`packages/cli/src/utils/command-options.ts:5`) blurs daemon/host; the app keeps them distinct.
 - `WorkspaceDescriptorPayloadSchema.workspaceKind` accepts legacy `"checkout"` on the wire (`packages/protocol/src/messages.ts:2187`) while `PersistedWorkspaceKind` does not (`packages/server/src/server/workspace-registry-model.ts:8`).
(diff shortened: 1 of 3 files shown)


nillikechatchat:
You're right, thanks for the catch @greptile-apps. Pushed `70dcce7` to fix it.

**What changed**

Removed the now-resolved entry from `docs/glossary.md` under "Inconsistencies (documented, not papered over)":

```
- CLI `--host <host>` description `"Daemon host target"` (`packages/cli/src/utils/command-options.ts:5`) blurs daemon/host; the app keeps them distinct.
```

The wording it flagged no longer exists in the code, so per the glossary's own rule ("obsolete sections go") the entry should come out. The other legacy entry on the same section (`workspaceKind` accepting `"checkout"` on the wire) is left alone — that's still a real inconsistency.

I had this called out the wrong way in the original PR body ("`docs/glossary.md` is not modified — it does not currently reference this CLI string"). Greptile is correct that it did, and I missed it. The PR's scope is still "limited to the --host help wording" per @boudra's guidance; updating the docs entry that tracks that exact wording is part of the same fix, not a new refactor.

**Final state**

- 2 commits, 3 files, +4 / −6
- `packages/cli/src/utils/command-options.ts` — constant + help text
- `packages/cli/src/output/with-output.ts` — JSDoc
- `docs/glossary.md` — drop resolved inconsistency entry

Ready for another pass when Greptile has a moment.

PR #4519, opened 2026-09-08
feat(desktop): add --daemon-only to run the app as a headless daemon service

josham:
## Problem

The desktop app starts its daemon when you open a window, and that daemon outlives the app once started — but nothing starts it before then. On a freshly booted machine, phone clients, browser clients, schedules, and the CLI have nothing to talk to until someone logs in and opens the GUI.

## What this adds

`Paseo --daemon-only` runs the packaged app as a headless daemon host: no window, no display needed.

It reuses the path the app already walks. `resolveDaemonRunnerEntrypoint()` + `createNodeEntrypointInvocation()` already re-exec the app's own Electron binary with `ELECTRON_RUN_AS_NODE=1` against the bundled `@getpaseo/server` supervisor every time the app starts its own daemon. The flag runs that same daemon in the foreground and keeps the app alive alongside it, so a service manager can supervise it.

Because the daemon ships inside the app, client and daemon are one artifact: no separate daemon install to keep in step, and an app update replaces both halves at once.

```ini
[Service]
Type=simple
ExecStart=%h/Applications/Paseo-x86_64.AppImage --daemon-only
EnvironmentFile=%h/.config/paseo/daemon.env
Restart=always
```

## Three things this had to get right

Each of these was found by running it as a real systemd user unit. None of them show up when you reason about the code, and the first two do not reproduce with a plain `kill -TERM` on the main pid.

1. **Node signal handlers never run.** Chromium's browser process handles `SIGTERM`/`SIGINT` itself and exits without dispatching to `process.on("SIGTERM")`. Verified by instrumenting the handler: `SIGTERM listeners=1` at registration, and the handler never fired while the process still exited 0. The stop therefore hangs off `app.on("before-quit")`, which does fire.

2. **`event.preventDefault()` on that event aborts the browser process.** Holding the quit open to wait for the daemon produced `Main process exited, code=dumped, status=5/TRAP` and left the unit `failed` after a normal `systemctl stop` — Chromium tears down its helper processes as soon as it handles the signal, so vetoing the quit leaves it running without them. The daemon is signalled and the quit is allowed to proceed. Nothing is lost: the daemon shuts down on the `SIGTERM` it just received, and a service manager waits for every process in the unit's cgroup rather than only the main one.

3. **A unit's environment must outrank the login shell.** The app inherits the login shell environment on startup so agents get the `PATH` a terminal would give them — which matters more for a service at boot than anywhere else, since there is no shell to inherit from. But `inheritLoginShellEnv()` applies that with `Object.assign(process.env, shellEnv)`, so a shell rc exporting `PASEO_HOME` for interactive CLI use would silently move the daemon to a different home and take every agent and project with it. `inheritLoginShellEnvForService()` re-applies the `PASEO_*` variables the service manager passed in.

## Interaction with a desktop app

The daemon is deliberately **not** marked `PASEO_DESKTOP_MANAGED`, so it is not recorded as `desktopManaged` in the PID lock. A desktop app that later attaches treats it as an external daemon and will not stop it on quit or to install an app update. The GUI quit lifecycle, auto-updater, and window management stay unregistered for this launch mode.

The web UI is left at the daemon's own default rather than forced off the way the desktop-managed daemon forces it, so browser and phone clients can reach a machine with no desktop session.

### Automated

22 new unit tests in `packages/desktop/src/daemon/daemon-only.test.ts` and `packages/desktop/src/desktop-startup.test.ts`, covering flag parsing (packaged and `defaultApp` argv shapes), exit-code resolution, stop forwarding and escalation, spawn failure, startup routing precedence, and environment precedence. No mocks — the child process, signal source, and environment are injected.

```
$ npx vitest run packages/desktop/src/daemon/daemon-only.test.ts \
    packages/desktop/src/desktop-startup.test.ts \
    packages/desktop/src/daemon/daemon-manager.test.ts \
    packages/desktop/src/daemon/node-entrypoint-launcher.test.ts --bail=1

 Test Files  4 passed (4)
      Tests  38 passed (38)

$ npm run typecheck   # clean
$ npm run lint -- packages/desktop/src
Finished in 190ms on 137 files with 177 rules using 16 threads.
```

### Manual — packaged app, real systemd user unit

Built with `electron-builder --linux --dir` so the asar and `app.asar.unpacked/dist/daemon/node-entrypoint-runner.js` paths are the ones under test. Run as a `systemctl --user` unit with `UnsetEnvironment=DISPLAY WAYLAND_DISPLAY`, a spare port, and a temporary `PASEO_HOME`.

| Check | Result |
| --- | --- |
| Starts with no `DISPLAY` and no `WAYLAND_DISPLAY` | `Active: active (running)`, `Main PID: … (Paseo)` |
| Honours `PASEO_HOME` | daemon state written to the unit's home, `paseo.pid` present |
| Honours `PASEO_LISTEN` | `LISTEN 127.0.0.1:6795 users:(("Paseo Daemon",…))` |
| Not marked desktop-managed | `desktopManaged` absent from the PID lock |
| `systemctl --user stop` | `ActiveState=inactive Result=success ExecMainCode=0 ExecMainStatus=0`; port and PID lock released |
| Graceful daemon shutdown | `supervisor_received_SIGTERM. Stopping worker...` → `Server closed` → `Worker exited (code 0). Supervisor shutting down.` |
| `systemctl --user restart` | back to `active`/`running` and listening |
| Daemon crash (`SIGKILL` the supervisor) | main process exits 1, `Restart=` starts a fresh invocation, daemon listening again |
| No orphans | unit cgroup empty after stop |

Environment precedence was proved end to end, not just in a unit test: the service was started with `PASEO_HOME` set by the unit and `SHELL` pointed at a stub login shell that exports `PASEO_HOME=/tmp/WRONG-HOME-FROM-SHELL`. The stub ran and its environment was applied (`afterPath: '/usr/bin:/bin'`, `pathChanged: true`), and the daemon still used the unit's home; `/tmp/WRONG-HOME-FROM-SHELL` was never created.

### Platform matrix

| Platform | Tested | Notes |
| --- | --- | --- |
| Linux (packaged, systemd user unit) | Yes | Arch, Electron 44. Full lifecycle above. |
| Linux (unpackaged dev Electron) | Yes | Same start/stop behaviour via the non-packaged runner path. |
| macOS | No | The flag is platform-neutral and the code path is shared, but launchd is untested. I do not have a Mac to verify on. |
| Windows | No | Untested. |
| iOS / Android / browser web | N/A | No app or protocol changes. |

Nothing outside `packages/desktop` changes, and every new branch is behind the flag: without `--daemon-only`, `runDesktopStartup` is byte-for-byte the previous path plus one predicate that returns `false`.

### Known cost

The Chromium browser process still initializes — it is the service's main process — and its zygote and network helpers stay resident. On this machine that measured ~430 MB RSS for the browser process plus ~200 MB across the helpers, beyond the daemon itself. A launcher-level wrapper that `exec`s the Node runner directly would avoid it, but only for the AppImage; this approach is the one that works for every packaging target and platform. Happy to follow up with a Linux-only optimisation if you'd rather have it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,6 +59,7 @@ At the start of non-trivial work, list `docs/` and skim anything relevant to the
 | [docs/browser-capture-harness.md](docs/browser-capture-harness.md)   | Real-Electron browser screenshot harness and compositor-surface gotcha                                                         |
 | [docs/android.md](docs/android.md)                                   | App variants, local/cloud builds, EAS workflows, version codes, F-Droid source builds and store metadata                       |
 | [docs/docker.md](docs/docker.md)                                     | Running the daemon and bundled web UI in Docker, volumes, agent images, security                                               |
+| [docs/daemon-service.md](docs/daemon-service.md)                     | Running the daemon as an OS service with `--daemon-only`, unit configuration, and env precedence                               |
 | [docs/release.md](docs/release.md)                                   | Release playbook, draft releases, completion checklist                                                                         |
 | [docs/terminal-activity.md](docs/terminal-activity.md)               | Terminal activity indicators — source-agnostic tracker, agent hook reporting, adding a new hook provider                       |
 | [SECURITY.md](SECURITY.md)                                           | Relay threat model, E2E encryption, DNS rebinding, agent auth                                                                  |
--- a/docs/daemon-service.md
+++ b/docs/daemon-service.md
@@ -1,0 +1,71 @@
+# Running the daemon as a service
+
+The desktop app starts its daemon when you open a window, and that daemon
+outlives the app once started. Nothing starts it before then, so on a freshly
+booted machine phone clients, browser clients, schedules, and the CLI have
+nothing to talk to until someone logs in and opens the GUI.
+
+`--daemon-only` runs the packaged app as a headless daemon host instead:
+
+```
+Paseo --daemon-only
+```
+
+No window is created and no display is needed. The app already re-execs its own
+Electron binary with `ELECTRON_RUN_AS_NODE=1` to run the bundled daemon every
+time it starts one; the flag runs that same daemon in the foreground and stays
+alive alongside it, so a service manager can supervise it.
+
+Because the daemon ships inside the app, the client and the daemon are one
+artifact. There is no separate daemon install to keep in step, and an app update
+replaces both halves at once.
+
+## systemd
+
+```ini
+[Unit]
+Description=Paseo daemon
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+ExecStart=%h/Applications/Paseo-x86_64.AppImage --daemon-only
+EnvironmentFile=%h/.config/paseo/daemon.env
+Restart=always
+RestartSec=3
+
+[Install]
+WantedBy=default.target
+```
+
+Run it as a user unit (`systemctl --user`). The daemon needs the user's home,
+SSH agent, and agent CLIs, so a system unit is the wrong scope.
+
+`systemctl stop` exits 0, and a daemon that dies on its own exits non-zero so
+`Restart=` brings it back.
+
+## Configuration
+
+Configure it through the environment, the same variables the daemon reads
+anywhere else — `PASEO_HOME`, `PASEO_LISTEN`, `PASEO_PASSWORD`.
+
+**A unit's environment outranks the login shell.** The app inherits the login
+shell's environment on startup so agents get the `PATH` a terminal would give
+them, which matters more here than usual because a service at boot has no shell
+to inherit from. Every `PASEO_*` variable the service manager passed in survives
+that inheritance. Without it, a shell rc that exports `PASEO_HOME` for
+interactive CLI use would move the daemon to a different home and take every
+agent and project with it.
+
+## What it is not
+
+A `--daemon-only` process is not a desktop-managed daemon. It is not recorded as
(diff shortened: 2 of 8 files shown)


josham:
Withdrawing this for now.

Running the flag as a real systemd unit showed the cost more clearly than I'd accounted for: because the flag is handled in `main.ts`, the service's main process is Chromium's browser process, and it brings its helpers with it. Measured on Linux for a mode that never opens a window: browser 426 MB, two zygotes 124 MB, gpu-process 198 MB, network service 81 MB, and a renderer 62 MB — about 890 MB RSS. The GPU process starts with `--ozone-platform=wayland --render-node-override=/dev/dri/renderD128`, so a headless boot service ends up opening the DRI device, and `--disable-gpu` does not prevent it.

That is a poor trade for the one use case the flag exists for, so I'd rather not put it in front of you in this shape. I'm going to carry it in my own fork's build and look at handling the flag before Electron starts, which is the only way to avoid the browser process entirely.

Two pre-existing things I ran into that may be worth a look independently of this PR:

- `startLocalDaemonForeground` spawns `process.execPath` without `ELECTRON_RUN_AS_NODE`, so `paseo daemon start --foreground` does not work from inside the packaged app.
- `registerExternalQuitSignals` installs `process.on("SIGTERM")` handlers, but Chromium's browser process handles SIGTERM itself and exits without dispatching to them. On Linux those listeners never fire; `app.on("before-quit")` does.

PR #4522, opened 2026-09-08
fix(app): keep terminal tab labels across terminals_changed pushes

### LINKED ISSUE #4521: bug: terminal tab labels fall back to Terminal after tab switches
### What's broken

Terminal tab labels intermittently fall back to the generic **"Terminal"** name after switching tabs/workspaces/agents, even though the PTY is still alive and previously showed a real title (for example an OSC title or renamed label).

### Why

1. Tab labels come from the React Query `listTerminals` cache (`title ?? name`).
2. `terminals_changed` pushes were filtered with strict `terminal.workspaceId === route.workspaceId`.
3. When a push omitted `workspaceId` (title/activity updates can), the terminal was wiped from the cache and the tab flashed the fallback label.

### Where did this happen

Desktop (Windows). Also expected on any client using the same push-router path.

### Paseo version

`0.8.0-beta.1` / current `main`

### Steps to reproduce

1. Open a workspace with multiple terminals that have non-default titles.
2. Switch between terminal tabs, agent tabs, and/or workspaces.
3. Observe terminal tab labels briefly or persistently becoming "Terminal".

### Fix direction

- Treat missing `workspaceId` on a push as belonging to the subscribed workspace (still exclude sibling workspace ids).
- Remember the last live title/name on the panel descriptor so a transient cache miss does not flash the fallback.

omorsi45:
## Summary
- `terminals_changed` used strict `workspaceId` equality, so pushes that omitted the field wiped terminals from the React Query cache and tab labels fell back to "Terminal".
- Keep terminals with a missing `workspaceId` for the subscribed workspace (still exclude sibling workspace ids).
- Remember the last live title/name on the terminal panel descriptor so a transient cache miss does not flash the fallback.

Fixes #4521

## Test plan
- [x] `npx vitest run packages/app/src/data/push-router.test.ts`
- [x] `npx vitest run packages/app/src/utils/terminal-tab-label.test.ts`
- [ ] Desktop: open terminals with custom/OSC titles, switch tabs/workspaces/agents, confirm labels stay instead of becoming "Terminal"

### DIFF
--- a/packages/app/src/data/push-router.test.ts
+++ b/packages/app/src/data/push-router.test.ts
@@ -50,7 +50,11 @@ function createFakeClient(config: { rejectCheckoutDiffSubscribe?: boolean } = {}
   emit: <K extends RouterMessageType>(message: Extract<RouterMessage, { type: K }>) => void;
   subscribeCheckoutDiffCalls: Array<{
     cwd: string;
-    compare: { mode: "uncommitted" | "base"; baseRef?: string; ignoreWhitespace?: boolean };
+    compare: {
+      mode: "uncommitted" | "base";
+      baseRef?: string;
+      ignoreWhitespace?: boolean;
+    };
     subscriptionId: string;
   }>;
   unsubscribeCheckoutDiffCalls: string[];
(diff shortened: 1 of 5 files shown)


boudra:
Closing because we haven't been able to reproduce the reported issue and there's no before/after QA evidence showing this patch fixes it. Speculative patches are not accepted. Please test the app before and after applying the patch, confirm that the issue is fixed, and reopen with that QA evidence.

boudra:

PR #4585, opened 2026-09-09
Preserve workspace selection when creation finishes late

boudra:
When workspace creation finishes after the user selects another workspace, completion now preserves that selection while still creating the requested agent, terminal, or empty workspace. Staying on the initiating New workspace screen keeps normal navigation. Revisiting the result does not submit a leftover draft.

This carries forward Christoph Leiter's contribution, rebased onto current main. Draft clearing uses the current composer interface and captures draft ownership before workspace creation starts, preserving a newer draft opened during that wait. Agent creation shares the existing request construction; the initiating host's runtime owns client lookup.

### Goals

- Preserve the workspace the user chose during delayed creation.
- Preserve submitted prompt, attachments, provider configuration, and exactly one requested resource.
- Keep normal navigation when the initiating creation screen remains active.

### Non-goals

Faster provisioning, completion notifications, and transport-retry deduplication. The latter remains separate work in #4442.

### Supersedes

- [#2986](https://github.com/getpaseo/paseo/pull/2986) — delivers Christoph Leiter's selected fix with current-main integration and the missing interaction recordings. His commits and attribution are retained. This maintainer branch avoids requiring changes to contributor fork permissions.

### QA

[Recordings, before/after screenshots, commands, and raw results](https://github.com/getpaseo/paseo/tree/qa/workspace-focus-2986).

- Current main 433e67b18: both original local-workspace journeys and all three explicit worktree journeys fail because completion selects the newly created workspace.
- Fix: seven local-workspace journeys and three real-worktree journeys pass. Covers chat/terminal/empty creation, leaving versus staying, preservation of a newer draft, and repeated visits without duplicate creation.
- Chat checks include the exact prompt, provider/model/mode/thinking settings, working directory, and uploaded file bytes. Terminal checks include the submitted command's real output.
- A draft-ownership regression fails before the repair and passes after it. Focused unit coverage: 14 tests passed across screen presence, background handoff, and the existing workspace draft tab.
- Build server, build app dependencies, full typecheck, lint, and format passed.

The recordings exercise Chromium and the real Expo app with isolated Linux daemons. Workspace creation, git worktrees, uploads, and shell terminals are real; the workspace response is held to control the race. Agents use the built-in simulated provider, which echoes only text. Attachment preservation is verified at the request/upload boundary. No live external provider or native/Electron platform was independently tested.

Risk surface: cross-platform creation navigation and draft cleanup. No protocol fields or server behavior change.


boudra:
Closing this duplicate: the verified changes were successfully pushed to Christoph Leiter’s original PR #2986, which remains the delivery PR. QA recordings and regression evidence: https://github.com/getpaseo/paseo/tree/qa/workspace-focus-2986. My earlier conclusion that the contributor branch could not be updated was incorrect.

Issue #4231, opened 2026-09-02
bug: finish notification lost after error->resume cycle via send_agent_prompt; attentionReason stuck at finished


cmulittlechild:
## Summary

When a child agent created with `notifyOnFinish: true` recovers from a provider error via an in-place `paseo_send_agent_prompt` resume, the finish notification for the eventual successful completion is never delivered to the parent. In one of two observed cases the child is additionally left with a permanently stuck `requiresAttention: true, attentionReason: "finished"`.

## Environment

- Paseo daemon on Linux, driven over MCP agent tools
- Parent agent creates child via MCP `paseo_create_agent` with `notifyOnFinish: true`
- Child provider: `opencode` (model `bjskcz/glm-5.3`)
- (Exact daemon version not captured in our notes; daemon logs available on request)

## Working lifecycle (verified multiple times)

1. Parent creates child with `notifyOnFinish: true`.
2. Child runs to completion without incident.
3. Parent receives the `<paseo-system> Agent ... finished.` notification with the agent response. Works reliably.

## Broken lifecycle

1. Child agent hits an upstream provider 503 (`APIError: Service temporarily unavailable, isRetryable: true`) and transitions to `error` state.
2. Parent sends a continue instruction to the **same agentId** via `paseo_send_agent_prompt`. The child returns to `running` and completes all of its work successfully (disk artifacts and logs complete, `get_agent_activity` timeline fully recorded).
3. Child finally transitions to `idle`, but the parent receives **no finish notification at all**.
4. Of two observed instances:
   - Child A: status permanently stuck at `requiresAttention: true, attentionReason: "finished"`, `attentionTimestamp: <completion time>` — the attention flag never clears.
   - Child B: attention flag cleared, but still no message delivered to the parent.

## Impact

Orchestration loops that rely on finish notifications (Brain–Worker supervision patterns) stall silently and must fall back to polling `get_agent_status` / `get_agent_activity`, which the notification system is supposed to make unnecessary.

## Suspected cause

The `notifyOnFinish` binding registered at `create_agent` time does not survive the error → resume cycle: the resumed run's completion either has no registered callback or the one-shot latch was already consumed by the `errored` lifecycle event (note the fix shape proposed in #3049 was "only latch on finished/errored" — if the error transition consumes the latch, a later successful finish after in-place resume would be silent, which matches what we see).

## Distinction from related issues

[...]

PR #4655, opened 2026-09-10
fix(server): keep the finish-notification watcher alive through a transient agent error

thomasvan:
Fixes the failure reported in #4231.

When a caller creates or prompts an agent with `notifyOnFinish: true`, the child-state subscription in `setupFinishNotification` maps `lifecycle === "error"` to `notifySafely("errored")`. `notifySafely` treats any call without `terminal: false` as terminal and calls `stop()`, which unsubscribes the watcher. A transient error therefore ends the watch permanently. When the same run then recovers — a provider 503, a resumed turn, an interrupted start that self-heals seconds later — the work completes and the caller is never told. The same path runs on the initial snapshot, so a watcher installed while the child is already in `error` from an earlier turn can spend its notification before the caller's own turn has begun.

#4231 reports this on the `opencode` provider and names the cause: the notification latch is consumed by the `errored` lifecycle event, so a later successful finish has no registered callback.

**What this changes.** An `error` is reported without ending the watch. Because a failed run must never be reported as that run *finishing*, the error report clears the run gate: only a genuinely fresh run can produce the next `finished`. The error report is spent per run and refreshed when a new run is observed, so a child that sits in error or flaps costs the caller one notification per run rather than one per event. Both error paths — the event handler and the initial snapshot — go through it.

**What this does not change.** The first `finished` still disarms the watcher. Every existing caller still receives exactly one `finished`, ever, and no default moves. This is deliberately not the re-arming behaviour proposed in #3455 and closed as a feature; nothing here asks for that discussion. Permanent disarm on `was closed` and on caller archival is untouched.

The tool schemas already promise notification when the created agent "finishes, errors, or needs permission" — a conjunction of three classes, not a single notification. This restores that for the case where the error is not the final outcome.

Tests cover: a transient error keeps the watch armed for the next run's finish; an error not followed by a fresh run never delivers a `finished`; a flapping child does not deliver a notification per event; a close after an error still disarms permanently; an initial-snapshot error behaves as the event-handler error does; and the first-finish disarm is unchanged.

`npx vitest run packages/server/src/server/agent/agent-prompt.test.ts` passes 20/20 on this branch. I have not run the full suite or CI.

### DIFF
--- a/packages/server/src/server/agent/agent-prompt.test.ts
+++ b/packages/server/src/server/agent/agent-prompt.test.ts
@@ -51,6 +51,8 @@ interface FinishNotificationScenarioOptions {
   childParentAgentId?: string | null;
   requireParentOwnership?: boolean;
   parentPromptError?: Error;
+  /** Lifecycle the child already holds when the watch is set up. */
+  initialChildLifecycle?: "idle" | "running" | "error";
   logger?: Logger;
 }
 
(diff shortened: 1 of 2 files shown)


thomasvan:
Withdrawing this: on our own measurement the change delivers two `finished` notifications for a single completion in the resume path, so the "exactly one `finished`" claim in the description above does not hold. Closing rather than leaving a draft that argues something untrue. Apologies for the noise.

PR #3826, opened 2026-08-24
Make Explorer a first-class pane host

boudra:
### Linked issue

Closes #2128
Closes #3096
Closes #3609

### Type of change

- [x] Bug fix
- [x] New feature
- [x] Enhancement
- [x] Refactor
- [x] Docs

### Reasoning

Files and Changes need a stable navigation surface while content opens into ordinary panes. The previous Side panel mixed those responsibilities with the main split tree, which made routing, drag/drop, resizing, window chrome, and keyboard behavior depend on shell-specific branches.

This replaces that surface with an Explorer sidebar backed by the same narrow panel contract as workspace panes. Panels declare supported hosts; hosts own their chrome, placement, resizing, and lifecycle. Files and Changes remain fixed Explorer views, while files, diffs, pull requests, terminals, and agents can use ordinary panes according to their manifest and source-specific layout preference. Wide native tablets retain the established resizable, combined Explorer dock without taking on the desktop split and drag system.

### Goals

- Make `Cmd/Ctrl+E` toggle a dedicated Explorer sidebar with Files and Changes by default.
- Keep the desktop Explorer outside the ordinary split tree while preserving its tab state and width.
- Keep phone navigation compact and restore the resizable combined Explorer dock on tablets.
- Normalize legacy empty Explorer state to Files and Changes instead of exposing a New Tab launcher.
- Route tree selections into main or side panes without leaking Explorer identity into panels.
- Split Changes into a tree surface and a diff surface, with coherent toolbars for tree, diff, and compact combined presentations.
- Support source-specific “open in main / open to the side” preferences and explicit “Open to the Side” actions.
- Share pane-host contracts, tab manifests, resize behavior, header controls, and toolbar chrome.
- Render native-style desktop window controls from one explicit macOS/Windows/Linux chrome mode, including a development override.
- Preserve mounted pane content during side-pane creation, tab replacement, and pane maximization.

### Non-goals

- Porting the desktop Explorer shell, pane splitting, or tab dragging to native tablets.
- Changing the phone Explorer’s full-screen, close-on-selection behavior.
- Making Explorer-only trees draggable into ordinary panes.
- Redesigning plugin panel APIs beyond the host capability contract required here.

### Supersedes

- #3567 — window controls are implemented through the shared desktop chrome mode and header ownership in this PR.
- #3536 — native control space is resolved by the same chrome presentation contract instead of a separate measurement path.
- #3393 — macOS traffic-light alignment and drag-region behavior are included in the unified titlebar implementation.

### QA

[...]

Issue #3806, opened 2026-08-24
bug: Desktop crashes reconciling a persisted layout with no visible pane


jezifm:
### What's broken

Paseo Desktop repeatedly opens its error boundary with `Error: Invariant failed` when restoring one workspace. Reloading or relaunching returns to the same error. The expected behavior is for the workspace to open with a usable pane.

The persisted layout for the affected workspace had this synthetic equivalent: one empty pane, that pane marked hidden and registered as the Side panel, while `focusedPaneId` referenced a missing `main` pane.

```json
{
  "layoutByWorkspace": {
    "server:workspace": {
      "root": {
        "kind": "pane",
        "pane": {
          "id": "pane-side-panel",
          "tabIds": [],
          "focusedTabId": null,
          "tabs": [],
          "hidden": true
        }
      },
      "focusedPaneId": "main"
    }
  },
  "explorerPaneIdByWorkspace": {
    "server:workspace": "pane-side-panel"
  }
}
```

No daemon health or connectivity failure was present.

### Steps to reproduce

1. Persist the synthetic layout above in the version 1 `workspace-layout-state` envelope.
2. Reload the Desktop renderer.
3. Let agent and terminal snapshots finish hydrating for an empty workspace.
4. Observe the invariant failure during tab reconciliation.

The original sequence that produced the invalid saved topology is not yet known. The saved state reproduces the startup failure consistently.

### Where did this happen

Desktop (Electron)

### Paseo version

0.5.1. The same persisted fixture also fails against `main` at `7db654aa` before the proposed regression fix.

### OS version

macOS 26.0.1 (25A362)

### Agent provider

Not relevant

### Logs

```text
Error: Invariant failed
    at n (paseo://app/_expo/static/js/web/index-1b3630bc4f2de9f19ff4ece083bfcac0.js:15254:220)
    at J (paseo://app/_expo/static/js/web/index-1b3630bc4f2de9f19ff4ece083bfcac0.js:15253:16640)
    at lt (paseo://app/_expo/static/js/web/index-1b3630bc4f2de9f19ff4ece083bfcac0.js:15253:21096)
    at pt (paseo://app/_expo/static/js/web/index-1b3630bc4f2de9f19ff4ece083bfcac0.js:15253:21944)
    at Nt (paseo://app/_expo/static/js/web/index-1b3630bc4f2de9f19ff4ece083bfcac0.js:15253:27777)
[...]

Issue #185, opened 2026-04-03
Custom opencode agents doesnt show


peyloride:
<img width="222" height="185" alt="Image" src="https://github.com/user-attachments/assets/1ce564e6-83da-44de-a8ca-ff7856e2340b" />

I have many agents defined in opencode (for example I have oh my opencode slim installed) but none of these non-standard agents pop-up.

PR #4682, opened 2026-09-10
fix: keep the last visible pane when its final tab closes

### LINKED ISSUE #4677: bug: closing the last tab of a split pane still persists a layout with no visible pane (React #185 in reconcileTabs, 0.8.0)
### What's broken

Closing a tab still persists a layout whose only remaining pane is hidden, and the
next render loop fills that pane with draft tabs until React dies with #185.

This is the "separate unresolved follow-up" referred to when #3806 was closed. The
persisted-state healing from #3826/#3861 is in 0.8.0, but the invalid topology is
still *written*, so the app produces it fresh at runtime rather than inheriting it.

What changed since #3806 is which pane survives. #3826 renamed `SIDE_PANEL_PANE_ID`
to `EXPLORER_SIDEBAR_PANE_ID` and the preserve rule kept following the constant:

https://github.com/getpaseo/paseo/blob/d7c7044dfc91d1d18721dc8757ac3bb913d8c232/packages/app/src/stores/workspace-layout-actions.ts#L1474-L1476

```ts
const preserveEmptyPaneId =
  input.preserveEmptyPaneId ??
  (pane.id === DEFAULT_PANE_ID || pane.id === EXPLORER_SIDEBAR_PANE_ID ? pane.id : null);
```

A pane that came out of a split carries a generated `pane_<uuid>` id, matches neither
constant, and `detachTabFromTree` drops it when its last tab goes. Previously the
collapsed side panel was left behind; now it is the hidden Explorer pane. Same root
cause, new survivor.

### What the data shows

App started 20:44:22, crashed 20:46:23, both on 0.8.0 — the state was written during
this session, not restored from an older one.

Persisted `workspace-layout-state` (envelope version 2) for the affected workspace,
IDs redacted, tab list truncated:

```json
{
  "layoutByWorkspace": {
    "server:workspace": {
      "root": {
        "kind": "pane",
        "pane": {
          "id": "explorer",
          "hidden": true,
          "focusedTabId": "draft_msg_1789065983486_7ocbm691u",
          "tabIds": ["files", "changes_tree", "draft_msg_1789065983242_…", "… 47 more drafts"],
          "tabs": [
            { "tabId": "files",         "target": { "kind": "files" },         "createdAt": 1788874898745 },
            { "tabId": "changes_tree",  "target": { "kind": "changes_tree" },  "createdAt": 1788874898745 },
            { "tabId": "draft_msg_1789065983486_7ocbm691u",
              "target": { "kind": "draft", "draftId": "draft_msg_1789065983486_7ocbm691u" },
              "createdAt": 1789065983486 }
          ]
        }
      },
      "focusedPaneId": null
    }
  },
  "explorerPaneIdByWorkspace": { "server:workspace": "explorer" }
}
```

Three things in there worth noting:

- The only pane is `hidden: true`, so the workspace has **zero visible panes**.
- `focusedPaneId` is `null` (in #3806 it was a dangling `"main"`).
- **49 draft tabs, all created within 244 ms** — first `createdAt` 1789065983242
  (2026-09-10T18:46:23.242Z), last 1789065983486 (…:23.486Z). That is the render loop
  writing state, not a user opening tabs.

`splitSizesByWorkspace` still holds `group_<uuid>` entries for this workspace, i.e. it
had been through a split before the crash — consistent with the surviving pane having
carried a generated id.

### Steps to reproduce

I did not catch the exact click, so treat the sequence as inferred from the persisted
state rather than confirmed. What the state supports:

1. Open a workspace and split a pane (the new pane gets a generated `pane_<uuid>` id).
2. Close the tabs in the generated pane until the last one is closed.
3. The generated pane is dropped; only the hidden Explorer pane remains.
4. The layout persists with no visible pane and `focusedPaneId: null`.
5. On the next reconcile, draft tabs are seeded into the hidden pane in a loop until
   React error #185.

#3806 has an organic report of the same shape from a different user on 0.7.2 (open a
workspace with one visible tab, close it, error boundary immediately), which suggests
the split is not strictly required — any pane whose id is neither `main` nor `explorer`
should do.

### Frequency

Fourth occurrence on this install: 2026-08-31, 2026-09-03 and 2026-09-10 with React
#185, plus the invariant-failure variant on 2026-08-24/25 that became #3806. Different
workspaces each time.

### Suggested fix

Same one-line guard proposed in #3806, never dropping the only pane the user can see:

```ts
const preserveEmptyPaneId =
  input.preserveEmptyPaneId ??
  (pane.id === DEFAULT_PANE_ID ||
   pane.id === EXPLORER_SIDEBAR_PANE_ID ||
   collectAllPanes(internalLayout.root).length === 1
    ? pane.id
    : null);
```

`collectAllPanes` already skips hidden panes. A regression test would close the last
tab of a `pane_<uuid>` pane in a layout whose other pane is a hidden Explorer, and
assert the resulting layout still has a visible pane.

Related: #4103 is the same crash on macOS 0.6.1 and is still open and unanswered.

### Where did this happen

Desktop (Electron)

### Paseo version

0.8.0

### OS version

Windows 11, daemon in WSL2 (Ubuntu)

### Agent provider

Not relevant

### Logs

```text
[2026-09-10 20:44:22.534] [info]  [desktop] app startup { version: '0.8.0', platform: 'win32', arch: 'x64', isPackaged: true }
[2026-09-10 20:44:24.654] [info]  [open-project] renderer requested pending path: { webContentsId: 1, pendingPath: null }
[2026-09-10 20:44:28.366] [info]  [auto-updater] check completed { currentVersion: '0.8.0', targetVersion: '0.8.0', releaseChannel: 'stable', intent: 'automatic', hasUpdate: false, readyToInstall: false, errorMessage: null }
[2026-09-10 20:46:23.534] [error] Error: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
[2026-09-10 20:46:23.534] [error] [RootErrorBoundary] Unhandled render error [object Object]
[2026-09-10 20:49:12.407] [info]  [desktop] app startup { version: '0.8.0', platform: 'win32', arch: 'x64', isPackaged: true }
```

Renderer stack from the error screen:

```text
Name: Error
Message: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Stack:
Error: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Qr (paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:1091:33378)
    at Vr (paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:1091:32951)
    at so (paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:1091:50584)
    at paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:1091:50462
    at paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:15048:361
    at Set.forEach (<anonymous>)
    at o (paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:15048:350)
    at paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:15050:6108
    at reconcileTabs (paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:15200:17819)
    at paseo://app/_expo/static/js/web/index-dc5d33886c66f1acb37bafb267496e5a.js:16215:32172
```

After deleting the affected workspace from `layoutByWorkspace`, `splitSizesByWorkspace`
and `explorerPaneIdByWorkspace`, the app started normally (20:49:12 above, no further
#185). An audit of the same store found 14 other workspaces, all currently healthy.

### Note on how this was investigated

The logs, the persisted state and the timestamps above are raw. The reading of
`closeTabInLayout` is my own from #3806 and unchanged since; I have not run a
debugger against 0.8.0 to confirm the code path executes as described.

L4XB:
### Linked issue

Closes #4677

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

`closeTabInLayout` only preserved an emptied pane when its id was the default pane or the explorer sidebar. A pane born from a split carries a generated `pane_<uuid>` id, so once the default pane had been dropped (its only tab moved into the split), closing the split's last tab removed the split as well and the layout was persisted with the hidden explorer as its only pane. The next render loop filled that hidden pane with draft tabs until React failed with #185. The healing from #3826/#3861 repairs such a layout on load but did not stop it from being written in the first place, which is the follow-up #3806 left open.

The fix treats "the only visible pane" like the default pane: it is kept and falls back to a New tab, mirroring the last-visible-pane gate that `setPaneHiddenInLayout` already applies to hiding. For users this means closing the last tab in the last pane always leaves a usable New tab, and the crash loop cannot be produced at runtime any more.

### Goals

- Closing the final tab of the only visible pane retains that pane with a New tab, regardless of the pane's id.
- A layout whose only pane is hidden is never produced by `closeTab`.
- Existing behaviour is unchanged when another visible pane exists: closing the sole New tab of a split next to a visible default pane still removes the split (existing test "closes a split through its sole New tab but keeps the final visible split" still passes).

### Non-goals

- Changing how tabs are moved between panes (`moveTabToPane` dropping the emptied source pane is untouched).
- Changing the load-time healing from #3826/#3861.

### Validation

```
npm ci && npm run build -w packages/protocol -w packages/plugin ...
cd packages/app && npx vitest run src/stores/workspace-layout-store.test.ts   # 125 passed (2 new)
cd packages/app && npx tsgo --noEmit   # 1 error in src/components/draggable-list.native.tsx, identical on main
```

New tests: a store-level flow that reproduces the report (open agent tab, split right, move the tab into the split so `main` is dropped, close the New tab, close the last tab) and a pure `closeTabInLayout` case on a `{explorer hidden, pane_split}` layout.

### DIFF
--- a/packages/app/src/stores/workspace-layout-actions.ts
+++ b/packages/app/src/stores/workspace-layout-actions.ts
@@ -1471,9 +1471,17 @@ export function closeTabInLayout(input: CloseTabInLayoutInput): WorkspaceLayout 
   if (!pane) {
     return null;
   }
+  // A workspace always has somewhere to look (see setPaneHiddenInLayout). A pane born
+  // from a split carries a generated id, so when it is the only visible pane left,
+  // dropping it with its last tab would persist a layout whose sole pane is the hidden
+  // explorer; keep it, like the default pane, and let it fall back to a New tab.
+  const visiblePaneIds = listPaneIds(internalLayout.root);
+  const isLastVisiblePane = visiblePaneIds.length === 1 && visiblePaneIds[0] === pane.id;
   const preserveEmptyPaneId =
     input.preserveEmptyPaneId ??
-    (pane.id === DEFAULT_PANE_ID || pane.id === EXPLORER_SIDEBAR_PANE_ID ? pane.id : null);
+    (pane.id === DEFAULT_PANE_ID || pane.id === EXPLORER_SIDEBAR_PANE_ID || isLastVisiblePane
+      ? pane.id
+      : null);
 
   const closeSuccessorTabId = getCloseSuccessorTabId({
     pane,
--- a/packages/app/src/stores/workspace-layout-store.test.ts
+++ b/packages/app/src/stores/workspace-layout-store.test.ts
@@ -16,6 +16,7 @@ vi.mock("@react-native-async-storage/async-storage", () => {
 });
 
 import AsyncStorage from "@react-native-async-storage/async-storage";
+import { closeTabInLayout } from "./workspace-layout-actions";
 import { buildWorkspaceTabPersistenceKey, type WorkspaceTab } from "@/workspace-tabs/model";
 import { defaultChangesState, type ChangesState } from "@/panels/changes/state";
 import { defaultFileState, type FileState } from "@/panels/file/state";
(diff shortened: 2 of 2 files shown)


L4XB:
#4844 fixed #4677 on `main` while this was open, and I checked it against this PR's tests rather than assuming. `closeTabInLayout never leaves a layout whose only pane is hidden` passes against `main`'s store unchanged, so the invariant this PR added is enforced there. The other test can't even reach its starting state on `main`: moving `main`'s only tab into a split no longer drops the `main` pane, so the split-born pane never ends up as the last visible one. Nothing is left for this PR to do, so I'm closing it. Thanks for the fix.

PR #4719, opened 2026-09-11
fix(acp): release the turn when an agent never answers a cancelled prompt

IsakPersson:
Relates to #3256 (case A: cancel then retry leaves the agent stuck with an ACP active-turn rejection)

## Problem

After `session/cancel` the spec requires the agent to answer the in-flight `session/prompt`, and compliant agents do so within a couple of seconds. An agent that acknowledges the cancel but never answers leaves `activeForegroundTurnId` set forever, so every later prompt on that session throws:

```
A foreground turn is already active
```

The session is then unusable until it is recreated. `cancelAgentRunNow` force-cancels at the manager level after its 2 second wait, which updates the manager's own bookkeeping, but nothing clears the bridge's turn slot.

Observed in the daemon log, followed by four failed sends:

```
cancelAgentRun: acknowledged turn still active after timeout, force-canceling
handleStreamEvent: turn_failed | err=A foreground turn is already active
```

## Fix

- Arm a 30 second watchdog when cancelling. If the answer never arrives, end the turn locally so the session stays usable.
- Ignore a settle that arrives for an already-ended turn, on both the resolve and reject paths, so the watchdog cannot produce a second terminal event for the same turn.

The timeout is deliberately well past normal settle time, including agents unwinding a long tool call. It exists to unstick a session that would otherwise stay broken forever, not to police slow agents.

## QA evidence

Reproduced against a real third-party ACP agent (Google's Antigravity ACP server, `agy_acp_server` 1.1.1) by cancelling in three situations and measuring when `session/prompt` settled:

| Cancel during | Result |
| --- | --- |
| plain text generation | settles in 2.5s, session keeps working |
| a pending permission request | settles in 2.5s with `stopReason: cancelled` |
| a running shell command | never settles, still pending 35s after the cancel |

The third case is what wedges the session.

Unit tests, `packages/server/src/server/agent/providers/acp-agent.test.ts`:

```
main (untouched):        Tests  102 passed (102)
this branch:             Tests  104 passed (104)
```

Two tests added: one driving a prompt that never settles, asserting the turn is released after the watchdog and that a subsequent `startTurn` succeeds rather than throwing; one asserting a late settle produces no second terminal event.

Checks on this branch:

```
npx oxlint <changed files>      Found 0 warnings and 0 errors.
npx oxfmt --check <changed>     All matched files use the correct format.
npm run typecheck -w @getpaseo/server
                                5 errors, identical to untouched main
                                (speech/providers/openai, terminal/node-pty:
                                 missing optional modules in my environment),
                                0 in acp-agent.ts
```

Tested on macOS 26.4 arm64, Node 26. Not tested on Linux or Windows. This does not address the other cases tracked in #3256 (concurrent send/fan-out, persistence across restart).

## Note

This touches `handlePromptResponse`, which is also touched by my other PR for #1390. Happy to rebase whichever lands second.

### DIFF
--- a/packages/server/src/server/agent/providers/acp-agent.test.ts
+++ b/packages/server/src/server/agent/providers/acp-agent.test.ts
@@ -2800,6 +2800,84 @@ describe("ACPAgentSession", () => {
     expect(asInternals<ACPSessionInternals>(session).activeForegroundTurnId).toBeNull();
   });
 
+  test("a cancel the agent never answers releases the turn instead of wedging the session", async () => {
+    vi.useFakeTimers();
+    try {
+      const session = createSession();
+      const events: AgentStreamEvent[] = [];
+      // An agent that ignores the spec and never settles the cancelled prompt.
+      const prompt = vi.fn(() => new Promise<PromptResponse>(() => {}));
+      const cancel = vi.fn(async () => {});
+
+      asInternals<ACPSessionInternals>(session).sessionId = "session-1";
+      asInternals<ACPSessionInternals>(session).connection = { prompt, cancel } as never;
+      session.subscribe((event) => {
+        events.push(event);
+      });
+
+      const { turnId } = await session.startTurn("hello");
+      await session.interrupt();
+
+      expect(cancel).toHaveBeenCalledOnce();
+      expect(asInternals<ACPSessionInternals>(session).activeForegroundTurnId).toBe(turnId);
+
+      await vi.advanceTimersByTimeAsync(30_000);
+
+      expect(events.find((event) => event.type === "turn_canceled")).toMatchObject({
+        type: "turn_canceled",
+        turnId,
+      });
+      expect(asInternals<ACPSessionInternals>(session).activeForegroundTurnId).toBeNull();
+
+      // The session is usable again rather than stuck on "A foreground turn is already active".
+      await expect(session.startTurn("second")).resolves.toMatchObject({
+        turnId: expect.any(String),
+      });
+    } finally {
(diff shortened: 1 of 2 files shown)


IsakPersson:
Thanks, all three were fair. Addressed in 70ffcfd.

**P1, cancel can bypass watchdog.** Correct, and the more serious of the two. The watchdog is now armed before the `session/cancel` request rather than after it, so an agent that leaves the cancel itself pending no longer keeps the turn slot occupied. Normal settlement clears the timer on either path.

**P2, late response corrupts usage.** Also correct. `handlePromptResponse` now returns before touching any state, so an abandoned turn's token counts can no longer be attributed to the turn that replaced it.

**P2, test depends on internals.** Removed. The cancellation event and the successful second `startTurn` already prove the caller-visible behaviour, so the private field reads were redundant.

Two regression tests added for the first two, and I checked they actually fail against the previous commit rather than just passing against the new one:

```
# tests from this commit, source reverted to the previous commit
× a cancel request that never settles still releases the turn
× a late response cannot attribute its usage to the turn that replaced it
  Tests  2 failed | 104 passed (106)

# with the fix
  Tests  106 passed (106)
```

Checks on the updated branch:

```
npx oxlint <changed files>              Found 0 warnings and 0 errors.
npx oxfmt --check <changed files>       All matched files use the correct format.
npm run typecheck -w @getpaseo/server   0 errors
```

Also dropped an unrelated `package-lock.json` change that my local install had pulled into the branch. The diff is now the two source files only.

Issue #4673, opened 2026-09-10
bug:  Timeline transformers only receive the latest tool call in Overview mode


mcowger:
### What's broken

When Paseo's Tool call detail setting is set to Overview, consecutive tool calls are grouped before plugin timeline transformers run.

A plugin that registers a tool_call transformer only receives the latest call in the group. The earlier calls are no longer available to the plugin, so a custom renderer cannot display one row per tool call.

The latest call can appear to replace the previous row rather than being inserted below it.

### Steps to reproduce

1. Install a plugin that registers a timeline transformer for tool_call items.
2. Have the transformer render the tool name, call ID, and details.
3. Set Tool call detail to Overview.
4. Run an agent that makes several consecutive tool calls, for example:
bash
bash
read
edit

5. Inspect the rows received by the plugin transformer.

## Expected behavior
A plugin that owns tool_call rendering should be able to receive each underlying tool call independently, including:
* Stable call ID
* Tool name
* Detail payload
* Status updates
* Error state
* Metadata
* Streaming-to-complete updates

Native Overview grouping should only apply when no plugin claims the individual tool calls, or the plugin API should expose the full grouped call list so the plugin can render it itself.

### Where did this happen

Web (browser)

### Paseo version

0.8.0

### Agent provider

None

### Logs

```text

```

PR #4820, opened 2026-09-14
fix(app): transform promoted assistant block groups as one message

### LINKED ISSUE #4789: bug(plugins): assistant_message timeline transformers receive promoted markdown blocks, not the message, in 0.8.0
### What's broken

A plugin timeline transformer registered for `assistant_message` no longer sees the whole message in 0.8.0. While a message streams, `promoteCompletedAssistantBlocks` (`packages/app/src/types/stream.ts`) splits the active assistant row into markdown blocks via `splitMarkdownBlocks` and promotes each completed block to its own stream item with a shared `blockGroupId`. `projectPluginTimelineItems` (`packages/app/src/plugins/timeline/projection.ts`) then runs transformers per stream item, so the transformer is called once per block with `item.text` set to that block's text alone.

The same transformer on 0.8.0-beta.1 received the complete message, so a plugin written against the documented contract ("Returning `items` replaces it") behaves differently between the two releases with no API change.

Concrete case: omp injects a `/skill:<name>` expansion as one custom message whose first line is `[IMPORTANT: User invoked the "<name>" skill; follow its instructions. Full skill below.]` followed by the SKILL.md body. A transformer that folds that row into a card gets:

- block 1: the marker line only, so the card renders with an empty body
- blocks 2..n: the skill body, rendered by Paseo as ordinary markdown below the card

The daemon emits the row as a single `assistant_message` timeline item. Captured through `@getpaseo/client` on 0.8.0 against a fresh omp agent: one `timeline` event, `assistant_message`, 1784 chars, marker on line 1. `agent.timeline.refetch()` on an older agent likewise returns the whole text as one entry. Only the app's render model is split.

This is the same class of problem as #4673 (tool-call transformers receive the Overview group instead of each call) and the same machinery as #4699 (streamed chunks rendering as separate blocks): transformers get a projected render row rather than the source timeline item.

### Steps to reproduce

1. Install a plugin with this client entry:

```ts
export default function contribute(client: PluginClientContext) {
  client.addTimelineTransformer({
    id: "probe",
    query: { itemType: "assistant_message" },
    transform: ({ item }) => ({
      items: [{ type: "plugin", kind: "probe", version: 1, data: { text: item.text } }],
    }),
  });
  client.addTimelineRenderer({
    kind: "probe",
    version: 1,
    schema: z.object({ text: z.string() }),
    Component: ({ item, theme }) => (
      <Text style={{ color: theme.colors.foreground }}>{`[${item.data.text.length} chars] ${item.data.text.slice(0, 40)}`}</Text>
    ),
  });
  return () => {};
}
```

2. Start any agent and send a prompt whose reply has several paragraphs (or, with omp, send `/skill:<any installed skill>`).
3. Observe one probe row per paragraph, each with that paragraph's length, instead of one row with the full text.
4. Reload the agent: the timeline refetch delivers whole rows and the same transformer now produces one probe row for the whole message.

Expected: the transformer receives the complete `assistant_message` once per source item, and block promotion, if it stays, happens after transformation or only for rows no transformer claimed.

### Where did this happen

Desktop (Electron)

### Paseo version

0.8.0 (works as expected on 0.8.0-beta.1)

### OS version

macOS 26.4 arm64

### Agent provider

omp

### Provider configuration

omp 18.1.19, `openai-codex/gpt-5.6-sol`; also reproduces with the paragraphs of an ordinary Claude reply, so it is not provider-specific.

### Logs

No daemon errors; the daemon side is correct. Client-side the transformer is simply invoked per block. Happy to add a screen recording if useful.

jegork:
Closes #4789

## Problem

Since 0.8.0, a plugin timeline transformer registered for `assistant_message` is called once per promoted markdown block while a message streams, with `item.text` set to that block's text. On 0.8.0-beta.1 the same transformer received the whole message. A plugin that folds a message into a card therefore gets a header-only card for the first block and the rest of the message rendered as plain markdown below it.

## Cause

`promoteCompletedAssistantBlocks` (`packages/app/src/types/stream.ts`) splits the active assistant row into block items that share a `blockGroupId`. `projectPluginTimelineItems` (`packages/app/src/plugins/timeline/projection.ts`) then ran `transformSourceItem` per stream item, so the transformer saw the render-model blocks instead of the source row the daemon emitted.

## Fix

Projection now groups consecutive `assistant_message` stream items that share a `blockGroupId` before transforming:

- The transformer is called once per group with `text` set to the blocks joined by `\n\n` (the separator `splitMarkdownBlocks` drops), `sourceId` set to the group's first block id, and the same `phase` rule as before (`"complete"` for assistant rows).
- If it returns items, they replace the whole group. If it returns `undefined`, the block items are returned by reference, so native rendering and the reducer's later reconciliation are unaffected.
- The per-item memoization is kept: a group is cached by its last block reference and validated member-wise, so an unchanged group does not re-run the transformer or remount its renderer. Non-grouped items go through the same path with a single member and behave exactly as before.

No changes to the reducer, block promotion, `view.tsx`, or the plugin SDK contract.

### Known limitation

`head` and `tail` are projected separately in `agent-stream/view.tsx`, and block promotion flushes completed blocks to `tail` while the live block stays in `head`. While a row is still streaming, the transformer therefore sees two groups: the completed blocks joined (from `tail`) and the live block alone (from `head`). As soon as the head is flushed (turn end, another lane starts, or reconciliation with the canonical row) the whole message is transformed once. Fixing the in-flight case needs projection to see head and tail together, which is a `view.tsx` change and out of scope here.

For the same reason the projection cannot tell a flushed group from a streaming one, so `phase` for assistant rows stays `"complete"`, matching the documented contract ("streaming" only for running tool calls and loading reasoning).

## Tests

New cases in `packages/app/src/plugins/timeline/projection.test.ts`:

- a 3-block group is transformed once with the full text and replaced by one plugin item
- a group left alone by the transformer is returned untouched (same references)
- a group whose last block still carries its streaming newline suffix keeps it in the joined text
- a lone live block is projected by its own id
- interleaving with user, reasoning, a second group, and a plain assistant row keeps order and separate calls
- an empty result removes every block of the group
- an unchanged group is memoized; promoting a new block re-runs the transformer once with the grown text

Five of the seven fail on the previous `projection.ts` (the lone-block and removal cases pass on both).

```
npm run test -w packages/app -- src/plugins/timeline   # 3 files, 24 tests passed
npm run typecheck -w packages/app                        # clean
npm run lint -- packages/app/src/plugins/timeline/projection.ts packages/app/src/plugins/timeline/projection.test.ts   # 0 warnings, 0 errors
npm run format:files -- <same two files>                 # no changes
```

## QA

- Exercised: unit tests above, typecheck, lint, format, pre-commit hooks (lint, format check, full workspace typecheck).
- Not yet exercised: the running app. Desktop (Electron) screenshots/recording of the probe plugin from the issue before and after are pending and will be added to this PR.
- Platforms not tested: iOS, Android, web browser. The change is in shared projection code with no platform-specific paths.


boudra:
Thanks @jegork for the report, reproduction, and proposed fix. I consolidated #4789 and #4673 in #4675 and credited you as a co-author in cfa26eb33. The combined implementation invokes transformers on whole accumulated messages before Markdown promotion, preserving streaming without reconstructing text from display blocks. Closing this PR as superseded by #4675; both issues remain linked there.

PR #4886, opened 2026-09-15
fix(app): keep composer tracks clear of timeline footers

### LINKED ISSUE #4885: bug: composer plugin pill overlaps timeline footer until keyboard opens
## What's broken

A composer plugin pill can overlap the final timeline footer on a compact web client. The composer track is added asynchronously after the stream's initial route anchor, but the memoized `AgentStreamView` ignores changes to its bottom-overlay clearance props. Even after the spacer renders, the viewport can remain one spacer-height short of the true bottom until another viewport change occurs.

Expected: the completed-turn footer remains fully above the composer track as soon as the plugin pill appears.

Actual: the pill covers the copy/fork/duration footer while the composer is unfocused. Focusing the composer and opening the iOS keyboard re-anchors the viewport and clears the overlap.

## Steps to reproduce

1. Run Paseo 0.8.0 with a client plugin that contributes an agent composer pill (reproduced with `paseo-defer`).
2. Open an agent containing a completed assistant response and its turn footer.
3. Fully close and reopen the native Paseo iOS app.
4. Wait for the plugin contribution to register.
5. Observe the composer pill overlapping the final turn footer.
6. Focus the composer so the iOS keyboard opens.
7. Observe that the viewport resize moves the footer above the pill.

## Where did this happen

Native Paseo iOS app.

## Paseo version

0.8.0

## Agent provider

Not relevant. The problem is in shared timeline/composer layout.

## Investigation evidence

The following two behaviors were independently reproduced:

- Adding `bottomOverlayTailClearance` and `bottomOverlayControlClearance` to the `AgentStreamView` memo comparison makes the late plugin registration render its intended spacer.
- Re-anchoring once when bottom-overlay clearance transitions from zero to positive makes that new protected tail visible without requiring a keyboard/viewport resize.

The re-anchor is skipped for detached historical timelines and does not repeat during ordinary scrolling.

## Screenshots or video

Unfocused and keyboard-focused screenshots were captured in the affected native iOS app and are attached in a follow-up comment.

sholiday:
## Summary

Fixes #4885.

- invalidate the memoized agent stream when composer-overlay clearance changes
- re-anchor once when a bottom composer track first appears
- leave detached historical timelines and ordinary scrolling untouched
- cover both layout decisions with focused regression tests

## Problem

Composer plugin contributions register asynchronously. The surrounding panel correctly recalculates the space required by the floating composer track, but `AgentStreamView` previously ignored those props in its memo comparator. Once the spacer was rendered, a compact web viewport could also remain one spacer-height short of the new bottom until the iOS keyboard caused a viewport resize.

## QA

Manual:

- Reproduced in the native Paseo iOS app connected to Paseo 0.8.0.
- Before: the Defer composer pill overlapped the copy/fork/duration footer while the composer was unfocused.
- After: the footer remains above the pill both unfocused and while typing.
- Confirmed the implementation skips the one-time anchor for detached historical timelines and viewports no longer following output.

Automated:

```text
npx vitest run packages/app/src/agent-stream/bottom-overlay-inset.test.ts packages/app/src/agent-stream/strategy-web.test.tsx packages/app/src/agent-stream/chat-outline/use-chat-outline.test.tsx --bail=1
Test Files  3 passed (3)
Tests       26 passed (26)

npm run lint -- <seven changed app files>

npm run build:app-deps
passed

npm run typecheck --workspace=@getpaseo/app
passed

npm run format
Finished on 4522 files.

pre-commit hook
format: passed
lint: passed
typecheck across all workspaces: passed
```

## Platforms

- Verified: native Paseo app on iOS
- Not manually verified: Android, desktop browser, Electron


sholiday:
Addressed in `0645bf1b`.

`StreamViewportHandle` now exposes the renderer's actual follow state: `followOutputRef` on web and the bottom-anchor controller's `sticky-bottom` mode on native. The delayed overlay anchor re-checks that state inside the animation-frame callback, so a user who has scrolled away before the callback is not moved. The focused regression coverage now includes the non-following case.

sholiday:
Addressed the object-parameter requirement in `5f8f5fbb`; the helper and all call sites now use a named input object. The focused 26-test set, lint, app typecheck, formatting, and full pre-commit hook all pass.

PR #4901, opened 2026-09-15
fix(app): retry terminal create after reconnect

ddvnguyen:
## Problem

Opening a new terminal right after idle (or a daemon restart) sometimes leaves a dead terminal tab: the create RPC goes out on a socket that looks connected but is already stale (idle cull, restart), fails, the failure is reported once via `onTerminalCreateFailed`, and nothing replays the user's intent when the connection heals. The user has to close the dead tab and retry by hand.

Server side was ruled out: PTY spawn holds 20/20, worker-IPC create holds 10/10 concurrent, and PROD logs show zero snapshot/worker failures around the drops. The gap is purely client-side recovery.

## Decision

`useWorkspaceTerminals` now holds a failed create input and replays it exactly once on the next disconnect→reconnect transition:

- No reconnect after the failure means a genuine error (bad cwd, unknown profile) — no retry.
- Any successful create, in-flight mutation, pending input, or fresh explicit `createTerminal` call supersedes the held failure, so a later reconnect can never spawn a surprise duplicate terminal.
- An immediate retry was deliberately not chosen: it would hit the same dead socket before liveness teardown.

## Check

- New `use-workspace-terminals.test.tsx`: replay-on-reconnect fires exactly once, no-retry without a drop, supersede by fresh intent, no duplicate on later reconnects — 3/3 pass.
- Adjacent suites (`state.test.ts`, `terminal-stream-controller.test.ts`) pass: 13/13.
- `oxlint` clean, `oxfmt` clean, app `tsc` shows only the 2 pre-existing errors also present on unmodified `upstream/main` (`workspace-labels/index.ts`, `workspace/mark-unread.ts`).

## Risk

Low: app-only change, no daemon/protocol impact. Worst case of a misfire is one duplicate terminal create on reconnect; the supersede rules above close that path and it is covered by test.


ddvnguyen:
Closing in favor of review at our fork; will upstream separately once accepted there.

PR #4911, opened 2026-09-15
feat(app): add recent-activity grouping to the sidebar

mayinghan:
## What

Adds a third grouping mode — **Recent activity** — alongside **Project** and **Status** in the sidebar's display preferences. It renders workspaces as one flat, cross-project list ordered most-recently-active first, so the session you touched last is always at the top.

## How

- `WorkspaceAgentActivity` gains `lastActivityAt`: the latest `lastActivityAt` among a workspace's root agents, tracked in the same index pass as the status bucket. It deliberately reads `lastActivityAt` rather than `attentionTimestamp ?? updatedAt` — recency should move on real activity, not stream ticks. A side effect is that the activity index (and every memoized subscriber) stays stable while an agent streams, since recency no longer churns on `updatedAt`.
- `SidebarWorkspaceEntry` threads it through; a new `sortSidebarWorkspacesByRecency` helper orders most-recent first, sinks workspaces with no recorded activity to the end in baseline order, and breaks ties by baseline position so the list doesn't reshuffle.
- The projection's `recentActivity` case emits a single flat group (no header label — the ordering carries the meaning), so the existing grouped-list rendering, pinned section, and keyboard shortcuts all work unchanged. Group headers gained a nullable leading visual for that.
- Persisted via `sidebar-view-store` with the group-mode schema extended and migration updated; menu option + icons added to display preferences; all 9 locales translated.

## Tests

- `workspace-agent-activity.test.ts`: recency tracked across root agents, index rebuild only when recency moves
- `sidebar-workspaces-view-model.test.ts`: `sortSidebarWorkspacesByRecency` ordering, no-activity sink, tie stability
- `sidebar-projection.test.ts`: recent-activity mode ordering, pinned chats excluded and kept above, project-icon coverage across all three modes

`npm run typecheck`, `npm run lint`, `npm run format:check` clean; targeted vitest suites (237 tests) green.


mayinghan:
cc @boudra — new sidebar grouping mode, would appreciate a review when you have a moment.

PR #4916, opened 2026-09-15
fix(app): keep measuring the autocomplete anchor until it is laid out

### LINKED ISSUE #4872: bug(app): slash autocomplete popover never renders on desktop in 0.8.0; Tab still completes, resize makes it appear (#4275 measurement lifecycle)
### What's broken

On Paseo Desktop 0.8.0 (macOS), the composer slash autocomplete popover never renders. The data is there: typing `/` then Tab completes the top-ranked command, Arrow keys move an invisible selection, but no popover is drawn. Worked on 0.7.2.

Resizing the app window makes the popover appear immediately, which points at the anchor-measurement lifecycle rather than the command list.

### Root cause

#4275 ("Smooth mobile keyboard transitions", in 0.8.0) rewrote the measurement effect in `packages/app/src/components/ui/autocomplete-popover.tsx`:

- 0.7.2 effect deps: `[visible, options.length, selectedIndex, anchorRef, portalHostName, measuredShift, shift, windowDimensions.width, windowDimensions.height]`, plus `Keyboard.addListener('keyboardDidShow'|'keyboardDidHide')` re-measure as a fallback.
- 0.8.0 effect deps: `[canMeasure, remeasure, windowDimensions.width, windowDimensions.height]` where `canMeasure = visible && (options.length === 0 || selectedIndex >= 0)` is a boolean. The keyboard listeners were replaced by a `useAnimatedReaction` on `isMoving`, which never fires on desktop (no keyboard motion).
- `remeasure()` silently returns when `measureFloatingPanelPortalHost()` yields `null` (`if (... || !hostRect) return`).

So on desktop, if the first measurement runs before the portal host / anchor has laid out, `relativeAnchorRect` stays `null`, the component returns `null`, and nothing ever re-triggers measurement except a window resize. `options` and `selectedIndex` are fully populated, hence Tab still completing.

### Steps to reproduce

1. Paseo Desktop 0.8.0, macOS, desktop-managed daemon 0.8.0.
2. Open a running agent (seen with the omp provider; the popover code is provider-agnostic).
3. Click the empty composer, type `/`. No popover.
4. Press Tab: top command is inserted. Or resize the window: popover appears.

### Fix that works locally

Patched the packaged renderer bundle with the equivalent of these two hunks and verified the popover renders on `/` without resizing:

```ts
// remeasure(): retry instead of dropping when host/anchor aren't laid out yet
.then(([anchorRect, hostRect]) => {
  if (generation !== measurementGeneration.current) return;
  if (!hostRect || !anchorRect.width) {
    setTimeout(remeasure, 50);
    return;
  }
  setRelativeAnchorRect({ ... });
  measuredShift.value = shift.value;
});

// effect deps: re-measure when options/selection change, as 0.7.2 did
}, [canMeasure, remeasure, windowDimensions.width, windowDimensions.height, options.length, selectedIndex]);
```

Either hunk alone may be enough; both together restore the 0.7.2 behaviour on desktop while keeping the `isMoving` reaction for mobile.

### Environment

- Desktop (Electron) 0.8.0, macOS 15 arm64
- Daemon 0.8.0, desktop-managed
- Provider: omp (daemon `list_commands_response` returns 158 commands for the agent, so this is not #4759's empty-palette race, and not #3074 since it fails on an empty composer)

L4XB:
### Linked issue

Closes #4872

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Docs

### Reasoning

Typing `/` in the composer on desktop shows no popover. The command list is there — Tab
completes the top command and the arrow keys move a selection nobody can see — and resizing
the window makes the popover appear at once. The reporter root-caused this to the measurement
lifecycle in `autocomplete-popover.tsx`, and that read holds up.

The popover only draws once it has measured its anchor against the floating-panel portal host.
Two things decide when that measurement happens:

- The effect's dependencies are `[canMeasure, remeasure, windowDimensions.width,
  windowDimensions.height]` (`autocomplete-popover.tsx:122`). `canMeasure` is a boolean
  (`:85`), and `remeasure` is a `useCallback` whose only unstable dependency is that same
  boolean, so in practice the effect runs when `canMeasure` flips or the window resizes.
  Nothing else.
- Each run measures twice, once immediately and once on the next frame.

`remeasure` then throws the result away when the host is not measurable
(`if (generation !== measurementGeneration.current || !hostRect) return undefined;`, `:96`).
`measureFloatingPanelPortalHost` answers `null` until `FloatingPanelPortalHost` has registered
its element from an effect (`floating-panel-portal.tsx:35`, registered at `:67`).

So when both attempts land before the host is registered, `relativeAnchorRect` stays `null`,
the component returns `null` (`:155`), and there is no third attempt. On mobile the
`useAnimatedReaction` on `isMoving` (`:124`) papers over this, because a keyboard transition
re-measures shortly afterwards — that reaction replaced the `keyboardDidShow`/`keyboardDidHide`
re-measure in #4275. Without keyboard motion nothing fires it, and a window resize is the only
thing left that re-runs the effect. That matches the report exactly.

Tab keeps working for an unrelated reason: `useAutocomplete` resolves Tab against a fallback
index when `selectedIndex` is still `-1` (`use-autocomplete.ts:97`–`:107`), which never consults
the popover. So the symptom is precisely "data fine, geometry missing".

This changes `remeasure` to treat an unusable measurement as *not yet measured* and try again on
the next frame, bounded to 12 frames and cancelled by the effect cleanup. Alongside a missing
host it also retries on a zero-height host and a zero-width anchor: those are accepted today and
position the popover against a host of height `0`, which puts it below the host's bottom edge and
hides it just as completely, again until a resize.

The reporter's own patch retried with `setTimeout(remeasure, 50)` and confirmed on a patched
bundle that the popover then renders on `/` without a resize. This is the same idea with a
frame-aligned clock, a bound, and cleanup.

### Goals

- Typing `/` renders the popover without a window resize when the portal host or the anchor need
  more than the two frames the popover currently allows.
- A measurement that yields no usable geometry is retried instead of discarded.
- The retry is bounded (12 frames) and cancelled by the effect cleanup, so closing the popover,
  reopening it, resizing, or unmounting cannot leave a frame callback behind.
- A host or anchor that measures as zero-sized counts as not-yet-laid-out rather than as a
  result, so the popover is not positioned off the bottom of its host.
- #4275's mobile behaviour is untouched: same `isMoving` reaction, same two attempts on open, no
  extra measurement once the geometry is in.

### Non-goals

- **Not restoring `options.length` and `selectedIndex` to the effect dependencies.** That is the
  reporter's second hunk and it would work, but it re-measures on every arrow-key press and on
  every keystroke that changes the filtered list — the measurement churn #4275 set out to remove.
  The retry fixes the reported failure without spending that on every platform.
- **Not polling.** The retry is frame-aligned and finite; it stops as soon as a measurement is
  usable, and gives up rather than running for as long as the popover is open.
- **Not touching the `isMoving` reaction, `useKeyboardShift`, or the mobile keyboard path.**
- **Not changing the anchor or the portal host.** Registering the host earlier, or handing the
  popover an `onLayout`, would both be larger changes to code that is not itself wrong; the
  popover is the component that needs to cope with measuring before layout has settled.
- No change to `canMeasure`, to the render guards, or to how the position is computed once a
  measurement lands.

### QA

Platform: macOS 26.6.2, arm64, Node 26.8.2. The app was built and run: Metro bundles it
(`HTTP 200`, 49,393,023 bytes) and `e2e/browser/composer-autocomplete.spec.ts` runs against a
real per-test daemon and a real Chromium.

**The existing popover e2e suite, on this branch:**

```
$ npx playwright test --project=browser e2e/browser/composer-autocomplete.spec.ts

  ✓  1 ... › stays visible after returning from app-wide routes (7.1s)
  ✓  2 ... › does not flash at the wrong position on the first slash command paint (4.3s)
  ✓  3 ... › does not jump when deleting a slash command search (3.9s)
  ✓  4 ... › shrinks to filtered slash command results without moving the bottom edge (3.9s)
  ✓  5 ... › stays visible while filtering slash command results (4.1s)
  ✓  6 ... › uses the live cursor when accepting a slash command (3.8s)
  ✓  7 ... › uses the live input when clicking a slash command (3.8s)
  ✓  8 ... › stays anchored to the composer when the desktop sidebar is open (3.8s)
  ✓  9 ... › compact sidebar layering › keeps the mobile agent sidebar above autocomplete (4.2s)
  9 passed (46.0s)
```

The popover as rendered on this branch, `/` typed into an empty composer:

The slash autocomplete popover rendering above the composer on this branch

<sub>Chromium 149.0.7827.55, web build, `/` typed into an empty composer. Hosted on a branch of my fork rather than committed here, so it stays out of the diff.</sub>

**Why there is no before/after pair, stated plainly.** I ran those same 9 tests on the merge
base and they also pass, `9 passed (51.7s)`, and a screenshot taken on the base at the same
point is indistinguishable from the one above. That is expected: in this harness the portal
host is registered and laid out well before `/` is typed, so the first measurement succeeds and
the regression never occurs. The bug needs the host to still be unmeasurable at that moment.

So a "popover absent on base" image would have to be produced by instrumenting the failure,
which would be evidence of the instrumentation and not of the bug. I have not staged one. This
is exactly why the regression test models the unmeasurable-host condition explicitly rather
than driving the UI.

**Browser deviation, disclosed.** Playwright 1.58.2 pins chromium and chromium-headless-shell
at revision 1208, and neither can be downloaded on this machine — Microsoft's CDN answers
`HTTP 400` with the body `GatewayExceptionResponse`. The runs above used revision 1228
(Chrome for Testing 149.0.7827.55), already present in the local Playwright cache, through a
throwaway config that spread the repo's `playwright.config.ts` and set only
`launchOptions.executablePath`. No repo file was changed and the throwaway config is not part
of this diff. A newer Chromium than pinned is the one way these runs differ from CI.

**The regression, as a test.** `autocomplete-popover.test.tsx` mounts the real component and
drives the prop order the composer actually produces (options land one render before
`useAutocomplete`'s effect picks an index, so `canMeasure` is false for that render and flips
true on the next). The portal host answers `null` for the first two measurements and real
geometry afterwards.

On the merge base, `7d74916f8e3c585bebffff809df0594a529806cd`, with only the test file added:

```
 ❯ |unit| src/components/ui/autocomplete-popover.test.tsx (4 tests | 2 failed) 118ms
     × keeps measuring until the portal host is laid out 15ms
     × gives up after a bounded number of frames when the host never becomes measurable 48ms

 FAIL  ... > keeps measuring until the portal host is laid out
AssertionError: expected null not to be null
 ❯ src/components/ui/autocomplete-popover.test.tsx:206:26

 FAIL  ... > gives up after a bounded number of frames when the host never becomes measurable
AssertionError: expected 2 to be greater than 2
 ❯ src/components/ui/autocomplete-popover.test.tsx:222:27

 Test Files  1 failed (1)
      Tests  2 failed | 2 passed (4)
```

`expected 2 to be greater than 2` is the defect measured directly: across 40 frames with a host
that never becomes measurable, the popover asks exactly twice and then never again.

On this branch:

```
 ✓ ... > anchors itself to the portal host once both are measurable 40ms
 ✓ ... > keeps measuring until the portal host is laid out 15ms
 ✓ ... > gives up after a bounded number of frames when the host never becomes measurable 74ms
 ✓ ... > does not keep scheduling measurements after it unmounts 16ms

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

The first case passes on both sides on purpose: it pins the geometry the popover computes from a
host rect (`left`, `width`, `bottom`) so the retry cannot "fix" the bug by rendering something
mispositioned.

**Whole `packages/app` unit suite.** Branch: `4 failed | 598 passed (602)` files. Merge base:
`5 failed | 596 passed (601)` files. The failing set on the branch is a strict subset of the
base's — `parse-changelog`, `svg-root-transform`, `utils/time`, `composer/draft/input-draft.live`,
plus `hooks/use-agent-history` on the base only. None of them touch this component, and the
counts move between runs, so they look order-dependent rather than related.

**Lint and format**, repo-wide on the branch:

```
> oxlint
Finished in 1.2s on 4254 files with 177 rules using 12 threads.

> oxfmt --check .
Finished in 1033ms on 4540 files using 12 threads.
```

**Typecheck** does not pass, and does not pass on the merge base either. With every workspace
package built, `tsgo --noEmit` in `packages/app` prints one error on both sides, byte for byte:

```
src/components/draggable-list.native.tsx(122,7): error TS2322: ...
  Property 'dragGestureHostPresented' does not exist on type 'IntrinsicAttributes & ...'
```

Nothing in this diff adds an error.

File and line references are against `main` @ `7d74916`.

### Checklist

- [ ] Plugin changes follow the [SDK import boundaries](../docs/plugins.md#sdk-import-boundaries) (if applicable)
- [x] One focused change
- [ ] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [ ] QA evidence

On the two unticked boxes: typecheck is red on the merge base for a reason this diff does not
touch, shown above. QA evidence is left unticked because it is incomplete against the bar in
CONTRIBUTING.md rather than absent — there is a screenshot and a green e2e run, but no
recording, and the only platform exercised is the web build under Chromium. The reported bug is
on Electron desktop, which I did not run, and the mobile path that #4275 was written for is
likewise unexercised here beyond the unit tests.


L4XB:
All three findings were right. Taken in full.

**Prohibited mocked component tests.** `docs/testing.md` allows unit tests with injected dependencies or real-browser tests and nothing in between, and names JSDOM and `@testing-library` component mounting explicitly. I should have read that before writing the suite, not after. The retry lifecycle is now its own module, `packages/app/src/components/ui/autocomplete-measurement.ts`, which takes the frame scheduler and the measurement as interfaces:

```ts
export interface FrameScheduler {
  request(callback: () => void): number;
  cancel(handle: number): void;
}
export interface MeasurementAttempt {
  measure(): Promise<readonly [Rect, Rect | null]>;
  onMeasured(rect: RelativeAnchorRect): void;
}
```

The test drives it with a frame queue it advances by hand and a scripted measurement. No JSDOM, no `vi.mock`, no mounting. The component keeps the real `requestAnimationFrame` adapter and is otherwise unchanged in shape. `autocomplete-popover.test.tsx` is deleted.

**Retries continue after success.** Correct, and worse than cosmetic in one direction you named: the two completions overlap, so an unusable one could queue a frame the usable one never cancelled. That frame measured again and re-rendered after the popover was already placed, and an unusable completion landing after a usable one could start the sequence over. Success now clears the budget and cancels the queued frame, so both are closed by construction rather than by ordering.

**Zero-size recovery is untested.** Also correct, and it was a real hole: deleting either check left every test green. There are now separate cells for a zero-height host and a zero-width anchor, each asserting nothing is reported until layout becomes usable and then the correct relative rect is.

While extracting it, two more gaps showed up that the old suite could not have caught, so they have cells too: two unusable measurements landing together must queue only one frame (otherwise the first handle is unreferenced and survives the stop meant to end it), and a run that starts after a stop must not retry.

Mutation battery over the loop, ten mutants, all killed:

| mutant | |
|---|---|
| host height check removed | killed |
| anchor width check removed | killed |
| null host accepted | killed |
| success no longer zeroes the budget | killed |
| success no longer cancels the queued retry | killed |
| retry budget ignored | killed |
| more than one retry in flight | killed |
| generation check removed | killed |
| stop refills instead of clearing the budget | killed |
| reset does not refill the budget | killed |

`oxlint` and `oxfmt --check` clean on the three files. `packages/app/src/components` is unchanged against the branch baseline: 18 failing tests before and after, all pre-existing in that path filter, with the 10 new ones added on top.

PR #4954, opened 2026-09-16
fix(nix): match node-pty prebuilds at either workspace hoisting location

CarsonF:
### Linked issue

Closes #

### Type of change

- [x] Bug fix

### Reasoning

Bumping node-pty to 1.2.0-beta.15 changed where npm hoists it, nesting
it under `packages/server/node_modules` instead of the workspace root.
`trace-daemon.mjs`'s glob for its prebuilt native addon still pointed
at the old root path, so it silently matched zero files and the Nix
daemon closure shipped without `pty.node`. Every terminal creation
then failed with "Terminal worker is not running," with no indication
the real cause was a missing native binary.

### Goals

- Ship `pty.node` in the Nix daemon closure regardless of which
  location npm hoists node-pty to.
- Fail the build loudly if an `additionalInputs` glob ever matches
  zero files, instead of silently shipping a broken daemon.

### Non-goals

- General node-pty hoisting/dependency-placement policy — this only
  fixes how `trace-daemon.mjs` finds the prebuilt addon.

### QA

Ran `nix build .#paseo` and confirmed `pty.node` is present in the
output closure at `packages/server/node_modules/node-pty/prebuilds/<platform>/`.

### Checklist

- [x] One focused change
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] QA evidence


cartercook:
+1 to get this merged

PR #4977, opened 2026-09-17
omp provider: mid-turn steering, paging drain, fast-mode passthrough, and lifecycle hardening

n24q02m:
## Summary
A batch of omp-provider improvements centered on long-running agent sessions:

**Features**
- **Mid-turn steering** (`steerActiveTurn`): input submitted while a turn is running is queued as steering input for the running turn instead of being dropped or implicitly starting a queued turn; covered by session-level tests including abort interaction.
- **Fast-mode passthrough**: the provider exposes the host's native Fast mode and keeps runtime model state in sync.

**Protocol/lifecycle fixes**
- **Messages paging drain**: `get_messages_page` drains with a legacy fallback when the daemon is busy or the cursor is stale, so history loads don't strand mid-page.
- **Abort escalation bound**: the RPC transport is terminated when an abort acknowledgement times out, instead of leaving a zombie transport.
- **Terminal settlement identity**: autonomous terminal settlement requires an identity check before finalizing.
- **Subagent announce ordering**: subagent lifecycle announcements are emitted before their metadata so consumers can register the agent before indexing it.
- **Completion gate deadline**: the turn completion gate is bounded by a deadline and custom notices stay non-terminal, preventing stuck "finishing" states.

## Test plan
Each bullet ships with provider-level tests (`agent.test.ts` + harness); full suite green on the provider package, including steer/abort interaction cases.


n24q02m:
Addressed both Greptile findings in 73e1ebe3c:

- **P1 (first history load fails)**: `getMessages` now falls back to `get_messages` when the runtime reports `unknown command` (protocol-v1), pinning `pagingUnsupported` so subsequent calls go straight to legacy. Test updated to assert the first load resolves via fallback.
- **P2 (3-positional-param rule)**: `request`/`startRequest` now take a single options object (`{ command, timeoutMs, requestOptions }`); all callsites updated (omp + pi providers, protocol-session, tests).

Local: 148/148 omp provider tests pass; typecheck clean on touched files.

n24q02m:
Addressed the new findings in cb8227298 + earlier commits:

- **P1 (first history load)**: real fix — `establishOmpProtocol` now returns whether RPC v2 was negotiated, and `OmpCliRuntimeSession` pins `pagingUnsupported` from it, so `get_messages_page` is never sent on protocol-v1 runtimes (which reply `Unknown command: <type>` with no id, previously hanging to timeout). The catch-fallback stays as a belt, broadened to a case-insensitive `unknown command` match.
- **P2 (duplicated options shape)**: extracted `JsonlRpcRequestArgs` — `startRequest`, `request`, and `OmpProtocolTransport.request` all share it.
- **P2 (timeout structure)**: `JsonlRpcTimeoutError` carries `operation`/`elapsedMs`/`timeoutMs`/`stderrTail` as fields.
- **P2 (fastModeEnabled/fastModeActive)**: not a bug — these are independent OMP protocol fields (enabled = user setting, active = currently applied); both optional because older OMP versions omit them entirely. Keeping as-is.

Local: 40/40 omp provider tests pass; typecheck clean on touched files.

PR #5040, opened 2026-09-18
Subscribe only the chats this session opened

boudra:
Reported in [Discord](https://discord.com/channels/1481169421832814616/1481169422990184542/1550336659382476974): *"this new 0.9 beta isnt properly keep track of when a workspace was last used. seems when i restart paseo they all show as recently accessed"*.

## What was happening

#4863 set out to stop evicting chats the user had open — the old hot set capped subscriptions at five and dropped the rest, so returning to a chat showed a stale timeline for a second. The cap was the thing to delete. Instead the membership **source** changed: it became `observeOpenWorkspaceAgentIds`, which walks the persisted workspace layout.

That layout is restored from disk at launch and `reconcileTabs` auto-opens an agent tab when you visit a workspace, so it carries a tab for every chat the user has ever opened on that host. At session start every one of those became a desired member, and `startAcknowledgedCatchUps()` ran an authoritative timeline fetch for each. `fetch_agent_timeline_request` calls `ensureAgentLoaded` → `client.resumeSession(...)`, so each fetch installed a provider session — a child process for CLI-backed providers, and nothing evicts them.

Each resume then did two more things:

- `registerSession` called `touchUpdatedAt` after the caller had explicitly passed the restored `updatedAt`, so the stored last-activity time became now;
- the resume path never read `requiresAttention` off the record, so the agent came back with `{requiresAttention: false}`.

Both were persisted. Workspace `statusEnteredAt` is re-derived from agent `updatedAt` on every daemon start, so the real last-used time was gone for good, and finished chats the user never looked at dropped out of **Ready to review**.

## Consumer-facing delta

| | before | after |
|---|---|---|
| Launch with 30 workspaces visited historically | 30 subscriptions, 30 resumes, 30 provider sessions | the chat on screen |
| Sidebar timestamps after a restart | all show the restart time, permanently | unchanged, whatever the agents last did |
| Ready to review after a restart | cleared on every restored tab | intact until the user reads the chat |
| Open 12 chats, switch workspaces, background, reconnect | 12 stay live | 12 stay live — unchanged, this is #4863's goal |
| Close a chat tab | released | released — unchanged |

## The changes

**Session owns the open set** (`viewed-timeline-sync.ts`, `session-context.tsx`). `openedAgentIds` starts empty and an agent joins it the first time `replaceVisibleAgentIds` reports it — the user opening it is the only way in. It survives hiding, workspace switching, retained-view eviction, backgrounding and reconnect, with no recent-agent limit. `replaceOpenAgentIds` becomes `replaceOpenTabAgentIds`: layout keeps its release role (closing a tab drops the chat) and loses its source role.

**Registration adopts restored timestamps** (`agent-manager.ts`). `touchUpdatedAt` now runs only when the caller supplied no `updatedAt`. Fresh creation and provider-session import are unaffected; resume and session reload keep the stored time.

**Resume carries stored attention** (`agent-loading.ts`, `persistence-hooks.ts`). New `extractAttention` sits beside `extractTimestamps`; `registerSession` already accepted an `attention` option. Reading the chat still clears attention through the client's `workspace.clear_attention`, which is properly gated on screen focus and app visibility. The duplicate inline projection in `dispatchStoredAgentState` now calls the shared one.

The daemon half is defence in depth — it makes a resume inert no matter who triggers it, including Hub follow-ups, MCP and notification deep-links, all of which lose the unread flag today.

## Behaviour changes worth a second look

- **Navigating chats while disconnected no longer resets the open set.** Pre-#4863 a disconnect reset the hot set to whatever was visible; that was paired with the five-cap. With the cap gone by design, dropping open chats on a socket blip reintroduces exactly the staleness #4863 removed. `navigation while disconnected reconnects only the currently visible agent` is rewritten accordingly.
- **Hiding a pane no longer evicts its chat.** Two tests conflated hiding with closing; they now close a tab to assert the release.
- **Session reload keeps its stored `updatedAt`** (it passes one). Reloading an agent after a config or provider change was never activity either.

## Verification

- `packages/app/src/timeline/viewed-timeline-sync.test.ts` — new: a restored layout subscribes nothing until the user opens a chat. 36 pass.
- `packages/server/src/server/agent/agent-loading.test.ts` — new: resuming keeps `requiresAttention`, `attentionReason`, `updatedAt` and `lastActivityAt`. Verified red before the fix (`expected false to be true`), green after.
- `packages/app/src/stores/workspace-layout-store.test.ts` — 169 pass with the timeline suite.
- End-to-end against a daemon, using a throwaway home whose agents finished five days ago: cold start, then a client catching up all three open chats, then a second cold start. Before, all three workspaces read `0.00d ago` from the catch-up onward and the agent records came back `attention=false` with a rewritten `updatedAt`. After, all three hold `statusEnteredAt=2026-09-13… (5.00d ago)` and `attention=true reason=finished` through every phase.

Not run locally: `packages/app/e2e/browser/viewed-agent-timelines.spec.ts`. Its scenarios open every chat through the UI, so they should be unaffected, but CI is the check. Separately, `src/server/agent/mcp-parity.e2e.test.ts` has four failing worktree-tool tests on this machine that this branch does not touch — please confirm against main rather than reading them as fallout here.


boudra:
Adversarial review (Opus) found one confirmed regression in the daemon half. Fixed in 663bb503b; two follow-ups noted below.

## Fixed — resume could move `updatedAt` backwards, and the client drops backwards updates

Dropping the registration stamp exposed a skew the stamp had been erasing. `extractTimestamps` read `lastActivityAt ?? updatedAt`, but the two diverge whenever something mutates an **unloaded** agent: `writeStoredMetadata` (rename, labels, detach), `markAgentUnread`, restore from archive, and the stored branch of `workspace.clear_attention` all move `updatedAt` and leave `lastActivityAt` alone. Not hypothetical — 8 of 854 agent records on one dev machine already carry that skew, the largest by 8 days.

`acceptAgentDirectoryUpdate` (`agent-directory-update-policy.ts:17`) accepts an update only when `incoming.updatedAt >= current.updatedAt` and otherwise keeps the current value, merging nothing but `lastUsage`. So:

> Chat finished 5 days ago (T1). You mark it unread from the sidebar without opening it — the agent is unloaded, so the record gets `updatedAt = T2 = now` and the client holds T2. Later you open the chat. The open resumes the agent with `updatedAt = lastActivityAt = T1`. Reading it fires `workspace.clear_attention`, the daemon clears attention and emits state with `updatedAt = T1 < T2` — and the client refuses the update. The unread dot stays in History, the agent list, Command Center and the tab until the agent does real work or a full `fetch_agents` replaces the entry.

The reviewer reproduced this against the real `AgentManager` + `AgentStorage`, then fed the resulting pair to the real `acceptAgentDirectoryUpdate` and got `requiresAttention: true` back. On `main` it cannot happen, because `touchUpdatedAt` forced `now`.

Fix: read both fields and take the newest — already the rule `resolveStoredAgentPayloadUpdatedAt` applies to the list payload. It moved beside the other stored-record readers as `resolveStoredAgentUpdatedAt` so one rule now covers both paths instead of two divergent re-derivations. The existing resume test now skews `updatedAt` ahead of `lastActivityAt`; verified red before the fix (`expected '2026-01-02…' to be '2026-01-09…'`), green after.

One consequence worth stating: because `toStoredAgentRecord` writes both stored fields from the single in-memory `updatedAt`, a resume now collapses them to their max. That loses a little `lastActivityAt` fidelity, but it is bounded by values already on the record and can never reach `now`, which is what mattered.

## Also changed — `restoring` instead of inferring from `updatedAt`

Replaced `!options?.updatedAt` with an explicit `restoring` flag on `registerSession`. The inference was correct for all four call sites, but it made a second meaning out of a field with nothing typing or testing it; the next caller to seed a timestamp for an unrelated reason would have opted out of activity stamping silently.

## Also added — a test at the store→sync seam

`session-context.tsx:500` is the single line keeping layout out of the membership source, and only a doc comment protected it. `workspace-layout-store.test.ts` now wires the real `observeOpenWorkspaceAgentIds` into a real sync exactly as session-context does and asserts a restored two-tab layout produces no subscription until a chat becomes visible. Repointing that callback at `replaceVisibleAgentIds` fails it with `expected [['agent-a','agent-b']] to deeply equal []`.

## Findings taken as follow-ups, not fixed here

**`ManagedAgent.updatedAt` is one field doing two jobs.** The record has two (`updatedAt`, `lastActivityAt`); `toStoredAgentRecord` writes both from the same value, and two call sites each paper over the collapse differently on the way back. That collapse is what produced the regression above. Giving `ManagedAgent` a real `lastActivityAt` and leaving `updatedAt` as monotonic record-mutation time would delete the class, but it is a larger change than this PR should carry.

**`ensureAgentLoaded`'s no-persistence-handle branch** still stamps activity and returns no attention. It is not a resume — there is no provider session to bring back, so it starts the agent's first one — and records without a handle have not accumulated attention. Narrowed the PR's "inert no matter who triggers it" claim and left a comment at the branch so it reads as deliberate.

## Two things the review confirmed that are worth recording

- **The cost was gigabytes, not megabytes.** Each Codex resume spawns its own `codex app-server`. Measured live on the dev machine: 61 of them at ~73 MB each (4.4 GB), plus 16 `claude` sessions at ~290 MB each (4.6 GB). `docs/agent-lifecycle.md:28` states idle agents stay resident indefinitely, so nothing reaped them. This PR changes *N = every chat ever opened on the host, every launch* to *N = chats opened this session*. The tail is still unbounded; bounding it belongs in daemon runtime residency, not in the app's subscription set.
- **A user-visible delta missing from the table above:** after an app restart, the first open of each chat catches up again, with a brief "Updating messages". #4863 had made that instant by pre-resuming everything.

Rerun green: 170 app tests (`viewed-timeline-sync`, `workspace-layout-store`), 210 server tests (`agent-loading`, `agent-projections`, `agent-manager`, `agent-archive`).

PR #5077, opened 2026-09-19
fix(app): reattach workspace labels when a connection is restored

morven-ai:
## Problem

On one host the workspace label picker renders every row disabled and stays that way for the
whole life of the connection. Other hosts in the same sidebar are fine, and everything else on
the affected host — agent streams, workspace updates — keeps working, so nothing looks wrong
until you try to assign a label. Relaunching the app does not help.

The rows are still visible, because `projectWorkspaceLabels()` merges the catalogs of every
online host while `disabled` is decided by the target host alone. So you get a list of labels, in
the right colors, that cannot be pressed — and no explanation, because `WorkspaceLabelPickerPage`
renders a `MenuHint` for `status === "unsupported"` but nothing for `offline`.

## Root cause

`DirectorySync.connectionChanged()` drops the replica on every connection change:

```ts
workspaceLabels.disconnect(this.serverId);   // index.ts:159
```

and never puts it back. The only caller of `connectWorkspaceLabels()` is `refreshAll()`, which is
one of the two branches of `requestDemandRefresh()`:

```ts
const refresh =
  this.fullDemandSources.size > 0
    ? this.refreshAll()                      // attaches labels
    : Promise.all([...]).then(() => undefined);   // does not
this.demandRefresh = refresh
  .then(() => {
    this.satisfiedDemandSource = source;     // marked satisfied either way
```

Both branches mark the connection epoch satisfied. So when route-only demand refreshes first —
the app reconnecting with a conversation on screen and no full-directory demand registered yet —
the later full demand early-returns against `satisfiedDemandSource`, `refreshAll()` never runs
for that connection, and the label channel stays detached.

Every other subscription torn down by `connectionChanged()` is restored inline in the same
method. The label replica is the only one left to a conditional path, which is why the host looks
healthy in every other respect.

## Fix

Attach the replica where the subscriptions are restored, so it no longer depends on which demand
branch wins the epoch.

## Tests

`packages/app/src/runtime/directory-sync/workspace-labels-attach.test.ts`, two cases:

- **route-only demand, then full demand** — fails on `main`:
  `AssertionError: expected 0 to be greater than 0` on
  `expect(client.listWorkspaceLabelsCalls).toBeGreaterThan(0)`. The client never lists labels for
  that connection.
- **full demand first** — passes on `main`, and is here so the fix cannot regress the path that
  already worked.

Both pass with the change, and `src/runtime/directory-sync/` is green.

## Open question

With this in place, a connection that does take the `refreshAll()` path attaches twice, costing
one extra `workspace.label.list.request` per connect. Two ways to settle that, happy to take
either: drop the call from `refreshAll()`, or make `WorkspaceLabelsController.connect()` a no-op
when the same client is already attached. Left alone here to keep the change minimal.

## Also worth a separate look

The picker explains `unsupported` but says nothing for `offline`. A disabled picker showing rows
borrowed from another host is unreadable — an offline hint would turn this class of report into a
five-second diagnosis.


morven-ai:
Superseded by #5079, which is the same change rebased onto current main (the branch here was cut from a stale main and conflicted with the connectionChanged refactor).

PR #5081, opened 2026-09-19
feat(plugins): pass the initial prompt to agent.create hooks

omerbentov:
## Problem

`server.before("agent.create")` can change an agent's provider, model, mode and thinking level before the session opens, but it cannot see the task the agent is created for: its request carries only `config` and `env`. A plugin that wants to choose the configuration from the task (for example, pick an agent profile per task) has nothing to decide with, and has to replace the normal New agent flow with its own slash command.

The daemon already has the prompt when it runs the hook: `createAgentCommand` sets `CreateAgentOptions.initialPrompt`, and `AgentManager.createAgent` drops it when building the hook request.

Discussion: #5080

## Change

- `agent.create` hook requests include `initialPrompt` when the creation request has one (`packages/server/src/server/agent/agent-manager.ts`).
- It is read-only. A hook that returns a different value fails creation, like changing `cwd`. A hook that returns a request without it, as hooks written before this field do, keeps it (`packages/server/src/server/plugins/lifecycle/index.ts`).
- Agents created through MCP pass their prompt too, so every creation path behaves the same (`create-agent/create.ts`).
- `PluginBeforeRequests["agent.create"]` gains `initialPrompt?: string`; the reference docs list the field.

No protocol change: `initialPrompt` is an existing optional field of `create_agent_request`.

## Automated tests

- `lifecycle/handlers.test.ts`: a hook reads the prompt; a hook that changes it is rejected; a hook that returns a request without it keeps it.
- `lifecycle.e2e.test.ts`: a plugin installed in a test daemon switches `claude` to `codex` when the prompt starts with "Review", and leaves an agent without a prompt alone.

Both new tests fail on `main`:

```
$ npx vitest run packages/server/src/server/plugins/lifecycle/handlers.test.ts --bail=1
 × agent.create hooks read the initial prompt but cannot change it
ZodError: [ … "message": "Unrecognized key: \"initialPrompt\"" ]
      Tests  1 failed | 4 passed (5)

$ npx vitest run packages/server/src/server/plugins/lifecycle.e2e.test.ts -t "initial prompt" --bail=1
 × agent creation hooks choose the provider from the initial prompt
AssertionError: expected 'claude' to be 'codex'
```

After the change:

```
$ npx vitest run packages/server/src/server/plugins/lifecycle/handlers.test.ts --bail=1
      Tests  6 passed (6)
$ npx vitest run packages/server/src/server/plugins/lifecycle.e2e.test.ts packages/server/src/server/plugins/agent-configuration.e2e.test.ts --bail=1
      Tests  6 passed (6)
$ npx vitest run packages/server/src/server/agent/create-agent/create.test.ts --bail=1
      Tests  10 passed (10)
$ npm run typecheck --workspace=@getpaseo/plugin   # ok
$ npm run typecheck --workspace=@getpaseo/server   # ok
$ npm run typecheck --workspace=@getpaseo/cli      # ok
$ npm run lint -- <changed files>
```

## QA

Dev daemon from this branch (`npm run dev:server`, 0.9.0-beta.2, `127.0.0.1:6768`, checkout-local home) with real Claude Code and Codex providers, and this plugin installed:

```ts
server.before("agent.create", ({ request }) => {
  const prompt = request.initialPrompt ?? null;
  const routeToCodex = prompt?.startsWith("Review") ?? false;
  console.log(JSON.stringify({ hook: "agent.create", provider: request.config.provider, initialPrompt: prompt, routeToCodex }));
  if (prompt?.startsWith("Quick")) {
    return { ...request, config: { ...request.config, model: "claude-sonnet-5" } };
  }
  if (!routeToCodex) return;
  const { model: _model, modeId: _modeId, thinkingOptionId: _thinking, ...config } = request.config;
  return { ...request, config: { ...config, provider: "codex" } };
});
```

Every agent was requested as Claude:

```
$ paseo run --provider claude "Review this empty folder. Reply with the single word OK."
{ "agentId": "e0cdf507-…", "status": "error", "provider": "codex", … }
$ paseo logs e0cdf507
[User] Review this empty folder. Reply with the single word OK.
[System Error] You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage …

$ paseo run --provider claude --model claude-opus-5 "Quick check: reply with the single word OK."
$ paseo inspect d5a6010b
Provider            claude
Model               claude-sonnet-5
Status              idle
[User] Quick check: reply with the single word OK.
OK

$ paseo run --provider claude --model claude-opus-5 "Say hello. Reply with the single word OK."
$ paseo inspect 150fe815
Provider            claude
Model               claude-opus-5
Status              idle
[User] Say hello. Reply with the single word OK.
OK

$ paseo plugin logs qa-prompt-router
{"hook":"agent.create","provider":"claude","initialPrompt":"Review this empty folder. Reply with the single word OK.","routeToCodex":true}
{"hook":"agent.create","provider":"claude","initialPrompt":"Quick check: reply with the single word OK.","routeToCodex":false}
{"hook":"agent.create","provider":"claude","initialPrompt":"Say hello. Reply with the single word OK.","routeToCodex":false}
```

The Codex agent was created on Codex with the prompt delivered; its turn stopped on my Codex account's usage limit, which is unrelated to this change. The provider switch itself is also covered by the e2e test.

| Desktop macOS   | Yes    | Daemon change; dev daemon + CLI on macOS 26.6 |
| Desktop Windows | No     | Daemon-only change, no platform-specific code |
| Desktop Linux   | No     | Same |
| iOS / Android / Web | N/A | No client change |

The app's New agent composer sends the same `create_agent_request` with `initialPrompt` as `paseo run`; I exercised it through the CLI, not the app UI.

Motivating plugin: https://github.com/omerbentov/paseo-jev-agent-router (today it needs a `/route` slash command for this).


omerbentov:
@boudra :)

PR #5085, opened 2026-09-19
fix(app): keep placement-less directory members across sequenced catch-ups

### LINKED ISSUE #5084: bug: sidebar drops conversations on sequenced catch-up after a reconnect
### What's broken

After the app reconnects to a daemon, the sidebar keeps only the conversations that changed since the client's sync cursor. Everything older disappears. The conversations are intact on the daemon and come back on a full app restart, then drain away again on the next reconnect. On a workstation where the socket lease expires roughly every 30 minutes, the directory empties down to the running agents several times a day, on desktop and on the phone alike.

### Steps to reproduce

Deterministic unit-level reproduction (`packages/app/src/runtime/directory-sync/agent-replica.test.ts`):

1. Hydrate an agent through its timeline: `replica.submitTimelineAgent(replica.captureTimeline("agent"), payload("timeline only"))`. It is a directory member from then on (`members.add` in `submitTimelineAgent`).
2. Apply an empty sequenced catch-up: `replica.commitChanges([], [], [])`.
3. The agent is gone from `sessions[serverId].agents`.

Cause: `commitChanges` rebuilds the merged directory from the previous store contents through `projectAgentDirectoryEntry`, which returns `null` for any agent without a `projectPlacement`. Those members never reach the merged map, so `commitSnapshot` treats them as absent and calls `removeFromDirectory` on them, and the replica cache is told to delete them. An agent hydrated from its timeline, or restored from the replica cache, carries no placement, so the first delta after a reconnect drops it. `reconcileAgentDirectory` already synthesizes a placement from `cwd` for deltas in the same flow, so the two halves disagree.

In the field this is what it looks like: on reconnect the app sends two fetches two milliseconds apart, and the cursored one is the sidebar's.

```
fetch_agents_request  scope=active  sync={generation:<uuid>, afterSeq:27}   -> 11 entries (changed since 27)
fetch_agents_request  scope=active  (no sync cursor)                        -> 26 entries (the full active set)
```

The daemon is correct in both: `directory-sync/index.ts` serves a delta when the cursor generation matches and a snapshot when it does not, and the full set is on the wire in the same capture. The 15 conversations that vanish are the ones the client dropped locally.

### Where did this happen

Desktop (Electron), same symptom on the iOS app against the same daemon.

### Paseo version

0.8.0 (app and daemon).

### OS version

Ubuntu, Linux 7.0.0-31-generic. Daemon runs under systemd --user.

### Agent provider

Not relevant. Affected conversations spanned Claude Code, Codex, OpenCode and OMP providers.

### Anything else

A fix with a regression test is in the linked PR.

bagutzu:
Fixes #5084.

## Problem

After a reconnect, the sidebar keeps only the conversations that changed since the client's sync cursor. The rest disappear until the app is restarted, then drain away again on the next reconnect. On a machine where the socket lease expires roughly every 30 minutes this empties the directory down to the running agents several times a day, on desktop and phone alike, while the daemon still holds and serves every conversation.

## Cause

`commitChanges` rebuilds the merged directory from the previous store contents through `projectAgentDirectoryEntry`, which returned `null` for any agent without a `projectPlacement`. Those members never reached the merged map, so `commitSnapshot` saw them as absent, called `removeFromDirectory` on them and persisted a delete into the replica cache.

An agent hydrated from its timeline is a directory member (`members.add` in `submitTimelineAgent`) but carries no placement, so the first sequenced catch-up after a reconnect drops it. `reconcileAgentDirectory` already synthesizes a placement from `cwd` for deltas in the same flow, so the two halves of `commitChanges` disagreed about what a member without a placement means.

## Fix

Derive the placement from the agent's `cwd` with the existing `resolveProjectPlacement` helper instead of dropping the member. 12 lines of production code, no protocol or daemon change.

## QA evidence

Regression test added to `agent-replica.test.ts`: hydrate an agent through its timeline, apply an empty sequenced catch-up, assert the agent is still in the directory.

Fails on `main`:

```
$ npx vitest run src/runtime/directory-sync/agent-replica.test.ts -t "timeline-hydrated"
 × keeps a timeline-hydrated agent when a sequenced catch-up carries no entries
AssertionError: expected undefined to be defined
 ❯ src/runtime/directory-sync/agent-replica.test.ts:321
```

Passes with the fix, together with the rest of the directory-sync suites:

```
$ cd packages/app && npx vitest run src/runtime/directory-sync/agent-replica.test.ts --bail=1
 Test Files  1 passed (1)
      Tests  11 passed (11)

$ npx vitest run src/runtime/directory-sync src/utils/agent-directory-sync.test.ts
 Test Files  5 passed (5)
      Tests  48 passed (48)

$ npm run typecheck --workspaces --if-present   # via the pre-commit hook
✔️ format  ✔️ lint  ✔️ typecheck
```

Field evidence from the daemon this was found on (desktop app 0.8.0 against daemon 0.8.0, Linux): a packet capture of the reconnect shows the app issuing both fetches two milliseconds apart, the cursored one returning 11 entries and the uncursored one returning the full 26, with all 26 present on the wire. The daemon side behaves correctly in both; the 15 conversations that vanish are dropped by the client.

Not verified: I did not rebuild and run the packaged Expo app against a live daemon after the fix, so there is no before/after recording of the sidebar itself. The reproduction and the verification here are at the replica level, which is where the conversations are lost.


muzhi1991:
I'm seeing a matching symptom on iOS 0.8.0 connected to a macOS daemon 0.8.0: the sidebar eventually shows only running workspaces, and restarting the daemon restores the list. My daemon has local timeline/tool-result size patches; I cannot confirm that my incident has exactly the same root cause as #5084.

After reviewing 28f22b36, this looks like the right fix for the reproduced placement-less-member bug:

- Missing placement no longer silently removes an existing member during incremental catch-up.
- The existing resolver preserves authoritative placement and derives a fallback only when it is missing.
- Server entries still replace prior entries, explicit removals are still applied after the merge, and full snapshots remain authoritative.

The regression test targets the right failure. Small test improvements would be to assert the retained agent's identity/content and resolved placement instead of only `toBeDefined()`, and pass `null` directly to `initializeSession` without the double cast.

This is a source review, not additional runtime verification: I have not rerun the tests or tested a patched iOS build. App-level reconnect QA is still outstanding, as noted in the PR.

Please consider prioritizing the remaining validation and merging this fix for an upcoming release, including iOS. Losing the sidebar entries until a restart is disruptive in everyday use. Thanks for investigating and keeping the fix focused!

boudra:
Closing in favor of [#5189](https://github.com/getpaseo/paseo/pull/5189), which carries this fix.

I diagnosed [#5084](https://github.com/getpaseo/paseo/issues/5084) independently and landed on the same change you made here: derive the placement from `cwd` through `resolveProjectPlacement` so `projectAgentDirectoryEntry` is total. Your commits are credited with a `Co-authored-by` trailer on that branch.

What #5189 adds is a second regression test at the `DirectorySync` level. It restores a cached agent and an agent cursor from the replica cache, then answers the cursored fetch with `sync: { mode: "changes", removals: [] }` and no entries, which is the shape captured on the wire in the issue. The existing workspace equivalent of that test already passed, so the agent version locks the reconnect path rather than the replica in isolation.

@muzhi1991 your iOS report is the same symptom. I have added a note to [#5084](https://github.com/getpaseo/paseo/issues/5084) and updates will go there.

PR #5123, opened 2026-09-21
fix(server): reset Codex default reviewer

big-branch:
## Summary
- send Codex `approvalsReviewer: "user"` for Default Permissions
- apply the reset on unloaded thread resume and loaded threads’ next turn
- cover fresh, unloaded-resume, and loaded-next-turn paths with zero-model-call app-server fixtures

## Verification
- `npx vitest run packages/server/src/server/agent/providers/codex-app-server-agent.test.ts --bail=1`
- `npm run typecheck`
- `npm run lint`
- `npm run format`


big-branch (commented):
verdict: ship

Independent terminating cold review bound to head `7656fd88d649377b373b124560d9d451f9ab2f25`. Worktree at that exact SHA, tree clean, nothing edited. No commits, pushes, merges, installs, or service restarts.

## Gates (all green, local)

| Gate | Result |
| --- | --- |
| `npx vitest run packages/server/src/server/agent/providers/codex-app-server-agent.test.ts --bail=1` | 1 file, **153 passed**, 3.34s |
| `PASEO_NATIVE_REVIEWER_QA=1 npx vitest run packages/server/src/server/agent/providers/codex-default-reviewer.native.real.e2e.test.ts --bail=1` | 1 file, **2 passed**, 1.80s (gate confirmed active — not silently skipped) |
| `npm run typecheck` | exit 0 |
| `npm run lint` (oxlint) | 0 warnings, 0 errors, 4271 files |
| `npm run format:check` (oxfmt) | all 4557 files correctly formatted |

## Out-of-band native protocol verification

I ran an independent raw-protocol probe against real `codex-cli 0.154.0` in an isolated `CODEX_HOME`, outside the repo, with **no `turn/start`** and therefore no model turn. It confirms the premises the PR rests on:

- `thread/start` with `approvalsReviewer: "auto_review"` reads back `approvalsReviewer: auto_review`, `approvalPolicy: on-request`, `sandbox: {type: workspaceWrite}`.
- After app-server restart, `thread/loaded/list` is `[]` and **`thread/resume` with no override returns `approvalsReviewer: "auto_review"`** — the persistence bug this PR fixes is real, not assumed.
- **`thread/resume` on an already-loaded thread with `approvalsReviewer: "user"` still reads back `auto_review`** — real Codex does ignore resume overrides on loaded threads, so the fixture's stubbed behaviour matches native reality.

Conclusion: the fix is correct and the mode matrix is right. `auto` → `user`, `auto-review` → `auto_review` (unchanged), and `user` is sent on all three paths — `thread/start` (fresh), `thread/resume` (unloaded), and `turn/start` (loaded next turn). `approvalPolicy` stays `on-request` and sandbox stays `workspaceWrite` throughout, asserted in both the unit tests and the native readback.

**Runtime-injected MCP configuration and tool policy are untouched.** `buildCodexInnerConfig()` (which assembles `mcp_servers` and applies `applyCodexToolPolicy`) is unmodified and still applied in `ensureThreadLoaded` after the new reviewer line; the new param writes a distinct key and clobbers nothing.

**Cleanup / leakage:** `CodexAppServerClient.dispose()` goes through `terminateWithTreeKill`, so the app-server process tree is reaped. `CODEX_HOME` is redirected to a per-test tmpdir, so the real `~/.codex` auth and session rollouts are never read or written; all fixture state and history lands under `mkdtemp` and is `rm -rf`'d in `finally`. No credentials or history reach the repo or test output.

## Non-blocking findings

1. **Test 1 doesn't actually prove persistence.** It resumes *with* `approvalsReviewer: "user"` and asserts `user`. That assertion passes identically if Codex merely echoes the param. The missing step is a no-override `thread/resume` asserting `auto_review` before the override resume. I verified that behaviour out-of-band (above), so the test isn't asserting something false — it is under-specified.
2. **Test 2 is a stub living in a `.native.real.e2e` file.** It spawns `persisted-reviewer-app-server.ts`, not `codex`. `expect(finalState.resumeOverridesIgnored).toBe(1)` re-asserts the stub's own hardcoded `thread/resume` handler rather than native behaviour. Real Codex does match it (verified above), but the file name promises native coverage this test doesn't provide.
3. **`turn/start` clearing a loaded native thread's reviewer is unverified natively** — that would require a real model turn. The `applyApprovalsReviewerParam` call at `turn/start` is pre-existing plumbing already exercised for `auto_review`, so the risk is low.
4. **Test 1 is not literally zero-network.** Real `thread/start` opens an outbound `wss://api.openai.com/v1/responses` connect (401 in my unauthenticated probe). No turn is started, so no model call and no billing — worth knowing before this runs on a credentialed QA box.
5. **Child-process leak on failure paths.** In `startNativeClient`, if `initialize` rejects, the spawned `codex` child is never killed — the client is constructed inline and the caller's `client` variable is never assigned, so `finally` can't dispose it. Same shape for the inline `new CodexAppServerClient(spawnFixture(), ...)` in test 2. Only bites when a test fails.
6. **`PASEO_NATIVE_REVIEWER_QA` is undocumented and not wired to CI**, so the native path runs only when someone sets it by hand. This matches the existing convention — `PASEO_NATIVE_ARCHIVE_QA` and `PASEO_NATIVE_RELOAD_QA` are equally undocumented — so it is not a regression. CI coverage for all three lifecycle paths does exist via the new unit tests, which do run unconditionally.
7. **Scope note, same class of bug.** `read-only` and `full-access` presets still carry no `approvalsReviewer`, so a thread that previously used Auto-review keeps `auto_review` natively after switching to either mode. Out of scope here; worth a follow-up.

## Blocker

**`action_required` on CI.** The `CI` workflow run for `7656fd88d` is `completed / action_required` — it requires maintainer approval to execute and has never actually run on this head. The only check that reported is `Greptile Review` (success). `mergeable_state` is `blocked` and the combined commit status is `pending`.

Nothing in the change is blocking; the local gates above stand in for CI, but full-suite green on this head cannot be attested until the workflow is approved.

